PR #11, opened 2017-11-08
Fix some formatting issues in docs/src/index.md

timholy:
Fixes the enumeration list in the [first docs page](http://www.glvisualize.com/MakiE.jl/stable/). Also fills out the `plotsbase` link.

You probably know this, but I think the docs could benefit from a bit more of a "tutorial" approach. Time permitting, as I learn this package I'll try to help out.


SimonDanisch:
Thanks a lot! I feel like I recently fixed the link already, so maybe this is an old version :( Let me check what I have locally...

PR #45, opened 2018-01-10
remove time + pessimistic_sleep

SimonDanisch:
this will remove the time attribute from a scene, since this was always waking up the whole event loop.
Removing it decreased the time spent on the cpu from ~20% to ~0  cpu usage for idle time.

I'd say, people should just implement their own loop for doing animations relying on the time.
cc @jpsamaroo
@ssfrr, what do you think, you used `scene[:time]` so far.

### DIFF
--- a/src/plotsbase/scene.jl
+++ b/src/plotsbase/scene.jl
@@ -201,12 +201,14 @@ function render_loop(tsig, screen, framerate = 1/60)
     while isopen(screen)
         t = time()
         GLWindow.poll_glfw() # GLFW poll
-        push!(tsig, t)
         if Base.n_avail(Reactive._messages) > 0
             render_frame(screen)
         end
         t = time() - t
-        GLWindow.sleep_pessimistic(framerate - t)
+        to_sleep = framerate - t
+        if to_sleep > 0.001 # minimal sleep time
+            sleep(to_sleep)
+        end
     end
     GLWindow.destroy!(screen)
     return


ssfrr:
I used it assuming it was doing something clever on the GPU, if that's not the case then there's less motivation for it.

It does seem like there should be an easy way to do realtime animations in a way that's synchronized with the frame-rate, and having a `scene[:time]` signal that's updated every frame seems like one reasonable way to do that. Honestly though I'm still trying to wrap my head around what idiomatic Makie code should look like, mostly with regards to the FRP bits (e.g. when to use a signal vs. a loop that just updates things in the scene).

SimonDanisch:
I think, the most idiomatic code should look like this:
```Julia
while isopen(scene) # or whatever loop, event callback one wants to chose
    scene[:value] = newvalue
   ...
# or recordframe! or actually just a yield() which should, but is not guaranteed, to yield to renderframe in the actual renderloop
   renderframe(scene) 
end
```

jpsamaroo:
I still need to try this out locally, but thanks for addressing this so quickly! I do feel that the burden of animations should be on either the user or some other package, since the decision of when to yield and render can vary based on the situation, so this makes sense to me!

SimonDanisch:
yeah it should be programmed to at least opt out in some way.. not sure if we could get the best of both otherwise

jpsamaroo:
Just confirmed that this branch brings down CPU load to about 1% or less in htop. Great work!

SimonDanisch:
I was thinking, I could check if time has any registered observers... But I'm not sure if I want this kind of special treatment and check every loop iteration ;)

PR #70, opened 2018-02-27
RFH: support polygon drawing

timholy:
I want to draw filled polygons in images like this:

image

(from [QuadDIRECT](https://github.com/timholy/QuadDIRECT.jl)). I have a sense that this is missing some important stuff, but because time is short I'm wondering if I could get some guidance. In exchange, I'd be happy to contribute to the devdocs a bit, once I understand more.


SimonDanisch:
Thanks! :) Let me have a look at it!

SimonDanisch:
See: https://github.com/JuliaPlots/Makie.jl/pull/71

PR #71, opened 2018-02-28
polys

SimonDanisch:
improves uppon #70


src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
-    for poly in polys
-        mesh = GLNormalMesh(poly) # make polygon
-        if !isempty(GeometryTypes.faces(mesh)) # check if polygonation has any faces
-            push!(result, GLVisualize.visualize(mesh, Style(:default), kw_args))
-        else
-            warn("Couldn't draw the polygon: $points")
-        end
+function _poly(scene::makie, attributes::Dict)
+    attributes = poly_defaults(scene, attributes)
+    println(typeof(attributes[:positions]))
```
timholy: Remove?

src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
+    bigmesh = lift_node(attributes[:positions]) do p
```
timholy: Any guidance on when `lift_node` is used? (I will contribute to docs.) This is one of the things I see everywhere in the GL* packages but don't really understand.

src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
+        polys = GeometryTypes.split_intersections(p)
+        merge(GLPlainMesh.(polys))
```
timholy: What's the `merge` doing here? This is from `GeometryTypes/src/meshes.jl`, right? Does this basically put multiple meshes into a single structure?

src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
     end
-    result
+    mesh(scene, bigmesh, color = attributes[:color])
+    line = lift_node(attributes[:positions]) do p
+        Point2f0[p; p[1:1]]
+    end
+    println(typeof(line))
```
timholy: Another thing to delete?

src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
-        end
+function _poly(scene::makie, attributes::Dict)
+    attributes = poly_defaults(scene, attributes)
+    println(typeof(attributes[:positions]))
+    bigmesh = lift_node(attributes[:positions]) do p
+        polys = GeometryTypes.split_intersections(p)
+        merge(GLPlainMesh.(polys))
     end
-    result
+    mesh(scene, bigmesh, color = attributes[:color])
```
timholy: So this basically does the `fill` part, right?

src/atomics/poly.jl
```diff
@@ -1,27 +1,39 @@
...
+    line = lift_node(attributes[:positions]) do p
+        Point2f0[p; p[1:1]]
+    end
+    println(typeof(line))
+    lines(scene, line,
```
timholy: And this the border?

src/plotsbase/atomics.jl
```diff
@@ -100,6 +100,25 @@ end
     )
 end
 
+@default function poly(scene, kw_args)
```
timholy: What happens to keywords that you don't handle in these `@default` functions?

timholy:
Would you like me to submit a PR to delete those `println` statements?

src/atomics/poly.jl
SimonDanisch: yes! will need something more sophisticated in the future!

src/atomics/poly.jl
SimonDanisch: woops, yes :-D

src/atomics/poly.jl
SimonDanisch: it's basically the same as Reactive.map(f, signal) and just registers a callback that gets called whenever 'signal' updates... it will return a new signal which holds 'f(value(signal))'

src/atomics/poly.jl
SimonDanisch: exactly!

src/atomics/poly.jl
SimonDanisch: yup :-D

src/plotsbase/atomics.jl
SimonDanisch: they will get passed through to the backend... in my new branch, i'm warning at a point where i can be sure that the additional kw args are not used!

SimonDanisch:
if you have the time :) i can also delete them tomorrow!

PR #77, opened 2018-03-22
One color per edge

edljk:
Hope it is in the right place..


SimonDanisch:
That took a while to incorporate ;) Sorry for not merging yours, but this PR was hopelessly out of date! Fix should be on master of Makie & AbstractPlotting now!

PR #92, opened 2018-06-17
PR to keep track of the refactoring work

louisponet:
This PR is the result of the work that was done before on the old `GLAbstraction.jl` package refactoring, and trying to implement those ideas into the  `Makie` glbackend. Hopefully this will result in a more well-defined boundary between `frontend` and `backend`, and allow for a more flexible use of the `GLAbstraction` module. 

This is very much a WIP and comments/requests/objections/discussions are more then welcome! 

The current goal is to migrate over everything from the other package, and integrate it. Afterwards a lot of cleanup will probably be necessary, on top of other (hopefully 0) unforeseen issues.


src/glbackend/GLAbstraction/vertexarray.jl
```diff
@@ -40,7 +40,10 @@ function attach2vao(buffers, attrib_location, kind)
 end
 
 @enum VaoKind simple elements elements_instanced
-
+#TODO speedup: Maybe it would be better for performance making our vertex arrays
+#              primitives, i.e. making the buffers an NTuple instead of a vector.
+#              I think that should be possible since nobody really wants to change
+#              the vao after it's made anyway?
```
SimonDanisch: sounds good! shouldn't change performance much though, since we just access the id!

src/glbackend/GLAbstraction/renderobject.jl
```diff
@@ -54,11 +54,31 @@ function RenderObject(data::Dict{Symbol, Any}, bbs=Signal(AABB{Float32}(Vec3f0(0
...
     end
+    #TODO renderobjectioncleanup: ugly crap here!
+
     bufferdict  = filter((key, value) -> isa(value, Buffer), data)
     #TODO shadercleanup. This needs to not be inside the robj, and also not done here very hacky!!!!!!
     merge!(data, passthrough) # in the end, we insert back the non opengl data, to keep things simple lelkek
+
+    facelen = 1
+    prim = haskey(data,:gl_primitive) ? data[:gl_primitive] : data[:primitive]
+    if prim == GL_POINTS
```
SimonDanisch: this should probably be a function?

PR #93, opened 2018-06-18
revamped plotting + abstractplotting

SimonDanisch:


maleadt:
You can remove `DISPLAY=:0`, is hard-coded in the OpenGL image now.

PR #149, opened 2018-09-26
Replaced `linspace` (deprecated) with `LinRange`

Tokazama:


mkborregaard:
I thought the recommended update of `linspace(0f0, 1f0, size(rotation_v, i))` was `range(0f0, stop = 1f0, length = size(rotation_v, i))`?

Tokazama:
After looking back through the deprecation Warnings and doing a little investigation it does specifically suggest `range` but it requires keyword arguments instead of just positional ones.

```
julia> x = range(0, step=10, length=10)
0:10:90

julia> y = LinRange(0, 10, 10)
10-element LinRange{Float64}:
 0.0,1.11111,2.22222,3.33333,4.44444,5.55556,6.66667,7.77778,8.88889,10.0

julia> typex = typeof(x)
StepRange{Int64,Int64}

julia> supertype(typex)
OrdinalRange{Int64,Int64}

jjulia> typey = typeof(y)
LinRange{Float64}

julia> supertype(typey)
AbstractRange{Float64}
```
Not sure if the subtypes will make a difference in the end though.

SimonDanisch:
Thank you! Sorry for the slow reply - on first sight I wasn't sure if LinRange is the best to use, so I wanted to give it a spin first...and then I forgot of course!

mschauer:
`range(x, stop=y, length=n)` has the advantage that it will be easy to translate into the (hopefully) future `range(x, y, length=n)`

PR #192, opened 2018-10-25
fix README.md confusion?

pgbarletta:
Let me know if I misunderstood the text, but it seemed like the "safe" option is the one that we should be careful about and could potentially break Julia.


SimonDanisch:
You're right, thanks!

PR #210, opened 2018-11-03
Add explanation to use images in scatter

MaximeRivest:
As a beginner following the example: image scatter. I would have like to know how to load any image, not Makie.logo(). Thus, I have added comments to explain that in the code. I think Makie.logo() is not representative of most image scatter that a user would want to create

### DIFF
--- a/examples/examples3d.jl
+++ b/examples/examples3d.jl
@@ -367,15 +367,17 @@
     end
     @cell "image scatter" [image, scatter] begin
         using LinearAlgebra
+        #using FileIO #Uncomment to use load()
         scatter(
             1:10, 1:10, rand(10, 10) .* 10,
             rotations = normalize.(rand(Quaternionf0, 10*10)),
             markersize = 1,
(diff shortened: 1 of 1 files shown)


examples/examples3d.jl
```diff
@@ -367,15 +367,17 @@
...
+        #using FileIO #Uncomment to use load()
         scatter(
             1:10, 1:10, rand(10, 10) .* 10,
             rotations = normalize.(rand(Quaternionf0, 10*10)),
             markersize = 1,
             # can also be an array of images for each point
             # need to be the same size for best performance, though
-            marker = Makie.logo()
-        )
+            marker = Makie.logo() # Replace Makie.logo() with load("MY_IMAGE.png") for anyother images. To use load you will need FileIO
```
mschauer:
```suggestion
            marker = Makie.logo() # Replace Makie.logo() with load("MY_IMAGE.png") for other images. To use load you will need FileIO.
```

PR #211, opened 2018-11-04
fix time conflict with Libc.time

MaximeRivest:
When using this example, assigning a value to `time` raised a `Libc.time` error.


SimonDanisch:
thank you!

PR #217, opened 2018-11-07
Update examples3D.jl with a graphene bandstructure

pablosanjose:
Produces a plot of the bandstructure of graphene with a random periodic potential 2x2 superlattice. It requires https://github.com/pablosanjose/QBox and https://github.com/pablosanjose/QBoxPlots, master branch.

Please let me know if it is preferable to spell out the `plot` commands used in QBoxPlots to produce the output. Also, can we do "using MyPackage" with an unregistered `MyPackage`, as is the case here?

### DIFF
--- a/examples/examples3d.jl
+++ b/examples/examples3d.jl
@@ -596,11 +596,11 @@
         Makie.GLMakie.destroy!(screen)
         scene
     end
-    @cell "Graphene bandstructure with QBox.jl" [surface, mesh] begin
-        using QBox, QBoxPlots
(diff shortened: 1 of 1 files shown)


pablosanjose:
I renamed my package from QBox to Elsa, and updated the PR accordingly

asinghvi17:
This should probably be moved to JuliaPlots/MakieGallery as that's where the examples and documentation are stored now.

pablosanjose:
Agreed

PR #221, opened 2018-11-08
Fixed gpu_setindex! writing to the wrong offset.

CarpeNecopinum:
In the old variant, e.g. writing 4-byte elements to index 1 (the beginning) of a GLBuffer, would put them at byte offset 4*1-1 = 3, rather than the correct 4*(1-1) = 0.

### DIFF
--- a/src/glbackend/GLAbstraction/GLBuffer.jl
+++ b/src/glbackend/GLAbstraction/GLBuffer.jl
@@ -110,7 +110,7 @@ end
 function gpu_setindex!(b::GLBuffer{T}, value::Vector{T}, offset::Integer) where T
     multiplicator = sizeof(T)
     bind(b)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Oh true :-O Weird, that this never caused an issue... How did you run into it?

CarpeNecopinum:
I had some geometry that I needed to re-upload every frame (but with the same topology, so I could re-use all the involved buffers), I found the `gpu_setindex!` looking at the code for GLBuffer, so I used that. 

Then I wondered why my render outputs didn't actually change. Then I noticed the wrong indices, which caused the `glBufferSubData` to error and do nothing.

I guess this didn't cause an issue with Makie as a whole is that (I assume) Makie creates a new buffer for updated geometry, rather than filling an old one with new data (?).

PR #226, opened 2018-11-11
Add GLScreen abstract type

barche:
This adds an abstract type to allow simpler sub typing of OpenGL-compatible screens, so a lot of the Makie OpenGL rendering code can be reused. With this type and by making the texture cache key an `Any`, Makie is simple to use from QML:

GLScreen implementation for QML:
https://github.com/barche/QML.jl/blob/master/src/makie_support.jl

Example:
https://github.com/barche/QML.jl/blob/master/example/makie.jl

Of course this example is still quite basic, I still need to add callbacks and selection support.

### DIFF
--- a/src/glbackend/drawing_primitives.jl
+++ b/src/glbackend/drawing_primitives.jl
@@ -100,7 +100,7 @@ function handle_intensities!(attributes)
     end
 end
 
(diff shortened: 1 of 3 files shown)


SimonDanisch:
That's awesome :) Glad this change turned out to be simple enough!

PR #235, opened 2018-11-18
volume

louisponet:
With the change of mutual exclusive flags on AbstractPlotting one can again use colors to define the volume color.

### DIFF
--- a/src/glbackend/drawing_primitives.jl
+++ b/src/glbackend/drawing_primitives.jl
@@ -239,7 +239,6 @@ function vec2color(colors, cmap, crange)
 end
 
 function get_image(plot)
-    println("getting_image")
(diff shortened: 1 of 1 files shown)


src/glbackend/drawing_primitives.jl
```diff
@@ -239,6 +239,7 @@ function vec2color(colors, cmap, crange)
 end
 
 function get_image(plot)
+    println("getting_image")
```
Evizero:
```suggestion
```

SimonDanisch:
merged in https://github.com/JuliaPlots/GLMakie.jl/commit/c205c629be2c55067b318cfcc94ec71bc53d6be3

PR #237, opened 2018-11-19
Sd factorout examples

SimonDanisch:


SimonDanisch:
One example fails - I guess that's fine for now ...

PR #251, opened 2018-12-03
update to Pkg3

SimonDanisch:


musm:
Why add the manifest?

SimonDanisch:
because I need it for static compilation

SimonDanisch:
ah, so the man manifest directly, I rather need for testing with the unregistered test dependencies.

PR #280, opened 2019-02-02
Update compilation instructions

asinghvi17:
Since `compile_package` is set to be deprecated in PackageCompiler, and after successfully static-compiling Makie, I feel that these instructions should be updated.  I also ran into an issue that it would have been nice to have foreknowledge of, hence the last line.


asinghvi17:
Also, is there a reason that Makie's `runtests.jl` needs an Internet connection?  Seems to me that we'd be better off bundling the images with the package itself, no?  That would probably significantly shorten the compile time with low internet speeds.

SimonDanisch:
Thank you!

PR #297, opened 2019-02-24
CI fix

asinghvi17:
Currently CI does not work due to some problems with the project directory itself.  I'm uploading this to try two different approaches:

1. use `Pkg.add` and manually parse `test/REQUIRE`
2. use `Pkg.resolve` with `--project=test/` and then evaluate normally

I'll submit this PR now for the CI to commence using the first method, and then try the second to see if it works.  In my opinion the second is far less 'hacky' and has less requirements, but I'm just curious to see if the first one works at all.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,5 +15,5 @@ job:
 
     script:
       - julia -e 'using InteractiveUtils; versioninfo()' # print version info
-      - julia -e 'using Pkg; open("test/REQUIRE", "r") |> readlines |> x -> filter(y -> y[1] != '#', x) .|> Pkg.add' # add packages to prevent annoying bugs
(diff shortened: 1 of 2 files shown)


asinghvi17:
I'm not sure that this is being CI'ed...am I doing something wrong here?

asinghvi17:
It seems the second method doesn't work.  I will find out how to embed testing dependencies in `Project.toml`.

asinghvi17:
Yeah, as it stands the CI fix should be ready to deploy.  Unfortunately, documentation is lacking for any of the more elegant methods I could think of, so it will take a lot more time.  Probably best that we get this fix out IMO, and fix any bugs this may cause on its own as I can't test this except from the command line on my computer.

SimonDanisch:
>julia -e 'using Pkg; open("test/REQUIRE", "r") |> readlines |> x -> filter(y -> y != "" && y[1] != '#', x) .|> Pkg.add'

I don't think this should be the solution to the problem! 
Sadly, the testing infrastructure is in flux right now, since quite a few things changed (e.g. I noticed that I wasn't really testing against the reference images) and a couple of bugs have sneaked in that need fixing before I can record a new set of test images...

PR #339, opened 2019-04-12
Add Makie dependencies to precompile instructions

dlfivefifty:
I needed to add Makie dependencies to get precompilation to work. This updates the README.

PR #364, opened 2019-06-08
Update PackageCompiler steps to include dependencies

dawbarton:
Generating a system image seems to require additional dependencies beyond just Makie - this adds the extra required packages to the README.md. I'm not entirely sure why this is the case but precompilation kept crashing for me until I added these packages explicitly to my default environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,8 @@ To do that, you need to check out the additional packages for precompilation.
 Then you can build a system image like this:
 
 ```julia
-# add PackageCompiler
-Pkg.add("PackageCompiler")
+# add PackageCompiler and other dependencies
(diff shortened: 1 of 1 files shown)


asinghvi17:
Did this happen on the latest Makie version as well?

dawbarton:
Just noticed #339 - I get the same error message on the latest version. My guess is that the newly created environments that are used for precompilation don't have a correct manifest (the project has the right dependencies but instantiate needs to be run).

dawbarton:
(I installed Makie this morning so I'm assuming I've got the latest version v0.9.4)

asinghvi17:
Hmm, could be the old PackageCompiler issue with test specific dependencies.  Thanks, will merge once I get to my laptop :)

PR #405, opened 2019-11-14
Update README.md

logankilpatrick:
Fixed the plot plan enabling/disabling description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,9 @@ Examples, and test infrastructure, are hosted at [`MakieGallery.jl`](https://git
 
 ## Using Juno with Makie
 
-The default OpenGL backend for Makie is not interactive in the Juno plotpane - it just shows a PNG instead.  To get full interactivity, you can disable the plotpane in Atom's settings by going to `Juno` - `Settings` - `UI Options` - Then, make sure `Enable Plot Plane` is __not__ checked. 
+The default OpenGL backend for Makie is not interactive in the Juno plotpane - it just shows a PNG instead.  To get full interactivity, you can run `AbstractPlotting.inline!(false).
+
+If that fails, you can disable the plotpane in Atom's settings by going to `Juno` - `Settings` - `UI Options` - Then, make sure `Enable Plot Plane` is __not__ checked. 
 
 ## Examples from the documentation: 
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
Why not both?  It's an internal API, true, but it's way easier, and allows you to get stuff in the plot pane if you need to.

README.md
```diff
@@ -52,7 +52,7 @@ Examples, and test infrastructure, are hosted at [`MakieGallery.jl`](https://git
 
 ## Using Juno with Makie
 
-The default OpenGL backend for Makie is not interactive in the Juno plotpane - it just shows a PNG instead.  To get full interactivity, you can disable the plotpane in Atom's settings, or execute `Atom.PlotPaneEnabled[] = false`.
+The default OpenGL backend for Makie is not interactive in the Juno plotpane - it just shows a PNG instead.  To get full interactivity, you can disable the plotpane in Atom's settings by going to `Juno` - `Settings` - `UI Options` - Then, make sure `Enable Plot Plane` is __not__ checked.
```
aminya:
```suggestion
, or execute `Atom.PlotPaneEnabled[] = false`.
```

logankilpatrick:
Sebastian stated it should not be used which is why I suggested to remove it

SimonDanisch:
By the way, I decoupled this from the plot pane being disabled or not with `AbstractPlotting.inline(false)`  - this will always open a window now... `AbstractPlotting.inline(true)` will only show something, when the plotpane is enabled, though.

README.md
```diff
@@ -52,7 +52,7 @@ Examples, and test infrastructure, are hosted at [`MakieGallery.jl`](https://git
...
```
asinghvi17:
```suggestion
The default OpenGL backend for Makie is not interactive in the Juno plotpane - it just shows a PNG instead.  To get full interactivity, you can run `AbstractPlotting.inline!(false).

If that fails, you can disable the plotpane in Atom's settings by going to `Juno` - `Settings` - `UI Options` - Then, make sure `Enable Plot Plane` is __not__ checked. 
```

asinghvi17: @logankilpatrick this should be gtm

logankilpatrick:
Done!

PR #450, opened 2020-03-17
New version with new compat

asinghvi17:


SimonDanisch:
Shouldn't we tag the versions first? And we should also not use `>=` anymore ;)

asinghvi17:
Good point, I'm just putting this there in advance because I had a couple minutes.  Won't merge it until we tag.

PR #455, opened 2020-03-31
do the release

SimonDanisch:


test/runtests.jl
```diff
@@ -1,15 +1,14 @@
 using MakieGallery, AbstractPlotting, GLMakie, Test
 using MakieGallery: @block, @cell
-empty!(MakieGallery.plotting_backends)
```
asinghvi17: Any particular reason to remove these?

Project.toml
```diff
@@ -7,58 +7,25 @@ AbstractPlotting = "537997a7-5e4e-5d89-9595-2241ea00577e"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
```
asinghvi17: The version needs to be bumped, unfortunately I can't suggest it :)

https://github.com/JuliaPlots/Makie.jl/pull/455/files#diff-910a7b3ca0075d545dcec45cb7335ca9R3

test/runtests.jl
```diff
@@ -1,15 +1,14 @@
 using MakieGallery, AbstractPlotting, GLMakie, Test
```
asinghvi17:
```suggestion
using Test
using Makie, AbstractPlotting, GLMakie, MakieGallery

File: test/runtests.jl [hunk 4]
```diff
@@ -1,15 +1,14 @@
...
 using MakieGallery: @block, @cell
```
asinghvi17: I don't think this is needed...could be wrong, though.

test/runtests.jl
SimonDanisch: I kinda deprecated this ;)

asinghvi17: I needed it for external package support - MakieGallery (surprisingly) works with Plots.jl with minimal tweaking, so it would be nice if we could keep it that way...

SimonDanisch: Oh...how are you actually using it?

asinghvi17:
```julia
empty!(MakieGallery.plotting_backends)
push!(MakieGallery.plotting_backends, "Plots")

function MakieGallery.save_media(example, plot::Plots.Plot, path::String)
    plotpath = joinpath(path, "image.svg")
    Plots.savefig(plot, plotpath)
    return [plotpath]
end
```
and it just works.

SimonDanisch: So...are you running the normal makie examples like that?Or your own?

asinghvi17: I've defined my own set of examples using Plots (and some other packages), which I then run using this.  Regression testing isn't really a concern for me, so SVG output is fine.

SimonDanisch: So can you just say `using Plots` ?

asinghvi17: In the toplevel code, yes, but then it gets mixed in with `using Makie` as well.

Project.toml
SimonDanisch: Which version?

PR #954, opened 2021-05-17
Merge AbstractPlotting into Makie

SimonDanisch:


SimonDanisch:
I think we need to merge + tag this to update the backends correctly ...

PR #957, opened 2021-05-20
fix theming in stats recipes

piever:
The theme is now sufficiently rich that it is possible to give reasonable defaults to all attributes in `boxplot` and `violin` without hardcoding values. The "hardcoded" white color to draw the median is no longer necessary, as they are cycling colors, so the default black works well on the "patchcolor" palette, for example:

violin

I'm not 100% sure whether the outliers in `boxplot` should inherit from `default_theme(scene, Scatter)[:strokewidth]` (and friends) or `theme(scene, :markerstrokewidth)` (and friends). I guess the latter is a bit cleaner.

Related question: what about the default `outliercolor`? The old behavior of matching the overall color makes less sense now that the default color has some transparency. I imagine it should inherit from `theme(scene, :markercolor)`, but should we also cycle it by default, or could we just keep default black outliers for boxplots?

Side note: kudos on the `theme`  redesign, it has exactly all attributes one might need!

### DIFF
--- a/src/stats/boxplot.jl
+++ b/src/stats/boxplot.jl
@@ -28,7 +28,6 @@ The boxplot has 3 components:
 - `show_outliers`: show outliers as points
 """
 @recipe(BoxPlot, x, y) do scene
-    scattertheme = default_theme(scene, Scatter)
     Theme(
         color = theme(scene, :patchcolor),
         colormap = theme(scene, :colormap),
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
Yeah I wondered the same about inheriting from a global `markerstrokewidth` or `Scatter.strokewidth`. I also settled on liking the global a bit better, but I guess there's pros and cons for both.

SimonDanisch (approved):

piever:
Ok, so I'll leave as is and inherit the attibutes from `markerstrokewidth` and co.

About the defaults for the `outliercolor`, on master by default it inherits from the main `color` of the boxplot, which has transparency, and markers with transparency are IMO a bad default. We could:

1. inherit from the main `color` but remove the transparency
2. inherit from `theme(scene, :markercolor)` and cycle (so by default they do end up matching the patchcolor but without transparency)
3. inherit from `theme(scene, :markercolor)` but do not cycle (so outliers are black by default)

Do you have preferences? I find 1 a bit odd and 2 or 3 equally reasonable.

SimonDanisch:
I'm slightly leaning towards 2, to have the colors matched... Although having outliers in a different color also seems like a good idea :D But I think 2 may be more visually pleasing and easier to group the outliers to the right plot, so I think 2 wins!

jkrumbiegel:
I don't think 1 is so bad, if you cycle two different cycles together it stops making sense if the cycles are not color-matched.

piever:
> I don't think 1 is so bad, if you cycle two different cycles together it stops making sense if the cycles are not color-matched.

I might actually go for 1 in the end, because I do agree that it's nicer to have the colors "synced" by default, and you are right that cycling both seems a bit brittle. I confess that option 1 also it has the nice side effect that one would get this syncing for free from AlgebraOfGraphics.

PR #959, opened 2021-05-20
Camera 3d fixes

ffreyer:
* fix `cam3d_cad` not zooming
* fix cleanup of camera nodes

The latter fixes issues where 3d scenes become non-interactive when another is added, e.g.

```julia
fig = Figure()
mesh(fig[1, 1], Rect3D(Point3f0(0), Point3f0(1)))
mesh(fig[1, 2], Rect3D(Point3f0(0), Point3f0(1)))
```

Note that I also removed `CameraLift` - I don't think it's needed anymore. I'm also not sure if this works with Observables 0.3, but WGLMakie and GLMakie restrict to 0.4 anyway...

### DIFF
--- a/src/camera/camera.jl
+++ b/src/camera/camera.jl
@@ -11,14 +11,10 @@ function Base.:(==)(a::Camera, b::Camera)
 end
 
 function disconnect!(c::Camera)
-    for node in c.steering_nodes
-        # remove all camera lifts
-        for f in listeners(node)
-            if f isa CameraLift
-                off(node, f)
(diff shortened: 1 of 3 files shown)


ffreyer:
I think I messed some stuff up by removing CameraLift...

ffreyer:
Yea we need it as a workaround for `onany` atm. I didn't implement that for PriorityObservables because it would replace the method from Observables.jl. (Which wouldn't be breaking, but it would warn)

PR #968, opened 2021-05-22
Fix #967

### LINKED ISSUE #967: DataInspector tooltip position does not switch to log scale
The tooltip value corresponds to the mouse position (also on log10 scale) but tooltip position moves on linear scale.
```
using GLMakie
x = range(0, 10, length=100)
y = sin.(x) .+ 1.1
f = Figure()
ax = Axis(f[1,1],yscale=log10)
lines!(ax,y, markersize=5)
Makie.inline!(false)
display(f)
DataInspector(f);
```

ffreyer:
Closes #967

### DIFF
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -203,11 +203,11 @@ end
 ## Shifted projection
 ########################################
 
-function shift_project(scene, pos)
+function shift_project(scene, plot, pos)
     project(
         camera(scene).projectionview[],
         Vec2f0(widths(pixelarea(scene)[])),
-        pos
+        apply_transform(transform_func_obs(plot)[], pos)
     ) .+ Vec2f0(origin(pixelarea(scene)[]))
 end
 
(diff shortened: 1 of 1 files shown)


ffreyer:
@SimonDanisch You may have removed a bit too much from the shaders?

```
ERROR: LoadError: program 79 not linked. Error in: 
/home/runner/.julia/packages/GLMakie/EojHm/assets/shader/fragment_output.frag or /home/runner/.julia/packages/GLMakie/EojHm/assets/shader/image.vert or /home/runner/.julia/packages/GLMakie/EojHm/assets/shader/texture.frag
error: fragment shader input `o_normal' has no matching output in the previous stage
error: fragment shader input `o_view_pos' has no matching output in the previous stage
```

SimonDanisch:
Oh I didn't realize, that this was used in fragment_output.frag... weird that it works at all, this should have broken all heatmap/image shaders?! Ah, I guess only when its actually used for ssao?

SimonDanisch:
I guess this should fix it: https://github.com/JuliaPlots/GLMakie.jl/pull/192

PR #969, opened 2021-05-22
Persistent tooltips

### LINKED ISSUE #962: Persistent Data Tips for Data Exploration and Comparison
Since `DataInspector` was merged we can explore data on plots at given in:

https://discourse.julialang.org/t/ann-makie-v-0-13/61522/7

I was wondering if we can add a persistent data tip like in MATLAB:

image

In MATLAB it is done by `Alt + Click` (Or choosing the `Data Tips` tool).

ffreyer:
First prototype. Creates a static tooltip with `Alt + Left click` and removes all with `escape`.

TODO:
- [x] get a basic prototype going
- [ ] figure out better depth order
- [ ] move tooltips when the view is adjusted
- [ ] figure out which attributes should be connected to DataInspector

Screenshot from 2021-05-22 17-27-24

Closes #962

### DIFF
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -359,6 +359,10 @@ end
 
 
 
+struct PersistentTooltip
+    plots::Vector{AbstractPlot}
+end
+
 # TODO destructor?
 mutable struct DataInspector
     root::Scene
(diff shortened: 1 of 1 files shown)


ffreyer:
So just as a minor update - getting this to react to zooming, translation etc takes some redesigning I think. Currently updates happen on mouse movement, and that is connected to selecting a new plot and value for the tooltip. 
For a persitent tooltip you'd want want the indicator and tooltip to move when the plot is moved, but you don't want it to change what is being displayed. So you want part of what the current implementation does. 

If this is something people desperately want/need I can make a quick implementation that duplicates a lot of code and maybe we can roll with that for a bit. Otherwise it's probably gonna take a while.

PR #974, opened 2021-05-23
use Callback type for easier dispatch

SimonDanisch:
And make it possible to use map/lift/onany with priority observables...
Open question: now, nobody gets a warning from the changed behaviour... 
Although I think the most annoying bugs would come from an `on` which already returns a bool, that wasn't actually intended to consume an event...
So maybe we really want to have people return something like `Consume()` to make this unambiguous.
Since only consuming an event changes behavior, it seems fair to not consume in all other cases.


ffreyer:
> And make it possible to use map/lift/onany with priority observables...

Just to mention it here too, it works on master as well, it's just spits out an angry warning. You're however stuck with the default priority, which is also true here.

> Open question: now, nobody gets a warning from the changed behaviour...

I guess it's not actually that important? You get pretty much the same behavior as before if you don't consume. 

> Although I think the most annoying bugs would come from an `on` which already returns a bool, that wasn't actually intended to consume an event...
> So maybe we really want to have people return something like `Consume()` to make this unambiguous.
> Since only consuming an event changes behavior, it seems fair to not consume in all other cases.

Yea that makes sense to me.

src/interaction/PriorityObservable.jl
```diff
@@ -77,36 +90,17 @@ end
...
-            "specify whether the update is consumed (true) or should " *
-            "propagate (false) to other observer functions. The given " *
-            "function has been wrapped to always return false."
-        ) 
-        Base.show_backtrace(stderr, backtrace())
-        println()
-        return sanitized_func
-    else
-        return f
-    end
```
ffreyer, line 95: We could add this again as a compat note. Note sure if it's worth it though

ffreyer:
We still need to update docs here

ffreyer:
I fixed #1007 as well.

If there are no complaints/test failures we can merge this I think.

ffreyer:
I forgot some docs changes actually

ffreyer:
Ok, nothing that's actually important. I just realized I didn't really check the section about keyboard and mouse interactivity. I changed `return false` to `return Consume(false)` there now, to be explicit and to push people towards writing type stable code.

SimonDanisch:
Thanks a lot for bringing this to the finish line! So I guess this is ready to merge? :)

ffreyer:
I'm not sure if I checked if everything is working, I can do that later and merge afterwards if you want.

ffreyer (approved):
My test plot with a bunch of interactive elements seems to be fine. I think we can merge this

SimonDanisch:
Thank you!

PR #979, opened 2021-05-24
Add ability to use vector path markers

jkrumbiegel:
This PR adds the ability to use vector path markers for Scatter. So far, we could only use glyphs as markers, however this is often not flexible enough. Even if a specific shape is represented in a font (we use DejaVu by default), that doesn't mean that the different shapes we need (square, circle, star, ngon, etc.) look well-matched to each other. The opposite is the case, the apparent sizes of our default markers were quite different.

For vector graphics, it's really useful to just specify some path operations that define a marker (as opposed to using bitmaps). Therefore, it was easy to implement this for CairoMakie, just store a vector of path operations and execute these when drawing the marker.

It was more difficult to implement support for GLMakie and WGLMakie, as these backends don't have support for drawing paths. Drawing paths is actually not a straight-forward thing in OpenGL, so here I've chosen to implement the same workaround as for glyphs. We render the paths to a small bitmap, convert it to a signed-distance-field and store it in the glyph atlas. For this, we use FreeType's outline drawing engine. It doesn't support all basic paths (e.g. EllipticalArcs) but we can replace them with cubic beziers on the fly.

The caveats to this technique are that the texture atlas could be filled if one was to generate tons of different marker shapes, because one atlas slot is needed for each marker. The other is that signed distance fields work only where there are filled areas, and not all paths are filled areas (the simplest one, a single line, is not, for example). This means that while CairoMakie will have no problem rendering unfilled markers (stroked paths only), GLMakie and WGLMakie are limited to filled markers. If it is absolutely essential to render something looking like an unfilled marker in GLMakie, a workaround could be to apply a stroke to the path and draw the resulting path. Depending on the stroke width used, and the resolution of the glyph atlas, this might not look great.

With the ability to specify our own markers, this PR also replaces the full set of markers accessible by symbols such as `:circle, :rect, :star4` etc. The goal was to make the default markers look well-matched amongst each other. The base size was chosen so that the markers do not look out-of-place compared with normal glyphs like `'x'`, or `'Y'`. Therefore the setting `markersize` is not a concrete metric of the side length of a marker, it just means "looks ok compared to glyphs at font size X".

Here is an overview of available markers before and after. For example, triangles are correctly centered now, stars and crosses are better matched in size to squares and circles.

Before
grafik
After
grafik


SimonDanisch:
I'm not sure if I 100% understand this PR, but looking at the corresponding CairoMakie PR, this looks like it's using a 1:1 Cairo representation of the bezierpath.
Makie shouldn't rely on any abstractions that only make sense for one backend and it would be better, if this can be translate to a generic bezierpath recipe.
Although, if the representation isn't too awkward for other backends and is still drawable there, we could adopt that representation in Makie too, I suppose ;) 
I'm not too worried about the details, but I want to make sure, that this is planned with the idea in mind, that this will be supported for all backends in the long run.

jkrumbiegel:
The bezier path operations are not really specific to cairo, they are the same operations that pdf and svg can handle. Bezier curves are a very useful thing to support because line approximations of such curves take much more storage space, especially if we're talking about many scatter markers. I understand that you want a consistent API and feature set across backends, but it would be sad not to be able to use really useful vector features. I guess it could also be ok to make bezier paths a separate package for cairomakie. But it would be a simple way to get perfectly controlled scatter markers of equal area.

SimonDanisch:
>I understand that you want a consistent API and feature set across backends, but it would be sad not to be able to use really useful vector features

That's not really in conflict, since we can draw beziers and vector graphics in all backends...
I just don't want to have very implementation specific things (like using Cairo draw commands) bleed into the generic, backend agnostic Makie API, if it's easy to avoid ;)

SimonDanisch:
I think there are two cases, where it's acceptable to use the "Cairo Language" for Beziers:
1) It's a very good representation for Beziers, and anything we would come up for a generic implementation would look similar
2) It's a lot of work to translate the svg based vector graphics into something more generic, and we currently don't have that time. In that case, I'd leave a todo, for whenever we have time to port/convert things to a representation, that can also be drawn by (W)GLMakie.

SimonDanisch:
Anyways, the more I look at this, the more I feel like the `BezierPath` struct is a fair way to represent a complex shape that we get from an SVG.
So maybe that should just be the most highlevel type to represent vector graphics that come from SVGs/PDFs.
To convert it to more low level types like actual bezier paths and lines can then be tackled when we implement it for the other backends.

hexaeder:
I'd like to see bezierpaths becoming a thing! Would it be possible to have backend depending argument conversion? I.e. CairoMakie could use the higher level Path struct while GL will just deconstruct them into nodes?
I've been [playing around](https://github.com/hexaeder/SplineMakie.jl) on how this could be used for spline drawing (context [self edges in GraphMakie](https://github.com/JuliaPlots/GraphMakie.jl/issues/23)). I've slightly changed the structs to allow for 3d pathes, though.

goretkin:
For the record, matplotlib does basically the same (use the AGG backend's path interface as its own generic interface). It also has a lot of back ends and converts between them. I include what I think is a fairly comprehensive overview of all the different path languages out there. I think Cairo is more or less "path complete", and if not it would be easy to add a few more path commands. Note that it does not include quadratric bezier curves, [only cubic](https://lists.cairographics.org/archives/cairo/2010-April/019691.html). Skia is perhaps the most complete of these path languages.

[matplotlib path interface](https://matplotlib.org/stable/api/path_api.html):
```
    The code types are:

    - ``STOP``   :  1 vertex (ignored)
        A marker for the end of the entire path (currently not required and
        ignored)

    - ``MOVETO`` :  1 vertex
        Pick up the pen and move to the given vertex.

    - ``LINETO`` :  1 vertex
        Draw a line from the current position to the given vertex.

    - ``CURVE3`` :  1 control point, 1 endpoint
        Draw a quadratic Bezier curve from the current position, with the given
        control point, to the given end point.

    - ``CURVE4`` :  2 control points, 1 endpoint
        Draw a cubic Bezier curve from the current position, with the given
        control points, to the given end point.

    - ``CLOSEPOLY`` : 1 vertex (ignored)
        Draw a line segment to the start point of the current polyline.
```

which is a subset directly from the [agg's path interface](https://github.com/matplotlib/matplotlib/blob/c06e8709dde6504d396349c0c80ef019c88c3927/extern/agg24-svn/include/agg_basics.h#L368-L380)

```
        path_cmd_stop     = 0,        //----path_cmd_stop    
        path_cmd_move_to  = 1,        //----path_cmd_move_to 
        path_cmd_line_to  = 2,        //----path_cmd_line_to 
        path_cmd_curve3   = 3,        //----path_cmd_curve3  
        path_cmd_curve4   = 4,        //----path_cmd_curve4  
        path_cmd_curveN   = 5,        //----path_cmd_curveN
        path_cmd_catrom   = 6,        //----path_cmd_catrom
        path_cmd_ubspline = 7,        //----path_cmd_ubspline
        path_cmd_end_poly = 0x0F,     //----path_cmd_end_poly
        path_cmd_mask     = 0x0F      //----path_cmd_mask    
```


[SVG path interface](https://www.w3.org/TR/SVG/paths.html#PathDataBNF)
```
drawto_command::=
    moveto
    | closepath
    | lineto
    | horizontal_lineto
    | vertical_lineto
    | curveto
    | smooth_curveto
    | quadratic_bezier_curveto
    | smooth_quadratic_bezier_curveto
    | elliptical_arc

[...]
```


[HTML Canvas path interface](https://html.spec.whatwg.org/multipage/canvas.html#building-paths)
```
path.moveTo(x, y)
path.closePath()
path.lineTo(x, y)
path.quadraticCurveTo(cpx, cpy, x, y)
path.bezierCurveTo(cp1x, cp1y, cp2x, cp2y, x, y)
path.arcTo(x1, y1, x2, y2, radius)
path.arc(x, y, radius, startAngle, endAngle [, counterclockwise ])
path.ellipse(x, y, radiusX, radiusY, rotation, startAngle, endAngle [, counterclockwise])
path.rect(x, y, w, h)
```

[cairo path interface](https://www.cairographics.org/manual/cairo-Paths.html)
```
void | cairo_close_path ()
void | cairo_arc ()
void | cairo_arc_negative ()
void | cairo_curve_to ()
void | cairo_line_to ()
void | cairo_move_to ()
void | cairo_rectangle ()
void | cairo_glyph_path ()
void | cairo_text_path ()
void | cairo_rel_curve_to ()
void | cairo_rel_line_to ()
void | cairo_rel_move_to ()
void | cairo_path_extents ()
```

[skia path interface](https://api.skia.org/classSkPath.html)
```
kMove_Verb |  
kLine_Verb |  
kQuad_Verb |  
kConic_Verb |  
kCubic_Verb |  
kClose_Verb |  
kDone_Verb
```
Some information about [Skia's conic curves](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/graphics/skiasharp/curves/beziers#the-conic-b%C3%A9zier-curve)

> The conic Bézier curve — also known as the rational quadratic Bézier curve — is a relatively recent addition to the family of Bézier curves. Like the quadratic Bézier curve, the rational quadratic Bézier curve involves a start point, an end point, and one control point. But the rational quadratic Bézier curve also requires a weight value. It's called a rational quadratic because the parametric formulas involve ratios.
> 
> The parametric equations for X and Y are ratios that share the same denominator. Here is the equation for the denominator for t ranging from 0 to 1 and a weight value of w:
> 
> d(t) = (1 – t)² + 2wt(1 – t) + t²
> 
> In theory, a rational quadratic can involve three separate weight values, one for each of the three terms, but these can be simplified to just one weight value on the middle term.
> 
> The parametric equations for the X and Y coordinates are similar to the parametric equations for the quadratic Bézier except that the middle term also includes the weight value, and the expression is divided by the denominator:
> 
> x(t) = ((1 – t)²x₀ + 2wt(1 – t)x₁ + t²x₂)) ÷ d(t)
> 
> y(t) = ((1 – t)²y₀ + 2wt(1 – t)y₁ + t²y₂)) ÷ d(t)
> 
> Rational quadratic Bézier curves are also called conics because they can exactly represent segments of any conic section — hyperbolas, parabolas, ellipses, and circles.
> 

I think postscript is the origin of all of these related path languages.
[postscript path interface, Section 4.3](https://www.adobe.com/content/dam/acom/en/devnet/actionscript/articles/psrefman.pdf)
image

## conversion

- [HTML canvas paths to SVG paths](https://github.com/d3/d3-path/blob/6230b3c3952011ec917fcf5ba50c837051dc9b78/src/path.js)

- matplotlib converts AGG to SVG:
[part 1ish](https://github.com/matplotlib/matplotlib/blob/ded49512f2851dafb2f1d51b54172aa8ab480da4/lib/matplotlib/backends/backend_svg.py#L652-L654)
[part 3ish](https://github.com/matplotlib/matplotlib/blob/2acc6b0f0311dd5553f8db5ebf82702c2df2e11e/src/_path.h#L1137-L1201)

- The [postscript backend](https://github.com/matplotlib/matplotlib/blob/5656d11562a49fbfeb8a7fbe6d46a9ed57f4b7c9/lib/matplotlib/backends/backend_ps.py#L418-L420) is basically the same but with different codes.

- [convert SVG to AGG](https://github.com/ghaerr/agg-2.6/blob/b9c42f82229bd349f71b2e40e4b09f6edf403554/agg-src/examples/svg_viewer/agg_svg_path_renderer.cpp#L305-L399)




Additional stuff:
maybe older than postscript: [G-code](https://marlinfw.org/docs/gcode/G005.html)
```
G5 creates a cubic B-spline in the XY plane with the X and Y axes only. P and Q parameters are required. I and J are required for the first G5 command in a series. For subsequent G5 commands, either both I and J must be specified, or neither. If I and J are unspecified, the starting direction of the cubic will automatically match the ending direction of the previous cubic (as if I and J are the negation of the previous P and Q

G5 [E<pos>] [F<rate>] I<pos> J<pos> P<pos> Q<pos> [S<power>] X<pos> Y<pos>
```

for completeness:
[HP PCL 5 or HPGL/2 ?? path interface](https://www.devenezia.com/docs/HP/index.html?1902)
```
This command, BR or Bezier Relative , draws bezier
     curves using relative coordinates. This command uses
     the current pen position as the first control point,
     and specifies the other three control points as
     relative increments from the first point.  For more
     information see the PCL 5 Printer Language Technical
     Reference Manual.
[...]
     BR0,3048,4572,0,3556,2032,
     -508,1016,2540,508,2540,-5080;     Draws a Bezier curve
                                   using the current
                                   position (1016,5080) as
                                   the first control point.
                                   The specified control
                                   points for the first
                                   curve are (0,3048),
                                   (4572,0), and
                                   (3556,2032). The second
                                   curve uses the last
                                   control point of the
                                   previous curve as the
                                   first control point
                                   (3554,2032). The other
                                   three control points for
                                   the second curve are (-
                                   508,1016), (2540,508) and
                                   (2540,-5080).
```

goretkin:
While I have the same initial inclination, I am not sure that it is wise to use Julia types to distinguish between different path commands. On one hand it is nice to have a polyline just e.g. a `Vector{LineTo}`, and in this case it would be efficient. However in general it would be `Vector{PathCommand}`, and I wonder how the performance would compare to encoding the path command in the "value domain" vs the "type domain".

hexaeder:
Unfortunately I don't have the time right now to look at the recent changes in detail but some thoughts on this from a different usecase: For curvy edges in GraphMakie I implemented a [`BezierSegments`](https://github.com/JuliaPlots/GraphMakie.jl/blob/649f1aec493bc57fdd3ca66e57d6b5eaaf946af4/src/recipes.jl#L475) recipe which is like `LineSegments` but plots arbitrary bezier paths. I know your coming from the low-level direction but I think it would be nice to have the same types for all kinds of path usages.

- I've found to be the type instability of `Vector{PathCommand}` in the `BezierPath` struct a real pain. Is the Union splitting approach (PathCommand as Union instead of abstract type) sufficient there? For my use case I mitigated this by defining `BezierPath <: AbstractPath`. For some basic paths, i.e. just lines, I defined `Line <: BezierPath` to be able to  have type stable collections of Pathes.
- 3D. BezierPathes totally make sense in 3 dimensions therefore it would be really nice to have this parametric

Fyi the graph makie version of this code can be found [here](https://github.com/JuliaPlots/GraphMakie.jl/blob/master/src/beziercurves.jl). On the long term I'd like to move the `BezierPaths` recipe to Makie. Currently i discretize all of the pathes and draw them as `lines`, long term it would be really cool if Cairo would draw them as "real" pathes while GL defaults to discretization.

jkrumbiegel:
I'm open to suggestions in terms of performance, my goal so far was to bring vector paths in so you can set them as markers (that's a bit different from beziersegments, which are complete on their own, the path versions always start at the previous point). I have only tried to get things working but didn't focus on efficiency, yet. My next focus will be to prerender arbitrary paths for GLMakie sprite markers.

I think these path primitives for 2d vector graphics are something separate from the 3d bezier generalization, so I'm not sure how much code can be effectively shared there.

GLMakie/src/GLVisualize/visualize/particles.jl
```diff
@@ -323,6 +346,22 @@ function sprites(p, s, data)
...
+            pop!(data, :scale, Observable(nothing)),
+            pop!(data, :scale_x, Observable(nothing)),
+            pop!(data, :scale_y, Observable(nothing)),
+            pop!(data, :scale_z, Observable(nothing))
+        )
+        # TODO 
+        # marker_offset should be Vec2f(0) by default for BezierPaths
+        # once that is the case we should switch to this offset
+        # offset = get(data, :offset, Observable(Vec2f(0)))
+        offset = Observable(Vec2f(0))
```
ffreyer, line 361: It would be good if BezierPath scatters would get some different defaults. `marker_offset` defaults to centering https://github.com/JuliaPlots/Makie.jl/blob/78d71c5609684ebbeaea7e073740582798d59330/src/interfaces.jl#L65-L68 which isn't what we want here and it might make sense to have `Markersize = 1` as the default?

ffreyer:
The alignment and size should be correct now. I tested with 
```julia
marker = BezierPath([
    MoveTo(Point(5, 5)),
    LineTo(Point(25, 5)),
    LineTo(Point(25, 25)),
    LineTo(Point(5, 25)),
    ClosePath()
])

scene = Scene()
scatter!(scene, Point2f(-1, -1), marker = marker, markersize = 1)
scene
```
which puts a 20px x 20px marker (5px, 5px) away from the bottom corner of the window.
Screenshot from 2022-01-14 15-04-51

jkrumbiegel:
Thank you, that looks promising! I will test it out later

jkrumbiegel:
Hm seems like it's not correct, yet. Maybe the square hid some aspect problem? Here's what I get with GLMakie for this
```julia
f = Figure()
ax = Axis(f[1, 1])

for (i, m) in enumerate([
        Makie.BezierCross,
        Makie.BezierSquare,
        Makie.BezierUTriangle,
        Makie.BezierDTriangle,
        Makie.BezierRTriangle,
        Makie.BezierLTriangle,
    ])

    scatter!(1:5, (1:5).+i, marker = m, markersize = 30)
end

f
```
grafik

Compared to CairoMakie:

grafik

ffreyer:
I relied on `bbox(bezierpath)` for the target scale. That seems to be wrong at least for `BezierRTriangle`:
```julia
> Makie.bbox(Makie.BezierRTriangle)
GeometryBasics.HyperRectangle{2, Float32}(Float32[-0.47140452, -1.7319121f-16], Float32[2.9802322f-8, 0.70710677])
```

ffreyer:
`union(bb1, bb2)` seems to not like negative widths. With the fix I get this now:
Screenshot from 2022-01-15 13-54-22

ffreyer:
And offsets fixed:
Screenshot from 2022-01-15 14-00-03

src/bezier.jl
```diff
@@ -0,0 +1,669 @@
...
+    BezierPath([
+        MoveTo(Point2(r, -r)),
+        LineTo(Point2(r, r)),
+        LineTo(Point2(-r, r)),
+        LineTo(Point2(-r, -r)),
+        ClosePath()
+    ])
+end
+
+BezierCross = let
```
SimonDanisch, line 162: should use const?

GLMakie/src/GLVisualize/visualize/particles.jl
```diff
@@ -220,16 +282,52 @@ function draw_scatter((marker, position), data)
...
+
+    elseif to_value(marker) isa AbstractArray
+        scale = map(combine_scales,
+            pop!(data, :scale, Observable(nothing)),
+            pop!(data, :scale_x, Observable(nothing)),
+            pop!(data, :scale_y, Observable(nothing)),
+            pop!(data, :scale_z, Observable(nothing))
+        )
+        offset = Observable(Vec2f(0))
+        _offset(x::Union{AbstractString, Char}, scale, offset) = rescale_glyph(x, font[], offset)
```
SimonDanisch, line 316: We should have just one function, that dispatches on all the various combinations of marker, vector of markers, with different element types (Bezier/Glyph) and remove the complexity from the drawing code.
It would also be nice to calculate scale + offset in one function, to only calculate the boundingbox once.

src/conversions.jl
```diff
@@ -1225,8 +1240,8 @@ to_spritemarker(marker::AbstractMatrix{<: Colorant}) = marker
 A `Symbol` - Available options can be printed with `available_marker_symbols()`
 """
 function to_spritemarker(marker::Symbol)
-    if haskey(_marker_map, marker)
-        return to_spritemarker(_marker_map[marker])
+    if haskey(default_marker_map(), marker)
```
SimonDanisch, line 1243: Hm, should we really always re-create the Dict when looking something like this up?
Now, that markers even get created in quite the complex way, this seems pretty expensive, to just be called in the default convert function?
I guess the best way would be to memoize `default_marker_map` ... 
Maybe just with `if isempty(DEFAULT_MARKER_MAP); ... create default markers...; end; return DEFAULT_MARKER_MAP;`

jkrumbiegel: ah yeah I think I did this at the time so I could Revise the markers

SimonDanisch:
Final changes of this PR:

* add `BezierPath` which can be constructed from SVG like command list, SVG string or from a `Polygon`. 
* add ability to use `BezierPath` and `Polgyon` as scatter marker
* replace default symbol marker like `:cross` with more precise `BezierPaths`
* deprecate string to specify multiple char markers (`scatter(1:4, marker="abcd")`)
* deprecate concrete geometries as marker like `Circle(Point2f(0), 1.5)`
* add single image marker support to WGLMakie

ffreyer:
>     * deprecate concrete geometries as marker like `Circle(Point2f(0), 1.5)`

Does this mean the in-shader generated signed distance fields are deprecated too? Or is there another way to cleanly reach those? (triangle, circle, rectangle and rounded_rectangle in distance_shape.frag)

SimonDanisch:
Just use `Circle` ;) From what I've seen, we didn't implement fetching the size from the concrete circle anyways...

ffreyer:
```julia
using GLMakie

t = RGBAf(0,0,0,0)
r = RGBAf(1,0,0,1)
pattern = Makie.ImagePattern([t t r r; t t r r; r r r r; r r r r])
scatter(rand(10), color = pattern)
```
doesn't work anymore, but it also doesn't work on master. Seems like the conversion pipeline for this got messed up at some point. I'd say that's something for a separate issue though.

Other example plots I've tried look fine to me. The square example I posted before still has the correct size and offset as well.

src/bezier.jl
```diff
@@ -0,0 +1,691 @@
...
+        m = Mat2f0(sin(a), cos(a), cos(a), -sin(a))
+        Ref(m) .* first_three
+    end |> x -> reduce(vcat, x)
+
+    BezierPath([
+        MoveTo(all[1]),
+        LineTo.(all[2:end])...,
+        ClosePath()
+    ])
+end
```
ffreyer, line 189: Why do these have slightly different sizes, and sizes smaller than 1x1 (or radius 0.5)?

For reference, Bezier circle and square on top of 100x100px shader square:
```julia
scene = Scene()
scatter!(scene, Point2f(0), marker = Rect, color = :green, markersize = 100)
scatter!(scene, Point2f(0), marker = Makie.BezierSquare, color = :blue, markersize = 100)
scatter!(scene, Point2f(0), marker = Makie.BezierCircle, color = :red, markersize = 100)
scene
```
Screenshot from 2022-09-10 20-17-14
and with the radii set to 0.5:
Screenshot from 2022-09-10 20-23-10

src/conversions.jl
```diff
@@ -1149,28 +1149,44 @@ function convert_attribute(value::Union{Symbol, String}, k::key"algorithm")
...
-    :star4 => '✦',
-    :star6 => '🟋',
-    :star8 => '✷',
-    :vline => '┃',
-    :hline => '━',
-    :+ => '+',
-    :x => 'x',
-    :circle => '●'
-)
+const DEFAULT_MAKER_MAP = Dict{Symbol, BezierPath}()
```
ffreyer, line 1152: Should be marker instead of maker?

src/conversions.jl
```diff
@@ -1198,11 +1214,24 @@ Note, that this will draw markers always as 1 pixel.
...
+Note that it needs to be a uniform vector with the same element type!
+"""
+to_spritemarker(marker::AbstractVector) = map(to_spritemarker, marker)
+to_spritemarker(marker::AbstractVector{Char}) = marker # Don't dispatch to the above!
 to_spritemarker(x::FastPixel) = x
 to_spritemarker(x::Circle) = x
-to_spritemarker(::Type{<: Circle}) = Circle(Point2f(0), 1f0)
-to_spritemarker(::Type{<: Rect}) = Rect(Vec2f(0), Vec2f(1))
+to_spritemarker(::Type{<: Circle}) = Circle
+to_spritemarker(::Type{<: Rect}) = Rect
```
ffreyer, line 1226: Should we also have conversions for rounded rectangle and triangle? Or perhaps allow `marker = Makie.ROUNDED_RECTANGLE`?

SimonDanisch:
> ```julia
> 
> t = RGBAf(0,0,0,0)
> ```

Hm, this seems to be already broken in the latest tagged version... I'll see if I can quickly fix it in this PR!

src/bezier.jl
jkrumbiegel: Ok so the reasoning was to create markers that have the same area, because I had assumed these would look similar. That's an important characteristic for markers because you don't really want to convey a difference in size when you use two different marker shapes with the same `markersize` setting.

I started with the circle, which I gave area = 1. That means the radius is about 0.56. Later I changed that because I wanted each shape to fit approximately within the 1x1 square. I then found out that not all shapes look like they have similar size if you correctly compute their area, probably because human visual perception introduces non-linearities when shapes differ. That's well known but I didn't exactly go on a literature search to do this (maybe I should).

So anyway, after I while I had shapes that looked good together, all approximately located in a 1x1 square. But these looked too big when mixing with character-based markers like `'X'` or `'a'`. That's because fonts at font-size X do not fill a square with size X * X with each character, they're effectively smaller. So I multiplied the whole range of markers with a factor that looked reasonably pleasing to me with both small and large letters. That is entirely subjective of course, but as markers are usually not meaningful in data space, it seemed reasonable to me to go this route. The benefit is that you can mix and match the normal markers with text/char at the same size, and won't have to adjust markersize each time. The drawback is that there's not a clean "theoretical" reason why they are of this size. I understand that making each marker fill 1x1 exactly sounds attractive theoretically, but to me it seemed to be less attractive in practice.

I find it confusing if char-based markers at size X are different than the same thing as `text` at size X, so that fixes those markers. Then the question is, should the default markers match those in size or not, I said yes they should. Hence the unusual size.

The nice thing about the new system is that it's easy for users to make their own markers, and those can conform to a base size of 1x1 or whatever, if that metric is important to the users. I can imagine it might be important when tiling markers in dataspace, but I'd argue that is sufficiently niche as to not influence the decision for a default.

jkrumbiegel: image

here you can see a radius 0.5 circle at the top, plus some other new default markers below, and two characters. I would find it pretty impractical if all default markers were of the size as the large circle.

ffreyer: I see, that makes sense.

There's a note on characters being of nonuniform size/area but as far as I can tell we don't explain how characters or other markers are actually scaled. I think it would be good to have that information there though. I.e. with `markerspace = :pixel`
- for characters `markersize` is equivalent to `textsize/fontsize` which is the pixel height of the bounding box a character is placed into, to my understanding
- default shape markers are scaled to character markers, meaning that `markersize = 10` results in smaller side length/diameters
- custom Bezier markers use a 1:markersize scaling between units of the marker and units in the plot. I.e. a Bezier marker filling a 0..1 x 0..1 square with `markerspace = 10` will fill a 0..10 x 0..10 pixel square on the screen exactly.
- Polygons, I assume, are handled the same

docs/examples/plotting_functions/scatter.md
```diff
@@ -105,6 +104,106 @@ f
 ```
 \end{examplefigure}
 
+### Bezier path markers
+
+You can also use bezier paths as markers.
+A bezier path marker should fit into the square from -1 to 1 in x and y to be comparable in size to other default markers and to be correctly rendered by GLMakie, because here the path has to be rendered to a bitmap first .
```
ffreyer, line 110: I didn't have any problems rendering a 5..25 x 5..25 square in GLMakie so I think this is wrong. Or are there issues with more complex shapes?

src/bezier.jl
jkrumbiegel: Yes it should be documented as such, I agree. And you're right, effectively the bezier coordinates are used as they are and just multiplied with the markersize, so you can get pixel-perfect output if you want.

docs/examples/plotting_functions/scatter.md
jkrumbiegel: Oh that's probably outdated, I started out implementing it that way, but later changed it so each marker is scaled to fit before rendering to the atlas.

SimonDanisch (approved):

PR #984, opened 2021-05-25
fix plotsym in Pluto + Documenter

SimonDanisch:


SimonDanisch:
Of course this doesn't work on 1.3... I'm not sure if this can even work on 1.3, since the alias aware printing happened later ...
See master and Julia 1.3:
```julia
julia> Makie.plotsym(Lines)
:Combined
```

PR #999, opened 2021-05-27
Fixing doc example recipe

jlchan:
Makes it so that the "simple" recipe in the docs runs. Should close https://github.com/JuliaPlots/Makie.jl/issues/771


SimonDanisch:
Sorry for letting this rot... I wanted to add some more recipes clean up to this, since it was still not fully correct... This got hopelessly out of date, and I think the most important changes found its way into the new documentation.
So I'll close this for now and hope I can find some time to polish up the recipe docs soon!

Issue #951, opened 2021-05-15
hist! in error for yscale=log10 


walra356:
hist! works fine for yscale=identity but errors for yscale=log10.  xscale=log10 works fine. A log of zero is always a problem but how to avoid it in this case? See example below: 
```
using GLMakie
GLMakie.activate!()
AbstractPlotting.inline!(true) 
data = abs.(randn(500))
f = Figure()
ax = Axis(f[1,1],yscale=log10)
hst = hist!(ax,data;bins=25)
f
```
ans = Found invalid direct y-limits (0.0f0, 63.0f0) for scale log10 which is defined on the interval 0.0..Inf (open)

walra356:
I made some progress: by setting the limits on axes I can suppress the error. I can nicely scroll the scale but the HISTOGRAM IS NOT DISPLAYED (see example below). I do not have this difficulty with xscale=log10. What to do to make the histogram visible?
```
using GLMakie
GLMakie.activate!()
data = abs.(randn(500))
f = Figure()
ax = Axis(f[1,1],yscale=log10)
ax.limits=(0.0011,3,0.0001,20)
hst = hist!(ax,data;bins=25)
AbstractPlotting.inline!(false) 
display(f)
``

**Comment by ffreyer** [comment]
The problem is that a histogram bar starts at 0 and `log10(0) = -Inf`, I assume. I'm not sure if this works with arbitrary scales, but if it does maybe `yscale = y -> log10(y+1)` would work?

**Comment by walra356** [comment]
Your suggestion did not work... I tend to agree with you. hist! cannot plot the lower end of the histogram bar. So, one has to set the lower end of the bar equal to the lower y-limit (for the log options). That would do the job. Do you know how to do that?

Maybe I should abandon the hist!  and do the binning myself. Then I can use a scatter plot...

As a physics user I would say that a y-log option is more or less indispensible.

Thanks for your comments and success with your studies!

**Comment by walra356** [comment]
I could not resist having a closer look at the code. The problem is that `hist!` depends on  `barplot`, which has an attribute called  `fillto` which is put on `automatic`. In `hist` `automatic` results in fillto=0.0,. So the histogram rectangle starts at 0 :

```
function bar_rectangle(x, y, width, fillto)
    # y could be smaller than fillto...
    ymin = min(fillto, y)
    ymax = max(fillto, y)
    w = abs(width)
    return FRect(x - (w / 2f0), ymin, w, ymax - ymin)
end
```
[...]

PR #1000, opened 2021-05-27
avoid truncated barplot with logscale

vchuravy:
With `(y|x)scale = log10` the first bar would be truncated.

### DIFF
--- a/src/basic_recipes/barplot.jl
+++ b/src/basic_recipes/barplot.jl
@@ -154,7 +154,7 @@ end
 
 function stack_from_to_sorted(y)
 	to = cumsum(y)
-	from = [0.0; to[firstindex(to):end-1]]
+	from = [eps(); to[firstindex(to):end-1]]
 
 	(from = from, to = to)
 end


vchuravy:
In general maybe we need a "zero" element for logscale?

vchuravy:
Example code:

```
using CairoMakie
function diff(x)
        res = [x[1]]
        for (i, v) in enumerate(x[2:end])
                push!(res, v-x[i])
        end
        res
end
rsbench = diff([4.4,   343.7,    1275.6, 6372.2, 10000])
rsbench_labels = [:AD, :CacheLICM, :RecomputeHeuristic, :Inlining, :PreOptimization]
rsbench_id = fill(1, length(rsbench))
rsbench_stack = collect(1:length(rsbench))
xsbench = diff([3.16, 9.61, 14.47, 47.40, 95.75, 10000])
xsbench_labels = [:AD, :Templating, :PHI, :AliasAnalysis, :RecomputeHeuristic, :PreOptimization]
xsbench_id = fill(2, length(xsbench))
xsbench_stack = collect(1:length(xsbench))
lbm = diff([6.39, 8.74, 19.87, 10000])
lbm_labels = [:AD, :Allocator, :RecomputeHeuristic, :InlineCacheABI]
lbm_id = fill(3, length(lbm))
lbm_stack = collect(1:length(lbm))
dgcuda = diff([17.85, 116.63, 1378.32, 10000])
dgcuda_labels = [:AD, :Unrolling, :MallocCoalesing, :PreOptimization]
dgcuda_id = fill(4, length(dgcuda))
dgcuda_stack = collect(1:length(dgcuda))
dgrocm = diff([5.4, 10000])
dgrocm_labels = [:AD, :Unrolling]
dgrocm_id = fill(5, length(dgrocm))
dgrocm_stack = collect(1:length(dgrocm))
unique_labels = sort([v for v in Set(vcat(rsbench_labels, xsbench_labels, lbm_labels, dgcuda_labels, dgrocm_labels))])
label_ids = Dict()
for (i, m) in enumerate(unique_labels)
        label_ids[m] = i
end
function label_id_map(m)
        [label_ids[v] for v in m]
end
colors = vcat([label_id_map(m) for m in [rsbench_labels, xsbench_labels, lbm_labels, dgcuda_labels, dgrocm_labels]]...)
fig = Figure()
ax = Axis(fig[1,1], yticks = (1:5, ["RSBench", "XSBench", "LBM", "DG (CUDA)", "DG (ROCM)"]),
        title = "Dodged bars with legend",
        xscale = log10)
xlims!(1, 10_000)
heights = vcat(rsbench, xsbench, lbm, dgcuda, dgrocm)
ids = vcat(rsbench_id, xsbench_id, lbm_id, dgcuda_id, dgrocm_id)
stacks = vcat(rsbench_stack, xsbench_stack, lbm_stack, dgcuda_stack, dgrocm_stack)
# Plot
barplot!(ax, ids, heights,
        stack = stacks,
        color = colors,
        direction = :x)
```

jkrumbiegel:
You usually have to handle this via the offset or fill_to parameter, have a look at the second plot in this section https://makie.juliaplots.org/stable/makielayout/axis.html#Log-scales-and-other-axis-scales

SimonDanisch:
Maybe we should set the fillto to that number for logscale plots then?

jkrumbiegel:
but what number do you set it to?

piever:
Is this something we want to support though? Naively, I would think that stacked bar plots on a log scale are a bad idea (subbars with the same height do not correspond to the same magnitude).

If one really wants them, maybe the "starting point" could be set using `fillto`, which is currently ignored in the stacked case, see https://github.com/JuliaPlots/Makie.jl/blob/fc769f59a31558b62aaf78210d7f5379dfa60912/src/basic_recipes/barplot.jl#L95

walra356:
For barplots manual `fillto` works fine for monolithic bars on log scales. Stacking is an issue. Only the bottom bar needs a `fillto` option. Using this value for all bars might be acceptable just because we are dealing with a log scale. An automatic is option would be a `fillto` to a value 10x smaller than the bottom `bar height`. Not good for performance. See also #951 and #1001  (adding `fillto` to `hist attributes`).

jkrumbiegel:
As this is a problem inherent to logscales, and I've added options for symmetric log-similar scales lately, which can deal with zero and negative values https://makie.juliaplots.org/dev/makielayout/axis.html#Pseudolog-and-symlog-scales, I'll close this.

Moelf:
for posterity, I want to point out in response to (no contention, just a FYI):
>Is this something we want to support though? Naively, I would think that stacked bar plots on a log scale are a bad idea (subbars with the same height do not correspond to the same magnitude).

that this is absolutely crucial for some fields:
https://atlas.cern/updates/briefing/fantastic-decays-and-where-find-them
https://cds.cern.ch/images/ATLAS-PHOTO-2020-010-5/file?size=large

Our hands are forced because, physics. (nature hates us?) The way these plots are made is we put the histogram with the smallest integral to the bottom, I know it's next to impossible to eye ball numbers, but at least you can point to a color when the stacked number doesn't agree with observed number and reason about distribution within the same color.

PR #1015, opened 2021-06-01
Delete title.jl

jlchan:
As suggested on Slack. `title` is deprecated, and there are [alternative ways to make titles](https://discourse.julialang.org/t/makie-add-title-to-plot/22340/7?u=jlchan).


SimonDanisch:
I guess this needs to delete the include for `title.jl` and any other mention of it (e.g. in the docs)

jlchan:
Gotcha - tracking it down in the docs now

jlchan:
Doesn't look like `title(args...)` shows up anywhere else in the docs.

SimonDanisch:
I bet that's an unused var, that got left over by a refactor, and only kept working because there was the title function around... :D Onany accepts any type, and the args aren't used, which is why nobody has noticed yet.
So I think you can just delete the title from the `onany` arguments!

jlchan:
Thanks for catching that stray `onany(title`!

SimonDanisch:
Thank you :)

Issue #996, opened 2021-05-27
Lightweight MakieRecipes.jl


rafaqz:
Recipes currently require depending on Makie.jl. It has a lot of dependencies compared to RecipesBase.jl:
https://github.com/JuliaPlots/Makie.jl/blob/master/Project.toml
https://github.com/JuliaPlots/RecipesBase.jl/blob/master/Project.toml

I'm keen to write recipes for packages like DimensionalData.jl and GeoData.jl. Adding Plots.jl recipes with a RecipesBase.jl dependency really had no downsides:

```julia
julia> @time using DimensionalData
  0.185980 seconds (528.76 k allocations: 37.096 MiB, 7.28% gc time, 49.88% compilation time)
```

But:
```julia
julia> @time using Makie
  2.849688 seconds (5.90 M allocations: 386.160 MiB, 10.64% gc time)
```

How feasible would it be to pull out method headers and abstract type definitions from Makie.jl to a base package we can depend on, similar to RecipesBase.jl?

SimonDanisch:
I wanted to transfer some core Makie functionality to MakieCore for some time and took the chance to see how minimal I can make it while still supporting creating recipes. It's pretty minimal now (only Observables as a dependency), and fairly quick to compile:
```julia
julia> @time using MakieCore
  0.024640 seconds (71.96 k allocations: 4.669 MiB, 6.46% compilation time)
```
https://github.com/JuliaPlots/MakieCore.jl
https://github.com/JuliaPlots/Makie.jl/pull/998

I'm not sure how usable this will be for defining lightweight recipes, but I intended to make this refactor anyways so 🤷 I guess we can experiment a bit and see in what direction MakieCore should go.
Initially I wanted to have MakieCore + Backend be a fully blown drawing solution, a.k.a the Makie core that does the graphic drawing without any plotting related APIs.
 
I'm happy to do the transformation step by step and start with an absolute minimal solution which we can use as a lightweight MakieRecipes.jl prototype.

I do want to emphasize though, that truly lightweight recipes, that can fully describe a Makie plot, will need some serialization / dict base format, to access all advanced recipes + layouting without having any dependencies...

rafaqz:
MakieCore looks good! Thanks for getting that moving. When I have time I'll try some GeoData.jl recipes using only MakieCore.

In terms of keeping it lightweight, should we define what lightweigth means to have a target? Maybe staying under 0.05 seconds on that machine.

[...]

PR #1020, opened 2021-06-03
turn Makie into a mono repo with all backends etc

SimonDanisch:
After realizing, that with MakieRecipes + MakieCore we'll have two new core packages, that will get releases in lockstep with the other packages, @jkrumbiegel and I realized that it's maybe time to create a Makie mono repo, which would solve lots of problems.
I didn't realize this was easily possible in Julia, but after some digging it seems like julia 1.5 has fairly solid support for it :) 
- [x] merge packages
- [x] update CI scripts to make sense
- [x] update readmes
- [x] update docs 
- [x] manually update urls in registry
- [ ] auto tag all touched sub repositories on merge to master


jlchan:
Just curious, what does this mean with respect to a lightweight recipe package, i.e., ￼￼https://github.com/JuliaPlots/Makie.jl/issues/996?

SimonDanisch:
Good news it means ;) You could almost say, that #996 triggered this, since I wanted to add MakieRecipes + MakieCore, without having greater maintenance burden for tagging packages etc.

PR #1036, opened 2021-06-09
Fix rotated font issue 

### LINKED ISSUE #1033: Cairo messing with fonts
Managed to create an MWE for Cairo messing with GLMakie's font atlas:

```julia
using CairoMakie
using FreeTypeAbstraction
using Colors
using Cairo 
dfont = findfont("Dejavu Sans")
bitmap, extent = renderface(dfont, '5', 64)
display(Gray.(bitmap./255))
```
image

```julia
function mwe(ctx, font)
    Cairo.save(ctx)
    cairoface = CairoMakie.set_ft_font(ctx, font)
    Cairo.rotate(ctx, 0.5)
    Cairo.text_path(ctx, "5")
    CairoMakie.cairo_font_face_destroy(cairoface)
    Cairo.rotate(ctx, 0.0)
    Cairo.restore(ctx)
end

surf = Cairo.CairoARGBSurface(500, 500)
ctx = Cairo.CairoContext(surf)
mwe(ctx, dfont)
```
```julia
dfont = findfont("Dejavu Sans")
bitmap, extent = renderface(dfont, '5', 64)
display(Gray.(bitmap./255))
```
image

But no idea how to cleanly stop cairo from doing this... Maybe easiest would be to copy the fonts?
Fun fact: my google skills aren't good enough to figure out how to copy fonts with FreeType, so the copy solution would mean, we need to somehow save the path with the font and load it from there as a copy...Or something like that.

SimonDanisch:
Fixes #1033

### DIFF
--- a/src/utilities/texture_atlas.jl
+++ b/src/utilities/texture_atlas.jl
@@ -263,7 +263,7 @@ function render(atlas::TextureAtlas, glyph::Char, font, downsample=5, pad=6)
     # the target pixel size of our distance field
     pixelsize = PIXELSIZE_IN_ATLAS[]
     # we render the font `downsample` sizes times bigger
-    # Make sure the font doesn't have a nutated font matrix from e.g. Cairo
+    # Make sure the font doesn't have a mutated font matrix from e.g. Cairo
     FreeTypeAbstraction.FreeType.FT_Set_Transform(font, C_NULL, C_NULL)
     bitmap, extent = renderface(font, glyph, pixelsize * downsample)
     # Our downsampeld & padded distancefield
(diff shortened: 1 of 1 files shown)


src/utilities/texture_atlas.jl
```diff
@@ -263,6 +263,8 @@ function render(atlas::TextureAtlas, glyph::Char, font, downsample=5, pad=6)
     # the target pixel size of our distance field
     pixelsize = PIXELSIZE_IN_ATLAS[]
     # we render the font `downsample` sizes times bigger
+    # Make sure the font doesn't have a nutated font matrix from e.g. Cairo
```
piever, line 266:
```suggestion
    # Make sure the font doesn't have a mutated font matrix from e.g. Cairo
```

Sorry for the not particularly insightful review and thanks a lot for fixing this!

PR #1064, opened 2021-06-24
port @ndinsmore osx retina fix

SimonDanisch:
This tries to port the fix from https://github.com/JuliaPlots/GLMakie.jl/pull/145 without any breaking changes.
Note, that retina rescaling needs to apply on all platforms (Windows/Linux as well), so that part of the pipeline should stay.
It's implementation should be refactored at some point though ;) 
I don't have a mac, so can @jkrumbiegel or @ndinsmore test this and make sure disabling it per default is ok?
I made it configurable, so the default could also be `enabled`.

### DIFF
--- a/.github/workflows/cairomakie.yaml
+++ b/.github/workflows/cairomakie.yaml
@@ -1,15 +1,15 @@
-name: CairoMakie CI
+name: CairoMakie
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - CairoMakie/**
+      - src/**
     branches:
       - master
   push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - CairoMakie/**
+      - src/**
     branches:
       - master
     tags: '*'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,15 @@
 name: Makie CI
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - MakieCore/**
+      - src/**
     branches:
       - master
   push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - MakieCore/**
+      - src/**
     branches:
       - master
     tags: '*'
--- a/.github/workflows/glmakie.yaml
+++ b/.github/workflows/glmakie.yaml
@@ -1,15 +1,15 @@
-name: GLMakie CI
+name: GLMakie
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - GLMakie/**
+      - src/**
     branches:
       - master
   push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - GLMakie/**
+      - src/**
     branches:
       - master
     tags: '*'
--- a/.github/workflows/wglmakie.yaml
+++ b/.github/workflows/wglmakie.yaml
@@ -1,15 +1,15 @@
-name: WGLMakie CI
+name: WGLMakie
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - WGLMakie/**
+      - src/**
     branches:
       - master
   push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths:
+      - WGLMakie/**
+      - src/**
     branches:
       - master
     tags: '*'
--- a/GLMakie/src/rendering.jl
+++ b/GLMakie/src/rendering.jl
@@ -56,7 +56,8 @@ function renderloop(screen; framerate=WINDOW_CONFIG.framerate[])
     end
 end
 
-const WINDOW_CONFIG = (renderloop = Ref{Function}(renderloop),
+const WINDOW_CONFIG = (
+    renderloop = Ref{Function}(renderloop),
     vsync = Ref(false),
     framerate = Ref(30.0),
     float = Ref(false),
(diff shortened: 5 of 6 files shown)


ffreyer:
Isn't the scaling the part that makes things slow though? And is it independent of `GLFW_COCOA_RETINA_FRAMEBUFFER`?

SimonDanisch:
🤷  I have no idea ... The scaling should just correct coordinates etc for the case `window_size(window) !== framebuffer_size(window)`...
So if those are not different, no scaling should happen. I assumed `GLFW_COCOA_RETINA_FRAMEBUFFER=false` would make those the same size, so no scaling needing to happen.

>GLFW_COCOA_RETINA_FRAMEBUFFER specifies whether to use full resolution framebuffers on Retina displays. Possible values are GLFW_TRUE and GLFW_FALSE. This is ignored on other platforms.

I'm not 100% sure what that means :D what does full resolution mean? I either get the pixels I request...or not :D 

Welp, I think someone on OSX will need to figure this all out 😓  
But the retina scaling (bad name) should be osx dependent and should really just make sure, that if I say I want a 800x800 PIXEL window, that I really get one and correct mouse positions etc. If it doesn't just do that, it's a bug, I suppose ^^

ffreyer:
Talking about mouse positions - I would like to see https://github.com/JuliaPlots/Makie.jl/blob/01369cbe36452d1554fc8decff8069a0bf02184f/GLMakie/src/events.jl#L180-L184 become independent of `GLFW.GetXSize` calls. As far as I remember that was a major performance sink and without those calls we don't need to throttle mouse position events. If the scaling only happens when `GLFW_COCOA_RETINA_FRAMEBUFFER=true` we could use that with `size(::GLFramebuffer)`.

I guess https://github.com/JuliaPlots/Makie.jl/blob/01369cbe36452d1554fc8decff8069a0bf02184f/GLMakie/src/glwindow.jl#L105-L106 should also depend on `osx_retina_scaling` then?

SimonDanisch:
>become independent of GLFW.GetXSize calls. As far as I remember that was a major performance sink and without those calls we don't need to throttle mouse position events. If the scaling only happens when GLFW_COCOA_RETINA_FRAMEBUFFER=true we could use that with size(::GLFramebuffer).

As noted as a comment in the function `retina_scaling_factor`, we best memoize the value of the scaling factor, and then just use that... I didn't realize that it's actually a performance issue, so we should immediately memoize it, I guess :D 

>should also depend on osx_retina_scaling then?

Good questions... I don't remember if the `2` was just lazyness, or if that's actually how one can calculate the value on OSX... Should have left a comment -.- 
One problem is definetly, that it seems to be only possible to get the scaling factor with an open window, while the monitor props can be queried all the time. Maybe not a real problem, since we only query them when we have an open window

ffreyer:
> > become independent of GLFW.GetXSize calls. As far as I remember that was a major performance sink and without those calls we don't need to throttle mouse position events. If the scaling only happens when GLFW_COCOA_RETINA_FRAMEBUFFER=true we could use that with size(::GLFramebuffer).
> 
> As noted as a comment in the function `retina_scaling_factor`, we best memoize the value of the scaling factor, and then just use that... I didn't realize that it's actually a performance issue, so we should immediately memoize it, I guess :D

I mentioned it in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/634 but I didn't dare to mess with mac related things... With some `TimerOutputs.@timeit_debug`s sprinkled in there, using an empty scene:

```
 ───────────────────────────────────────────────────────────────────────────────────────
                                                Time                   Allocations      
                                        ──────────────────────   ───────────────────────
            Tot / % measured:                15.9s / 0.87%           7.63MiB / 0.34%    

 Section                        ncalls     time   %tot     avg     alloc   %tot      avg
 ───────────────────────────────────────────────────────────────────────────────────────
 mouse_position on rendertick      325    139ms   100%   426μs   26.5KiB  100%     83.5B
   correct_mouse                   325   87.5ms  63.2%   269μs   1.66KiB  6.25%    5.22B
     window_size                   325   44.4ms  32.0%   137μs     0.00B  0.00%    0.00B
     framebuffer_size              325   41.1ms  29.7%   127μs     0.00B  0.00%    0.00B
   GLFW.GetCursorPos               325   47.9ms  34.6%   147μs     0.00B  0.00%    0.00B
 ───────────────────────────────────────────────────────────────────────────────────────
 ```
Somehow this worse than I remember...

Using a callback without scaling factors/GLFW.GetXSize:

```julia
function cursorposition(window, w::Cdouble, h::Cdouble)
    @timeit_debug "cursorposition" begin
    @print_error begin
        # pos = correct_mouse(window, w, h)
        fb_w, fb_h = size(screen.framebuffer)
        pos = (w, fb_h - h)
        e.mouseposition.val = pos
        return
    end
    end
end
disconnect!(window, mouse_position)
GLFW.SetCursorPosCallback(window, cursorposition)

```

results in

```
 ─────────────────────────────────────────────────────────────────────────
                                  Time                   Allocations      
                          ──────────────────────   ───────────────────────
     Tot / % measured:         13.6s / 0.03%           7.56MiB / 2.63%    

 Section          ncalls     time   %tot     avg     alloc   %tot      avg
 ─────────────────────────────────────────────────────────────────────────
 cursorposition    6.51k   3.63ms   100%   557ns    203KiB  100%     32.0B
 ─────────────────────────────────────────────────────────────────────────
```

So yea, seems like this is almost 1000x faster :sweat_smile:

SimonDanisch:
Damn :D Ok, so maybe we should immediately make a new PR to fix at least that...

ffreyer:
I don't think this is really bothering anyone. This should be a flat per-frame cost, maybe even relatively hardware independent? I guess it would help people who have trouble getting to 30/60fps. There might also be harm in having things that rely on mouse position trigger more frequently...

ffreyer:
Maybe we could have the callback just affect `event.val` and continue to trigger `on(rendertick)`. Though it probably makes more sense to throttle in Makie. Maybe we should move rendertick over there. (Also just noticed that I'm updating `e.mouseposition.val` in my test, but it doesn't really matter for just `Scene()`)

SimonDanisch:
Going to close this until someone with a Mac is willing to debug this...

PR #1065, opened 2021-06-24
fix Rect marker

SimonDanisch:
Somehow, `scatter(1:4, marker=Rect)` has been plain broken... 
`:rect` as markertype is rendering '■', which turns out smaller than the markersize, and also doesn't scale correctly with `transform_marker=true, markerspace=SceneSpace`.

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -208,38 +208,18 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Scatter)
 
         Cairo.set_source_rgba(ctx, rgbatuple(col)...)
         m = convert_attribute(marker, key"marker"(), key"scatter"())
+        Cairo.save(ctx)
         if m isa Char
             draw_marker(ctx, m, best_font(m, font), pos, scale, strokecolor, strokewidth, offset, rotation)
         else
             draw_marker(ctx, m, pos, scale, strokecolor, strokewidth, offset, rotation)
         end
+        Cairo.restore(ctx)
     end
     nothing
 end
 
-
-function draw_marker(ctx, marker::Circle, pos, scale, strokecolor, strokewidth, marker_offset, rotation)
-
-    marker_offset = marker_offset + scale ./ 2
-
-    pos += Point2f0(marker_offset[1], -marker_offset[2])
-
-    # Cairo.scale(ctx, scale...)
-    Cairo.move_to(ctx, pos[1] + scale[1]/2, pos[2])
-    Cairo.arc(ctx, pos[1], pos[2], scale[1]/2, 0, 2*pi)
-    Cairo.fill_preserve(ctx)
-
-    sc = to_color(strokecolor)
-
-    Cairo.set_source_rgba(ctx, rgbatuple(sc)...)
-    Cairo.set_line_width(ctx, Float64(strokewidth))
-    Cairo.stroke(ctx)
-end
-
 function draw_marker(ctx, marker::Char, font, pos, scale, strokecolor, strokewidth, marker_offset, rotation)
-
-    Cairo.save(ctx)
-
     # Marker offset is meant to be relative to the
     # bottom left corner of the box centered at
     # `pos` with sides defined by `scale`, but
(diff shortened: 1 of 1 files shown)


piever:
I'm traveling at the moment so can't really double check, but I think a similar issue also exists for `:circle` (at least in CairoMakie). The marker you get with `:circle` is smaller than the one you get with `Circle`.

This is a bit problematic for cycling, because the first option of the cycle is smaller than the overall default option.

jkrumbiegel:
We should really define what the markersize means. I like proportional area as that's perceptually relevant. Right now it's pretty random.

SimonDanisch:
Yeah, at some point we really need to pin down the meaning and write pixel perfect tests for it ...

CairoMakie/src/primitives.jl
```diff
@@ -287,32 +267,32 @@ function draw_marker(ctx, marker::Char, font, pos, scale, strokecolor, strokewid
...
+    marker_offset = marker_offset + scale ./ 2
+    pos += Point2f0(marker_offset[1], -marker_offset[2])
+    Cairo.arc(ctx, pos[1], pos[2], scale[1]/2, 0, 2*pi)
+    Cairo.fill_preserve(ctx)
 
-    offset = marker_offset .+ scale ./ 2
+    Cairo.set_line_width(ctx, Float64(strokewidth))
 
-    pos += Point2f0(offset[1], -offset[2])
+    sc = to_color(strokecolor)
```
jkrumbiegel, line 281: I wanted to go more into the direction of converting these things before going into the drawing functions. It would probably be good if colors had to be RGBA already etc to avoid conversion costs in these tight loops.

CairoMakie/src/primitives.jl
```diff
@@ -287,32 +267,32 @@ function draw_marker(ctx, marker::Char, font, pos, scale, strokecolor, strokewid
...
 
-function draw_marker(ctx, marker::Union{Rect, Type{<: Rect}}, pos, scale, strokecolor, strokewidth, marker_offset, rotation)
-    s2 = if marker isa Type{Rect}
-        Point2(scale[1], -scale[2])
-    else
-        Point2((widths(marker) .* scale .* (1, -1))...)
-    end
+    marker_offset = marker_offset + scale ./ 2
+    pos += Point2f0(marker_offset[1], -marker_offset[2])
+    Cairo.arc(ctx, pos[1], pos[2], scale[1]/2, 0, 2*pi)
```
jkrumbiegel, line 276: So for Circle and Rect to be the same apparent size at the same markersize, the scale shouldn't be the diameter of the circle and the side length of the rect. I think if you go with the area approach it becomes clear that rect should be 1 side length for area 1, and circle should be whatever you need there, don't know from the top of my head but calculated it for a bunch of shapes in the bezierpath branch.

CairoMakie/src/primitives.jl
SimonDanisch: Yeah, that's what I started implementing in the MakieCore refactor ;)

CairoMakie/src/primitives.jl
SimonDanisch: I think we should do any changes like this in a new PR ;)

CairoMakie/src/primitives.jl
jkrumbiegel: ok great then this is fine. I also made a `TorVector` like data structure in the mathtexengine pr, probably should merge that with your approach. It did make for a nice integration with `broadcast_foreach` I thought. We should probably make the attribute conversions explicit to one of these types. I find `convert_attribute(attr, key"color"())` a bit opaque, what about `convert(TorVector{RGBAf0}, attr)`? I called the data structure ScalarOrVector I think.

SimonDanisch: the plan is to refactor all the `convert_attribute(x, key"color")` to get called by `to_color(x)` ... I already started this, but only for the most used once (e.g. `to_color, to_font` etc).

>what about convert(TorVector{RGBAf0}, attr)?

I guess when we have the new, fully typed plot types, we can do:
`field_convert(fieldtype(plot, key), x)` ... I don't think we'll be able to use `convert`, since that would need a lot of type piracy.
Also, I'm a bit afraid, that it will be harder to implement, since the unions in the field types get pretty complex, so may just stick with `to_color` etc. Also, we may still need some fields of type any...

SimonDanisch: Btw, `convert_attribute(attr, key"color"())` is really meant as an internal interface, for operations like:
```julia
for (key, value) in attributes
   convert_attribute(value, Key{key}())
end
```

jkrumbiegel: ok but the output of these `to_color` etc. conversions could be one of these TorVectors I guess, so that the caller has to handle both possibilities correctly?

SimonDanisch: Yeah... the main change for the new Plot types will be, that at least the set of result types should be easily visible.

PR #1066, opened 2021-06-24
allow colorrange as an attribute for Colorbar, which is consistent with plot types

SimonDanisch:


jkrumbiegel:
In principle I don't mind this, but the current interface for colors / colormaps could be improved overall. Currently, we only support linear color lookup. You can of course transform the colormap but you lose the necessary precision and get banding issues etc. So instead of color plus colorrange, it would be better to pass colors as a sampler type directly, where array values and colormap go together with a possible scale function.

SimonDanisch:
Yeah, I've actually be wanting to do that for quite some time, and also get rid of the colormaps argument and always just use color, for anything color related...

src/makielayout/layoutables/colorbar.jl
```diff
@@ -70,11 +70,26 @@ function layoutable(::Type{<:Colorbar}, fig_or_scene; bbox = nothing, kwargs...)
...
+                return colorrange
+            end
+        else
+            if !isnothing(colorrange)
+                error("Both colorrange + limits are set, please only set one, they're aliases. colorrange: $(colorrange), limits: $(limits)")
+            end
+            return limits
+        end
+    end
+
```
jkrumbiegel, line 92: this could be shortened to

```
if !any(isnothing, (limits, colorrange))
    error()
end
something(limits, colorrange, (0, 1))
```

SimonDanisch: Nice! Didn't know about `something`..

SimonDanisch:
I'm going to merge this and opened this issue to later decide on a consistent name: https://github.com/JuliaPlots/Makie.jl/issues/1106

PR #1068, opened 2021-06-25
only update axis limits on real changes

jkrumbiegel:


mschauer:
Could even use approximate equality?

SimonDanisch (approved):

PR #1086, opened 2021-06-30
Add stairs plot

sethaxen:
This PR adapts @jkrumbiegel's implementation of a stairs plot in https://gist.github.com/jkrumbiegel/529936a3a6f158715343601d9487880b.

Example:
```julia
f = Figure()
x = 1:10
y = 11:20
f[1, 1] = Axis(f; title="step=:pre")
stairs!(f[1, 1], x, y)
f[2, 1] = Axis(f; title="step=:post")
stairs!(f[2, 1], x, y; step=:post)
f[3, 1] = Axis(f; title="step=:center")
stairs!(f[3, 1], x, y; step=:center)
f
```
tmp


sethaxen:
Additional options we _could_ add:
- ability to style vertical and horizontal segments differently
- ability to provide a marker and marker style options to simultaneously scatter

SimonDanisch (approved):

lazarusA:
this other case will also be so useful... (histstep)

https://gist.github.com/jkrumbiegel/5baa32d0c010cb842e037603a18663cd

sethaxen:
Probably the simplest thing to do for `histstep` would be to factor the histogram creation code in `hist` into its own helper function, and then the difference between `hist` and `histstep` is only the attributes and whether it calls `barplot!` or `stairs!` internally.

PR #1094, opened 2021-07-05
Avoid same name for lower and upper

touste:
This change avoids an error where `lower` and `upper` defined below as Node (lines 89-90) are change to Point2f0, resulting in: `type Array has no field val`


SimonDanisch:
Thanks!

Issue #1102, opened 2021-07-09
CairoMakie implicit keyword syntax used


theogf:
The implicit keyword argument used here (`step`) is not compatible with Julia 1.3 (I forgot in which version it was introduced though) while CairoMakie's Project.toml assumes compatibility from 1.3

https://github.com/JuliaPlots/Makie.jl/blob/c8c973c22bc5df3f3897f575dc12b706633d53f1/CairoMakie/src/primitives.jl#L446

PR #1103, opened 2021-07-09
Remove implicit keyword syntax in `range`

theogf:
Solves #1102
(Caught this bug in https://github.com/theogf/Turkie.jl/pull/32/checks?check_run_id=3029084908)

EDIT: The implicit keyword syntax is from 1.5 : https://julialang.org/blog/2020/08/julia-1.5-highlights/#implicit_keyword_argument_values
EDIT:EDIT: I see CairoMakie is only tested from 1.5 actually :sweat_smile:


SimonDanisch:
thx

PR #1111, opened 2021-07-12
prerender with better highlight.js and ansicolor

jkrumbiegel:


jkrumbiegel:
I literally wanted to write you a message this second about those errors. Thanks I'll have a look if that's the case.

PR #1118, opened 2021-07-13
Conditional scene update

t-bltg:
This PR allows saving a figure without updating the current scene:
```julia
# save a zoomed figure
save("test.png", fig; update=false)

# changing the resolution with update=false also works
w, h = size(get_scene(fig))
save("test.png", fig; update=false, resolution=(2w, 2h))
```
Required when saving a zoomed figure.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -48,17 +48,15 @@ function backend_display(::Missing, ::Scene)
     """)
 end
 
-Base.display(fap::FigureAxisPlot) = display(fap.figure)
-Base.display(fig::Figure) = display(fig.scene)
+Base.display(fap::FigureAxisPlot; kw...) = display(fap.figure; kw...)
+Base.display(fig::Figure; kw...) = display(fig.scene; kw...)
 
-function Base.display(scene::Scene)
+function Base.display(scene::Scene; update=true)
 
     if !use_display[]
         return Core.invoke(display, Tuple{Any}, scene)
     else
-        # set update to true, without triggering an event
-        # this just indicates, that now we may update on e.g. resize
-        update!(scene)
+        update && update!(scene)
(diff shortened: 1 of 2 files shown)


src/display.jl
```diff
@@ -48,17 +48,17 @@ function backend_display(::Missing, ::Scene)
...
+Base.display(fig::Figure; kw...) = display(fig.scene; kw...)
 
-function Base.display(scene::Scene)
+function Base.display(scene::Scene; update=true)
 
     if !use_display[]
         return Core.invoke(display, Tuple{Any}, scene)
     else
         # set update to true, without triggering an event
         # this just indicates, that now we may update on e.g. resize
```
jkrumbiegel, line 60: is this still correct?

src/display.jl
```diff
@@ -209,6 +209,7 @@ Save a `Scene` with the specified filename and format.
 ## All Backends
 
 - `resolution`: `(width::Int, height::Int)` of the scene in dimensionless units (equivalent to `px` for GLMakie and WGLMakie).
+- `update`: Update the scene before saving.
```
jkrumbiegel, line 212: Maybe an explanation is needed what updating means (why would you not want it)

t-bltg: Update the scene and its children (`update_limits!` and `center!`) before saving. One might want to set `update=false` when saving a zoomed scene.

Something like that ?

src/display.jl
t-bltg: It think not since `update!` is in an event loop: https://github.com/JuliaPlots/Makie.jl/blob/33e600665a48f4a26924cdd3fd0c7471379f039d/src/scenes.jl#L115

Should we remove ?

t-bltg:
Thanks !

PR #1123, opened 2021-07-14
Update volumeslices - docs

t-bltg:
vol-slices

### DIFF
--- a/docs/src/plotting_functions/volumeslices.md
+++ b/docs/src/plotting_functions/volumeslices.md
@@ -39,6 +39,7 @@ set_close_to!(sl_yz, .5length(x))
 set_close_to!(sl_xz, .5length(y))
 set_close_to!(sl_xy, .5length(z))
 
-cam3d!(ax.scene, projectiontype=Makie.Orthographic)
+# cam3d!(ax.scene, projectiontype=Makie.Orthographic)
+
 fig
 ```
--- a/docs/src/plotting_functions/volumeslices.md
+++ b/docs/src/plotting_functions/volumeslices.md
@@ -1,0 +1,44 @@
+# volumeslices
+
+```@docs
+volumeslices
+```
+
+### Examples
+
+```@example
+using GLMakie
+GLMakie.activate!() # hide
+Makie.inline!(true) # hide
+
+fig = Figure()
+ax = LScene(fig[1, 1], scenekw=(show_axis=false,))
+
+x = LinRange(0, π, 50)
+y = LinRange(0, 2π, 100)
+z = LinRange(0, 3π, 150)
+
+lsgrid = labelslidergrid!(
+  fig,
+  ["yz plane - x axis", "xz plane - y axis", "xy plane - z axis"],
+  [1:length(x), 1:length(y), 1:length(z)]
+)
+fig[2, 1] = lsgrid.layout
+
+vol = [cos(X)*sin(Y)*sin(Z) for X ∈ x, Y ∈ y, Z ∈ z]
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Cool, thank you! :)  
This was just a random recipe that I added at some point, but I guess it's a useful enough for us to maintain it :) 

image
This actually looks wrong - those circles should line up...

t-bltg:
> This was just a random recipe that I added at some point, but I guess it's a useful enough for us to maintain it :)
I think this is pretty useful for visualizing volumetric data :)

> This actually looks wrong - those circles should line up...
```julia
x = LinRange(0, π, 50)
y = LinRange(0, 2π, 100)
z = LinRange(0, 3π, 150)
vol = [cos(X)*sin(Y)*sin(Z) for X ∈ x, Y ∈ y, Z ∈ z]
```
That does seem right to me, doesn't it ?

SimonDanisch:
The input isn't wrong, the planes are flipped ;)

t-bltg:
> The input isn't wrong, the planes are flipped ;)

That still looks right to me, compared to:
```julia
using GLMakie

main() = begin
  fig = Figure(); axes = [Axis(fig[i, 1]) for i ∈ 1:3]

  x = LinRange(0, π, 50)
  y = LinRange(0, 2π, 100)
  z = LinRange(0, 3π, 150)
  v = [cos(X)*sin(Y)*sin(Z) for X ∈ x, Y ∈ y, Z ∈ z]

  ax = axes[1, 1]
  heatmap!(ax, y, z, v[length(x)÷2, :, :])
  ax.aspect = DataAspect()
  ax.title = "yz plane - x axis"
  ax.xlabel = "y"
  ax.ylabel = "z"
  ax = axes[2, 1]
  heatmap!(ax, x, z, v[:, length(y)÷2, :])
  ax.aspect = DataAspect()
  ax.title = "xz plane - y axis"
  ax.xlabel = "x"
  ax.ylabel = "z"
  ax = axes[3, 1]
  heatmap!(ax, x, y, v[:, :, length(z)÷2])
  ax.aspect = DataAspect()
  ax.title = "xy plane - z axis"
  ax.xlabel = "x"
  ax.ylabel = "y"

  save("heatmap.png", fig)
  return
end

main()
```
heatmap

With the `show_axis=true`:
volslices

SimonDanisch:
Where the planes meet, they need to go through the same points in the volume, so they need to have have the same color where they meet... That's quite clearly not happening:
image
That happens either because they're flipped, or because the colorrange is different, so the color gets mapped differently.
The colorrange should actually be global and passed to the heatmap calls, and default to `extrema(volume)`.

SimonDanisch:
This looks correct:
image
So I guess it's the z plane that's wrong

t-bltg:
> Where the planes meet, they need to go through the same points in the volume, so they need to have have the same color where they meet... That's quite clearly not happening

It depends on the location though:
ss-1

:thinking:

t-bltg:
> This looks correct:
> image
> So I guess it's the xy plane that's wrong

That seems to fix the issue `hattributes[:colorrange] = plot[:colorrange][]`.

SimonDanisch:
Yeah been there ;) these cos/sin plots are full of symmetries, so there are lots of configurations where they look correct even when wrong!

t-bltg:
> Yeah been there ;) these cos/sin plots are full of symmetries, so there are lots of configurations where they look correct even when wrong!

Harmonic functions are really useful though :wink:

t-bltg:
New output:

sl

PR #1143, opened 2021-07-18
Docs - Specify how to add a title to a LScene

t-bltg:


SimonDanisch:
I'll turned it into an actual example to make sure that it works and people can see how it looks!

Issue #1142, opened 2021-07-18
New cam3d! and scale! issue


t-bltg:
Since `Makie@0.15.0` the example at https://github.com/JuliaPlots/Makie.jl/issues/1115#issuecomment-879070279 stopped working:

1. `scale!` doesn't scale the axis labels and ticks anymore:
![bug](https://user-images.githubusercontent.com/13423344/126064293-acf0956d-771d-471d-8c5e-f8ba0d4239ad.png)

2. new `cam3d!` with `projectiontype=Makie.Orthographic` behaves badly (without the `scale!` in the previous example): I'm trying to un-zoom and drag the scene but it obviously fails (whereas using `old_cam3d!` or `projectiontype=Makie.Perspective` is OK):
![bug](https://user-images.githubusercontent.com/13423344/126064805-a916f321-86dd-46b2-b291-47339fc71715.gif)

t-bltg:
Replacing https://github.com/JuliaPlots/Makie.jl/blob/d5eb92de73fa3f014abad4ca22102e5af789f676/src/camera/camera3d.jl#L571
with `camera.zoom_mult[] = norm(middle - neweyepos)` fixes at least the zoom issue.

PR #1062, opened 2021-06-23
Ffreyer camera

SimonDanisch:
https://github.com/JuliaPlots/Makie.jl/pull/1024 without merge conflicts and old_cam3d! fix

PR #1144, opened 2021-07-18
Fix projection matrix computation when using an orthographic projection

t-bltg:
Partial correction for https://github.com/JuliaPlots/Makie.jl/issues/1142.

### DIFF
--- a/ReferenceTests/src/tests/examples3d.jl
+++ b/ReferenceTests/src/tests/examples3d.jl
@@ -40,15 +40,15 @@ end
     dir_scaled = Vec3f0(
         dir[1] * scene.transformation.scale[][1],
         0.0,
-        dir[3] * scene.transformation.scale[][2],
+        dir[3] * scene.transformation.scale[][3],
     )
     cam.upvector[] = (0.0, 0.0, 1.0)
     cam.lookat[] = minimum(scene_limits(scene)) + dir_scaled
     cam.eyeposition[] = (cam.lookat[][1], cam.lookat[][2] + 6.3, cam.lookat[][3])
     cam.attributes[:projectiontype][] = Makie.Orthographic
-    cam.zoom_mult[] = 0.61f0
+    cam.zoom_mult[] = .097
     update_cam!(scene, cam)
-    # stop scene display from centering, which would overwrite the camera paramter we just set
+    # stop scene display from centering, which would overwrite the camera parameter we just set
     scene.center = false
     fig
 end
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -418,6 +418,7 @@ function translate_cam!(scene, cam, t)
 
     cam.eyeposition[] = eyepos + trans
     cam.lookat[] = lookat + trans
+    update_cam!(scene, cam)
     return
 end
 
@@ -480,13 +481,14 @@ function rotate_cam!(scene, cam::Camera3D, angles::VecTypes, from_mouse=false)
     else
         cam.lookat[] = eyepos + viewdir
     end
+    update_cam!(scene, cam)
     return
 end
 
-zoom!(scene::Scene, zoom_step) = zoom!(scene, cameracontrols(scene), false, false)
+zoom!(scene::Scene, zoom_step) = zoom!(scene, cameracontrols(scene), zoom_step, false, false)
 function zoom!(scene::Scene, cam::Camera3D, zoom_step, shift_lookat = false, cad = false)
     if cad
-        # move exeposition if mouse is not over the center
+        # move eyeposition if mouse is not over the center
         lookat = cam.lookat[]
         eyepos = cam.eyeposition[]
         up = cam.upvector[]         # +y
@@ -520,16 +522,15 @@ function zoom!(scene::Scene, cam::Camera3D, zoom_step, shift_lookat = false, cad
     end
 
     # apply zoom
-    cam.zoom_mult[] = cam.zoom_mult[] * zoom_step
-
+    cam.zoom_mult[] *= zoom_step
+    update_cam!(scene, cam)
     return
 end
 
 
 function update_cam!(scene::Scene, cam::Camera3D)
-    @extractvalue cam (lookat, eyeposition, upvector)
+    @extractvalue cam (lookat, eyeposition, upvector, near, far)
 
-    near = cam.near[]; far = cam.far[]
     aspect = Float32((/)(widths(scene.px_area[])...))
 
     if cam.attributes[:projectiontype][] == Perspective
@@ -537,8 +538,9 @@ function update_cam!(scene::Scene, cam::Camera3D)
         cam.fov[] = fov
         proj = perspectiveprojection(fov, aspect, near, far)
     else
-        w = 0.5f0 * (1f0 + aspect) * cam.zoom_mult[]
-        h = 0.5f0 * (1f0 + 1f0 / aspect) * cam.zoom_mult[]
+        distance = norm(lookat - eyeposition)
+        w = 0.5f0 * (1f0 + aspect) * cam.zoom_mult[] * distance
+        h = 0.5f0 * (1f0 + 1f0 / aspect) * cam.zoom_mult[] * distance
         proj = orthographicprojection(-w, w, -h, h, near, far)
     end
 
@@ -547,7 +549,7 @@ function update_cam!(scene::Scene, cam::Camera3D)
     scene.camera.projection[] = proj
     scene.camera.view[] = view
     scene.camera.projectionview[] = proj * view
-    scene.camera.eyeposition[] = cam.eyeposition[]
+    scene.camera.eyeposition[] = eyeposition
 end
 
 function update_cam!(scene::Scene, camera::Camera3D, area3d::Rect)
@@ -555,11 +557,11 @@ function update_cam!(scene::Scene, camera::Camera3D, area3d::Rect)
     bb = FRect3D(area3d)
     width = widths(bb)
     half_width = width/2f0
-    lower_corner = minimum(bb)
     middle = maximum(bb) - half_width
     old_dir = normalize(eyeposition .- lookat)
     camera.lookat[] = middle
-    neweyepos = middle .+ (1.2*norm(width) .* old_dir)
+    fact = (camera.attributes[:projectiontype][] == Perspective) ? 1.2 : .5
+    neweyepos = middle .+ (fact * norm(width) .* old_dir)
     camera.eyeposition[] = neweyepos
     camera.upvector[] = Vec3f0(0,0,1)
     if camera.attributes[:near][] === automatic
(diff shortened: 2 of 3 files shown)


t-bltg:
`examples3d_20_OrthographicCamera` failed, I'm on it.

t-bltg:
@SimonDanisch, running the test locally produces:

**new**
examples3d_20_OrthographicCamera

**ref**
examples3d_20_OrthographicCamera

which shows imho a difference in colors, and not a regression.

Where can I submit a PR to change the ref image ?

ffreyer:
This is probably a lighting thing. By default the light source is at `scene.camera.eyeposition`, so you should get maximally bright shading. If that's not the case anymore `camera.eyeposition` is either out of sync with Camera3D's `eyeposition` or both are out of sync with where the camera actually is.

ReferenceTests/src/tests/examples3d.jl
```diff
@@ -40,15 +40,15 @@ end
     dir_scaled = Vec3f0(
         dir[1] * scene.transformation.scale[][1],
         0.0,
-        dir[3] * scene.transformation.scale[][2],
+        dir[3] * scene.transformation.scale[][3],
     )
     cam.upvector[] = (0.0, 0.0, 1.0)
     cam.lookat[] = minimum(scene_limits(scene)) + dir_scaled
-    cam.eyeposition[] = (cam.lookat[][1], cam.lookat[][2] + 6.3, cam.lookat[][3])
```
ffreyer, line 47: Ah actually it's probably just because the eyeposition was moved closer here. But I realize now that the orthographic camera should adjust `eyeposition` on zoom...

t-bltg: Ok, I've restored the previous setting and changed `zoom_mult` instead.

t-bltg:
@SimonDanisch the CI seems broken, and unrelated to this PR:

```
 text_252_latexstrings.png: Test Failed at /home/runner/work/Makie.jl/Makie.jl/ReferenceTests/src/runtests.jl:106
  Expression: score < difference
   Evaluated: 0.037494160234979264 < 0.03
```

t-bltg:
I fixed the default eyeposition for Orthopragphic projection:

```julia
using GLMakie
using Makie

main(proj) = begin
  xs = LinRange(0, 10, 100)
  ys = LinRange(0, 15, 100)
  zs = [cos(x)*sin(y) for x in xs, y in ys]

  fig, ax, p = fap = surface(xs, ys, zs)
  cam3d!(ax.scene, projectiontype=proj)

  save("sf-$proj.png", fap)
  display(fap)
  return
end

main(Makie.Orthographic)
main(Makie.Perspective)
```

Orthographic
---------
<kbd>

</kbd>

Perspective
---------
<kbd>

</kbd>

src/camera/camera3d.jl
```diff
@@ -418,6 +418,7 @@ function translate_cam!(scene, cam, t)
 
     cam.eyeposition[] = eyepos + trans
     cam.lookat[] = lookat + trans
+    update_cam!(scene, cam)
```
ffreyer, line 421: These will run twice on interaction if you put them there - once in `on_pulse` or `on(mouseevent)` and once in these functions. I originally put them in `on_pulse` so I could merge rotations translations and zoom updates when they happen simultaneously. It's probably not a big deal to move these, but we should at least avoid duplicate calls.

t-bltg: The problem is that if triggered programmatically using e.g. `translate_cam!` then `update_cam!` is not called.

ffreyer:
> This is probably a lighting thing. By default the light source is at `scene.camera.eyeposition`, so you should get maximally bright shading. If that's not the case anymore `camera.eyeposition` is either out of sync with Camera3D's `eyeposition` or both are out of sync with where the camera actually is.

I thought a bit more about this - with orthographic projection there is no clear position for `eyeposition` because depth has no effect. There is no difference between `eyeposition = lookat - normalize(viewdir)` and `eyeposition = lookat - 1000 * normalize(viewdir)` when it comes to the camera. Well I suppose there would be after this pr, but it's artificial. (Which doesn't mean we shouldn't have it. If it makes things easier to use I'm all for it.)

t-bltg:
> > This is probably a lighting thing. By default the light source is at `scene.camera.eyeposition`, so you should get maximally bright shading. If that's not the case anymore `camera.eyeposition` is either out of sync with Camera3D's `eyeposition` or both are out of sync with where the camera actually is.
> 
> I thought a bit more about this - with orthographic projection there is no clear position for `eyeposition` because depth has no effect. There is no difference between `eyeposition = lookat - normalize(viewdir)` and `eyeposition = lookat - 1000 * normalize(viewdir)` when it comes to the camera. Well I suppose there would be after this pr, but it's artificial. (Which doesn't mean we shouldn't have it. If it makes things easier to use I'm all for it.)

While I trust you on the maths of an orthographic projection, I disagree, there is an effect on the camera.

It I restore back to:
```
# restore previous behavior
w = 0.5f0 * (1f0 + aspect) * zoom_mult
h = 0.5f0 * (1f0 + 1f0 / aspect) * zoom_mult

[...]

# only update eyeposition on Perspective projection
if camera.attributes[:projectiontype][] == Perspective
    camera.eyeposition[] = middle .+ (1.2 * norm(width) .* old_dir)
end
```

I obtain this:
out

Please provide a reproducible example.

ffreyer:
```julia
fig, ax, p = scatter(rand(Point3f0, 100))
cam = Camera3D(ax.scene, projectiontype = Makie.Orthographic)
```

And then try various factors in

```julia
update_cam!(ax.scene, cam, cam.lookat[] - 10 * normalize(cam.lookat[] - cam.eyeposition[]), cam.lookat[], cam.upvector[])
```

This shouldn't change the plot at all as long as it's between near and far clip. Clipping happens around 1 and 100 for me, between that nothing happens.

Conceptually the `view` matrix should just position the camera. Move it to `eyeposition` and rotate it so that `upvector` is upwards and `lookat` is what the camera is pointed at. The projection matrix does the scaling + translation to set up the corrdinate space that will be rendered. The orthographic projection scales x, y, z independently, so it doesn't matter for x and y what z is - there is no depth scaling. If the matrix always scales x by 1 it doesn't matter if an object is 1 unit or 100 units away, assuming they're not clipped. Perspective projection on the other hand includes scaling of x and y based on z, so an object 1 unit away might be scaled by a factor of 10 and an object 100 units away by 1.

But I think this is getting a bit off-topic. This was meant to fix translation speed, right? I think the goal should be that if you start a translation with a point under your cursor, it should still be under the cursor when you finish, regardless of zoom etc for an orthographic camera. As far as I can tell we're not there yet?

t-bltg:
> ```julia
> fig, ax, p = scatter(rand(Point3f0, 100))
> cam = Camera3D(ax.scene, projectiontype = Makie.Orthographic)
> ```
> 
> And then try various factors in
> 
> ```julia
> update_cam!(ax.scene, cam, cam.lookat[] - 10 * normalize(cam.lookat[] - cam.eyeposition[]), cam.lookat[], cam.upvector[])
> ```
> 
> This shouldn't change the plot at all as long as it's between near and far clip. Clipping happens around 1 and 100 for me, between that nothing happens.

I will try that now.

> Conceptually the `view` matrix should just position the camera. Move it to `eyeposition` and rotate it so that `upvector` is upwards and `lookat` is what the camera is pointed at. The projection matrix does the scaling + translation to set up the corrdinate space that will be rendered. The orthographic projection scales x, y, z independently, so it doesn't matter for x and y what z is - there is no depth scaling. If the matrix always scales x by 1 it doesn't matter if an object is 1 unit or 100 units away, assuming they're not clipped. Perspective projection on the other hand includes scaling of x and y based on z, so an object 1 unit away might be scaled by a factor of 10 and an object 100 units away by 1.

Thanks for the detailed explanation.

> But I think this is getting a bit off-topic. This was meant to fix translation speed, right?

This PR was originally aimed at fixing the second issue mentioned in https://github.com/JuliaPlots/Makie.jl/issues/1142. I understand that I can use `old_cam3d!` to restore previous behavior, but I prefer to understand why the new `cam3d!` or `Camera3D` that you wrote in https://github.com/JuliaPlots/Makie.jl/pull/1062 is failing this particular [mwe](https://github.com/JuliaPlots/Makie.jl/issues/1115#issuecomment-879070279).

ffreyer:
> > But I think this is getting a bit off-topic. This was meant to fix translation speed, right?
> 
> This PR was originally aimed at fixing the second issue mentioned in #1142. I understand that I can use `old_cam3d!` to restore previous behavior, but I prefer to understand why the new `cam3d!` or `Camera3D` that you wrote in #1062 is failing this particular [mwe](https://github.com/JuliaPlots/Makie.jl/issues/1115#issuecomment-879070279).

Oh so this is about translations going wild on zoom? That's the `shift_lookat` branch in `zoom!`. Seems like I tunneled on getting perspective projection to work and forgot about orthographic entirely... If you want to fix that you should try splitting that into an `Orthographic` branch and a `Perspective` branch. You can also just turn off those translations with `cam3d!(..., zoom_shift_lookat = false)`.

I am trying to get orthographic drag-translations to be exact atm, I can look into getting the zoom-shift to work too.

t-bltg:
Ok so reverting this PR and using
```
using Random; Random.seed!(1234)
using LinearAlgebra
using GLMakie

main() = begin
  fig, ax, p = scatter(rand(Point3f0, 100))
  cam = Camera3D(ax.scene, projectiontype=Makie.Orthographic)

  for fact ∈ (2, 10, 50, 100, 200)
    eyeposition = cam.lookat[] - fact * normalize(cam.lookat[] - cam.eyeposition[])
    @show eyeposition
    update_cam!(ax.scene, cam, eyeposition, cam.lookat[], cam.upvector[])

    # important to set update=false, else `update_cam!(scene::Scene, camera::Camera3D, area3d::Rect)` resets the camera
    save("cam-$fact.png", fig; update=false)
  end

  display(fig; update=false)
  return
end

if @__FILE__() ∈ ("string", abspath(PROGRAM_FILE))
  main()
end
```

does not modify the output figures, apart from the clipping that you mentioned.

> Oh so this is about translations going wild on zoom? That's the `shift_lookat` branch in `zoom!`. Seems like I tunneled on getting perspective projection to work and forgot about orthographic entirely... If you want to fix that you should try splitting that into an `Orthographic` branch and a `Perspective` branch. You can also just turn off those translations with `cam3d!(..., zoom_shift_lookat = false)`.
> 

Thanks, I will try that.

> If you want to fix that you should try splitting that into an `Orthographic` branch and a `Perspective` branch.
Ok, will modify the PR it that works :wink:

t-bltg:
> Oh so this is about translations going wild on zoom? That's the `shift_lookat` branch in `zoom!`. Seems like I tunneled on getting perspective projection to work and forgot about orthographic entirely... If you want to fix that you should try splitting that into an `Orthographic` branch and a `Perspective` branch. You can also just turn off those translations with `cam3d!(..., zoom_shift_lookat = false)`.

Ok, so `zoom_shift_lookat=false` works on getting data visible.

> I am trying to get orthographic drag-translations to be exact atm, I can look into getting the zoom-shift to work too.

Yes, drag under `orthographic` proj is a bit too fast ...

t-bltg:
Superseded by https://github.com/JuliaPlots/Makie.jl/pull/1159.

PR #1146, opened 2021-07-18
Test: skip downloading ref imgs tar if env var REUSE_IMAGES_TAR is set to 1

t-bltg:
a.k.a. "spare my bandwidth"

(and `Base.download` replaced (deprecated))


SimonDanisch:
Should we pair this with actually caching the images in the github action?

t-bltg:
> Should we pair this with actually caching the images in the github action?

It think it's ok if we cleanup and extract all the time, it's not a `700Mb` archive. I didn't want to propagate a flag all the way down to the test, I think that using an env var is reasonable here.

PR #1150, opened 2021-07-20
make hist more flexible

SimonDanisch:
With `scale_to`, `offset`, `fillto` and `flip`, one can put multiple histograms into the same plot:

```@example hist
fig = Figure()
ax = Axis(fig[1, 1])
for i in 1:5
     hist!(ax, randn(1000), scale_to=0.6, offset=i, fillto=i, direction=:x, flip=true)
end
fig
```

image


SimonDanisch:
@piever @jkrumbiegel do you have any concerns about this, or can we merge?

src/stats/hist.jl
```diff
@@ -72,12 +82,19 @@ function Makie.plot!(plot::Hist)
...
-    points = lift(edges, plot.normalization) do edges, normalization
+    points = lift(edges, plot.normalization, plot.offset, plot.scale_to, plot.flip) do edges, normalization, offset, scale_to, flip
         h = StatsBase.fit(StatsBase.Histogram, values[], edges)
         h_norm = StatsBase.normalize(h, mode = normalization)
         centers = edges[1:end-1] .+ (diff(edges) ./ 2)
         weights = h_norm.weights
-        return Point2f0.(centers, weights)
+        if !isnothing(scale_to)
+            min, max = extrema(weights)
+            weights .= ((weights .- min) ./ (max - min)) .* scale_to
```
piever, line 92: This seems dangerous to me. Have you tried setting `scale` on `hist(rand(1_000_000))` with the current implementation?

I think `weigths .= weights ./ max .* scale_to` is more reasonable. Morally, you always have weight `0` outside your data range anyways, so `min` "should" always be `0`. 

In general, I think the bar length should be proportional to the number of entries in that bin, which you respect with `weigths .= weights ./ max .* scale_to`, but not with the current implementation.

src/stats/hist.jl
```diff
@@ -28,6 +28,13 @@ can be normalized by setting `normalization`. Possible values are:
    norm 1.
 *  `:none`: Do not normalize.
 
+The following attributes can move the histogram around,
+which comes in handy when placing multiple histograms into one plot:
+* offset = 0.0: adds an offset to every value
+* fillto = 0.0: defines where the bar starts
```
piever, line 34: I'm not too sure I like this API, in that for things to work you always need to set both to the same value. Wouldn't it be easier to just add `offset` to `barplot`, which would make the bar start at `offset` but (unlike `fillto`) have length `ys`? That way, users only need to set `offset` in histograms.

If we are unsure that `offset` makes sense for `barplot`, we can add it only for `hist`, but I think "shift bar while preserving the length" is a clearer meaning.

piever (changes requested):

piever:
Nice! I've added a couple of comments, based on the observation that the API should try to respect as much as possible the proportionality between bar length and number of entries in bin.

jkrumbiegel:
We wanted to find a consistent "interface" for these flipping keywords. (Also because AlgebraOfGraphics should be able to use this information). We currently have `direction` and `orientation`, this would add `flip`.

piever:
> We currently have `direction` and `orientation`, this would add `flip`.

IIUC, `flip` here is not x vs y, but rather "bars go towards the right vs left (or down vs up)". That being said, if we wanted to use `flip` for "flip x and y axis" in the future, this use here would conflict.

src/stats/hist.jl
SimonDanisch: Oh good catch! That's absolutely wrong :D

SimonDanisch:
Alright, I removed `flip` since one can just do that via `scale_to=-x`.
I also moved `offset` to the barplot recipe, and made it set `fillto` if `fillto` is automatic.
So now the example is a bit simpler:
```julia
fig = Figure()
ax = Axis(fig[1, 1])
for i in 1:5
     hist!(ax, randn(1000), scale_to=-0.6, offset=i, direction=:x)
end
fig
```
image

src/basic_recipes/barplot.jl
```diff
@@ -208,7 +209,7 @@ function Makie.plot!(p::BarPlot)
 
         if stack === automatic
             if fillto === automatic
-                fillto = 0.0
+                fillto = offset
             end
         elseif eltype(stack) <: Integer
             fillto === automatic || @warn "Ignore keyword fillto when keyword stack is provided"
```
piever, line 215:
```suggestion
            if !iszero(offset)
                @warn "Ignore keyword offset when keyword stack is provided"
                offset = 0.0
            end
```

I think in the future we should support `offset` in combination with `stack` (just move the whole stacked bar up or down), but the current implementation doesn't work, so I think it's best to just throw a warning and remove the offset in that case. We can add the feature in a future PR.

piever (approved):

piever:
LGTM! I've just added a small comment to disable `offset` when `stack` is also given. We may want to implement `offset` + `stack` in the future, but that doesn't need to happen in this PR.

src/basic_recipes/barplot.jl
piever: Actually, I wonder if just changing line 219 to `y, fillto = to, from .+ offset` is enough to get `offset` and `stack` to play nice together, maybe that's the best solution.

PR #1158, opened 2021-07-21
Modify drag speed when using Orthographic projection

t-bltg:
Addresses https://github.com/JuliaPlots/Makie.jl/pull/1144#issuecomment-884135316.


ffreyer:
There's a bit more than just the factor wrong here. #1159 should fix it

PR #1160, opened 2021-07-21
Use MathTeXEngine support for latex-tex mixing

Kolaru:
I added support for mixing text and latex in MathTeXEngine, and this makes use of it.


SimonDanisch:
Great, thanks! :)

PR #1162, opened 2021-07-22
Fix edge case in z vector conversion for SurfaceLike

### LINKED ISSUE #1161: Segfault in surface call
The following code gives a segfault with Makie 0.15 and Julia 1.6.1 (also reproduced with 1.7-beta2):

```julia
using Makie

xs = range(150, 160, length=100)
ys = range(7, 9, length=100)
points = [(x, y, 0.0) for x in xs for y in ys]
x, y, z = map(i -> getindex.(points, i), 1:3)

flags = (153 .< x .< 153.5) .& (7 .< y .< 7.3)
surface(y[flags], x[flags], z[flags]);
```

Starting Julia with `--check-bounds=yes` gives the following error:

```
ERROR: BoundsError: attempt to access 14×5 Matrix{Float32} at index [1, 6]
Stacktrace:
  [1] setindex!
    @ ./array.jl:841 [inlined]
  [2] (::Makie.var"#180#182"{Matrix{Float32}, Vector{Float32}, Vector{Float32}})(::Tuple{Float64, Float64, Float64})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/conversions.jl:335
  [3] foreach(f::Makie.var"#180#182"{Matrix{Float32}, Vector{Float32}, Vector{Float32}}, itr::Base.Iterators.Zip{Tuple{Vector{Float64}, Vector{Float64}, Vector{Float64}}})
    @ Base ./abstractarray.jl:2141
  [4] convert_arguments(SL::ContinuousSurface, x::Vector{Float64}, y::Vector{Float64}, z::Vector{Float64})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/conversions.jl:332
  [5] convert_arguments(::Type{Surface{Tuple{Vector{Float64}, Vector{Float64}, Vector{Float64}}}}, ::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N; kw::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/conversions.jl:10
  [6] convert_arguments(::Type{Surface{Tuple{Vector{Float64}, Vector{Float64}, Vector{Float64}}}}, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/conversions.jl:8
  [7] plot!(::Scene, ::Type{Surface{ArgType} where ArgType}, ::Attributes, ::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N; kw_attributes::Base.Iterators.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:show_axis,), Tuple{Bool}}})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/interfaces.jl:321
  [8] plot(::Type{Surface{ArgType} where ArgType}, ::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N; axis::NamedTuple{(), Tuple{}}, figure::NamedTuple{(), Tuple{}}, kw_attributes::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/.julia/packages/Makie/xbI6d/src/figureplotting.jl:28
  [9] plot(::Type{Surface{ArgType} where ArgType}, ::Vector{Float64}, ::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N)
    @ Makie ~/.julia/packages/Makie/xbI6d/src/figureplotting.jl:18
 [10] surface(::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N; attributes::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ MakieCore ~/.julia/packages/MakieCore/S8PkO/src/recipes.jl:31
 [11] surface(::Vector{Float64}, ::Vararg{Vector{Float64}, N} where N)
    @ MakieCore ~/.julia/packages/MakieCore/S8PkO/src/recipes.jl:31
 [12] top-level scope
    @ REPL[15]:1
```

(I don't know if this use of `surface` is supposed to work: the documentation only mentions a matrix for z while here x,y,z are all vectors of the same length. In other cases I did get a surface, although an ugly one with many holes, for examples when using all points instead of the subset selected with `flags`.)

### LINKED ISSUE #1163: SurfaceLike plots looking weird when giving a vector for z
Giving same-length vectors for x,y,z is [supported](https://makie.juliaplots.org/stable/plotting_functions/heatmap.html#Three-vectors) at least for `heatmap`, but the result sometimes look weird, with holes.

Compare this:

```julia
using CairoMakie

xs = LinRange(0, 10, 100)
ys = LinRange(0, 15, 50)
zs = [cos(x) * sin(y) for x in xs, y in ys]
heatmap(xs, ys, zs, figure=(resolution=(200, 150),))
```

image

with this:

```julia
points = [(x, y, cos(x) * sin(y)) for x in xs for y in ys]
x, y, z = map(i -> getindex.(points, i), 1:3)
heatmap(x, y, z, figure=(resolution=(200, 150),))
```

image

The problem also affects `surface` and `contour`:

|`contour`|`surface`|
|-|-|
|```julia
points = [(x, y, cos(x) * sin(y)) for x in xs for y in ys]
x, y, z = map(i -> getindex.(points, i), 1:3)
heatmap(x, y, z, figure=(resolution=(200, 150),))
``` | ![image](https://user-images.githubusercontent.com/2412819/126657218-13908dc5-3aa8-43e4-9841-2dc34aaf586d.png)
|

### PR DESCRIPTION
**Pull request by knuesel**
Closes #1161

Closes #1163

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -297,7 +297,7 @@ end
 """
     convert_arguments(P, Matrix)::Tuple{ClosedInterval, ClosedInterval, Matrix}
 
-Takes an `AbstractMatrix`, converts the dimesions `n` and `m` into `ClosedInterval`,
+Takes an `AbstractMatrix`, converts the dimensions `n` and `m` into `ClosedInterval`,
 and stores the `ClosedInterval` to `n` and `m`, plus the original matrix in a Tuple.
 
 `P` is the plot Type (it is optional).
@@ -317,6 +317,9 @@ function convert_arguments(SL::SurfaceLike, x::AbstractVector{<:Number}, y::Abst
         error("x, y and z need to have the same length. Lengths are $(length.((x, y, z)))")
     end
 
+    # Convert to Float32 before checking for uniqueness
+    x, y, z = map(el32convert, (x, y, z))
+
     xys = tuple.(x, y)
     if length(unique(xys)) != length(x)
         c = StatsBase.countmap(xys)
@@ -324,10 +327,10 @@ function convert_arguments(SL::SurfaceLike, x::AbstractVector{<:Number}, y::Abst
         error("Found duplicate x/y coordinates: $cdup")
     end
 
-    xs = Float32.(sort(unique(x)))
+    xs = sort(unique(x))
     any(isnan, xs) && error("x must not have NaN values.")
-    ys = Float32.(sort(unique(y)))
-    any(isnan, ys) && error("x must not have NaN values.")
+    ys = sort(unique(y))
+    any(isnan, ys) && error("y must not have NaN values.")
     zs = fill(NaN32, length(xs), length(ys))
     foreach(zip(x, y, z)) do (xi, yi, zi)
         i = searchsortedfirst(xs, xi)
--- a/test/conversions.jl
+++ b/test/conversions.jl
@@ -205,4 +205,10 @@ end
     @test_throws ErrorException convert_arguments(Heatmap, x, y, z)
     x = [NaN, 1, 2]
     @test_throws ErrorException convert_arguments(Heatmap, x, y, z)
+
+    # https://github.com/JuliaPlots/Makie.jl/issues/1161
+    x = [2, 1, 2];
+    y = [2, 2, nextfloat(3.0)];
+    z = [1, 2, 3];
+    @test isequal(convert_arguments(Surface, x, y, z), ([1,2], [2,3], [2 NaN; 1 3]))
 end

### CODE REVIEW

**Comment by knuesel** [comment]
I just realized this *does* fix the weird looking plots 🙂.

Then it might make sense to document that "vector for z" is supported for `surface`, `contour` and other `SurfaceLike` plots... maybe in another PR?

File: src/conversions.jl [hunk 1]
```diff
@@ -317,17 +317,20 @@ function convert_arguments(SL::SurfaceLike, x::AbstractVector{<:Number}, y::Abst
         error("x, y and z need to have the same length. Lengths are $(length.((x, y, z)))")
     end
 
+    # Convert to Float32 before checking for uniqueness
+    x, y, z = map(el32convert, (x, y, z))
```
jkrumbiegel, line 321: I think this doesn't need to include z, right? Every element from z will be converted to Float32 anyway when assigning it to zs later.

jkrumbiegel (changes requested):
I think we should do this a bit differently, now that I'm aware of the Float32 issue. I suggest we take the original unique elements, then convert that vector to float32 and check if all elements are still unique, because the behavior will be really weird if some of the z values are binned together. So in that case we should error I think, saying that Float32 precision is not enough to represent the chosen bins. This should happen for all conversions to Float32 ranges I think, not just in this special case.

knuesel:
@jkrumbiegel  that makes sense! at least for `SurfaceLike` and `VolumeLike`... I'm happy to try and do this properly but I'd need to better understand the flow of conversions:

1. Some methods return values without conversion to Float32, e.g.:

   ```julia
   convert_arguments(::SurfaceLike, x::AbstractMatrix, y::AbstractMatrix) = (x, y, zeros(size(y)))
   
   function convert_arguments(P::SurfaceLike, x::ClosedInterval, y::ClosedInterval, z::AbstractMatrix)
       convert_arguments(P, to_linspace(x, size(z, 1)), to_linspace(y, size(z, 2)), z)
   end
   ```

2. Some methods don't return vectors, e.g.:

   ```julia
   function convert_arguments(::SurfaceLike, data::AbstractMatrix)
       n, m = Float32.(size(data))
       (0f0 .. n, 0f0 .. m, el32convert(data))
   end
   
   function convert_arguments(::DiscreteSurface, data::AbstractMatrix)
       n, m = Float32.(size(data))
       (0.5f0 .. n+0.5f0, 0.5f0 .. m+0.5f0, el32convert(data))
   end
   ```

(These are also subject to the issue of Float64 values merging into the same Float32.)

I guess eventually all these values get converted to vectors of Float32? Could you outline how that happens for these cases?

src/conversions.jl
knuesel: Indeed!

jkrumbiegel:
i've fixed this for now in the pr linked above, the issue what happens with merging bins is punted to the other conversion methods though

PR #1177, opened 2021-07-29
Minor typos in the recipes docs

theogf:
Here I come again with minor fixes no one care about :laughing: 
`convert_single_argument` appears as convert _single_ argument without the ticks


jkrumbiegel:
I care! Thanks :) I've actually just fixed the same thing on my Franklin branch but that will take a bit more time before it's ready.

theogf:
Ah nice! Feel free to close this PR then :)

Issue #1180, opened 2021-07-29
Error when showing surface in 2D using GLMakie


wkearn:
I am basically trying to plot heatmaps on curvilinear grids. If there is a better way to do this now, please let me know, but prior to GLMakie v0.4.3, I was able to use `surface` as follows

```julia
N = 100
r = range(1.0,6.0,length=N)
θ = range(0,2π,length=N)

X = [r * cos(θ) for r in r, θ in θ]
Y = [r * sin(θ) for r in r, θ in θ]

Z = zeros(N,N)
W = [sin(θ) for r in r, θ in θ]

surface(X,Y,Z,color=W,shading=false)
```

As of GLMakie v0.4.3, this errors with the following error

```julia
ERROR: LoadError: Dimensionality: 4, not supported for OpenGL texture
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] default_texturetype(ndim::Int64)
    @ GLMakie.GLAbstraction C:\Users\wkearney\.julia\packages\GLMakie\rXGL8\src\GLAbstraction\GLTexture.jl:450
  [3] GLMakie.GLAbstraction.Texture(image::Array{Float32, 4}; kw_args::Base.Iterators.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:minfilter,), Tuple{Symbol}}})
    @ GLMakie.GLAbstraction C:\Users\wkearney\.julia\packages\GLMakie\rXGL8\src\GLAbstraction\GLTexture.jl:114
  [4] GLMakie.GLAbstraction.Texture(x::Observable{Array{Float32, 4}}; kw::Base.Iterators.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:minfilter,), Tuple{Symbol}}})
    @ GLMakie.GLAbstraction C:\Users\wkearney\.julia\packages\GLMakie\rXGL8\src\GLAbstraction\AbstractGPUArray.jl:210
  [5] (::GLMakie.var"#164#171")(arg::Observable{Array{Float32, 4}})
    @ GLMakie C:\Users\wkearney\.julia\packages\GLMakie\rXGL8\src\drawing_primitives.jl:536
  [6] map(f::GLMakie.var"#164#171", t::Tuple{Observable{Array{Float32, 4}}, Observable{Array{Float32, 4}}, Observable{Matrix{Float32}}})
    @ Base .\tuple.jl:215
[...]

PR #1183, opened 2021-07-30
fix matrix surface

SimonDanisch:


wkearn:
This does fix #1180 for me. Thanks, @SimonDanisch!

PR #1193, opened 2021-08-04
add multiargument method to annotations

piever:
This way, users can do

```julia
annotations(rand(10), rand(10), rand(["a", "b"], 10))
```

which IMO is a bit simpler and plays a bit nicer with AoG (even though the previous method is now [supported](http://juliaplots.org/AlgebraOfGraphics.jl/dev/generated/gallery/#Pre-scaled-data)).

I confess I could not find the `annotations` doc page to add some example with this syntax, so I'm not too sure how to document it.

### DIFF
--- a/src/basic_recipes/annotations.jl
+++ b/src/basic_recipes/annotations.jl
@@ -11,6 +11,25 @@ $(ATTRIBUTES)
 end
 
 function convert_arguments(::Type{<: Annotations},
+                           xs::AbstractVector{<: Number},
+                           ys::AbstractVector{<: Number},
+                           strings::AbstractVector{<: AbstractString})
+    return (map(strings, xs, ys) do str, x, y
+        (String(str), Point{2, Float32}(x, y))
+    end,)
+end
+
+function convert_arguments(::Type{<: Annotations},
+                           xs::AbstractVector{<: Number},
+                           ys::AbstractVector{<: Number},
+                           zs::AbstractVector{<: Number},
+                           strings::AbstractVector{<: AbstractString})
+    return (map(strings, xs, ys, zs) do str, x, y, z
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
annotations is actually just a wrapper for `text` now. I like the idea of better conversions, although I've wondered if the text shouldn't be relegated to attribute status somehow. The positions should be the positional arguments, I agree.

piever:
Ah, I see, so I should just add this conversion to `Text`? In that case maybe better to just deprecate `Annotations` for `Text` to avoid confusion.

I also like the keyword argument idea. It feels similar to `scatter(x, y, marker=markers)` for example. So instead of `text(x, y, labels)`, we should maybe do `text(x, y, text=labels)`? What would be a good keyword name here?

From a practical point of view, how does one "consume" an attribute in a `convert_arguments` call?

SimonDanisch:
>Ah, I see, so I should just add this conversion to Text? In that case maybe better to just deprecate Annotations for Text to avoid confusion.

Yes, and yes ;) 

>From a practical point of view, how does one "consume" an attribute in a convert_arguments call?

Hm, what do you mean?

piever:
> > From a practical point of view, how does one "consume" an attribute in a convert_arguments call?
> 
> Hm, what do you mean?

Never mind, I found it in the mean time, I was looking for the `@keyword` macro, to convert the value of a keyword argument as well during `convert_argument`.

Btw, do we have consensus as to whether we prefer `text(x, y, labels)` or `text(x, y, text=labels)` (or yet another keyword name)? I'm mildly in favor of the latter, but would be fine with either.

SimonDanisch:
I think we should close this and concentrate on deprecating annotations, right?

PR #1194, opened 2021-08-04
Remove reference to AbstractPlotting in JOSS paper

greimel:


jkrumbiegel:
oh good catch, thanks!

PR #1195, opened 2021-08-04
smarter defaults for tellheight and tellwidth

piever:
This updates the defaults for `tellheight` and `tellwidth` if the orientation of the legend is horizontal. In the current implementation, if the user does not set `tellheight` and `tellwidth`, they update with the legend orientation. If we don't like that, we can set them just the first time using `to_value`.

I was thinking `Colorbar` also needed this, but it actually already "tells" both dimensions by default, so flipping that doesn't make a difference.

### DIFF
--- a/src/makielayout/layoutables/colorbar.jl
+++ b/src/makielayout/layoutables/colorbar.jl
@@ -58,7 +58,6 @@ end
 
 function layoutable(::Type{<:Colorbar}, fig_or_scene; bbox = nothing, kwargs...)
     topscene = get_topscene(fig_or_scene)
-    attrs = merge!(Attributes(kwargs), default_attributes(Colorbar, topscene).attributes)
 
     default_attrs = default_attributes(Colorbar, topscene).attributes
     theme_attrs = subtheme(topscene, :Colorbar)
--- a/src/makielayout/layoutables/legend.jl
+++ b/src/makielayout/layoutables/legend.jl
@@ -6,8 +6,14 @@ function layoutable(::Type{Legend},
     topscene = get_topscene(fig_or_scene)
 
     default_attrs = default_attributes(Legend, topscene).attributes
+    attrs = Attributes(kwargs)
+    orientation = get(attrs, :orientation, nothing)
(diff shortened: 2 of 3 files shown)


src/makielayout/layoutables/colorbar.jl
```diff
@@ -58,7 +58,6 @@ end
 
 function layoutable(::Type{<:Colorbar}, fig_or_scene; bbox = nothing, kwargs...)
     topscene = get_topscene(fig_or_scene)
-    attrs = merge!(Attributes(kwargs), default_attributes(Colorbar, topscene).attributes)
```
piever, line 61: Unrelated, but it seems like this line doesn't do anything and was probably forgotten in a previous PR.

SimonDanisch (approved):

src/makielayout/layoutables/legend.jl
```diff
@@ -6,8 +6,14 @@ function layoutable(::Type{Legend},
     topscene = get_topscene(fig_or_scene)
 
     default_attrs = default_attributes(Legend, topscene).attributes
+    attrs = Attributes(kwargs)
+    orientation = get(attrs, :orientation, nothing)
+    if !isnothing(orientation)
+        default_attrs[:tellheight] = lift(==(:horizontal), orientation)
+        default_attrs[:tellwidth] = lift(==(:vertical), orientation)
+    end
```
jkrumbiegel, line 14: I think I have done this a bit differently so far with other attributes, where I've set them to `automatic` by default and then checked the other values. Something like
```julia
real_tellwidth = lift(tellwidth, orientation) do tw, o
    tw === automatic ? o == :vertical : tw       
end
``` 

One would then pass `real_tellwidth` to the LayoutObservables at the bottom

piever: Ah I see. I've changed it with your suggestion, so now `attrs` stores a value of `tellheight` and `tellwidth` which could be `automatic`, and `LayoutObservables` receives the "computed version" (must be boolean).

PR #1207, opened 2021-08-07
Backtick escape function names in headers.

Krastanov:
Done in order to avoid incorrectly parsing them as italics when rendering.


Krastanov:
Here is an example of bad parsing in the current doc render:
image

PR #1208, opened 2021-08-08
Fix bug in the function inv_symlog10

Ezequiel92:
Fix bug in the definition of the inv_symlog10 function. s is not defined, the correct variable is low.


jkrumbiegel:
thank you!

PR #1222, opened 2021-08-15
add logical expressions for `ispressed`

ffreyer:
Adds the possibility to pass logical expressions to `ispressed`, for example

```julia
scene = Scene()

trigger = Node(nothing)
on(x -> trigger[] = nothing, events(scene).keyboardbutton)
on(x -> trigger[] = nothing, events(scene).mousebutton)

hotkey = Node(Makie.@logical Keyboard.left_shift && !Keyboard.left_control && Mouse.left)
on(trigger) do _
    println(ispressed(scene, hotkey[]))
end
scene
```

where `hotkey` lowers to `And(Keyboard.left_shift, And(Not(Keyboard.left_control), Mouse.left))`. So you could set up key combinations (a && b), alternative hotkeys (a || b) and everything in between. Performance is the same if not a bit better (BenchmarkTools is given me less frequent and lower max times for the case above).

With these changes things like
https://github.com/JuliaPlots/Makie.jl/blob/194e767e207b8b67796412b1fe40fd7fd18f54c6/src/camera/camera3d.jl#L275
could be simplified to `if dragging[] && ispressed(scene, button[])`.

I also added a `ispressed(scene, ::Bool)` so you can have a hotkey be always active (true) or always inactive (false). This would add a way to disable an interaction that relies on `ispressed`.

### DIFF
--- a/src/Makie.jl
+++ b/src/Makie.jl
@@ -195,7 +195,7 @@ export pixelarea, plots, cameracontrols, cameracontrols!, camera, events
 export to_world
 
 # picking + interactive use cases + events
-export mouseover, ispressed, onpick, pick, Events, Keyboard, Mouse, mouse_selection
+export mouseover, ispressed, onpick, pick, Events, Keyboard, Mouse, mouse_selection, @logical
 export register_callbacks
 export window_area
 export window_open
--- a/src/interaction/events.jl
+++ b/src/interaction/events.jl
@@ -89,7 +89,7 @@ ispressed(scene, result::Bool) = result
 
 ispressed(scene, op::And) = ispressed(scene, op.l) && ispressed(scene, op.r)
 ispressed(scene, op::Or)  = ispressed(scene, op.l) || ispressed(scene, op.r)
-ispressed(scene, op::Not) = !ispressed(scene, op.l)
+ispressed(scene, op::Not) = !ispressed(scene, op.x)
 
 ispressed(scene, set::Set) = all(x -> ispressed(scene, x), set)
 ispressed(scene, set::Vector) = all(x -> ispressed(scene, x), set)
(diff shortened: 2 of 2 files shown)


SimonDanisch:
Hm, I guess that's a nice idea!
I think we could also overload `&` `|` and `!` instead of using a macro:
```julia
struct And
 a::Keyboard.Button
 b::Keyboard.Button
end
Base.:(&)(a::Keyboard.Button, b::Keyboard.Button) = And(a, b)
julia> Keyboard.left_alt & Keyboard.left_shift
And(Makie.Keyboard.left_alt, Makie.Keyboard.left_shift)
````

ffreyer:
Changes:
- removed macro in favor of & | and ! overloads
- added `Exclusively` to do hard matches as described above
- updated cam2d! and cam3d! to use ispressed. So now you can disable interactions via the keys you pass and you can modify trigger them with complex combinations from the outside
- added tests for `ispressed`
- added docs for `ispressed`

SimonDanisch:
Awesome! :) This is ready to go right?

ffreyer:
I did a quick merge to see if tests fail. I think the docs I wrote might be broken now that we moved to Franklin.

I also had a couple of concerns that I didn't mention yet. 

I'm wondering if it would be good to have variants of `ispressed` to detect when a button or combination has just become active or has just become inactive. Because currently you have to go back to backend events to deal with that. I played around with a `waspressed(scene, keys, removed_key)` function but I can't find that code anymore.

Since `ispressed` works with mouse and keyboard buttons I'm wondering whether we should add an `onyany` that works with just PriorityObservables. You'd generally want to trigger on both and `onany` would simplify that.

When I thought about this last I decided that I didn't want to make this pr more complicated. Maybe it's better to leave this as is and make another pr later adding that stuff?

SimonDanisch:
>When I thought about this last I decided that I didn't want to make this pr more complicated

Yeah, I think that's a good idea!

SimonDanisch:
Merged in https://github.com/JuliaPlots/Makie.jl/pull/1393

PR #1238, opened 2021-08-19
workaround for 1 elem hists

SimonDanisch:
`hist([8,8,8,8,8])` would yield histograms with nothing.
With this PR, they get one bar at 8:
image

Not sure if it's the nicest way to set edges to nothing in that case, but since the edges are kind of undefined in that case, most other values would not work out?

### DIFF
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -85,8 +85,8 @@ function Makie.plot!(plot::Hist)
             return bins
         end
     end
-    points = lift(edges, plot.normalization, plot.scale_to) do edges, normalization, scale_to
 
+    points = lift(edges, plot.normalization, plot.scale_to) do edges, normalization, scale_to
         h = StatsBase.fit(StatsBase.Histogram, values[], edges)
         h_norm = StatsBase.normalize(h, mode = normalization)
         centers = edges[1:end-1] .+ (diff(edges) ./ 2)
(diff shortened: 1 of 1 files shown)


src/stats/hist.jl
```diff
@@ -72,6 +72,9 @@ function Makie.plot!(plot::Hist)
     edges = lift(values, plot.bins) do vals, bins
         if bins isa Int
             mi, ma = float.(extrema(vals))
+            if mi == ma
+                return nothing
```
piever, line 76: Can't we just set

```julia
mi, ma = mi - 0.5, ma + 0m5
```

and keep all the remaining logic the same? We still get the arbitrary bar of width `1` centered around the unique value, but avoid having to keep two different code paths.

In particular, I think having 2 separate code paths can cause issues with normalization options (for example with `normalization = :pdf` weights should be `1` instead of `values[][1]`).

SimonDanisch: Good idea, thanks :)

src/stats/hist.jl
```diff
@@ -81,8 +85,8 @@ function Makie.plot!(plot::Hist)
             return bins
         end
     end
-
     points = lift(edges, plot.normalization, plot.scale_to) do edges, normalization, scale_to
+
```
piever, line 89:
```suggestion

    points = lift(edges, plot.normalization, plot.scale_to) do edges, normalization, scale_to
```

Just to not add unrelated formatting changes to the PR

piever (approved):

PR #1244, opened 2021-08-21
Fix precompilation errors with Julia 1.3

devmotion:
Currently precompilation on Julia 1.3 errors even though Makie 0.15.0 states that it is compatible with Julia 1.3. This PR fixes the issue which was introduced in https://github.com/JuliaPlots/Makie.jl/pull/1118 and adds CI tests with Julia 1.3 to avoid regressions in the future.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Makie"
 uuid = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 authors = ["Simon Danisch", "Julius Krumbiegel"]
-version = "0.15.0"
+version = "0.15.1"
```
SimonDanisch, line 4: Best to leave this out, since with the backends & mono repo, things got a bit more complicated ;) 
We want to tag the current master anyways, so just leave it to us :)

devmotion:
The test errors are caused by GridLayoutBase, more concretely by https://github.com/jkrumbiegel/GridLayoutBase.jl/blob/b514330a891bfaabc809d8742ee596f9206c0e78/src/gridlayout.jl#L1346.

devmotion:
I opened a PR: https://github.com/jkrumbiegel/GridLayoutBase.jl/pull/22

.github/workflows/ci.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.3'
```
devmotion, line 24: I am wondering if maybe you want to just test Julia '1.3' (the lower bound) and '1' (the latest stable version)?

SimonDanisch: no I decided against '1', because I find it confusing and not very helpful... The latest stable version doesn't change often enough, that its really worth having such a short cut ;)

devmotion: Fair enough :) But is it actually helpful to test '1.5'?

devmotion:
Hmm, it seems GridLayoutBase 0.6.1 which contains the fix is not used in the tests.

.github/workflows/ci.yml
SimonDanisch: Ah, if we test 1.3, probably not ;)

devmotion:
Makie tests pass with Julia 1.3 :tada: It depends on GeometryBasics 0.4 (and compatible versions of GridLayoutBase and MeshIO) though, which causes many deprecation warnings in the tests. Even though they are not user facing, we should probably fix them first, either in this PR or in https://github.com/JuliaPlots/Makie.jl/pull/1132.

SimonDanisch:
Cool, yeah we should really merge that PR!

SimonDanisch:
@knuesel is on vaccation, but maybe we could also renew that PR with the given script

devmotion:
I opened https://github.com/JuliaPlots/Makie.jl/pull/1252.

SimonDanisch:
Oh damn... Seems like GeometryBasics also isn't 1.3 ready anymore? :D

devmotion:
It is, and it works successfully in the Makie, CairoMakie, and WGLMakie tests. I am a bit surprised about these test failures with GLMakie - it worked before we eliminated the deprecation warnings yesterday :smile:

SimonDanisch:
I think this line somehow triggers to fall back to the latest registered Makie version, and not the locally dev'ed one:
https://github.com/devmotion/Makie.jl/blob/dw/precompile_julia13/GLMakie/test/runtests.jl#L10

```julia
ERROR: LoadError: Unsatisfiable requirements detected for package GeometryBasics [5c1252a2]:
 GeometryBasics [5c1252a2] log:
 ├─possible versions are: [0.1.0-0.1.3, 0.2.0-0.2.15, 0.3.0-0.3.13, 0.4.0-0.4.1] or uninstalled
 ├─restricted to versions * by ReferenceTests [d37af2e0], leaving only versions [0.1.0-0.1.3, 0.2.0-0.2.15, 0.3.0-0.3.13, 0.4.0-0.4.1]
 │ └─ReferenceTests [d37af2e0] log:
 │   ├─possible versions are: 0.1.0 or uninstalled
 │   └─ReferenceTests [d37af2e0] is fixed to version 0.1.0
 ├─restricted to versions 0.4.1-0.4 by GLMakie [e9467ef8], leaving only versions 0.4.1
 │ └─GLMakie [e9467ef8] log:
 │   ├─possible versions are: 0.4.5 or uninstalled
 │   └─GLMakie [e9467ef8] is fixed to version 0.4.5
 └─restricted by compatibility requirements with Makie [ee78f7c6] to versions: 0.3.5-0.3.13 — no versions left
   └─Makie [ee78f7c6] log:
     ├─possible versions are: [0.9.0-0.9.6, 0.10.0, 0.11.0-0.11.2, 0.12.0, 0.13.0-0.13.14, 0.14.0-0.14.2, 0.15.0-0.15.1] or uninstalled
     ├─restricted to versions * by ReferenceTests [d37af2e0], leaving only versions [0.9.0-0.9.6, 0.10.0, 0.11.0-0.11.2, 0.12.0, 0.13.0-0.13.14, 0.14.0-0.14.2, 0.15.0-0.15.1]
     │ └─ReferenceTests [d37af2e0] log: see above
     └─restricted to versions 0.15.1 by GLMakie [e9467ef8], leaving only versions 0.15.1
       └─GLMakie [e9467ef8] log: see above
```
Note the:
```
restricted by compatibility requirements with Makie [ee78f7c6] to versions: 0.3.5-0.3.13 — no versions left (GeometryBasics)
```
But I'm not really sure why this works for the other backends, which should do exactly the same...
Maybe the registry is queried by the Pkg command and its this bug that it only finds older versions?

devmotion:
I made the GLMakie test environment similar to the one of WGLMakie and removed the submodule references to Makie, GeometryBasics, and FileIO and instead loaded them explicitly. It seems this fixed the test error (at least no errors so far).

SimonDanisch:
Yay, thanks for all the effort :D

devmotion:
Is it possible to release a new version of Makie with the fixes for Julia 1.3? I would like to support Makie 0.15 in AbstractGPsMakie without dropping 1.3 :slightly_smiling_face:

Issue #1236, opened 2021-08-19
Documentation on `Zoom!` 


SimonCoste:
Hello, I am trying to use the `zoom!` function, see [here](http://makie.juliaplots.org/stable/makie_api.html#Makie.zoom!-Tuple{Any,%20Union{Tuple{Vararg{T,%20N}},%20StaticArrays.StaticVector{N,%20T}}%20where%20{N,%20T},%20Any,%20Bool}), but there is no usable doc and no examples.

More generally, the doc is extremely hard to use: there are many functions who use `Vec3f0`, but I found no reference for this. For instance, `zoom(scene, Vec3f0(0,0,0), 2)` outputs a 
> no method matching zoom!(::Scene, ::Vec{3, Float32}, ::Int64)

so that I have no clue on how to use these Vec3f0 objects. 

I also have no clues on what are those `5f0` [often found in examples](https://lazarusa.github.io/BeautifulMakie/surfWireLines/), and I could not find something in the doc on them.

ffreyer:
`5f0` is a way to construct Float32, equivalent to `Float32(5 * 10^0)`. `Vec3f0` is just a (static) vector of 3 Float32's.

The function `zoom!(scene, point, zoom_step, shift_lookat::Bool)` applies to the old 3D camera, which is not used by default anymore. The arguments here are the `scene` it applies to (i.e. `lscene.scene`), the `point` which the zoom focuses, the amount of zooming that's done is `zoom_step` and `shift_lookat = true` adds translation when zooming to keep `point` at the same position.

The new camera has a method `zoom!(scene, zoom_step)` which does an fov based zoom where `zoom_step` again controls the amount of zooming. Depending on what kind of zoom you want. Here you need to call `update_cam!(scene, cameracontrols(scene))` to apply the changes at the moment.

If you don't want an fov-based zoom you may want to try translating the camera instead. You can use `translate_cam!(scene, Vec3f0(0, 0, +-zoom))` to translate in/out of the screen. (Again with a update_cam! afterwards) The `Vec3f0` passed here is in camera space, so it's (left/right, up/down, in/out). 

Another option is to adjust the camera position (eyeposition) and/or focal point (lookat) via `update_cam!(scene, cameracontrols(scene), eyeposition, lookat)`.

[...]

PR #1247, opened 2021-08-21
Fix `zoom!` method, cleanup

ffreyer:
See #1236 

* fix `zoom!(scene, zoom_step)` method for `Camera3D`
* call `update_cam!` after `translate_cam!`, `rotate_cam!` and `zoom!`
* cleanup/update some dosctrings


SimonDanisch (approved):
Looks good to me if it works :)

PR #1248, opened 2021-08-21
add Cycled object for direct cycler access

jkrumbiegel:


Datseris:
I've realized that without this building my docs with Makie.jl feels a bit unatural. Because I always need to use the colors of my color cycle directly in every code block and becomes tedius after a while. What's the status quo here, any chance we get this feature in soon? I recall we had a discussion on slack with other usesr as well and the majority was fine with current syntax.

Datseris:
woohoo!!!

Issue #1249, opened 2021-08-21
`contour(::Matrix)` and `contour!(::Axis, ::Matrix)` are undocumented


kimlaberinto:
It seems like `contour(::Matrix)` and `contour!(::Axis, ::Matrix)` are undocumented on the stable and dev branches.
http://makie.juliaplots.org/stable/plotting_functions/contour.html

I'm planning to do a PR on this to document this as well as include examples in the docs if it would be helpful

kimlaberinto:
See PR #1250

ffreyer:
Not anymore

PR #1250, opened 2021-08-21
Document `contour(::Matrix)` in docstring and with example

kimlaberinto:
Ref #1249 

This PR adds a bit of documentation to `contour`'s docstring and to its docs page, to document `contour(::Matrix)` and usage of `contour!(::Axis, ::Matrix)`.

### DIFF
--- a/docs/examples/plotting_functions/contour.md
+++ b/docs/examples/plotting_functions/contour.md
@@ -22,3 +22,22 @@ contour!(xs, ys, zs)
 f
 ```
 \end{examplefigure}
+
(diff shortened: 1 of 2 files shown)


kimlaberinto:
I now think that the two examples in `contour` are very similar. The difference being if the x and y locations are the indices or not. Let me know if there's any problem with the two examples being extremely similar.

SimonDanisch:
Thanks :) 
Yeah, more diverse examples would be helpful - especially for the volume case `contour(rand(4, 4, 4))`... But looks like a solid improvement

Issue #1259, opened 2021-08-25
Implement the generalized Arrays interface (OffsetArrays)


yakir12:
Makie should be able to plot the generalized Arrays interface to be compatible with e.g. OffsetArrays.

Currently, 
```julia
using OffsetArrays, GLMakie
xy = OffsetArray(rand(100), -50)
lines(xy)
```
results in:
```julia
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 100 and 100")
Stacktrace:
  [1] _bcs1(a::Base.OneTo{Int64}, b::OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}})
    @ Base.Broadcast ./broadcast.jl:516
  [2] _bcs
    @ ./broadcast.jl:510 [inlined]
  [3] broadcast_shape
    @ ./broadcast.jl:504 [inlined]
  [4] combine_axes
    @ ./broadcast.jl:499 [inlined]
  [5] instantiate
    @ ./broadcast.jl:281 [inlined]
  [6] materialize
    @ ./broadcast.jl:904 [inlined]
  [7] convert_arguments(P::PointBased, x::UnitRange{Int64}, y::OffsetVector{Float64, Vector{Float64}})
    @ Makie ~/.julia/packages/Makie/NL7Xw/src/conversions.jl:128
  [8] convert_arguments
    @ ~/.julia/packages/Makie/NL7Xw/src/conversions.jl:139 [inlined]
  [9] convert_arguments(T::Type{Lines{Tuple{OffsetVector{Float64, Vector{Float64}}}}}, args::OffsetVector{Float64, Vector{Float64}}; kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/.julia/packages/Makie/NL7Xw/src/conversions.jl:10
 [10] convert_arguments
    @ ~/.julia/packages/Makie/NL7Xw/src/conversions.jl:8 [inlined]
 [11] plot!(scene::Scene, P::Type{Lines}, attributes::Attributes, args::OffsetVector{Float64, Vector{Float64}}; kw_attributes::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:show_axis,), Tuple{Bool}}})
    @ Makie ~/.julia/packages/Makie/NL7Xw/src/interfaces.jl:321
[...]

PR #1260, opened 2021-08-25
allow lines of offsetarrays

yakir12:
Super simple tiny change that allows plotting lines of OffsetArrays, suggested by @SimonDanisch on slack. It solves #1259. 

This needs tests (it passes locally), but I'm not sure exactly how to properly test it. My naive approach would be to:
1. add OffsetArrays dependency to the Project file in the test repo
2. add a new test file for OffsetArrays
3. include said file in runtests.jl

Would that be good enough...?


yakir12:
Not sure what to do with this. My needs to plot offset arrays have changed, but I bet someone will need it at some point. So if there is anything else I can/should do to get this merged let me know.

SimonDanisch:
I see... I think this is mergeable right?

yakir12:
Technically it is mergeable, but while this works for image(some_offsetarray), it does not work for heatmap. So not sure it's in a fantastic state.

SimonDanisch:
Lets merge this... We can polish it later - the offsetarray depedency seems fine, since several packages we depend on already depend on it

PR #1277, opened 2021-08-29
use standard normal for QQNorm

palday:
In usual statistical practice, qqnorm plots are plotted against the quantiles of the standard normal and not against a normal distribution fitted to the data. This changes the behavior of `qqnorm` to use the standard normal as the reference distribution. I'm somewhat unsure at the moment about the tick labels, but perhaps @dmbates can comment (who brought this issue to my attention).

There is an additional friction point introduced by this change: the default `y=x` line shown in the qqnorm plot is now ridiculous for empirical distributions that aren't unit scaled. I realize that this uses `qqbuild` from Distributions, but perhaps changing the values to be on the quantile/CDF scale instead of the raw scale might help (that ties into the axis labeling as well). 

This type of plot also greatly benefits from a default 1:1 aspect, but I don't know how to make that change quickly.

I'm unsure whether this counts as a breaking change or just a bugfix.

### DIFF
--- a/src/stats/distributions.jl
+++ b/src/stats/distributions.jl
@@ -34,8 +34,8 @@ end
     default_theme(scene, Scatter)
 end
 
-convert_arguments(::Type{<:QQNorm}, args...; kwargs...) =
-    convert_arguments(QQPlot, Distributions.Normal(0, 1), args...; kwargs...)
+convert_arguments(::Type{<:QQNorm}, args...; qqline=:R, kwargs...) =
+    convert_arguments(QQPlot, Distributions.Normal(0, 1), args...; qqline=qqline, kwargs...)
 
 convert_arguments(::Type{<:QQPlot}, args...; kwargs...) =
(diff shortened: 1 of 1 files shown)


piever:
That makes sense, I guess plotting against a standard normal adds some extra information (makes it easier to get a sense of the standard deviation of the empirical distribution).

Re changing the fit line, I think you can just do something like

```julia
convert_arguments(::Type{<:QQNorm}, v; qqline=:fit) =
    convert_arguments(QQPlot, Distributions.Normal(), v; qqline=qqline)
```

so that `QQNorm` has a different default than `QQPlot`, which I think is reasonable. I imagine the typical use case of `QQPlot` is to test whether two distributions are equal, rather than equal up to shift and rescale (I may be wrong though).

dmbates:
Thanks for raising the issue, @palday.  I am still a bit confused about the structure of the code but we can get a chance to discuss it over the next couple of weeks.

piever (approved):

PR #1283, opened 2021-08-31
Allow IntervalsBetween(1)

jtrakk:
IntervalsBetween(1) lets the minor ticks use the same spacing as the major ticks.
This is useful when the major ticks don't cover the whole axis space. For example, 
this plots y major grid lines on -2, 0, 2 and minor grid lines on -6, -4, 4, 6.

```jl
draw(spec, axis=(; 
yticks=[-2,0,2], 
limits=(nothing, (-7, 7)), 
yminorticks=IntervalsBetween(1), 
yminorgridvisible=true, 
yminorgridcolor=:purple
))
```

### DIFF
--- a/src/makielayout/types.jl
+++ b/src/makielayout/types.jl
@@ -102,10 +102,6 @@ Indicates to create n-1 minor ticks between every pair of adjacent major ticks.
 struct IntervalsBetween
     n::Int
     mirror::Bool
-    function IntervalsBetween(n::Int, mirror::Bool)
-        n < 2 && error("You can't have $n intervals (must be at least 2 which means 1 minor tick)")
-        new(n, mirror)
-    end
 end
 IntervalsBetween(n) = IntervalsBetween(n, true)
 


jtrakk:
Counterpoint: The deleted error is easy to work around so the change may be more trouble than it's worth.

SimonDanisch:
What's the state of this PR?

SimonDanisch:
I'm closing this, since the state isn't clear to me ;) Feel free to re-open!

jtrakk:
@SimonDanisch From my perspective it's ready to merge. I don't have permissions to reopen though.

SimonDanisch:
But what about:

>Counterpoint: The deleted error is easy to work around so the change may be more trouble than it's worth if the absence of an error confuses some users who mean to use IntervalsBetween(2).

That suggests we shouldn't merge the PR?

jkrumbiegel:
From my perspective IntervalsBetween(1) doesn't make sense as that means there are no minor ticks in the current logic. The major ticks are not shadowed. If it's really wanted to shadow the minor ticks with major ticks, then one could just add a different tick type for that.

jtrakk:
Here is the product. Note that the minor grid covers the top and bottom, which are not covered by the major grid, even though they have the same spacing.


image


```jl
using CairoMakie, AlgebraOfGraphics

spec = data((;a=1:10, b=-4:5)) * mapping(:a, :b) * visual(Scatter)

struct IntervalsBetweenX
    n::Int
    mirror::Bool
end
IntervalsBetweenX(n) = IntervalsBetweenX(n, true)

function MakieLayout.get_minor_tickvalues(i::IntervalsBetweenX, scale, tickvalues, vmin, vmax)
    vals = Float64[]
    length(tickvalues) < 2 && return vals
    n = i.n

    if i.mirror
        firstinterval = tickvalues[2] - tickvalues[1]
        stepsize = firstinterval / n
        v = tickvalues[1] - stepsize
        prepend!(vals, v:-stepsize:vmin)
    end

    for (lo, hi) in zip(@view(tickvalues[1:end-1]), @view(tickvalues[2:end]))
        interval = hi - lo
        stepsize = interval / n
        v = lo
        for i in 1:n-1
            v += stepsize
            push!(vals, v)
        end
    end

    if i.mirror
        lastinterval = tickvalues[end] - tickvalues[end-1]
        stepsize = lastinterval / n
        v = tickvalues[end] + stepsize
        append!(vals, v:stepsize:vmax)
    end

    vals
end


draw(spec, axis=(;
yticks=[-2,0,2],
limits=(nothing, (-7, 7)),
yminorticks=IntervalsBetweenX(1),
yminorgridvisible=true,
yminorgridcolor=:purple,
ygridcolor=:green,
))
```

Issue #1303, opened 2021-09-08
Lift calls callback twice on construction


j-fu:
Hi, 
when constructing an observable via `lift`, the callback function `f` is called twice:

https://github.com/JuliaPlots/Makie.jl/blob/84d552d6470fd5480baff0d89d932687e8b1cc09/src/interaction/nodes.jl#L9

however, the calculation behind may be expensive. Is there way to prevent this ?

j-fu:
I replaced `Makie.lift(f,o)` by `map(f,o)` in my code. As I understand, 
`Base.map(f,::AbstractObservable)` acts as a constructor:
https://juliagizmos.github.io/Observables.jl/stable/#Base.map-Union{Tuple{F},%20Tuple{F,Observables.AbstractObservable,Vararg{Any,N}%20where%20N}}%20where%20F 
and it seems that this would spare the call currently used for observable construction in `lift`.

ffreyer:
Has been fixed

PR #1307, opened 2021-09-09
change Node to Observable, link video

Datseris:
Can someone tell me whether the way I've embedded html is correct...?

### DIFF
--- a/docs/documentation/nodes.md
+++ b/docs/documentation/nodes.md
@@ -1,34 +1,37 @@
 # Observables & Interaction
 
 Interaction and animations in Makie are handled using [`Observables.jl`](https://juliagizmos.github.io/Observables.jl/stable/).
-`Observable`s are called `Node`s in Makie for historical reasons and the two terms are used interchangeably.
 An `Observable` is a container object whose stored value you can update interactively.
 You can create functions that are executed whenever an observable changes.
 You can also create observables whose values are updated whenever other observables change.
 This way you can easily build dynamic and interactive visualizations.
 
-On this page you will learn how the `Node`s pipeline and the event-based interaction system work.
+On this page you will learn how the `Observable`s pipeline and the event-based interaction system work. Besides this, there is also a video tutorial on how to make interactive visualizations (or animations) with Makie.jl and the `Observable` system:
 
-## The `Node` structure
+```@raw html
+<iframe width="560" height="315" src="https://www.youtube.com/embed/L-gyDvhjzGQ" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+```
 
-A `Node` is an object that allows its value to be updated interactively.
+## The `Observable` structure
+
+A `Observable` is an object that allows its value to be updated interactively.
 Let's start by creating one:
 
 ```julia:code1
 using GLMakie, Makie
 
-x = Node(0.0)
+x = Observable(0.0)
 ```
 \show{code1}
 
-Each `Node` has a type parameter, which determines what kind of objects it can store.
+Each `Observable` has a type parameter, which determines what kind of objects it can store.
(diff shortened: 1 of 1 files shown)


Datseris:
cc @SimonDanisch this should be the start of getting rid of `Node`.

Datseris:
Okay, so now the question is: should I do a massive search and replace all Node -> Observable?

ffreyer:
If we're getting rid of `Node` should we also get rid of `lift`? Maybe I'm missing something but my impression is that it's the same as `map`/`map!`.

Datseris:
wait WHAT????? All this time I've been living in a LIE??? I thought they were different functions!!! Oh my god yes PLEASE let's remove all these duplicate names!!!!!!

jkrumbiegel:
The difference is that lift executes once I think, so it creates a typed observable, while map creates an untyped one. At least I think it used to be like that.

ffreyer:
They both do that now.

From [Makie](https://github.com/JuliaPlots/Makie.jl/blob/77a655d63504fe3e82e9bb165873442cf97d10da/src/interaction/nodes.jl#L9):
```julia
function lift(f, o1::Observables.AbstractObservable, rest...; kw...)
    # compat...
    init = f(to_value(o1), to_value.(rest)...)
    typ = typeof(init)
    result = Observable{typ}(init)
    map!(f, result, o1, rest...)
    return result
end
```

From [Observables](https://github.com/JuliaGizmos/Observables.jl/blob/0f94e69cabddd87fcf057fb95ed5dc64be9661de/src/Observables.jl#L477):
```julia
@inline function Base.map(f::F, arg1::AbstractObservable, args...; kwargs...) where F
    # compat...
    map!(f, Observable(f(arg1[], map(to_value, args)...)), arg1, args...; update=false)
end
```

Those are the same except for `update = false` as far as I can tell. 

> By default `observable` gets updated immediately, but this can be suppressed by specifying `update=false`.

I guess that's be why `lift` triggers twice on creation, i.e. #1303?

SimonDanisch:
Seems like this broke the docs:
```
→ Initial full pass...
┌ Franklin Warning: in <nodes.md>
│ Encountered an issue processing 'nodes.md' in l/docs/documentation.
│ Verify, then re-start the Franklin server.
│ The error is displayed below:
│ Franklin.OCBlockError("I found the opening token 'CODE_TRIPLE' but not the corresponding closing token.", "Context:\n\t...fullscreen></iframe>\n```\n\n## The `Observabl... (near line 10)\n\t                        ^---\n")
└
ERROR: LoadError: I found the opening token 'CODE_TRIPLE' but not the corresponding closing token.
Context:
	...fullscreen></iframe>
```

## The `Observabl... (near line 10)
	                        ^---
```

File: docs/documentation/nodes.md [hunk 1]
```diff
@@ -1,42 +1,45 @@
...
 An `Observable` is a container object whose stored value you can update interactively.
 You can create functions that are executed whenever an observable changes.
 You can also create observables whose values are updated whenever other observables change.
 This way you can easily build dynamic and interactive visualizations.
 
-On this page you will learn how the `Node`s pipeline and the event-based interaction system work.
+On this page you will learn how the `Observable`s pipeline and the event-based interaction system work. Besides this, there is also a video tutorial on how to make interactive visualizations (or animations) with Makie.jl and the `Observable` system:
 
-## The `Node` structure
+```@raw html
```
SimonDanisch, line 11: @jkrumbiegel do you know how to properly inline this with Franklin?

Datseris: I have seen you guys have already put the tutorial in the home page under tutorials, so maybe we can just crosslink there? (I don't know how)

jkrumbiegel: Cross-linking would also work, although there isn't the perfect thing to link to I guess

SimonDanisch:
Merged in https://github.com/JuliaPlots/Makie.jl/pull/1393

PR #1314, opened 2021-09-13
Add mixed math- and text-mode LaTeX docs example

briochemc:


docs/documentation/latex.md
```diff
@@ -30,6 +30,24 @@ f
...
 \end{examplefigure}
 
+You can also mix math-mode and text-mode:
+
+\begin{examplefigure}{svg = true}
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+
+f = Figure(fontsize = 18)
```
jkrumbiegel, line 40: I guess the fontsize is superfluous here

docs/documentation/latex.md
```diff
@@ -30,6 +30,24 @@ f
...
+
+ax = f[1,1] = Axis(f,
```
jkrumbiegel, line 42: I would use the `Axis(f[1, 1])` syntax instead, I've been phasing out use of the other one over time

docs/documentation/latex.md
briochemc: Yes, I just wanted to see it big! Let me edit

Issue #1280, opened 2021-08-30
Incorrect depth-ordering in `arrows()`


staticfloat:
MWE:

```julia
using WGLMakie
function cartesian(r, θ, ϕ)
    return Point3.(
        r .* cos.(ϕ) * sin.(θ),
        r .* sin.(ϕ) * sin.(θ),
        r .* cos.(θ),
    )
end
cartesian(spherical::Point3) = cartesian(spherical.data...)

ndiv = 4
θs = (0:(1*ndiv-1)).*1π./(1*ndiv - 1)
ϕs = (0:(2*ndiv-1)).*2π./(2*ndiv-1)
arrows(
    cartesian.(Point3(0.1, θ, ϕ) for θ in θs, ϕ in ϕs)[:],
    cartesian.(Point3(0.1, θ, ϕ) for θ in θs, ϕ in ϕs)[:],
    arrowsize = 0.1,
    arrowcolor = :blue,
)
```

Example output:

https://user-images.githubusercontent.com/130920/131387970-82b84c09-c233-4d5c-a2ae-70a97e158607.mov

PR #1330, opened 2021-09-21
Fix depth buffer issue in WGLMakie

### LINKED ISSUE #1294: Lighting and Z-Order issues in WGLMakie
I'm using this example from the docs:

```julia
using IntervalSets
using WGLMakie

N = 60
function xy_data(x, y)
    r = sqrt(x^2 + y^2)
    r == 0.0 ? 1f0 : (sin(r)/r)
end
l = range(-10, stop = 10, length = N)
z = Float32[xy_data(x, y) for x in l, y in l]
surface(
    -1..1, -1..1, z,
    colormap = :Spectral
)
```

Package Info:
```
(@v1.6) pkg> st
      Status `~/.julia/environments/v1.6/Project.toml`
  [e9467ef8] GLMakie v0.4.6
  [8197267c] IntervalSets v0.5.3
  [80ea8bcb] PortAudio v1.1.1 `~/Dropbox/juliadev/PortAudio`
  [276b4fcb] WGLMakie v0.4.6
```

I'm seeing a couple issues and I'm not sure if any of them are intertwined. LMK which are independent and I can split out different issues.

1. The lighting seems to lag behind the camera motion. Sometimes it's just a slight lag, but sometimes the light spends a couple seconds catching up and following my path after I've changed my viewpoint
2. the light seems to be in the wrong place after I make a full circle - it seems like it moves to behind the surface
3. The z-order seems to get messed up when I rotate to some angles

None of these issues happen for me using the GL backend.

I've currently tested this from within VSCode, using Arch Linux.

I tried to make a GIF but my screencap software just gave a black box for the webgl window. Here are some still images.

This is the original view, where things look as expected:
expected plot

This is after 1/4turn counterclockwise - notice the z-ordering issue (I think?):
after  25 turn counterclockwise

This is after a full turn counterclockwise - notice the light is coming from the right now:
after 1 turn counterclockwise

rlschuller:
Make the attributes `transparency` and `overdraw` `Bool` instead of `Observable` in the return value of 'serialize_three(scene::Scene, plot::AbstractPlot)' (`serialization.jl` file).

In `create_material(program)` (`wglmakie.js` file), these attributes are passed directly to their corresponding javascript variables. This was setting `depthTest` and `depthWrite` to `false` regardless of the actual values of `transparency` and `overdraw`.

Fix #1294 and #1280 .

An alternative solution would be to create an update function that could recompile the shader whenever `transparency` or `overdraw` changed. However, **GLMakie** also doesn't seem to change these values dynamically.

The attached images show Pluto sessions after the fix, for the issues #1294 and #1280 respectively.

image
image

### DIFF
--- a/WGLMakie/src/serialization.jl
+++ b/WGLMakie/src/serialization.jl
@@ -292,8 +292,8 @@ function serialize_three(scene::Scene, plot::AbstractPlot)
     mesh[:name] = string(Makie.plotkey(plot)) * "-" * string(objectid(plot))
     mesh[:visible] = plot.visible
     mesh[:uuid] = js_uuid(plot)
-    mesh[:transparency] = plot.transparency
-    mesh[:overdraw] = plot.overdraw
+    mesh[:transparency] = plot.transparency[]
+    mesh[:overdraw] = plot.overdraw[]
     uniforms = mesh[:uniforms]
     updater = mesh[:uniform_updater]
 
--- a/WGLMakie/src/serialization.jl
+++ b/WGLMakie/src/serialization.jl
@@ -292,8 +292,8 @@ function serialize_three(scene::Scene, plot::AbstractPlot)
     mesh[:name] = string(Makie.plotkey(plot)) * "-" * string(objectid(plot))
     mesh[:visible] = plot.visible
     mesh[:uuid] = js_uuid(plot)
-    mesh[:transparency] = plot.transparency[]
-    mesh[:overdraw] = plot.overdraw[]
+    mesh[:transparency] = plot.transparency
+    mesh[:overdraw] = plot.overdraw
     uniforms = mesh[:uniforms]
     updater = mesh[:uniform_updater]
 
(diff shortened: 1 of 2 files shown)


rlschuller:
**GLMakie** deals with `transparency` and `overdraw` in the pre-render phase, see code bellow:

https://github.com/JuliaPlots/Makie.jl/blob/f761ec5e9af676d5ddd15a9ce41937c05f289fe4/GLMakie/src/GLAbstraction/GLRenderObject.jl#L29-L51

It's possible that I'm missing something, but I couldn't find any update functions that change the shaders / materials in **GLMakie**.

SimonDanisch:
Oh, good catch... 
The correct fix would be to use: `JSServe.get_observable(program.overdraw)` on the JS side, though ;)

rlschuller:
Thanks for the suggestion. :)

rlschuller:
Any thoughts?

SimonDanisch:
Perfect, thanks for the ping!

PR #1336, opened 2021-09-23
Improve contour documentation

behinger:
it took me quite a while (and asking @SimonDanisch) to figure out, that I can simply put an array for `levels` in contour.
This is now clarified (I then noticed it was already clarified in `countourf`).


jkrumbiegel:
thank you!

PR #1340, opened 2021-09-24
hide franklin's output macro from readers

jkrumbiegel:


SimonDanisch (approved):

jkrumbiegel:
There's different functionality in Franklin now so we don't need this hack anymore

Issue #1342, opened 2021-09-25
`stairs` fails when using an array of colors


Krastanov:
```julia
f = Figure()
Axis(f[1, 1])

xs = 0:0.5:10
ys = 0.5 .* sin.(xs)

color = ys
#color = :black # this would work fine

m.lines!(xs, ys .- 0; color)    # works
m.scatter!(xs, ys .- 1; color)  # works
m.stairs!(xs, ys .- 2; color)   # fails

f
```

The plot obtained with only `lines` and `scatter` is bellow. I hope to get the same behavior out of `stairs`
![image](https://user-images.githubusercontent.com/705248/134787395-5e2b37d0-8e79-4839-9805-e3d2b6d11283.png)

Here is the error message

```
Failed to show value:

AssertionError: length(positions) == length(colors)

    draw_multi(::MakieCore.Lines{Tuple{Vector{GeometryBasics.Point{2, Float32}}}}, ::Cairo.CairoContext, ::Vector{GeometryBasics.Vec{2, Float32}}, ::Vector{ColorTypes.RGB{Float32}}, ::Vector{Float32}, ::Nothing)@primitives.jl:126
    draw_multi(::MakieCore.Lines{Tuple{Vector{GeometryBasics.Point{2, Float32}}}}, ::Cairo.CairoContext, ::Vector{GeometryBasics.Vec{2, Float32}}, ::Vector{ColorTypes.RGB{Float32}}, ::Float32, ::Nothing)@primitives.jl:114
    draw_atomic(::Makie.Scene, ::CairoMakie.CairoScreen{Cairo.CairoSurfaceBase{UInt32}}, ::MakieCore.Lines{Tuple{Vector{GeometryBasics.Point{2, Float32}}}})@primitives.jl:57
    draw_plot(::Makie.Scene, ::CairoMakie.CairoScreen{Cairo.CairoSurfaceBase{UInt32}}, ::MakieCore.Lines{Tuple{Vector{GeometryBasics.Point{2, Float32}}}})@infrastructure.jl:240
    draw_plot(::Makie.Scene, ::CairoMakie.CairoScreen{Cairo.CairoSurfaceBase{UInt32}}, ::MakieCore.Combined{Makie.stairs, Tuple{Vector{GeometryBasics.Point{2, Float32}}}})@infrastructure.jl:244
[...]

PR #1345, opened 2021-09-28
Stairs needs to modify the color attribute (new contributor)

Krastanov:
If the color attribute is a vector, its components need to be
repeated in order to contain a color for each line segment.

Fixes https://github.com/JuliaPlots/Makie.jl/issues/1342

I marked it as "new contributor", because I am completely unaware of Makie code and testing conventions and would need some "handholding".

### DIFF
--- a/src/basic_recipes/stairs.jl
+++ b/src/basic_recipes/stairs.jl
@@ -63,7 +63,23 @@ function plot!(p::Stairs{<:Tuple{<:AbstractVector{<:Point2}}})
         end
     end
 
-    lines!(p, steppoints; [x for x in pairs(p.attributes) if x[1] != :step]...)
+    if isa(p.color[], AbstractVector)
+        stepcolor = lift(p.color, p.step) do color, step
+            if step == :pre
+                c = [color';color'][2:end]
+            elseif step == :post
+                c = [color';color'][1:end-1]
+            elseif step == :center
+                c = [color';color'][1:end]
+            else
+                error("Invalid step $step. Valid options are :pre, :post and :center")
+            end
+            c
+        end
+        lines!(p, steppoints; color=stepcolor, [x for x in pairs(p.attributes) if x[1] != :step && x[1] != :color]...)
+    else
+        lines!(p, steppoints; [x for x in pairs(p.attributes) if x[1] != :step]...)
+    end
     p
 end
 


sethaxen:
I agree it would be nice to be able to set the colors of line segments. In particular, I could see a user wanting vertical and horizontal line segments to have different colors, alphas, styles, etc.

But it seems this particular approach allows for setting the colors of horizontal line segments only, while the vertical line segments end up being gradients.

```julia
stairs(1:10, 1:10; color=iseven.(1:10))
```


I don't know enough about Makie's color system to know if it's possible to set the colors of specific segments.

Krastanov:
I guess this is a decision to be made. What should be the interface? A couple of options:

- the suggestion in this PR
- the verticals are the same as the pre/post horizontal
- separate colors for the verticals defaulting to one of the previous options

All three, provided in separate flags?

If one of the devs decides on an interface, I would be happy to update this PR appropriately.

In the meantime, two questions:
- How is makie tested? Do I write a test that just checks the plot does not raise an error? Do I test both backends? What is a canonical example of a good test for this?
- Why is this not a problem in GLMakie (it is only CairoMakie that needs it as discussed in #1342)

Krastanov:
Actually, anything more sophisticated than the current implementation would probably be fairly invasive. For what is worth, the current one seems to be the closest to what Line and Scatter do.

SimonDanisch:
>I don't know enough about Makie's color system to know if it's possible to set the colors of specific segments.

Lines share vertices, you need to use linesegments to color separate segments!

Krastanov:
@SimonDanisch, are you saying that this should be converted to linesegments and that the extra functionality should be implemented? I think I understand how to do that, but a "decision from above" for what the interface should be would be helpful, so that I do not write multiple version that would get declined. With the current suggested behavior (coloring the horizontal parts and gradients for the vertical parts) one can just use lines and no changes are necessary. For most of the other behaviors, linesegments would be necessary, but it would be helpful which other behavior is preferred by the devs:

- the suggestion in this PR
- the verticals are the same as the pre/post horizontal
- separate colors for the verticals defaulting to one of the previous options
- All three, provided in separate flags?

jkrumbiegel:
The question is what should the colors correspond to in a step plot, and I'm not sure what it would be. The step plot visualizes n points, so if it makes sense to visualize one value each per one of these points, then one would have to go with the gradient version. But gradients on the other hand run counter to what the step plot is about, namely not interpolating between two adjacent values but stepping. So stepping color would make more sense to me. What the color between two step points should mean, I'm not sure. It would help to collect examples from real plots so we can see what use cases there are.

Krastanov:
@jkrumbiegel , with `:pre` and `:post` the location of the vertical line corresponds to a point, so it makes sense for that vertical line to have the same color as the point. With `:center`, I guess there should either be a gradient on the horizontal line or the horizontal line should be split in half in two separate colors. And separately, there can be an overwrite for the color of both the vertical and horizontal lines. Would that be agreeable?

Krastanov:
and here is a stackoverflow question about matlab with similar goals https://stackoverflow.com/questions/38843676/adding-colors-to-lines-of-a-stairstep-plot

SimonDanisch:
Should we just merge this? It seems like a solid improvement...

Krastanov:
I think there are just a couple of "behavior" questions that need to be decided upon by a benevolent dictator:

- Is it ok that the connecting lines a gradients instead of steps (or this can be disregarded by claiming it can become an option in the future without breaking changes)

SimonDanisch:
I see... I think steps would be better than a gradient.
You can use one color per segment by supplying `n_points / 2` colors, e.g.:
```julia
linesegments(1:4, color=[:red, :green])
linesegments(1:4, color=[1, 2]) # or with colormap
```

src/basic_recipes/stairs.jl
```diff
@@ -62,24 +62,6 @@ function plot!(p::Stairs{<:Tuple{<:AbstractVector{<:Point2}}})
...
-            else
-                error("Invalid step $step. Valid options are :pre, :post and :center")
-            end
-            c
-        end
-        lines!(p, steppoints; color=stepcolor, [x for x in pairs(p.attributes) if x[1] != :step && x[1] != :color]...)
-    else
-        lines!(p, steppoints; [x for x in pairs(p.attributes) if x[1] != :step]...)
-    end
-    p
```
asinghvi17, line 67: strange - looks like the merge commit took this code out.  Was that intentional @SimonDanisch?  (Was going through the PR backlog and saw this).

SimonDanisch: Not sure to be honest...

Issue #442, opened 2020-03-02
time axis support


visr:
For plotting time series, it would be nice to support `Date` and `DateTime` axes.

A minimal example:
```julia
using Makie
x = DateTime(2020):Hour(1):DateTime(2020, 1, 2)
y = randn(25)
scatter(x, y)  # -> MethodError: no method matching push!(::Annotations{...}, ::Array{String,1}, ::Tuple{Int64,Float64}; rotation=0.0, textsize=0.8251367139816284, align=(:center, :top), color=:black, font="Dejavu Sans")
```

I currently work around this by making an "hours since" `Float64`, using `hours(times::Vector{DateTime}) = Dates.value.(times - first(times)) / 3_600_000` , but you easily lose track of time.

Using Plots this works:
```julia
using Plots: Plots
Plots.gr()
Plots.scatter(x, y)
```
Resulting in
![image](https://user-images.githubusercontent.com/4471859/75660974-c2a46b00-5c6c-11ea-9ff6-c8dbb3dc93b9.png)

mkborregaard:
@daschw is this functionality not all in PlotUtils so it would mostly trivial to add to Makie?

daschw:
The functionality for calculating ticks for Date(Time) objects is in PlotUtils `optimize_datetime_ticks`.
Plotting the objects is implemented in Plots via a type recipe. I guess that should be fairly trivial for Makie.

SimonDanisch:
Just seems to be a bug ;)

asinghvi17:
Definitely a bug on our end:
```julia
ERROR: MethodError: no method matching push!(::Annotations{...}, ::Array{String,1}, ::Tuple{Int64,Float64}; rotation=0.0, textsize=0.7853086853027342, align=(:center, :top), color=:black, font="Dejavu Sans")
Closest candidates are:
  push!(::Any, ::Any, ::Any) at abstractarray.jl:2158 got unsupported keyword arguments "rotation", "textsize", "align", "color", "font"
  push!(::Any, ::Any, ::Any, ::Any...) at abstractarray.jl:2159 got unsupported keyword arguments "rotation", "textsize", "align", "color", "font"
[...]

PR #1347, opened 2021-09-29
Implement time axis

SimonDanisch:
I noticed, that we can hack together a simple time aware axis as a small wrapper type around `Axis`...
We may even be able to integrate this into the `Axis` code itself, so that we don't need a wrapper, and the axis will just change to the correct conversion + formatting functions, depending on what you plot into it.

This would basically add a pre `convert_arguments` conversion, that is axis aware and can store axis local state (e.g. target unit, target range, etc).

We could also use this for categorical values and Float32 conversion, I think.
Still pretty early to tell, but I will use the `TimeAxis` wrapper to figure out the interface and add tests + proper error handling for the myriad of time conversions we may run into.
If this works out nicely, we can think about removing `TimeAxis` and add it to the standard `Axis` code.

But the simple mvp already works as intended:
```julia
f = Figure()
ax = TimeAxis(f[1,1])
scatter!(ax, rand(Hour(1):Hour(1):Hour(20), 10), 1:10)
f
```
image

```julia
scatter!(ax, rand(Second(1):Second(60*60):Second(20*60*60), 10), 1:10)
```
image

This currently uses promote_type to figure out the conversion target, which isn't ideal since we should rather chose the unit from the range of time values we want to display...


ffreyer:
Ultrafast laser pulses seem to be resolved on a fs scale, so maybe it's worth going down to that level?
For reference:
https://www.newport.com/n/ultrafast-pulse-characterization
https://www.dickinson.edu/download/downloads/id/25/ultrafastlaserpulses

SimonDanisch:
Yeah makes sense - that was just an oversight ... Still getting familiar with Unitful.jl and how to discover things...

src/makielayout/layoutables/time_axis.jl
```diff
@@ -1,32 +1,42 @@
...
 end
-const UnitfulTimes = (
-    typeof(1.0*u"d"), typeof(1.0*u"hr"), typeof(1.0*u"minute"),
-    typeof(1.0*u"s"), typeof(1.0*u"ms"), typeof(1.0*u"μs"), typeof(1.0*u"ns"))
 
-TimeUnits = Union{
-    Number,  # No time units
-    Nothing, # nothing yet
-    UnitfulTimes...}
+TimeUnits = (:yr, :wk, :d, :hr, :minute, :s, :ms, :μs, :ns, :fs)
```
ffreyer, line 9: This should go ns (10⁻⁹), ps (10⁻¹²), fs (10⁻¹⁵)

SimonDanisch: Yeah, I know, see: https://github.com/JuliaPlots/Makie.jl/blob/20734db18e1f7e2efb89ebae2ec0b373120ffa09/src/makielayout/layoutables/time_axis.jl#L219
This was a commit in the middle of refactoring the code, just to save the progress ;)

visr:
Great to see this! (I assume this) fixes #442.

SimonDanisch:
Fixes: https://github.com/JuliaPlots/Makie.jl/issues/442

ValentinKaisermayer:
I'm really looking forward to this. When do you expect this to be ready for release?

rapus95:
Will this also allow setting an axis to categorical data in recipes?

SimonDanisch:
>Will this also allow setting an axis to categorical data in recipes?

Yes it should!

laborg:
Great work! I consider dropping Plots for Makie if this lands. Any time prediction on when this will happen?

SimonDanisch:
In principle this PR is feature complete and works, but I will need some more time to finish up developer docs and clean up some hacks.
E.g., I noticed that this PR doubles CI times for docs and adds at least 10min to the tests...Which is pretty unacceptable and I need to debug why that happens...
Also, the conversion pipeline got even more confusing and I still hope to find a cleaner implementation for the axis converts.

SimonDanisch:
Although, I just noticed, that RPRMakie goes from 5min to 23min... Which is a bit suspicious, since the one test right now in RPRMakie is super simple, and seeing that Makie CI is even faster than in other PRs, this all doesn't make much sense :D Maybe it's just a fluke of the CI.
I also just tested this code piece, which should give a good idea of any large scale regression, and that comes out with exactly the same number as on master:
```julia
 @time begin
       using GLMakie
       display(scatter(1:4))
       end
```
So maybe this was really just a CI fluke, or there is one super slow code path - both should be good news, since those should then be much easier to fix than all of Makie becoming much slower (by the new deps, or new type unstable code).

rapus95:
If I wanted to use/test this already, what would be the easiest way to do so? It'd be so cool if there was an insider/experimental release

jeremiedb:
@SimonDanisch Sorry for the bump, I was wondering if there was anything left blocking? Having direct support for dates would be great feature!

uniment:
> Ultrafast laser pulses seem to be resolved on a fs scale, so maybe it's worth going down to that level? For reference: https://www.newport.com/n/ultrafast-pulse-characterization https://www.dickinson.edu/download/downloads/id/25/ultrafastlaserpulses

Considering that quantities measured in femtoseconds generally don't require use of the Julian calendar, you can probably make do without femtosecond support here. fwiw

SimonDanisch:
Closing in favor of https://github.com/MakieOrg/Makie.jl/pull/3226

PR #1355, opened 2021-10-01
refactor layoutables

jkrumbiegel:
- layoutables have a base layer scene which allows translating in z (useful for inset axes for example) and decouples from Figure better
- layoutable attributes are hardcoded fields with mostly constrained types
- layoutables are created with only a few fields preinitialized, the remaining fields must be filled in separate functions. this allows to use the objects themselves for dispatch for the organization of all that helper code. this way, the huge Axis constructor function can be broken up, etc, which also might help compilation latency in conjunction with the attribute typing


jklymak:
@jkrumbiegel if you ever want to talk about layout, I authored the matplotlib implementation of constrained_layout, and happy to chat. We just did a refactor about a year ago as well...

SimonDanisch:
I think this can be closed in favor of https://github.com/JuliaPlots/Makie.jl/pull/1796

PR #1365, opened 2021-10-05
fix floating point bug in surfacelike conversion

jkrumbiegel:


SimonDanisch:
This is a related fix right? https://github.com/JuliaPlots/Makie.jl/pull/1162

jkrumbiegel:
> This is a related fix right? #1162

Yeah same issue, fix is also almost the same, I just don't convert to Float32 directly anymore for that step but use the original data

Issue #1356, opened 2021-10-02
Label box not complete in subplots


bardo84:
See example in
[https://discourse.julialang.org/t/plots-and-subplots-with-makie/69113](https://discourse.julialang.org/t/plots-and-subplots-with-makie/69113)

The first label box misses the left side.

ffreyer:
Same thing as #1349?

bardo84:
Yes, same effect.
In my case happens only on the first subplot.

Issue #1349, opened 2021-09-30
Legend left frame spine is invisible


yakir12:
Running the first example in the [legend examples](https://makie.juliaplots.org/stable/examples/layoutables/legend/#legend_element_attributes) works but the left spine in the frame of the legend doesn't plot. This seems to be consistent in many other cases.
![a](https://user-images.githubusercontent.com/3281957/135458098-666db7ce-880a-46b3-99c5-ecf5c8b673b9.png)

I'm on `GLMakie v0.4.6`.

jkrumbiegel:
Can you check if master has this problem? I fixed something related a couple weeks ago here https://github.com/JuliaPlots/Makie.jl/commit/dfb44f1bd1b3916b260f5011a135c65ed83bc1f0

yakir12:
Yes, it does:
![a](https://user-images.githubusercontent.com/3281957/135459711-39355f08-6b4d-4027-8a1f-15dc09ac68b1.png)

This is from GLMakie v0.4.6 `https://github.com/JuliaPlots/Makie.jl.git:GLMakie#master`

jkrumbiegel:
it's Makie that needs to be on master, not GLMakie

yakir12:
Still a problem:
![a](https://user-images.githubusercontent.com/3281957/135463020-5974f2bc-64c2-49f2-a276-b5a80eef8083.png)

I'm on:
```julia
(jl_srBZ7r) pkg> st
      Status `/tmp/jl_srBZ7r/Project.toml`
  [e9467ef8] GLMakie v0.4.6
  [ee78f7c6] Makie v0.15.2 `https://github.com/JuliaPlots/Makie.jl.git#master`
```

ffreyer:
See also #865.

Does the full frame become visible if you increase the linewidth?
```julia
f.content[2].elements[:frame].strokewidth[] = 2
```
What about pushing the lines forward?
```julia
translate!(f.content[2].elements[:frame].plots[2], 0, 0, 1)
```
Or setting overdraw true for the lines?
```julia
f.content[2].elements[:frame].plots[2].overdraw[] = true
```

yakir12:
> See also #865.

Wow, yeah, seems very relevant. My WM is `i3`. I don't think `i3wm` is considered headless, but I don't have gnome or anything like that installed.

> Does the full frame become visible if you increase the linewidth?

no:
[...]

Issue #536, opened 2020-02-10
Rect polys are not closed correctly 


jkrumbiegel:
```julia
poly(Rect(0, 0, 100, 100), strokewidth = 10,
    color = (:blue, 0.3), strokecolor = :black)
```

![grafik](https://user-images.githubusercontent.com/22495855/74140333-2a7b0f00-4bf5-11ea-978a-e80d82e28d56.png)

ffreyer:
#3907 fixed this

PR #1374, opened 2021-10-08
Fix point duplication in `convert_arguments(::PointBased, ::Rect2)`

### LINKED ISSUE #865: GLMakie doesn't plot left-side legend outline when headless
On a headless system, GLMakie doesn't plot the left-side outline for legends. Normally, it does (obviously). Here is an example of what I am talking about:

Normal:
plot-with-head

Headless:
plot-headless

I am using Julia 1.6.0-rc1, FileIO 1.6, and GLMakie 0.1.28. My OS is 5.10.19-200.fc33.x86_64.

You can find a Docker image that can reproduce the issue in my repository here: https://github.com/Octogonapus/glmakie-legend-bug

### LINKED ISSUE #1366: Lines do not plot if a coordinate is repeated
I was surprised by this:
```julia
xy = rand(Point2f, 5) # five random coordinates
insert!(xy, 3, xy[3]) # repeat one of them
lines(xy) # behaves as if there was a NaN there?
```
a

ffreyer:
This is an attempt to fix #865, #1349 and #1356. Maybe also #536 since I got something similar by translating a poly plot. This does however not fix #1366 which I assume to be the cause.

~~I haven't checked if this actually works so I'll leave this as a draft for now.~~ Fixes the one case I could reproduce (#1356)

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -160,7 +160,7 @@ Takes an input `Rect` `x` and decomposes it to points.
 """
 function convert_arguments(P::PointBased, x::Rect2)
     # TODO fix the order of decompose
-    return convert_arguments(P, decompose(Point2f, x)[[1, 2, 4, 3, 1]])
+    return convert_arguments(P, decompose(Point2f, x)[[1, 2, 4, 3]])
 end
 
 function convert_arguments(P::PointBased, mesh::AbstractMesh)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
closed in favor of: https://github.com/JuliaPlots/Makie.jl/pull/1386

Issue #1378, opened 2021-10-13
Edges/strokes not reliably visible


Kevin-Mattheus-Moerman:
The below example (I created with the help of @SimonDanisch) and the gif animation show that for certain view angles the edges (or strokes) become invisible. 

Perhaps this is an "in front / behind" or ray tracing error? Or is the stroke width defined in a funny way with respect to the camera angle? 

```julia
] add Makie@0.15.2 GLMakie@0.4.6
```

```julia
using GLMakie, GeometryBasics

V=Point3f[(0, 0, 0), (1, 0, 0), (1, 1, 0), (0, 1, 0), (2, 1, 1), (2, 0, 1)]
F=QuadFace{Int}[(1, 2, 3, 4), (2, 3, 5, 6)] 

f = Figure()
ax=Axis3(f[1, 1], aspect = :data, xlabel = "x label", ylabel = "y label", zlabel = "z label", title = "Title")
poly!(GeometryBasics.Mesh(V, F), strokewidth=5,shading=true)
f
```

Ubuntu 20.04
Julia 1.6.3

![Peek 2021-10-13 09-54](https://user-images.githubusercontent.com/8392709/137100954-69a20016-69c1-4440-a63a-141a22391cb0.gif)

jkrumbiegel:
Oh that would be good, wireframe over mesh is pretty bad looking right now

ffreyer:
Another thing we could do is add a `depth_shift` acting on clip space coordinates (i.e. when the +z points out of the screen) in vertex shaders:

```glsl
uniform float depth_shift;
...
gl_Position = projection * view * model * world_pos; // or equivalent
gl_Position.z += gl_Position.w * depth_shift; 
// If I understand correctly gl_Position will get divided by gl_Position.w later.
// We probably want to counter that to keep depth_shift on a normalized scale
```

With that you could bump a plot object to the front or back. (Just to be extra clear - this would not affect the position whatsoever. It just affects the z-order/depth of the plot.) With that added to `line_segments.vert`:

```julia
let
    fig = Figure()
    left = LScene(fig[1, 1])
    poly!(left, GeometryBasics.Mesh(V, F), strokewidth=5, shading=true)
    
    right = LScene(fig[1, 2])
    p = poly!(right, GeometryBasics.Mesh(V, F), strokewidth=5, shading=true)
    p.plots[2].plots[1].attributes[:depth_shift] = Node(-0.01f0)
    
    fig
end
```

[...]

PR #1382, opened 2021-10-13
Add depth_shift

ffreyer:
This would add a `depth_shift` attribute that applies after the model, view (camera) and projection matrices to adjust an objects depth. It would allow you to push one render object/plot towards the front or back in terms of render order.

For example, if you had two copies of the same mesh at the same position, you could adjust which mesh is drawn in front of the other by setting `front_mesh_plot.depth_shift = Node(-0.001f0)`. 

TODO:
- [x] add this to WGLMakie (CairoMakie?)
- [ ] docs
- [x] tests

See #1378

@SimonDanisch I think I talked about something like this with you before, but back then I tried doing it on a fragment shader level. Iirc you said that it would be bad for performance to mess with fragment coordinates/depth because it stops opengl from discarding occluded fragments. But if we do things on the vertex shader level we should be fine, right?


ffreyer:
Merged in #1393

PR #1386, opened 2021-10-13
Fix repeated points in line

### LINKED ISSUE #865: GLMakie doesn't plot left-side legend outline when headless

SimonDanisch:
... on Headless/ bad GPU drivers...
Not 100% sure why this is so driver dependent, but the normals became NaN and poisened the line like that...

Fixes #865 #1349 #1356  #1366 ...
@ffreyer, I think we shouldn't remove the duplicated points, since in CairoMakie, that closes the line 😅

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -160,7 +160,6 @@ Takes an input `Rect` `x` and decomposes it to points.
 """
 function convert_arguments(P::PointBased, x::Rect2)
     # TODO fix the order of decompose
-    # repeat endpoint to close the line in cairomakie at least
     return convert_arguments(P, decompose(Point2f, x)[[1, 2, 4, 3]])
 end
 


SimonDanisch:
I tested those :) Turns out, I can reproduce the issue with WSL2 + headless GLMakie...

PR #1401, opened 2021-10-23
added rotation arg for labels

lazarusA:
this allows to rotate barplot labels.


Flausi:
Hey, I think the rotation function is great, but I think at the same time you should be able to change the orientation/alignment. Maybe in the style of `label_align = (:left, :center)`. 

If the attribute is already there, sorry. :)
hist1

jkrumbiegel:
The align can probably be made automatic for most cases following the logic of Axis ticklabels.

Issue #1368, opened 2021-10-06
Mesh does not interpolate using the colormap


mmorligh:
Hello

first, thanks so much for maintaining this package, it is the best! This is more of a feature request I guess. If I use a jet colormap and use mesh to create a plot, I get this:
```
using GLMakie
import ColorSchemes.jet
Makie.mesh( [0.0 0.0; 1.0 0.0; 1.0 1.0; 0.0 1.0;], [ 1 2 3], shading = false, color = [1;2;3;3], colormap=jet)
```

![Screen Shot 2021-10-06 at 3 51 32 PM](https://user-images.githubusercontent.com/43073461/136274116-7f6eda1d-449a-4703-b3a1-d897532f2d17.png)

But I would expect something like this (interpolate using the full spectrum of the colormap instead of the between the colors of each vertex):

![Screen Shot 2021-10-06 at 3 53 28 PM](https://user-images.githubusercontent.com/43073461/136274145-b4a5aacf-d123-4a04-bd41-8e8ad5c8c4f8.png)

which would be consistent with MATLAB's patch function:

```
patch('Vertices',[0.0 0.0; 1.0 0.0; 1.0 1.0; 0.0 1.0;],'Faces',[ 1 2 3],'FaceVertexCData',[1;2;3;3],'FaceColor','interp');
colormap jet
```

As a side note, I am not sure why this does not work:
```
using GLMakie
import ColorSchemes.jet
Makie.mesh( [0.0 0.0; 1.0 0.0; 1.0 1.0], [ 1 2 3], shading = false, color = [1;2;3;3], colormap=jet)
```

Thank you for considering this request!
Mathieu

PR #1406, opened 2021-10-24
Add `uv_transform` attribute to mesh, meshscatter, surface and image

ffreyer:
# Description

I'm updating this pr to be more general and of better quality. The old description is still at the end of this post.

This pr (now) adds a `:uv_transform` attribute to allow transforming the uv coordinates passed with a mesh/generated by surface/image. The original motivation was to allow matching a color gradient in `meshscatter` with the height of differently scaled bars, see https://discourse.julialang.org/t/3d-bars-makie-colors-and-meshes/70283.

```julia
using GLMakie, GeometryBasics, ColorSchemes
texture = reshape(get(colorschemes[:Spectral_11], 0:0.01:1), 101, 1)

# create fitting mesh
geom = Rect3f(Point3f(-0.5, -0.5, 0), Vec3f(1, 1, 1))
rect_mesh = GeometryBasics.normal_mesh(geom)
# use uv.u = z as a measure for mesh height
uvs = [Point2f(p[3], 0) for p in coordinates(rect_mesh)]
rect_mesh = GeometryBasics.Mesh(
    meta(
        decompose(Point3f, rect_mesh), 
        normals = decompose_normals(rect_mesh),
        uv = uvs
    ), 
    faces(rect_mesh)
)

pos = [Point3f(i, j, 0) for i in 1:10 for j in 1:10]
z = rand(10,10)

fig = Figure()
ax = Axis3(fig[1, 1]; aspect=(1, 1, 1), elevation=π / 6, perspectiveness=0.5)
meshscatter!(
    ax, pos, marker = rect_mesh, markersize = Vec3f.(1, 1, z[:]),
    uv_transform = Vec2f.(z[:], 1), 
    color = texture, shading = NoShading
)
ax.limits[] = ((0, 11), (0, 11), (0, 1.2))
  
fig
```
Screenshot 2024-07-30 142441

Another motivation to pick this back up and generalize is [Tyler](https://github.com/MakieOrg/Tyler.jl) which I've learned at Juliacon relies on multiple mesh! plots. With this pr it could rely on a meshscatter plot instead, as uv transforms allow you to target different sections of an image.

```julia
using FileIO, GLMakie, GeometryBasics
cow = FileIO.load(joinpath(pkgdir(Makie), "assets", "cow.png"))

f, a, p = meshscatter(
    [Point3f(x, y, rand()) for x in 1:10 for y in 1:10],
    color = cow,
    uv_transform = [(Vec2f(x, y), Vec2f(0.1, -0.1)) for x in 0.0:0.1:0.9 for y in 0.1:0.1:1],
    markersize = Vec3f(0.9, 0.9, 1),
    marker = uv_normal_mesh(Rect2f(-0.5, -0.5, 1, 1))
)
cameracontrols(a).settings[:projectiontype] = Makie.Orthographic
f
```
image

This may also become useful in the future if we take loading obj files more seriously. They can define an offset and scale for textures used in materials, which may connect nicely to `uv_transform`.

## TODO

- [x] add `uv_transform` to mesh, surface to fix and generalize `uv_scale`
- [x] export `UVTransform`
- [x] maybe update WGLMakie to allow textures meshscatter (and then add `uv_transform`)
- [x] maybe update CairoMakie
- [x] consider allowing rotation/generalizing to `Mat{2, 3, Float32}` and encoding the `uv = (1.0 - uv.y, uv.x)` here

## Changes

- Added `uv_transform` attribute for mesh, meshscatter, surface and image in GLMakie, WGLMakie and CairoMakie.
- Updated WGLMakie to allow textures in `meshscatter`
- Fixed incorrect order of matrix sizes in GLMakie/GLAbstractions (i.e. our `Mat{2, 3, Float32}` is equivalent to `mat3x2` in glsl)

## Type of change

A bit of everything...
- breaking because it replaces and generalizes `uv_scale`
- bug fix because `uv_scale` is not easily accessible on master due to attribute validation
- new feature: uv_transform attribute (generalizing scale to include rotation, transpose, translations, ...)
- new feature: textured meshscatter in WGLMakie

## Checklist

---

<details>
<summary>Old Description</summary>
In response to https://discourse.julialang.org/t/3d-bars-makie-colors-and-meshes/70283.

This enables passing of textures and per instance/marker `uv_scale`s to `meshscatter` with GLMakie. With that and the trick from #1368 you can do

```julia
using GLMakie, GeometryBasics, ColorSchemes
texture = reshape(get(colorschemes[:Spectral_11], 0:0.01:1), 1, 101)

rectMesh = FRect3D(Vec3f0(-0.5, -0.5, 0), Vec3f0(1, 1, 1))
recmesh = GeometryBasics.normal_mesh(rectMesh)
uvs = [Point2f(p[3], 0) for p in coordinates(recmesh)] # normalize this so zmax = 1
recmesh = GeometryBasics.Mesh(
    meta(coordinates(recmesh); normals=normals(recmesh), uv = uvs), 
    faces(recmesh)
)

pos = [Point3f(i, j, 0) for i in 1:10 for j in 1:10]
z = rand(10,10)

fig = Figure(resolution=(1200, 800), fontsize=26)
ax = Axis3(fig[1, 1]; aspect=(1, 1, 1), elevation=π / 6, perspectiveness=0.5)
meshscatter!(
    ax, pos, marker = recmesh, markersize = Vec3f.(1, 1, z[:]),
    uv_scale = Vec2f.(1, z[:]), color = texture, shading = false
)
ax.limits[] = ((0, 11), (0, 11), (0, 1.2))
  
fig
```

Screenshot from 2021-10-24 17-25-01
</details>


lazarusA:
don't forget this PR, it's a nice addition 😄

ffreyer:
#1436 would also add this but it'll take me some time to figure what each shader/plot primitive should include

asinghvi17:
Hey, just wanted to bump this - I'm working on some plots using meshscatter and this would definitely be nice to have!

asinghvi17:
Hey, wanted to pick this up again.  Is there anything stopping this from being merged?

lazarusA:
bump

ffreyer:
The branch should be up to date with master again.

It seems that something about `coordinates(mesh)` broke in GeometryBasics, so example now needs to be:
```julia
using GeometryBasics, ColorSchemes, GLMakie
texture = reshape(get(colorschemes[:Spectral_11], 0:0.01:1), 1, 101)
    
prim = FRect3D(Vec3f0(-0.5, -0.5, 0), Vec3f0(1, 1, 1))
rectmesh = GeometryBasics.normal_mesh(prim)
uvs = [Point2f(p[3], 0) for p in coordinates(rectmesh)] # normalize this so zmax = 1
rectmesh = GeometryBasics.Mesh(
    meta(collect(coordinates(prim)); normals=normals(rectmesh), uv = uvs), 
    faces(rectmesh)
)

pos = [Point3f(i, j, 0) for i in 1:10 for j in 1:10]
z = rand(10,10)

fig = Figure(resolution=(1200, 800), fontsize=26)
ax = Axis3(fig[1, 1]; aspect=(1, 1, 1), elevation=π / 6, perspectiveness=0.5)
meshscatter!(
    ax, pos, marker = rectmesh, markersize = Vec3f.(1, 1, z[:]),
    uv_scale = Vec2f.(1, z[:]), color = texture, shading = false
)
ax.limits[] = ((0, 11), (0, 11), (0, 1.2))
  
fig
```

@SimonDanisch Do you think it would be a good idea to generalize `uv_scale` to a `uv_transform::Mat3f`? That would encode rotation + translation + scale just like a model matrix, but for uv coordinates.

ffreyer:
> consider allowing rotation/generalizing to Mat{2, 3, Float32} and encoding the uv = (1.0 - uv.y, uv.x) here

@SimonDanisch What do you think about this? I.e. replacing the `o_uv = vec2(1.0 - tex_uv.y, tex_uv.x);` we have in a couple of shaders with
```julia
# attribute default
uv_transform = Mat{2, 3, Float32}(0, 1.0, -1.0, 0.0, 1.0, 0.0)
# 2×3 StaticArraysCore.SMatrix{2, 3, Float32, 6} with indices SOneTo(2)×SOneTo(3):
#  0.0  -1.0  1.0
#  1.0   0.0  0.0

# shader
o_uv = uv_transform * vec3(tex_uv, 1.0);
```

SimonDanisch:
I thought about adding `uv_remap=uv->reverse(uv)`  , since that's more flexible and I don't think we have a high need to run this on the shader?
It could also have a few predefined ones like (uv_remap=flipx) or so

SimonDanisch:
Ah I guess you'd need this to upload them per instance for meshscatter... We could allow function and matrix?

ffreyer:
I don't really get the point of passing a Function. When would you ever want to do more than a 2x3 Matrix allows for (which allows scaling, translation, rotation, mirroring, swapping axes, etc)? This isn't particularly expensive (negligible memory since it's per instance or const, negligible performance difference with 10 ops for Mat23 vs 7ish ops current vs 5ish ops on master per vertex shader run. (uv_scale already exists))

SimonDanisch:
>When would you ever want to do more than a 2x3 Matrix allows for

I don't really, but it's much easier to flip a coordinate or reverse it than constructing the corresponding matrix. 
So it's a usability feature to have e.g. `uv_mapping=reverse`.
I'm open to other solutions, but I'd love to have a solution that is both generic and user friendly.

SimonDanisch:
I'm happy to have an enum for the most common operations that either get converted to a matrix or get handled in the shader directly (could be more performant to flip something in the shader, or do no transformation at all, than always multiplying my a 3x2 matrix).

ffreyer:
I think worrying about performance of a matrix multiplication is one of this early optimization traps. With SIMD `if enum == option1 ... ` has a tough battle to fight...

I'm also don't think this is a particularly user-facing feature. Seems more like something a power user may rarely need, at least if we handle defaults well. But I also think we can make this fairly simple by just providing some good functions for generating matrices. Or maybe making `UVTransform` just magical enough (e.g. pre-applying permutations and mirroring).
Also the current solutions to fixing issues with uvs/textures are still just as accessible (manipulating the image or the mesh)

SimonDanisch:
>With SIMD if enum == option1 ...  has a tough battle to fight...

I thought we'd just `{{interpolate}}` it into the shader like we do with buffers etc, so that we don't always have to pay the overhead.
Our `meshscatter` _is_ in fact already quite slow, so I'm happy to add the shader interpolation to when performance problems actually come up which may then need some more tricks anyways.

>I'm also don't think this is a particularly user-facing feature.

I thought you were thinking to add this to other plot functions? E.g. flipping images is quite common: image(img; uv_transform=flipx)`. Although, for image i probably wouldn't like to call it uv_transform...?

Anyways, as I said I'm quite happy to have other solutions, e.g. making `UVTransform` more magical ;) 
I'd add aliases though for the most common operations, just like the enum ;)

If this is ballooning this PR, I'm happy to cut down on usability for now, and just keep those things in mind to be added another time.

asinghvi17:
If we do switch to meshscatter for Tyler, I highly recommend keeping the API - I would need individual meshes for Tyler to work seamlessly with GeoMakie transforms (which are not translateable to GPU).

SimonDanisch:
@asinghvi17 this should be discussed at Tyler.jl?

ffreyer:
I was planning on mesh and surface because they already had uv_scale. I didn't plan for heatmap and image, but since image relies on the mesh pipeline (I think?) I guess I should add it there too. Shouldn't be much work

In terms of bloating I'm more worried about WGLMakie because that doesn't allow textured meshscatter as far as I can tell.

asinghvi17:
Fair, will open an issue over there to track whether we want the change or not.   

As for UV modification, I agree that complex changes (setting UV based on some vertex computation for example) should be done by directly changing the mesh UVs, and we can point to that in the docs.  

I can see some value in having direct functions or enums - I could use UV remappers in GeoMakie's `meshimage` recipe, to preserve transformation fidelity/locality when zooming in (at least in 2D axes).  I'd only need offset + scale for that, though.

The reversing / flipping usecase is also pretty useful if you have a super large image (say 5000x10000) and want to plot that as an image without issues.  The current approach means you have to `rotr90` such an image, which is already in memory and thus might actually overload your RAM if you do this for a sufficiently large image.  You _could_ modify mesh UVs yourself, which is not an end user thing, to display it correctly in Makie.

That being said, some of my usecases do have ~300,000 to ~1,000,000 vertices, so I'm not sure how inefficient it would be, relatively, to go with simple matrix multiplication, as opposed to some enum (or subtype + interface) approach.  The advantage of a subtype + interface thing is that you could theoretically encode arbitrarily complex transformations so long as they were written in OpenGL - not sure if that's a thing we want to support, but that is a possibility.

asinghvi17:
@ffreyer I found a few discourse posts indicating that you could use a shader-level UV modification, same as OpenGL, to be able to do this in Three.js:

https://discourse.threejs.org/t/how-to-change-texture-per-instance-in-instancedmesh/36304
https://discourse.threejs.org/t/how-to-set-different-textures-on-each-instancedmesh/29433
https://discourse.threejs.org/t/how-to-apply-offsets-for-texture-atlas-in-instancedmesh/33191/4
https://jsfiddle.net/prisoner849/g2cpv675/

There are a lot more accessible with a quick google, this was just the few I could find which seemed reasonable to me.

lazarusA:
ahh... for this one, I think the most useful thing I was thinking back then was to do a 3d `Colorbar` for surface plots using RPRMakie :D, now is so much more.  Hopefully now some version of this will land in master 😄 .

GLMakie/src/glshaders/particles.jl
```diff
@@ -75,6 +75,24 @@ function draw_mesh_particle(screen, p, data)
...
+                    output[3 * (i-1) + 1] = transforms[i][Vec(1, 2)]
+                    output[3 * (i-1) + 2] = transforms[i][Vec(3, 4)]
+                    output[3 * (i-1) + 3] = transforms[i][Vec(5, 6)]
+                end
+                return output
+            end => TextureBuffer
+        end
+    else 
+        # handled automatically
+    end
```
ffreyer, line 95: This is a bit more complicated because we use textures rather than instance buffers. Textures don't support matrix inputs and only up to 4 component vectors, so I'm repacking the matrix into Vec2f's here. Those should match up with memory order of mat3x2 in glsl if I got things correct.

src/conversions.jl
```diff
@@ -895,6 +895,54 @@ to_3d_scale(x::VecTypes) = to_ndim(Vec3f, x, 1)
...
+    if action == :rotr90
+        return uv_transform(pi/2)
+    elseif action == :rotl90
+        return uv_transform(-pi/2)
+    elseif action == :swap_xy
+        return Mat3f(0,1,0, 1,0,0, 0,0,1)
+    else
+        error("Transformation :$action not recognized.")
+    end
+end
```
ffreyer, line 944: My current setup for the conversion pipeline allows:
- direct `Mat{2, 3, Float32}`
- `Mat3f` because these are easier to combine
- via `uv_transform()`:
  - `scale::VecTypes{2, <: Real}`
  - `(translation::VecTypes{2, <: Real}, scale::VecTypes{2, <: Real)`
  - `rotation::Real` (2D rotation, single angle in radians)
  - `(translation, scale, rotation)`
  - `named::Symbol`
- vectors of any of the above
- `nothing` which should remove the matmult in shaders

I'm wondering if 
- `uv_transform` should just be a method of `transformationmatrix()`
- we should allow a tuple of multiple symbols which then combine operations (e.g. `(:flip_x, :rotr90)` combining to do both) but I'm worried that the order of operations will be confusing

Also feel free to suggest some more named operations.

ffreyer:
Maybe this is also a good point to reassess the default orientation of images in plots. Calling `image(cow)` and equivalent we have two groups on master:

### Image, Heatmap and Surface (in GLMakie) produce:
Screenshot 2024-07-30 210317

### And mesh, meshscatter (with a Rect2f) produce
image

Should I just reproduce these or do we want to change them?

SimonDanisch:
>Should I just reproduce these or do we want to change them?

Lets reproduce them for now so we can get this in the next non breaking release, and then make a small PR for the next breaking release to change the orientation :)

src/conversions.jl
```diff
@@ -895,6 +895,68 @@ to_3d_scale(x::VecTypes) = to_ndim(Vec3f, x, 1)
...
+function uv_transform(translation::VecTypes{2, <: Real}, scale::VecTypes{2, <: Real})
+    return Mat3f(
+        scale[1], 0, 0,
+        0, scale[2], 0,
+        translation[1], translation[2], 1
+    )
+end
+function uv_transform(translation::VecTypes{2, <: Real}, scale::VecTypes{2, <: Real}, angle::Real)
+    return Mat3f(
+         scale[1] * cos(angle), scale[2] * sin(anlge), 0,
```
asinghvi17, line 934:
```suggestion
         scale[1] * cos(angle), scale[2] * sin(angle), 0,

File: src/conversions.jl [hunk 4]
```diff
@@ -895,6 +895,68 @@ to_3d_scale(x::VecTypes) = to_ndim(Vec3f, x, 1)
...
+        -scale[1] * sin(angle), scale[2] * cos(angle), 0,
+        translation[1], translation[2],                1
+    )
+end
+function uv_transform(angle::Real)
+    return Mat3f(
+         cos(angle), sin(anlge), 0,
```
asinghvi17, line 941:
```suggestion
         cos(angle), sin(angle), 0,

File: src/conversions.jl [hunk 5]
```diff
@@ -895,6 +895,68 @@ to_3d_scale(x::VecTypes) = to_ndim(Vec3f, x, 1)
...
+# convert_attribute(::Automatic, ::key"uv_transform", ::key"image") = Mat{2, 3, Float32}(0, 1, 1, 0, 0, 0)
+
+# defaults matching master
+# Note - defaults with Patterns should be identity (handled in backends)
+convert_attribute(::Automatic, ::key"uv_transform", ::key"meshscatter") =  Mat{2, 3, Float32}(0, 1, -1, 0, 1, 0)
+convert_attribute(::Automatic, ::key"uv_transform", ::key"mesh") =  Mat{2, 3, Float32}(0, 1, -1, 0, 1, 0)
+convert_attribute(::Automatic, ::key"uv_transform", ::key"surface") = nothing
+convert_attribute(::Automatic, ::key"uv_transform", ::key"image") = Mat{2, 3, Float32}(1, 0, 0, -1, 0, 1)
+
+convert_attribute(x::Vector, k::key"uv_transform") = convert_attribute.(x, (k,))
```
asinghvi17, line 913:
```suggestion
convert_attribute(x::AbstractVector, k::key"uv_transform") = convert_attribute.(x, (k,))
```
This might make sense if someone passes in a view or something

src/conversions.jl
asinghvi17: Maybe also something like `transpose`?  Given how flexible this is, it would be pretty easy to have a series of structs with stored values (e.g. `RestrictTo(bbox)` if you want to use only some subsection of a texture, etc), which you then implement `uv_transform(::RestrictTo)` for.  Is that understanding of the pipeline correct?

src/conversions.jl
ffreyer: This one would collide with with the uv_transform() default, since Vec3f <: AbstractVector

src/conversions.jl
ffreyer: I added transpose.
Instead of `RectrictTo` we could just use `Rect2f`` directly. Though you can also use (translation, scale) already.

> uv_transform should just be a method of transformationmatrix()

No, because some of the transformations assume $0 \le uv \le 1$ and don't make sense in a general matrix context

ffreyer:
I dropped rotations from `uv_transform` because it's not clear to me how to handle them. For 0..1 uvs you'd need to translate them to -0.5..0.5, then rotate, then translate back to rotate the texture. For patterns you'd want them to just rotate. If you think about uvs directly, or about combining operations as `uv_transform(op2) * uv_transform(op1)` it also makes more sense for it to just rotate. So the common use case is rather at odds with all the less common ones.

I also added a refimg now, testing a couple of transforms for mesh, surface, meshscatter and image:
grafik

ffreyer:
WGLMakie should be working now too.

For per-element uv_transforms in meshscatter I added the piece-by-piece cow as a refimg:
Screenshot 2024-08-02 134803

PR #1421, opened 2021-11-01
fix camera replacement

ffreyer:
On master
```julia
scene = Scene()
cam3d!(scene)
cam3d!(scene)
```
errors because there is no `disconnect!(scene.camera_controls)`. That's by design - all the camera observables are held and cleaned up by `scene.camera`.

### DIFF
--- a/src/scenes.jl
+++ b/src/scenes.jl
@@ -301,7 +301,6 @@ cameracontrols(scene::Scene) = scene.camera_controls
 cameracontrols(scene::SceneLike) = cameracontrols(scene.parent)
 
 function cameracontrols!(scene::Scene, cam)
-    disconnect!(scene.camera_controls)
     scene.camera_controls = cam
     return cam
 end


SimonDanisch:
Hm, I think that's the correct behaviour - `disconnect!` just bitrot -.- 
You shouldn't be able to connect a new camera without disconnecting the old one.

ffreyer:
The way it's designed atm is that `Camera` holds on to all the observables related to camera controls and `disconnect!` is called on it.
https://github.com/JuliaPlots/Makie.jl/blob/c3bfb04afd20b350dc82aef7fbb6b03f7987f226/src/camera/camera.jl#L13-L19

The call to `disconnect!` already happens in the relevant constructors:
https://github.com/JuliaPlots/Makie.jl/blob/c3bfb04afd20b350dc82aef7fbb6b03f7987f226/src/camera/camera2d.jl#L31
https://github.com/JuliaPlots/Makie.jl/blob/c3bfb04afd20b350dc82aef7fbb6b03f7987f226/src/camera/camera3d.jl#L151
https://github.com/JuliaPlots/Makie.jl/blob/c3bfb04afd20b350dc82aef7fbb6b03f7987f226/src/camera/old_camera3d.jl#L45

SimonDanisch:
Huh, seems like you're right :D Guess I haven't touched that code in a long time...

PR #1423, opened 2021-11-02
Fix #1209

### LINKED ISSUE #1209: `not_implemented_for` is not defined where it is used
This code fails:

```julia
@recipe(AAA) do scene
    Theme(backgroundcolor=:gray90)
end

function Makie.plot!(p::AAA)
    lines!(p,rand(10))
end

aaa()
```

I am not sure whether my code is wrong or not, but the actual reported error is `UndefVarError: not_implemented_for not defined` seemingly raised from https://github.com/JuliaPlots/Makie.jl/blob/ab654211c3d67e584aca783d8bc339e25189c045/MakieCore/src/recipes.jl#L170 because the definition in https://github.com/JuliaPlots/Makie.jl/blob/9268b7551aea81069b5a6451d8f3ab5b7975de56/src/interfaces.jl#L1 does not seem available to it.

lassepe:
This moves `not_implemented_for` to `MakieCore` to fix #1209.

I'm not sure if the way I imported it in `Makie` is the correct/preferred way. I failed to understand the pattern that is used to organize the imports.


lassepe:
Also, it may be good to add a test for this but I'd need a pointer where that should go.

ffreyer:
If that's something you can test after `using Makie` I'd put it in Makie's tests (i.e. somewhere in `Makie.jl/test/`)

lassepe:
The test failures of the docs seem related.  Looks like some macro hygene issue; i.e., `not_implemented_for` seems to be undefined at the callsite of the macro (?).

SimonDanisch:
Alright, the docs project just didn't dev MakieCore...

Issue #1055, opened 2021-06-17
Document `dodge` keyword argument to boxplot, etc


CameronBieganek:
It would be nice to add some documentation on the `dodge` keyword argument to the docstrings for `boxplot`, `barplot`, `violin`, etc.

CameronBieganek:
Maybe also document the `dodge_gap` keyword argument?

PR #1428, opened 2021-11-04
Add dodge in boxplot docstring

HenriDeh:
I think it is a useful addition to the docstring.

### DIFF
--- a/src/stats/boxplot.jl
+++ b/src/stats/boxplot.jl
@@ -27,6 +27,7 @@ The boxplot has 3 components:
 - `whiskerwidth`: multiplier of `width` for width of T's on whiskers, or
     `:match` to match `width`
 - `show_outliers`: show outliers as points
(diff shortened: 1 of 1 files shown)


HenriDeh:
This closes #1055

src/stats/boxplot.jl
```diff
@@ -27,6 +27,7 @@ The boxplot has 3 components:
 - `whiskerwidth`: multiplier of `width` for width of T's on whiskers, or
     `:match` to match `width`
 - `show_outliers`: show outliers as points
+- `dodge`: vector (length of data) of grouping variable to create multiple side-by-side boxes at the same `x` position
```
piever, line 30: Could be worth to mention that it has to be a vector of integers (at the moment there is no handling of categorical data). To close #1055 this should also document `dodge_gap`, but that can be done in a later PR

Issue #1388, opened 2021-10-15
Clean up plot attributes


SimonDanisch:
- [ ] go through all basic plot types and make sure they have a consistent, well defined set of attributes
- [ ] move all color related attributes to `color` (e.g. colormap)
- [ ] consistently allow for the same attributes in colormap (e.g. `high_clip`, etc)
- [ ] add alpha attribute #84
- [ ] move all shading attributes to e.g. `material`
- [ ] move all light attributes to scene
- [ ] make sure all attributes are documented and tested
- [ ] make sure it's clear what types are allowed for each attribute
- [ ] rename backend attributes like fxaa, to make the cross backend, or move them to something like backend-attributes
- [ ] good error messages for wrong plot attributes
- [ ] better API for backends. Right now, it's pretty fuzzy to see what the basic plot types support, leading to bugs + making it more difficult to write new backends/maintain them

ffreyer:
We should also clean up the Attribute handling in recipes. Currently some recipes inherit attributes, some don't, some are probably missing attributes and passing them along is often very explicit.

For example `arc` inherits with `default_theme(scene, Lines)` and passes `Attributes(p)`, whereas `spy` doesn't inherit any attributes (which might be reasonable), is missing some general attributes like `visible` and passes explicitly with `name = p.name`. 

In that context it might also be good to cleanup/change `default_theme`. Currently it has some defaults for attributes that belong to every plot like `visible`, some that apply to a category like `diffuse` (only relevant for 3D plots) and some that target one or two specific plot primitive like `linewidth`. Having `default_theme(scene)` return just generally applicable attributes and `default_theme(scene, Plot2D/Plot3D)` for 2D and 3D specific attributes would make writing recipes easier I'd say. A constructor like `Attributes(parent...; kwargs...) = merge(parents..., kwargs)` would also be useful I think.

ffreyer:
checked off some points here (none were checked before)

PR #1436, opened 2021-11-07
Clean up color handling in shaders

ffreyer:
My goal with this is to move most if not all the color fetching code into one shader so we don't have as much duplicate code. 

To avoid having tons of methods differentiating `Nothing image` from `sampler2D image` etc, I'm also proposing that we multiply all the different sources. That shouldn't change anything in most of the current use cases but it would reduce the amount of code in shaders and allow, well, mixing of textures, colors etc. If we go through with these changes it may also make sense to also separate textures, colormap indices and patterns from `color`.

Some wild mixing examples:
```julia
img = FileIO.load(Makie.assetpath("cow.png"))
c1 = RGBAf(1, 1, 1, 1)
c2 = RGBAf(0, 0, 0, 0)
pattern = [
    c1 c1 c2; 
    c1 c1 c1; 
    c1 c1 c1
]
colors = [:red, :green, :blue, :gray, :white]
fig, ax, p = scatter(
    rand(Point2f, 5), 
    color = colors, 
    image = img, 
    pattern = pattern, 
    markersize=80, strokewidth=2, uv_scale = Vec2f(0.2)
)
```

Screenshot from 2021-11-07 20-06-35

```julia
cat = FileIO.load(Makie.assetpath("cat.obj"))
img = FileIO.load(Makie.assetpath("diffusemap.tga"))
matcap = FileIO.load(download("https://raw.githubusercontent.com/nidorx/matcaps/master/thumbnail/54584E_B1BAC5_818B91_A7ACA3.jpg"))
c1 = RGBAf(1, 1, 1, 1)
c2 = RGBAf(0.5, 1.5, 1, 1)
pattern = [
    c1 c1 c2; 
    c1 c1 c1; 
    c1 c1 c1
]
mesh(
    cat, 
    color = RGBAf(1, 0.9, 0.4, 1),
    image = img,
    matcap = matcap,
    pattern = pattern,
    uv_scale = Vec2f(0.3),
    lightposition = Vec3f(0, 2, 0)
)
```

Screenshot from 2021-11-07 20-29-28

Ref #1388

### DIFF
--- a/GLMakie/assets/shader/colors.frag
+++ b/GLMakie/assets/shader/colors.frag
@@ -1,0 +1,98 @@
+{{GLSL_VERSION}}
+{{GLSL_EXTENSIONS}}
+{{SUPPORTED_EXTENSIONS}}
+
+struct Nothing{ //Nothing type, to encode if some variable doesn't contain any data
+    bool _; //empty structs are not allowed
+};
+
+{{image_type}} image;
+{{pattern_type}} pattern;
+{{matcap_type}} matcap;
+{{color_map_type}}  color_map;
+{{color_range_type}} color_range; 
+
+uniform vec4 highclip;
+uniform vec4 lowclip;
+uniform vec4 nan_color;
+uniform vec2 uv_scale;
+
+
+////////////////////////////////////////////////////////////////////////////////
+/// fragment shader functions
+////////////////////////////////////////////////////////////////////////////////
+
+
+// implementations 
+vec4 _texture_color(sampler1D t, float u){return texture(t, u);}
+vec4 _texture_color(sampler2D t, vec2 uv){return texture(t, uv);}
+vec4 _texture_color(samplerBuffer t, int i){return texelFetch(t, i);}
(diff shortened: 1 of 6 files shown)


ffreyer:
There are two major TODOs here:

1. Figure out a way to set and mix all the different color uniforms from Makie. Currently most go through the `color` attribute and get mapped to various uniforms based on their type. So if you want to throw a pattern on your cow markers you need to pass on of them with a uniform name. That's kind of a hack and (if I recall correctly) resulted in some issues setting defaults in this pr. Simon's suggestion was to allow passing multiple different uniforms through `color`, e.g. something like  `color = (color = :red, texture = cat_texture)`. The remaining uniforms would then get filled out with default colors or `nothing`, depending on what makes sense.
2. There are a couple of vertex shaders that set colors. These should also run via a common `colors.vert` to simplify/unify their behavior.

Also it looks like the merge got messed up

SimonDanisch:
Should we close this until we have a better idea about what we really want here?

PR #1444, opened 2021-11-12
ecdf plot plus internal fixes

piever:
@sethaxen, I'm continuing #1310 here, because it looks like it uncovered a small issue in Makie's conversion mechanism that should be addressed at the same time.


src/interfaces.jl
```diff
@@ -150,7 +150,7 @@ function apply_convert!(P, attributes::Attributes, x::PlotSpec{S}) where S
     for (k, v) in pairs(kwargs)
         attributes[k] = v
     end
-    return (plottype(P, S), args)
+    return (S, args)
```
piever, line 153: @SimonDanisch, I'm trying your suggestion here (which fixes the `ECDFPlot` bug) but I was wondering: should this actually be `(plottype(S, P), args)`, so that the plot type suggested by the conversion pipeline, i.e. `S`, takes precedence, but if that is `Any`, it falls back to `P`?

SimonDanisch: Ah... maybe that was the intention... Yeah, if `plottype(S, P)` does that, it sounds like a plan ;)

sethaxen:
> @sethaxen, I'm continuing #1310 here, because it looks like it uncovered a small issue in Makie's conversion mechanism that should be addressed at the same time.

@piever it would be cleaner to separate the ECDF plot PR, which does not touch Makie's internals, from this PR that changes the internals of Makie. I'd prefer if you made the necessary changes separately and then I use them in that PR.

piever:
> @piever it would be cleaner to separate the ECDF plot PR, which does not touch Makie's internals, from this PR, which does change the internals of Makie. I'd prefer if you made the necessary changes separately and then I use them in that PR.

Sure! Closing this in favor of #1310 and #1445 then

sethaxen:
Thanks!

Issue #771, opened 2020-12-02
Example recipe in docs doesn't work


yha:
The docs at http://makie.juliaplots.org/stable/recipes.html#Full-recipes-with-the-@recipe-macro suggest that this should work (although it's not presented as a complete example in a single block, but in two parts):
```
# arguments (x, y, z) && theme are optional
@recipe(MyPlot, x, y, z) do scene
    Theme(
        plot_color => :red
    )
end

function plot!(plot::MyPlot)
    # normal plotting code, building on any previously defined recipes
    # or atomic plotting operations, and adding to the combined `plot`:
    lines!(plot, rand(10), color = plot[:plot_color])
    plot!(plot, plot[:x], plot[:y])
    plot
end

```
There are two trivial problems: `plot_color=>:red` should be `plot_color=:red`, and `plot!` should be fully qualified to extend it (or there should be `import AbstractPlotting: plot!`). 
After fixing these, it still fails:
```
julia> myplot(1:4,1:4,1:4)
MethodError: no constructors have been defined for Any
plot!(::Combined{myplot,Tuple{UnitRange{Int64},UnitRange{Int64},UnitRange{Int64}}}, ::Type{Any}, ::Attributes, ::Observable{UnitRange{Int64}}, ::Observable{UnitRange{Int64}}) at interfaces.jl:722
plot!(::Type{Any}, ::Combined{myplot,Tuple{UnitRange{Int64},UnitRange{Int64},UnitRange{Int64}}}, ::Observable{UnitRange{Int64}}, ::Vararg{Observable{UnitRange{Int64}},N} where N; kw_attributes::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at interfaces.jl:565
plot!(::Type{Any}, ::Combined{myplot,Tuple{UnitRange{Int64},UnitRange{Int64},UnitRange{Int64}}}, ::Observable{UnitRange{Int64}}, ::Observable{UnitRange{Int64}}) at interfaces.jl:564
plot!(::Combined{myplot,Tuple{UnitRange{Int64},UnitRange{Int64},UnitRange{Int64}}}, ::Vararg{Any,N} where N; attributes::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at recipes.jl:16
[...]
myplot(::UnitRange{Int64}, ::Vararg{UnitRange{Int64},N} where N) at recipes.jl:12
[...]

PR #1448, opened 2021-11-16
Fix recipe docs

jonas-schulze:


docs/documentation/recipes.md
```diff
@@ -99,7 +99,9 @@ myplot!(args...; kw_args...) = ...
 A specialization of `argument_names` is emitted if you have an argument list
 `(x,y,z)` provided to the recipe macro:
 
-    `argument_names(::Type{<: MyPlot}) = (:x, :y, :z)`
+```julia
+argument_names(::Type{<: MyPlot}) = (:x, :y, :z)
+```
 
 This is optional but it will allow the use of `plot_object[:x]` to
```
jonas-schulze, line 106: Later on in the stocks example, the arguments are fetched like `plot_object.x`. Is both supported or you should the above be changed to match the example?

jonas-schulze:
Just found out about #771 and friends. So maybe this isn't needed anymore?

docs/documentation/recipes.md
SimonDanisch: I prefer `plot.x` and would like to deprecate `plot[:x]` at some point, or change its meaning...

jonas-schulze: Should I change this in the docs to bridge the gap while you are working on the overhaul? I couldn't find where `plot.x` get's wired into `getproperty`, so I didn't feel comfortable doing the change right away. Here are some other occurences:

* https://github.com/jonas-schulze/Makie.jl/blob/patch-1/MakieCore/src/recipes.jl#L121-L122
* https://github.com/jonas-schulze/Makie.jl/blob/patch-1/MakieCore/src/recipes.jl#L126
* https://github.com/jonas-schulze/Makie.jl/blob/patch-1/docs/documentation/recipes.md?plain=1#L112

SimonDanisch: Sorry, somehow missed your comment... Yeah I think that's a good idea!

Its here: https://github.com/JuliaPlots/Makie.jl/blob/78820311c9b45347f0937caf8b31a25bbc27dd73/MakieCore/src/attributes.jl#L189

And then getproperty uses getindex here:
https://github.com/JuliaPlots/Makie.jl/blob/78820311c9b45347f0937caf8b31a25bbc27dd73/MakieCore/src/attributes.jl#L83
(Not very straightforward to find, especially with the `Transformable` supertype)

PR #1452, opened 2021-11-16
disable rectangle zoom if we don't have inverse

SimonDanisch:
The warning was there correctly, but somehow `inverse_transform(s) = nothing` got lost...
Also improves the warning, by only displaying it one time.


Datseris:
Is this in a tagged release? It is necessary for GeoMakie to be tagged, right? What's the version up limit I should put to GeoMakie?

PR #1469, opened 2021-11-23
extensible axistheme function

jkrumbiegel:


jkrumbiegel:
I'll close this for now, we can revisit the implementation idea when we get to Axis recipes.

PR #1474, opened 2021-11-24
Create PULL_REQUEST_TEMPLATE.md

jkrumbiegel:


SimonDanisch:
This seems somewhat redundant:
> This change requires a documentation update

> Added or changed relevant sections in the documentation

I think we should just keep the latter?

PR #1481, opened 2021-11-26
start adding tests for primitives

jkrumbiegel:


SimonDanisch:
Should we merge this? I guess we can extend this a lot later on, but doesn't really need to be in one go!

SimonDanisch:
Going to merge this and integrate with the tests better, when I work on updating all tests

Issue #1476, opened 2021-11-25
Rotating Scene in CairoMakie is not working properly


davibarreira:
I was trying to rotate a plot (and axis) to produce a visualization like this one below:
![image](https://user-images.githubusercontent.com/6407557/143333206-cf0b4206-8a6f-4d2f-864c-08b232fcd80d.png)
I've posted the question and the following answer was given:
```julia
fig = Figure()
b = Axis(fig[1, 1])
scene = Scene(fig.scene)
campixel!(scene)
ax = Axis(scene, bbox = MakieLayout.BBox(-100, 100, -100, 100))

for s in (scene, ax.scene)
    Makie.rotate!(s, Vec3f(0,0,1), -pi/4)
    translate!(s, Vec3f(300, 300, 100))
end
fig
```
Unfortunately, this does not work on CairoMakie. As pointed out by @ffreyer who posted this solution: "This works for me in GLMakie but CairoMakie still has two problems - text gets transformed differently and disappears, and the background pane of the rotated axis doesn’t draw over the normal axis. I think these are things that need to fixed in CairoMakie though."

ffreyer:
Text rotation has been fixed so I assume this can be closed

PR #1482, opened 2021-11-27
fix model transformation of heatmaps and text in CairoMakie

### LINKED ISSUE #1298: CairoMakie rotated heatmap does not match GLMakie
On the left is the CairoMakie output, on the right is GLMakie
image

The code

```julia
using Makie: Makie, RGBA
using GLMakie: GLMakie
using CairoMakie: CairoMakie

GLMakie.activate!()

fig = Makie.Figure()
ax = Makie.Axis(fig[1, 1])
ax.title = "title"

lsgrid = Makie.labelslidergrid!(
    fig,
    ["x", "y", "z", "deg"],
    [0:0.01:1, 0:0.01:1, 0:0.01:1, 0:1.0:360];
)
Makie.set_close_to!(lsgrid.sliders[3], 1.0) # lsgrid.sliders[3].value[] = 1.0
fig[2, 1] = lsgrid.layout

sliderobservables = [s.value for s in lsgrid.sliders]
quat = Makie.lift(sliderobservables...) do (arg...)
    (x, y, z, deg) = arg
    Makie.qrotation(Makie.SVector(x, y, z), deg2rad(deg))
end

ax.autolimitaspect = 1
# Makie.hidespines!(ax)
# Makie.hidedecorations!(ax) # remove ticks and tick labels
Makie.heatmap!(ax, rand(Bool, 10, 10); colormap=[RGBA(0, 0, 0, 0), RGBA(0, 0, 0, 1)])

Makie.on(quat) do q
    Makie.rotate!(fig.scene.children[1], q)
end

display(fig)

function save_pdf(fig)
    CairoMakie.activate!()
    Makie.save("/tmp/out.pdf", fig)
    GLMakie.activate!()
end

```

```
  [13f3f980] CairoMakie v0.6.5
  [e9467ef8] GLMakie v0.4.6
```

ffreyer:
Fixes #1298

GLMakie vs CairoMakie:

Screenshot from 2021-11-28 00-07-32

The main change here is using a polygon given by corner points rather than a rectangle that is always screen aligned. I tried adjusting things via a Cairo transform matrix but that does not work without a bunch of extra effort. (I assume the problem is that the model matrix should be applied first but is applied last if passed as a Cairo transform matrix. ) 
I also changed the bulging to be applied in every direction because with rotations gaps may open up otherwise.


ffreyer:
Test failures are unrelated (some latex tests fail which also fail in other prs atm)

ffreyer:
Also fixes #1476, i.e. that the model matrix is ignored in text.

Screen space: (GLMakie, CairoMakie)

Screenshot from 2021-11-28 12-16-54

Data space: (GLMakie, CairoMakie)

Screenshot from 2021-11-28 13-12-27

Issue: (CairoMakie)

Screenshot from 2021-11-28 13-14-22

jkrumbiegel:
Very cool, thanks! Does bulging on every side mean that the heatmap grows larger than its actual extent? Also, I don't quite understand what the problem about the rectangle was. Was it that you needed to rotate it manually given the model matrix, then draw it? (at which point it's not a rectangle anymore) The only problem I see with this is that it probably increases the size of an exported svg a bit, because a polygon with four corners needs more data than a rectangle. Maybe it's also a bit slower?

ffreyer:
> Very cool, thanks! Does bulging on every side mean that the heatmap grows larger than its actual extent?

I suppose. But now that I think about it again rotations shouldn't matter for shaving off the outside bulging. 

> Also, I don't quite understand what the problem about the rectangle was. Was it that you needed to rotate it manually given the model matrix, then draw it? 

As far as I know Cairo only has a `rectangle(x0, y0, width, height)`, so that is always axis-aligned. So the options for applying a model matrix with rotations are:
1. Derive some matrix T such that `T * project(point, model = I) = project(point, model = model)` and apply that as a transformation matrix for each rectangle. I'm not sure if this can even work.
2. Separate the rotation from the model matrix and apply it after `project` via `Cairo.rotate!`. I think this would be difficult if we assume `model` to be some general matrix. (I.e. if the user sets it as `model = ...` rather than via Makie transformations)
3. Include `model` in `project`. Now the rectangle might be rotated so giving it as origin + widths doesn't work anymore. Though I guess we could also calculate a rotation at this point, draw normal rectangles and apply it via `Cairo.rotate`...

> (at which point it's not a rectangle anymore) 

Rotation, translation and scale shouldn't change that. Though I guess a general model matrix can also shear...

> The only problem I see with this is that it probably increases the size of an exported svg a bit, because a polygon with four corners needs more data than a rectangle. Maybe it's also a bit slower?

Yea that's what I'd assume. Defining things with transformation matrices probably just trades space for time here. We could add another fast path here if that's actually a problem. (I.e. if model [1:3, 1:3] is diagonal and model[1:3, 4] == 0 we should only have scaling and translation so we can use rectangles.)

ffreyer:
> But now that I think about it again rotations shouldn't matter for shaving off the outside bulging.

Nope, using
```julia
if alpha(colors[i, j]) == 1
    center = 0.25 * (p1 + p2 + p3 + p4)
    p1 += sign.(p1 - center) .* Point2f(0.5(i==1),  0.5(j==1))
    p2 += sign.(p2 - center) .* Point2f(0.5(i==ni), 0.5(j==1))
    p3 += sign.(p3 - center) .* Point2f(0.5(i==ni), 0.5(j==nj))
    p4 += sign.(p4 - center) .* Point2f(0.5(i==1),  0.5(j==nj))
end
```
should trim the outer edges but results in some gaps depending on the rotation:
Screenshot from 2021-11-29 20-35-23

ffreyer:
I messed up the comparison, it works fine with `!=`. Now the heatmap should be as large as requested again.

goretkin:
> > The only problem I see with this is that it probably increases the size of an exported svg a bit, because a polygon with four corners needs more data than a rectangle. Maybe it's also a bit slower?
> 
> Yea that's what I'd assume. Defining things with transformation matrices probably just trades space for time here. We could add another fast path here if that's actually a problem. (I.e. if model [1:3, 1:3] is diagonal and model[1:3, 4] == 0 we should only have scaling and translation so we can use rectangles.)

I imagine that it's possible to group together all the rectangles, and then transform the group with the [SVG transform attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform)

PR #1487, opened 2021-12-01
fix log ticklabels with very small or large exponents

jkrumbiegel:


glwagner:
Might close https://github.com/JuliaPlots/Makie.jl/issues/1490 ?

PR #1493, opened 2021-12-02
Added axes-linking to layout tutorial

eirikbrandsaas:
a) This is a very useful command
b) The finished plot clearly implies the `axtop` and `axmain` have the same x-axis and  `axright` and `axmain` the same y-axis.

### DIFF
--- a/docs/tutorials/layout-tutorial.md
+++ b/docs/tutorials/layout-tutorial.md
@@ -91,6 +91,18 @@ f
 ```
 \end{examplefigure}
 
+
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
I happenend to do the same thing a couple days ago :) https://github.com/JuliaPlots/Makie.jl/pull/1477/files

You probably didn't notice because it's only on `dev` so far, not `stable`, but still, thanks!

eirikbrandsaas:
Good to know that the mediocre minds can think alike to great minds, just with a delay of a few days! 

Thank you for the great package.

PR #1494, opened 2021-12-02
Fix blurry text

### LINKED ISSUE #1470: Blurry menus / scaling issues using multiple monitors
I'm working on a 13" 2018 MacBook Pro having a WQHD monitor connected via HDMI (over a USB dock). When I try the [menu example](https://makie.juliaplots.org/stable/examples/layoutables/menu/), the menu looks very blurry on the external monitor:

monitor

When I move the window to the laptop's screen, the menus look much smaller and therefore sharper as well:

macbook

#1186 is also visible when opening the upper menu.

ffreyer:
Closes #1470

The problem here was that the glyph padding did not make it into the texture atlas. 

Makie uses a signed distance field approach, which essentially encodes the distance to the closest edge at each pixel. From that you can recover the original symbol by evaluating `color = text_color * (distance > 0)` at each pixel (assuming positive distances inside the symbol). Antialiasing essentially swaps the step function `(distance > 0)` for a smoothed step function (i.e. sigmoid). This means some small negative distances outside the original shape are included. We need to make sure that these are available, otherwise we get artifacts like in #1470.
On current master this is not the case. Padding is included in the texture generation but is discarded when the signed distance fields gets inserted into the texture atlas. This pr changes that - with it the padding is included during insertion and it's increased so that we have 1.5px padding at a textsize of 8px. (1px padding still looked blurry.)

Note that this chunks up the glyph size quite a bit. If we run into "texture atlas is too small. Resizing not implemented yet. Please file an issue at Makie if you encounter this" errors we should be fine decreasing the default glyph resolution to 32px. (In case you find this pr because of that error please do report it. You can clear your cache by deleting the files at `Makie.get_cache_path()` and restarting Julia or `empty!(Makie.global_texture_atlas)`.)

Before and after:
(Textsizes are 8, 10, 16, 20, 32, 64, 100)
master
Screenshot from 2021-12-02 21-19-08

And specifically the figure from #1470 
Screenshot from 2021-12-02 21-48-35

### DIFF
--- a/GLMakie/assets/shader/distance_shape.frag
+++ b/GLMakie/assets/shader/distance_shape.frag
@@ -41,7 +41,7 @@ flat in vec4            f_uv_texture_bbox;
 // These versions of aastep assume that `dist` is a signed distance function
 // which has been scaled to be in units of pixels.
 float aastep(float threshold1, float dist) {
-    return min(1.0, f_viewport_from_u_scale)*smoothstep(threshold1-ANTIALIAS_RADIUS, threshold1+ANTIALIAS_RADIUS, dist);
+    return smoothstep(threshold1-ANTIALIAS_RADIUS, threshold1+ANTIALIAS_RADIUS, dist);
 }
 float aastep(float threshold1, float threshold2, float dist) {
     return smoothstep(threshold1-ANTIALIAS_RADIUS, threshold1+ANTIALIAS_RADIUS, dist) -
--- a/GLMakie/src/gl_backend.jl
+++ b/GLMakie/src/gl_backend.jl
@@ -17,7 +17,6 @@ using .GLAbstraction
 const atlas_texture_cache = Dict{Any, Tuple{Texture{Float16, 2}, Function}}()
 
 function get_texture!(atlas)
-    Makie.set_glyph_resolution!(Makie.High)
     # clean up dead context!
     filter!(atlas_texture_cache) do (ctx, tex_func)
         if GLAbstraction.context_alive(ctx)
--- a/src/layouting/layouting.jl
+++ b/src/layouting/layouting.jl
@@ -310,12 +310,14 @@ function text_quads(positions, glyphs::AbstractVector, fonts::AbstractVector, te
     offsets = Vec2f[]
     uv = Vec4f[]
     scales = Vec2f[]
+    pad = GLYPH_PADDING[] / PIXELSIZE_IN_ATLAS[]
     broadcast_foreach(positions, glyphs, fonts, textsizes) do offs, c, font, pixelsize
     # for (c, font, pixelsize) in zipx(glyphs, fonts, textsizes)
         push!(uv, glyph_uv_width!(atlas, c, font))
         glyph_bb, extent = FreeTypeAbstraction.metrics_bb(c, font, pixelsize)
-        push!(scales, widths(glyph_bb))
-        push!(offsets, minimum(glyph_bb))
+
+        push!(scales, widths(glyph_bb) .+ pixelsize * 2pad)
+        push!(offsets, minimum(glyph_bb) .- pixelsize * pad)
     end
     return positions, offsets, uv, scales
 end
@@ -326,13 +328,14 @@ function text_quads(positions, glyphs, fonts, textsizes::Vector{<:ScalarOrVector
     offsets = Vec2f[]
     uv = Vec4f[]
     scales = Vec2f[]
+    pad = GLYPH_PADDING[] / PIXELSIZE_IN_ATLAS[]
 
     broadcast_foreach(positions, glyphs, fonts, textsizes) do positions, glyphs, fonts, textsizes
         broadcast_foreach(positions, glyphs, fonts, textsizes) do offs, c, font, pixelsize
             push!(uv, glyph_uv_width!(atlas, c, font))
             glyph_bb, extent = FreeTypeAbstraction.metrics_bb(c, font, pixelsize)
-            push!(scales, widths(glyph_bb))
-            push!(offsets, minimum(glyph_bb))
+            push!(scales, widths(glyph_bb) .+ pixelsize * 2pad)
+            push!(offsets, minimum(glyph_bb) .- pixelsize * pad)
         end
     end
 
--- a/src/utilities/texture_atlas.jl
+++ b/src/utilities/texture_atlas.jl
@@ -14,23 +14,29 @@ Base.size(atlas::TextureAtlas) = size(atlas.data)
 
 @enum GlyphResolution High Low
 
-const TEXTURE_RESOLUTION = Ref((2048, 2048))
-const CACHE_RESOLUTION_PREFIX = Ref("high")
-
 const HIGH_PIXELSIZE = 64
 const LOW_PIXELSIZE = 32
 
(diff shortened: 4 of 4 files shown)


jkrumbiegel:
I'm quite happy the bad font quality turned out to be a bug after all, thanks for having a look at this

GLMakie/src/gl_backend.jl
```diff
@@ -17,7 +17,6 @@ using .GLAbstraction
 const atlas_texture_cache = Dict{Any, Tuple{Texture{Float16, 2}, Function}}()
 
 function get_texture!(atlas)
-    Makie.set_glyph_resolution!(Makie.High)
```
ffreyer, line 20: https://github.com/JuliaPlots/Makie.jl/blob/8bab35403db92bb921738380d4c5ecff8df6ec37/WGLMakie/src/WGLMakie.jl#L49

Is this necessary? If it is we should probably have GLMakie set the glyph resolution on init, not here.

ffreyer:
The test failures are the animated surface. I think we can ignore those

GLMakie/src/gl_backend.jl
ffreyer: > If it is we should probably have GLMakie set the glyph resolution on init, not here.

That's already the case.

SimonDanisch:
Amazing!
I was pretty convinced that this is a bug, since in the early days it looked just as crisp, but never got around finding the bug!

GLMakie/assets/shader/distance_shape.frag
```diff
@@ -41,7 +41,7 @@ flat in vec4            f_uv_texture_bbox;
 // These versions of aastep assume that `dist` is a signed distance function
 // which has been scaled to be in units of pixels.
 float aastep(float threshold1, float dist) {
-    return min(1.0, f_viewport_from_u_scale)*smoothstep(threshold1-ANTIALIAS_RADIUS, threshold1+ANTIALIAS_RADIUS, dist);
```
ffreyer, line 44: I should probably check if that changed anything. For text it was irrelevant, but maybe it matters for scatter?

ffreyer:
The failing test is

```julia
lines(Rect(0, 0, 1, 1), linewidth=4)
scatter!([Point2f(0.5, 0.5)], markersize=1, markerspace=SceneSpace, marker='I')
current_figure()
```

which yields

short_tests_13

which matches what I'd expect the marker to look like. Here it is with the render from FreeTypeAbstraction below it:
```julia
using FreeTypeAbstraction
lines(Rect(0, 0, 1, 1), linewidth=4)

font = findfont("Dejavu Sans")
char = 'I'
img, fe = renderface(font, char, 400)
w, h = size(img)
l = 0.5 - w/800
r = 0.5 + w/800
b = 0.5 - h/800
t = 0.5 + h/800
image!(l..r, b..t, img[:, end:-1:1], colormap=(:white, :black))

scatter!(
    [Point2f(0.5, 0.5)], markersize=1, markerspace=SceneSpace, marker=char, 
    color = (:orange, 0.4)
)
current_figure()
```
Screenshot from 2021-12-04 12-52-37

GLMakie/assets/shader/distance_shape.frag
ffreyer: I don't see a difference in characters scatters. I guess it's fine

SimonDanisch:
Yeah it's not unlikely, that we should just update the reference images for that .. we should take the chance and pin down the expected behavior

ffreyer:
The short version of my changes is that a real numbered markersize with a character or string marker is now drawn as scaled version of the respective character. I.e. `scatter` matches in size with `text` if the markersize matches textsize. The old version scaled the quad without caring about what the aspect ratio the glyph has, stretching things like `I`.

The scaling changes result in this:
```julia
scene = Scene()
campixel!(scene)
scatter!(scene, Point2f[(50, 100), (100, 100), (150, 100), (200, 100)], marker="^l_j", markersize=50)
scatter!(scene, Point2f[(50, 100), (100, 100), (150, 100), (200, 100)], color = :red, markersize=4)
text!(scene, ["^", "l", "_", "j"], position = Point2f[(50, 200), (100, 200), (150, 200), (200, 200)], textsize = 50, align=(:center, :center))
scene
```
Screenshot from 2021-12-04 11-15-33

jkrumbiegel:
Perfect, I think that matches how I implemented it in Cairomakie

ffreyer:
Up to maybe a pixel or two the "^l_j" example is the same between CairoMakie, GLMakie and WGLMakie now. Though CairoMakie can't do String markers, only characters so I switched to 4 individual scatter plots.

GLMakie/src/gl_backend.jl
SimonDanisch: Uh, I think I added this, to make sure Makie gets the correct texture atlas - sadly this is a Makie global variable... E.g., when using GLMakie, then WGLMakie and then switching back to GLMakie you'll get pretty weird artifacts

SimonDanisch: Ok, nvm, this works on an existing texture atlas...
`activate!` is the right place: https://github.com/JuliaPlots/Makie.jl/blob/master/GLMakie/src/GLMakie.jl#L42

SimonDanisch:
Sweet!!!!
Do you want to add the test that generated 
image
to maybe:
https://github.com/JuliaPlots/Makie.jl/blob/master/ReferenceTests/src/tests/text.jl
?
I'll update the reference images after we merged everything!

GLMakie/src/gl_backend.jl
ffreyer: I used to get artifacts when doing `empty!(Makie.global_texture_atlas)` and removing the cached files. But I'm not getting them anymore. I don't think I got when switching between WGLMakie and GLMakie either.

ffreyer:
I was using https://github.com/JuliaPlots/Makie.jl/blob/cb9087ee9055313260611e7eafa8e7a0a0184024/GLMakie/assets/shader/distance_shape.frag#L158-L160
for those, so getting the bounding boxes for a ref image test isn't really possible. We could do it without, but there's already  https://github.com/JuliaPlots/Makie.jl/blob/cb9087ee9055313260611e7eafa8e7a0a0184024/ReferenceTests/src/tests/short_tests.jl#L13-L17
which more or less checks the same thing?

Without bounding boxes:
Screenshot from 2021-12-06 14-13-53

SimonDanisch:
Ah I see... Would be nice to test those boundingboxes somewhere still, to see if they match what freetype expects...
Lets merge this regardless! Thanks for the great effort & detective work to fix this!! :)

ffreyer:
They don't because they have padding :p

I guess you could run a character through the text machinery and then check if the `text_quads` output matches the size from FreeTypeAbstraction with `pixelsize * 2 * GLYPH_PADDING[] / PIXELSIZE_IN_ATLAS[]` padding?

jonas-schulze:
Nice work, thank you!

SimonDanisch:
Well, we do need a function that correctly gives us the rendered boundingbox of a glyph for a certain font & fontsize ... I lost a bit track which one that is right now, but maybe  that's a good chance to find the correct one and document & test it!

ffreyer:
Ok I'll give it a try

PR #1499, opened 2021-12-06
Text layouting tests

ffreyer:
I don't think these tests are all that useful since they mostly repeat what Makie does internally. The problem with comparing to `FreeTypeAbstraction` directly (i.e. letting it render at the correct fontsize or the glyph fontsize) is that it doesn't scale characters in float steps but integer steps. So the difference between a 32px and 64px fontsize isn't a factor of 2:

```julia
img32, ext32 = renderface(font, '^', 32)
img64, ext64 = renderface(font, '^', 64)
ext32
# FontExtent{Float64}([-11.0, 11.0], [3.0, 23.0], [27.0, 32.0], [21.0, 9.0])
ext64
# FontExtent{Float64}([-21.0, 23.0], [6.0, 47.0], [54.0, 64.0], [41.0, 17.0])
```

It also seems to do more than just rounding up sometimes:

```julia
renderface(font, '_', 20)[2].scale
#2-element Vec2{Float64} with indices SOneTo(2):
# 12.0
#  2.0

ceil.(Int, FreeTypeAbstraction.get_extent(font, '_').scale * 20)
#2-element Vec2{Int64} with indices SOneTo(2):
# 11
#  2
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -28,4 +28,5 @@ using Makie: volume
     include("transformations.jl")
     include("stack.jl")
     include("events.jl")
+    include("text.jl")
 end
--- a/test/text.jl
+++ b/test/text.jl
@@ -1,5 +1,5 @@
 @testset "Glyph Collections" begin
-    using FreeTypeAbstraction
+    using Makie.FreeTypeAbstraction
 
     # Test whether Makie's padded signed distance field text matches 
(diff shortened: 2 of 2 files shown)


SimonDanisch:
Thank you! I guess we need to overthink the text boundingbox API some more after the next release ^^

PR #1504, opened 2021-12-08
Disconnect other cameras when switching to PixelCamera

ffreyer:
If you do something like this
```julia
fig = Figure()
scene = LScene(fig[1, 1], show_axis = false)
campixel!(scene.scene)
scatter!(scene, 400rand(10), 400rand(10))
fig
```
and then click (and drag) the plot area everything will disappear. That happens because `campixel!` does not disconnect the previous camera, in this case `Camera3D`. The pr fixes that.


ffreyer:
Failing tests are from missing/old refimages. We can merge this

Issue #1511, opened 2021-12-14
Precompiling error


Mo8it:
I get an error after running `Pkg.update()`. The updated packages are:
- Tables
- Compat
- Static
- Interpolations

Makie is dependent on all of them.

---

# Error

julia> using CairoMakie
[ Info: Precompiling CairoMakie [13f3f980-e62b-5c42-98c6-ff1f3baf88f0]
WARNING: Method definition quantile(AbstractArray{var"#s261", N} where N where var"#s261"<:Real) in module StatsBase at deprecated.jl:70 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition stdm(AbstractArray{T, N} where N where T<:Real, AbstractArray{T, N} where N where T<:Real, Int64) in module StatsBase at /home/mo/.julia/packages/StatsBase/CuCbN/src/moments.jl:141 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition stdm##kw(Any, typeof(Statistics.stdm), AbstractArray{T, N} where N where T<:Real, AbstractArray{T, N} where N where T<:Real, Int64) in module StatsBase at /home/mo/.julia/packages/StatsBase/CuCbN/src/moments.jl:141 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition describe(Any) in module StatsBase at /home/mo/.julia/packages/StatsBase/CuCbN/src/scalarstats.jl:716 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition describe(IO, AbstractArray{T, N} where N where T) in module StatsBase at /home/mo/.julia/packages/StatsBase/CuCbN/src/scalarstats.jl:721 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition describe(IO, AbstractArray{T, N} where N) where {T<:Union{Base.Missing, Real}} in module StatsBase at /home/mo/.julia/packages/StatsBase/CuCbN/src/scalarstats.jl:717 overwritten in module StatsBase on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **

[...]

PR #1512, opened 2021-12-14
Remove redundant imports

ffreyer:
Removes a bunch of imports that are already imported elsewhere.

Some other things we may want to remove/clean up
- src/scene.jl (testing file?)
- src/utilities/quaternions.jl (not used anymore?)
- GLMakie/src/experiments (testing?)
- src/jl_rasterizer (WIP?)

See also #1511

### DIFF
--- a/GLMakie/src/GLVisualize/visualize/particles.jl
+++ b/GLMakie/src/GLVisualize/visualize/particles.jl
@@ -43,7 +43,6 @@ function to_mesh(mesh::TOrSignal{<: GeometryBasics.Mesh})
     return NativeMesh(mesh)
 end
 
-using Makie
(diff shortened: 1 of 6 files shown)


Mo8it:
Thank you for the quick merge request! It has to be merged quickly and released because Makie is now broken after an env update.

Is there a way to automatically find redundant imports?

ffreyer:
I just looked for import and using with grep.

I also don't think these duplicates are really the reason you got an error. Many of them have been around for a long time. Did you try restarting your Julia session and importing Makie again? Or restarting, building and them importing?

SimonDanisch (approved):

PR #1526, opened 2021-12-21
Add documentation for attributes of primitive plots

ffreyer:
This is based on the attributes each plot function generates with GLMakie active. I skipped:

In general:
- transformations
- diffuse, specular, shininess, shading, ssao for plots that aren't 3D
- linewidth for all but lines/linesegments (Why does every plot have this?)

scatter:
- transform_marker
- uv_offset_with

heatmap:
- levels

text:
- lineheight
- offset

mesh:
- interpolate
- cycle


MakieCore/src/basic_plots.jl
```diff
@@ -24,7 +24,7 @@ Plots an image on range `x, y` (defaults to dimensions).
         colormap = [:black, :white],
         colorrange = automatic,
         interpolate = true,
-        fxaa = false,
+        fxaa = true,
```
SimonDanisch, line 27: hm, I think that was on purpose false, because fxaa messes with the image sampling

SimonDanisch:
Really nice, thank you :)

docs/examples/plotting_functions/mesh.md
```diff
@@ -2,6 +2,33 @@
 
 {{doc mesh}}
 
+## Attributes
+
+### Generic
+
+- `visible::Bool = true` sets whether the plot will be rendered or not.
```
SimonDanisch, line 9: I think ideally, we could put these in doc strings in code somewhere, so it's easier to keep it synchronized with the code

MakieCore/src/basic_plots.jl
ffreyer: Guess that makes sense, I'll undo this

docs/examples/plotting_functions/mesh.md
ffreyer: I think so too, but I'd say we should put in a bit more effort than just a raw docstring. Lots of recipes inherit attributes and it would make sense if they inherit docstrings as well. (It would also be nice to have some filtering, e.g. filtering attributes applicable to `lines` to more easily pass attributes. But that's kinda beside the point.)

I didn't want to bother with this here though. I just wanted to make a quick pr documenting things since I keep seeing people ask about it.

SimonDanisch: Fair enough, lets leave it like that then ;)

SimonDanisch (approved):

PR #1540, opened 2022-01-04
Some more clean ups

SimonDanisch:
- [x] fix docs
- [x] use Mesh instead of Poly for Axis background (more efficient)
- [x] fix transparency


ffreyer:
> fix transparency

For a bit of context the selection rectangle was invisible in an `Axis` with GLMakie. The reason for that is render order. The selection rectangle is rendered with `transparency = true`, i.e. with OIT, which runs before `fxaa = false` plots. The background of an axis used `poly!` which by default has `fxaa = false`. (Should it?) Now that the background uses `mesh!` it defaults to `fxaa = true` and the selection rectangle is visible.

But this raises a question about render order for OIT. In short OIT accumulating a depth and alpha weighted color average (to the HDR color buffer) and then normalizing and adding it to an opaque background. If you split up the color accumulation part, e.g. by accumulating `fxaa = true` colors, normalizing, adding to the opaque color, running fxaa and then continuing with `fxaa = false` colors you will not get the desired mixing between `fxaa = true` and `fxaa = false` colors. As far as OIT is concerned `fxaa = true` colors are part of the background, opaque and at the backgrounds depth. So splitting OIT won't work.

I also doubt that running fxaa directly on the accumulation buffer would work. Though I never took the time to understand the fxaa shader so maybe I'm wrong. If it does we would probably need another HDR color buffer.

Another possibility would be to let OIT run after all opaque passes, without fxaa. I didn't check how much of a difference that makes, I just assumed you'd generally want fxaa. As long as everything OIT is moved as one block this shouldn't cause any issues.

Personally I think we should keep the render order as it is and just stear away from mixing OIT with `fxaa = false`.

SimonDanisch:
Fxaa is now getting masked, instead of having its own render pass: https://github.com/JuliaPlots/Makie.jl/pull/1544
This solves the order problem as well...

PR #1547, opened 2022-01-07
Clarify remote servers page

rikhuijzer:
I was going to the "remote servers" page and expected some information about running Makie on a server and that kind of stuff. Instead, it was only about running on headless systems. To make this clearer, I've added one sentence.

Also, I've moved the CairoMakie section up because it makes sense to show the easiest option first. It would be unfortunate if someone goes for a difficult option only to find out later that it could have been much easier.

### DIFF
--- a/docs/documentation/headless.md
+++ b/docs/documentation/headless.md
@@ -1,6 +1,6 @@
-# Remote servers
+# Headless
 
-Makie can be used remotely and it supports working headless systems (such as CI servers).
+Makie can be used on headless systems (such as CI servers).
 This page describes what is required to get different back ends working in headless systems.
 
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Thanks! Makes sense. Maybe the page should be split into `headless operation` and `connecting via ssh` or something like that, seems like different scenarios to me.

rikhuijzer:
> connecting via `ssh` or something like that

I don't think that covers all bases of setting up a server. One could also create a back end server and connect via `json`. But, is that something Makie would need to document? Seems very specific.

As a suggestion, I've renamed the page to headless in 00c1530

PR #1548, opened 2022-01-07
remove all mentions of node

SimonDanisch:


src/interaction/liftmacro.jl
```diff
@@ -3,7 +3,7 @@ Returns a set of all sub-expressions in an expression that look like \$some_expr
 """
 function find_node_expressions(obj::Expr)
```
ffreyer, line 4: This still uses "node"

PR #1556, opened 2022-01-08
Revert campixel and camrelative to old near/far values

ffreyer:
Fixes the missing frame (and ticks) in Axis3, e.g. the animation on the front page of the documentation. https://makie.juliaplots.org/dev/

Alternatively we could also adjust

https://github.com/JuliaPlots/Makie.jl/blob/fd54f014fb1c4a3d72a40f1d46b019efa6990404/src/makielayout/layoutables/axis3d.jl#L447-L453

and probably

https://github.com/JuliaPlots/Makie.jl/blob/fd54f014fb1c4a3d72a40f1d46b019efa6990404/src/makielayout/layoutables/axis.jl#L127-L142

### DIFF
--- a/src/camera/camera2d.jl
+++ b/src/camera/camera2d.jl
@@ -308,7 +308,7 @@ struct PixelCamera <: AbstractCamera end
 
 Creates a pixel-level camera for the `Scene`.  No controls!
 """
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Oh boi... Ok, lets merge this - I think the Axis3/Axis cameras need some rework at some point anyways.

PR #1573, opened 2022-01-09
Fix `cam3d` update

t-bltg:
Fix point 1. of https://github.com/JuliaPlots/Makie.jl/issues/1572.


SimonDanisch:
I think the `update_cam!` should go into the constructor of Camera3D, here: https://github.com/JuliaPlots/Makie.jl/blob/651a8b8501f1c300b7b2299cf03d1c0a284e3577/src/camera/camera3d.jl#L225
@ffreyer what do you think?

ffreyer (approved):

PR #1582, opened 2022-01-11
make it possible to have one color per glyph

SimonDanisch:


src/layouting/layouting.jl
```diff
@@ -216,10 +228,10 @@ function glyph_collection(str::AbstractString, font_per_char, fontscale_px, hali
         reduce(vcat, charorigins),
         [x.extent for x in charinfos],
         [Vec2f(x.scale) for x in charinfos],
-        [rotation for x in charinfos],
-        [color for x in charinfos],
-        [strokecolor for x in charinfos],
-        [strokewidth for x in charinfos],
+        per_char_info(rotation, charinfos),
```
jkrumbiegel, line 231: hm neither what I had before nor this seems ideal. there is already a function called `attribute_per_char` at the top, this is currently used for font and textsize, but similar logic could apply to color. anyway, it feels weird to do one expansion before the glyph_collection function and one here.

jkrumbiegel (commented):
this is one of these places in the code base where it's difficult to know what type `color` etc could have, and therefore how the code will execute. also, I guess we recompile this expensive looking function for a color given as a Symbol vs. as an RGBf etc. I would like clearer logic, or does that tie into the attribute refactor idea?

SimonDanisch:
>I would like clearer logic, or does that tie into the attribute refactor idea?

To separate the concerns a bit, I would leave this for the attribute refactor PR, as long as this PR doesn't introduce any regressions

src/layouting/layouting.jl
SimonDanisch: Ah, I somehow didnt find that function... So I guess I should just use that?

SimonDanisch: Ok, this uses `attribute_per_char` now!

PR #1585, opened 2022-01-12
fix autolimits for linked axis

jqfeld:
# Description

Fixes a problem with `autolimits` for linked axes.
Before `autolimits` only checked for linked x-axes, even if dim=2. This probably has many side effects. For me it produces an error, when `autolimits` is called for an axis with `yscale=log10` which has an linked x-dimension to another axis with negative y-limits.
The intendend behaviour is probably to check for linked x-axes if dim=1 and for linked y-axes if dim=2. This is now implemented in the PR.

## Type of change

### DIFF
--- a/src/makielayout/layoutables/axis.jl
+++ b/src/makielayout/layoutables/axis.jl
@@ -834,7 +834,8 @@ function autolimits(ax::Axis, dim::Integer)
     # try getting x limits for the axis and then union them with linked axes
     lims = getlimits(ax, dim)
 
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Oh damn... Didn't realize that made it into 0.16 already!
This was meant to stay in https://github.com/JuliaPlots/Makie.jl/pull/1574, where I wanted to add some more tests to make sure that I don't regress axis linking by the refactor...
Thanks for finding and fixing this!

PR #1603, opened 2022-01-19
Font caching info message fix

IanButterworth:
This doesn't make sense to the end user, and I assume this was just a product of a find-and-replace for `AbstractPlotting`


ffreyer (approved):

SimonDanisch:
Ha :D I feel like we've fixed this message a couple of times already :D

PR #1612, opened 2022-01-21
Memoize `get_ticks`

IanButterworth:
# Description

Profiling the updates on an interactive plot I'm working on was showing that a large proportion of time was spent in `PlotUtils.optimize_ticks_typed`, but most of the axes didn't actually have ticks changing, so the computation seems like it could be cached.

Memoizing this higher level `get_ticks` function with a LRU cache makes a huge improvement in interactivity & framerate for me.
If this kind of approach is acceptable, using a LRU cache should make this robust to too much memory usage, I think.

I left the cache size customizable though. 

## Type of change

## Checklist


IanButterworth:
Just a friendly bump. I've been running this locally and it's a nice render speed improvement

src/makielayout/MakieLayout.jl
```diff
@@ -20,11 +20,15 @@ using GridLayoutBase
...
 import Showoff
 using Colors
+using Memoization: @memoize
+using LRUCache: LRU
 
 const FPS = Observable(30)
 const COLOR_ACCENT = Ref(RGBf(((79, 122, 214) ./ 255)...))
 const COLOR_ACCENT_DIMMED = Ref(RGBf(((174, 192, 230) ./ 255)...))
 
+const TICK_CACHE_SIZE = Ref{Int}(100)
```
IanButterworth, line 30: It just occurred to me that maybe this can't be dynamic given the `@memoize` is evaluated at precompile time?

SimonDanisch: Yeah, it should be evaluated when the macro gets expanded..Could we make the LRU a const global, which can then be resized?

jkrumbiegel:
While this seems to fix the symptom of superfluous tick updating, the real problem is that tick computations are retriggered when an axis changes its area. And that happens when other axes are zoomed in and out, because with the tick size updating, the layout changes ever so slightly. The fix (which I tried before but didn't finish because it was a bit tricky) is to fix the update chain of observables within Axis, so that the px_area trigger doesn't kick off limit calculation anymore. The web of observables did get quite complicated after a certain point, so there are bound to be some inefficiencies in there

SimonDanisch:
Should we still merge it to get the speed up, while trying to figure out a better solution in the future?

jkrumbiegel:
Maybe there is a simpler form of this "hack" then, where somewhere in the chain, we check if the triggering limits are the same as before, and if so, don't update.

IanButterworth:
Yeah, that sounds better

IanButterworth:
If Makie isn't already using `Observables.observe_changes` maybe that's the fix, and it goes beyond the ticks issue?
https://github.com/JuliaGizmos/Observables.jl/blob/0f94e69cabddd87fcf057fb95ed5dc64be9661de/src/Observables.jl#L481

IanButterworth:
I just tested this ^ with `ChangeObservable(x) = observe_changes(Observable(x))` as a drop-in replacement throughout the stack and it appears to work with my relatively complex interactive plot.

I guess a more nuanced approach might be needed, deciding where caching values is memory efficient? But here's the blanket approach https://github.com/IanButterworth/Makie.jl/commit/4d580b28a7cab6b2d7d20b8de766eefb993a025e

However, it didn't save any of the `get_ticks` time. That appears to still be triggered.

But... That change, plus this PR makes for quite a nice speed improvement.

After loading GLMakie

This PR
TTFP:  84.581741 seconds (343.27 M allocations: 19.546 GiB, 13.47% gc time, 88.27% compilation time)
TTSP:  11.346520 seconds (86.50 M allocations: 5.703 GiB, 28.82% gc time)

Master
TTFP: 119.780434 seconds (1.17 G allocations: 34.355 GiB, 14.79% gc time, 63.43% compilation time)
TTSP:  49.060643 seconds (916.26 M allocations: 20.599 GiB, 22.32% gc time)

t-bltg:
@IanButterworth, can you try out profiling with latest `PlotUtils@1.1.4` ? It should be much faster now.

SimonDanisch:
I think we can close this for now and pick it up if there are still performance problems left after merging the PlotUtils prs.

IanButterworth:
@t-bltg sorry for the delay. I've not been near Makie recently, but when I next do I'll profile. Thanks all!

PR #1613, opened 2022-01-21
GLMakie: in render_frame store zvals for sort rather than use `by` for efficiency

IanButterworth:
# Description

This appears to speed up rendering because when using a function for `by` in a sort, the function is usually called many more times than the number of elements in the vector. 

Storing the vals then sorting via that vector appears to speed up the sort and thus the render loop.

Here are some profiles (on the top of #1612 ), The sort is the lower busy region just to the right of the middle in the first one.
On the 2nd it's a lot smaller (it has the mouse hovering over it with the line info)

Before
Screenshot from 2022-01-21 02-15-10

This PR
Screenshot from 2022-01-21 02-26-56

Note that the way I profiled this was to launch the interactive window normally, then in the repl ran `@profile sleep(10)` then interacted with the plot. It captures the async tasks and avoids anything to do with plot startup.

Also with ProfileView you can ignore the other threads.

## Type of change

- [x] Optimization ?

## Checklist

- [x] Tested locally

### DIFF
--- a/GLMakie/src/rendering.jl
+++ b/GLMakie/src/rendering.jl
@@ -126,7 +126,7 @@ function render_frame(screen::Screen; resize_buffers=true)
         return Makie.zvalue2d(plot)
     end
     zvals = sortby.(screen.renderlist)
-    screen.renderlist = screen.renderlist[sortperm(zvals)]
+    permute!(screen.renderlist, sortperm(zvals))
 
     # NOTE
     # The transparent color buffer is reused by SSAO and FXAA. Changing the
--- a/GLMakie/src/rendering.jl
+++ b/GLMakie/src/rendering.jl
@@ -125,7 +125,8 @@ function render_frame(screen::Screen; resize_buffers=true)
         # TODO, use actual boundingbox
         return Makie.zvalue2d(plot)
     end
-    sort!(screen.renderlist; by=sortby)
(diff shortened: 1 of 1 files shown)


GLMakie/src/rendering.jl
```diff
@@ -125,7 +125,8 @@ function render_frame(screen::Screen; resize_buffers=true)
         # TODO, use actual boundingbox
         return Makie.zvalue2d(plot)
     end
-    sort!(screen.renderlist; by=sortby)
+    zvals = sortby.(screen.renderlist)
+    screen.renderlist = screen.renderlist[sortperm(zvals)]
```
ffreyer, line 129:
```suggestion
    permute!(screen.renderlist, sortperm(zvals))
```
To avoid some unnecessary allocations. 

Though I think it would be best to cache the z values instead (probably in plot?) and continue using `sort!`. I think @SimonDanisch wants to bring back cached bounding boxes - maybe caching z values is better done alongside that?

IanButterworth: Even if caching of z-values is on the roadmap like you say, could this be merged before that? It's quite nice as a render speed improvement

SimonDanisch:
Ah interesting, didn't realize that's how sort works... Thank you, lets merge it for now and think about further improvements later! :)

PR #1615, opened 2022-01-21
fix `add_cycle_attributes!` for nested attributes

hexaeder:
Custom recipes with nested attributes won't work otherwise, because
`getindex(::Attributes)` fails.

Fixes https://github.com/JuliaPlots/GraphMakie.jl/issues/50

## Type of change


jkrumbiegel:
I think this could maybe also be done with `to_value(allattrs[key]) isa Cycled`, saves one additional type unstable lookup

hexaeder:
good suggestion, I've tested and changed it :)

jkrumbiegel (approved):

PR #1619, opened 2022-01-25
replace all Observable -> ChangeObservable

SimonDanisch:
..turned into PR to let CI run

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -404,7 +404,7 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
             glyphpos = let
                 # project without yflip - we need to apply model before that
                 p = project_position(scene, position, Mat4f(I), false)
-
+                
                 # flip for Cairo
                 p += (p3_to_p2(glyphoffset .+ p3_offset))
                 p = (_deref(model) * Vec4f(p[1], p[2], 0, 1))[Vec(1, 2)]
@@ -597,11 +597,11 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Union{Heatmap
 
             # Rectangles and polygons that are directly adjacent usually show
             # white lines between them due to anti aliasing. To avoid this we
-            # increase their size slightly.
+            # increase their size slightly. 
 
             if alpha(colors[i, j]) == 1
-                # sign.(p - center) gives the direction in which we need to
-                # extend the polygon. (Which may change due to rotations in the
+                # sign.(p - center) gives the direction in which we need to 
+                # extend the polygon. (Which may change due to rotations in the 
                 # model matrix.) (i!=1) etc is used to avoid increasing the
                 # outer extent of the heatmap.
                 center = 0.25 * (p1 + p2 + p3 + p4)
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -634,7 +634,7 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Makie.Mesh)
         draw_mesh2D(scene, screen, primitive)
     else
         if !haskey(primitive, :faceculling)
-            primitive[:faceculling] = ChangeObservable(-10)
+            primitive[:faceculling] = Observable(-10)
         end
         draw_mesh3D(scene, screen, primitive)
     end
@@ -820,7 +820,7 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Makie.Surface
         primitive[:color] = primitive[3]
     end
     if !haskey(primitive, :faceculling)
-        primitive[:faceculling] = ChangeObservable(-10)
+        primitive[:faceculling] = Observable(-10)
     end
     draw_mesh3D(scene, screen, primitive, mesh=mesh)
     primitive[:color] = old
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -404,7 +404,7 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
             glyphpos = let
                 # project without yflip - we need to apply model before that
                 p = project_position(scene, position, Mat4f(I), false)
-                
+
                 # flip for Cairo
                 p += (p3_to_p2(glyphoffset .+ p3_offset))
                 p = (_deref(model) * Vec4f(p[1], p[2], 0, 1))[Vec(1, 2)]
@@ -597,11 +597,11 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Union{Heatmap
 
             # Rectangles and polygons that are directly adjacent usually show
             # white lines between them due to anti aliasing. To avoid this we
-            # increase their size slightly. 
+            # increase their size slightly.
 
             if alpha(colors[i, j]) == 1
-                # sign.(p - center) gives the direction in which we need to 
-                # extend the polygon. (Which may change due to rotations in the 
+                # sign.(p - center) gives the direction in which we need to
+                # extend the polygon. (Which may change due to rotations in the
                 # model matrix.) (i!=1) etc is used to avoid increasing the
                 # outer extent of the heatmap.
                 center = 0.25 * (p1 + p2 + p3 + p4)
@@ -634,7 +634,7 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Makie.Mesh)
         draw_mesh2D(scene, screen, primitive)
     else
         if !haskey(primitive, :faceculling)
-            primitive[:faceculling] = Observable(-10)
+            primitive[:faceculling] = ChangeObservable(-10)
         end
         draw_mesh3D(scene, screen, primitive)
     end
@@ -820,7 +820,7 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Makie.Surface
         primitive[:color] = primitive[3]
     end
     if !haskey(primitive, :faceculling)
-        primitive[:faceculling] = Observable(-10)
+        primitive[:faceculling] = ChangeObservable(-10)
     end
     draw_mesh3D(scene, screen, primitive, mesh=mesh)
     primitive[:color] = old
--- a/GLMakie/src/glwindow.jl
+++ b/GLMakie/src/glwindow.jl
@@ -10,7 +10,7 @@ end
 
 
 mutable struct GLFramebuffer
-    resolution::Observable{NTuple{2, Int}}
+    resolution::AbstractObservable{NTuple{2, Int}}
     id::GLuint
 
     buffer_ids::Dict{Symbol, GLuint}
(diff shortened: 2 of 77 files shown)


ffreyer:
For reference: #1612

I don't think this actually does anything like this. Writing `obs = ChangeObservable(1.0)` lowers to `obs = observe_changes(Observable(1.0))` which becomes
```julia
obs = let
    obs = Observable(1.0)
    T = eltype(obs)
    eq = ==

    out = Observable{T}(obs[])
    on(obs) do val
        if !eq(val, out[])
            out[] = val
        end
    end
    out
end
```
so you just end up with a normal observable whose parent is forgotten.
```julia
obs = Makie.ChangeObservable(nothing)
Observable{Nothing} with 0 listeners. Value:
nothing

julia> on(_ -> println("tick"), obs)
(::Observables.ObserverFunction) (generic function with 0 methods)

julia> obs[] = nothing
tick
```

We will also run into some issues if we just replace all Observables like this. For example `render_tick` in GLMakie can't be replaced like this. It holds `nothing` and gets triggered once per frame - or never with what this change intends to do. It's used to trigger mouseposition and screen resize events at a reduced rate.

IanButterworth:
Ah true. There might still be a way to overload it so it's a simple replace.

Like make `ChangeObservable` contain both observables, and define the set and get methods for ChangeObservable to forward to the underlying value Observable, but define the `on` etc. methods to forward to the change observable.

As an aside, I think that means the improved TTFP was all the `get_ticks` memoization.. That's promising for the idea to make that more efficient

ffreyer:
Skimming through these changes it seems like a lot of them are redundant. Backends often generate defaults with something like `get(attributes, name, Observable(default_value)`. There are also a bunch cases with backend specific observables, e.g. `uniforms[name] = Observable(default)`. I don't think either of those signatures ever get updates. I've also seen a couple of changes in docs and tests which should be skipped. 

I think it should be fine to completely exclude backends from this kind of change. I can't think of anything other than events (which already use PriorityObservables) that gets triggered from backends. On the other hand GridLayoutBase might also benefit from it?

IanButterworth:
Yeah, this was a blind find and replace. (Didn't intend it to be a PR 😅)

I'm just trying to make my suggestion above work. If I get there I'll also try to revert some of the needless changes.

IanButterworth:
I've pushed what I tried, but I got stuck around the `map!` implementation, with
```
ERROR: LoadError: MethodError: no method matching (GeometryBasics.Recti)(::Int64, ::Int64, ::MakieCore.ChangeObservable{Tuple{Int64, Int64}})
Closest candidates are:
  (GeometryBasics.HyperRectangle{N, T} where N)(::Number, ::Number, ::Number, ::Number) where T at ~/.julia/packages/GeometryBasics/WMp6v/src/primitives/rectangles.jl:107
  (GeometryBasics.HyperRectangle{N, T} where N)(::Number, ::Number, ::Union{Tuple{T, T}, StaticArrays.StaticVector{2, T}} where T) where T at ~/.julia/packages/GeometryBasics/WMp6v/src/primitives/rectangles.jl:140
  (GeometryBasics.HyperRectangle{N, T} where N)(::Union{Tuple{T, T}, StaticArrays.StaticVector{2, T}} where T, ::Number, ::Number) where T at ~/.julia/packages/GeometryBasics/WMp6v/src/primitives/rectangles.jl:136
Stacktrace:
  [1] (::Makie.var"#93#96")(res::MakieCore.ChangeObservable{Tuple{Int64, Int64}})
    @ Makie ~/Documents/GitHub/Makie.jl/src/scenes.jl:171
  [2] lift(::Function, ::Observables.Observable{Any}; kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/Documents/GitHub/Makie.jl/src/interaction/observables.jl:30
  [3] lift(::Function, ::Observables.Observable{Any})
    @ Makie ~/Documents/GitHub/Makie.jl/src/interaction/observables.jl:27
  [4] Makie.Scene(; px_area::Nothing, events::Makie.Events, clear::Bool, transform_func::Function, camera::typeof(Makie.campixel!), camera_controls::Makie.EmptyCamera, transformation::Makie.Transformation, plots::Vector{MakieCore.AbstractPlot}, theme::MakieCore.Attributes, children::Vector{Makie.Scene}, current_screens::Vector{MakieCore.AbstractScreen}, parent::Nothing, visible::MakieCore.ChangeObservable{Bool}, ssao::Makie.SSAO, lights::MakieCore.Automatic, theme_kw::Base.Pairs{Symbol, Tuple{Int64, Int64}, Tuple{Symbol}, NamedTuple{(:resolution,), Tuple{Tuple{Int64, Int64}}}})
    @ Makie ~/Documents/GitHub/Makie.jl/src/scenes.jl:170
  [5] Makie.Figure(; kwargs::Base.Pairs{Symbol, Tuple{Int64, Int64}, Tuple{Symbol}, NamedTuple{(:resolution,), Tuple{Tuple{Int64, Int64}}}})
    @ Makie ~/Documents/GitHub/Makie.jl/src/figures.jl:69
  [6] initialize_plot!(rp::LeukoPlots.ResultPlot; resolution::Tuple{Int64, Int64}, yrange::Tuple{Int64, Int64}, zoomspan::Float64, interactive::Bool, show_sliders::Bool)
```

CairoMakie/src/primitives.jl
```diff
@@ -404,7 +404,7 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
             glyphpos = let
                 # project without yflip - we need to apply model before that
                 p = project_position(scene, position, Mat4f(I), false)
-                
+
```
IanButterworth, line 407:
```suggestion
                
```

CairoMakie/src/primitives.jl
```diff
@@ -597,11 +597,11 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Union{Heatmap
 
             # Rectangles and polygons that are directly adjacent usually show
             # white lines between them due to anti aliasing. To avoid this we
-            # increase their size slightly. 
+            # increase their size slightly.
```
IanButterworth, line 600:
```suggestion
            # increase their size slightly. 
```

CairoMakie/src/primitives.jl
```diff
@@ -597,11 +597,11 @@ function draw_atomic(scene::Scene, screen::CairoScreen, primitive::Union{Heatmap
...
 
             if alpha(colors[i, j]) == 1
-                # sign.(p - center) gives the direction in which we need to 
-                # extend the polygon. (Which may change due to rotations in the 
+                # sign.(p - center) gives the direction in which we need to
+                # extend the polygon. (Which may change due to rotations in the
```
IanButterworth, line 604:
```suggestion
                # sign.(p - center) gives the direction in which we need to 
                # extend the polygon. (Which may change due to rotations in the 
```

GLMakie/src/glwindow.jl
```diff
@@ -96,7 +96,7 @@ function GLFramebuffer(fb_size::NTuple{2, Int})
 
     fb_size_node = Observable(fb_size)
 
-    # To allow adding postprocessors in various combinations we need to keep 
+    # To allow adding postprocessors in various combinations we need to keep
```
IanButterworth, line 99:
```suggestion
    # To allow adding postprocessors in various combinations we need to keep 
```

src/interaction/observables.jl
```diff
@@ -3,7 +3,7 @@
 """
     lift(f, o1::Observables.AbstractObservable, rest...) -> w
 
-Create a new `w::Observable` by applying `f` to the _values_ of all observables 
+Create a new `w::Observable` by applying `f` to the _values_ of all observables
```
IanButterworth, line 6:
```suggestion
Create a new `w::Observable` by applying `f` to the _values_ of all observables 
```

src/interaction/observables.jl
```diff
@@ -27,6 +27,7 @@ function lift(f, o1::Observables.AbstractObservable, rest...; kw...)
     if !isempty(kw)
         error("lift(f, obs...; init=f.(obs...), typ=typeof(init)) is deprecated. Use lift(typ, f, obs...), or map!(f, Observable(), obs...) for different init.")
     end
+    # @show typeof(o1), typeof(to_value(o1))
```
IanButterworth, line 30:
```suggestion
```

src/makielayout/layoutables/menu.jl
```diff
@@ -338,7 +338,7 @@ function _reassemble_menu(
             padding = textpadding, visible = is_open
         )
 
-        # translate dropdown elements in the foreground 
+        # translate dropdown elements in the foreground
```
IanButterworth, line 341:
```suggestion
        # translate dropdown elements in the foreground 
```

src/makielayout/layoutables/menu.jl
```diff
@@ -355,10 +355,10 @@ function _reassemble_menu(
 
     resize!(mouseeventhandles, length(alltexts))
     map!(mouseeventhandles, eachindex(allrects), allrects) do i, r
-        # Use base priority for [Menu   v] and high priority for the dropdown 
+        # Use base priority for [Menu   v] and high priority for the dropdown
```
IanButterworth, line 358:
```suggestion
        # Use base priority for [Menu   v] and high priority for the dropdown 
```

src/makielayout/layoutables/menu.jl
```diff
@@ -355,10 +355,10 @@ function _reassemble_menu(
...
         # elements that may overlap with out interactive layoutables.
         addmouseevents!(
-            scene, r.layoutobservables.computedbbox, 
+            scene, r.layoutobservables.computedbbox,
```
IanButterworth, line 361:
```suggestion
            scene, r.layoutobservables.computedbbox, 
```

test/liftmacro.jl
```diff
@@ -2,7 +2,7 @@
     x = Observable(1.0)
     y = Observable(2.0)
     z = (x = x, y = y)
-    
+
```
IanButterworth, line 5:
```suggestion
    
```

IanButterworth:
Ok, I took the other approach.. I reverted and went as far as I could without breaking it.

`get_ticks` is still dominating, and I think the trick there is to change the internal lift Observable to a ChangeObservable, but that seemed to break a lot

IanButterworth:
Now dependent on https://github.com/JuliaGizmos/Observables.jl/pull/79

Notes:
- I tried to fix any use of Observables internals that got in the way of using the new `ChangeObservable <: AbstractObservable` like `obs[] = obs[]` which IIUC should be replaceable with `notify(obs)`
- I broadened the `Observables` arg typedefs to `AbstractObservable` where I thought it appropriate
- Lift now has an optional `change_only` mode, which is only enabled for two lifts in `lineaxis.jl`


I've only been testing this with `using GLMakie; scatter(rand(2,10))` and it's not rendering the axis margin & spines properly.

There seems to be some additional no-change observable triggers needed. Or perhaps the two lifts need some additional things to watch?


I think I've got as far as I understand, so would appreciate if you could see what you think @jkrumbiegel

PR #1626, opened 2022-01-28
try out githubactionsutils and disable all but cairomakie ci

jkrumbiegel:


jkrumbiegel:
We go with the local flow using ReferenceTests for now and might return to this automated posting of images at a later point

PR #1631, opened 2022-01-31
Use julia-actions/cache

rikhuijzer:
Reduces lines of code and also `julia-actions/cache` caches `~/.julia/packages` by default too. (I haven't tested how much time that actually saves, but well, it doesn't hurt at least.)

I also moved the position of the cache call forward in `wglmakie.yaml`. I don't think that the old position was useful because the `github-actions/cache` would put everything back when a cache hit occurs after Julia was done running.


SimonDanisch:
Awesome, thank you :)

PR #1596, opened 2022-01-15
Add projection `space` option as a generic attribute

ffreyer:
# Description

This pr adds `space` as a general attribute to switch between camera space, pixel space, unit space (0..1) and maybe clip space (-1..1). That way you can adjust the units of the plot without creating another scene or messing with camera matrices.

Currently works in GLMakie:

```julia
using GLMakie
fig = Figure()
# ax = LScene(fig[1, 1])
# ax = Axis(fig[1, 1])
ax = Axis3(fig[1, 1])
meshscatter!(ax, Point3f.(1:10, 1:10, 1:10), space = :data, color = :blue)
# scatter!(ax, Point2f.(1:10, 1:10), space = :data, color = :blue)
scatter!(ax, Point2f.(200, 100:100:500), space = :pixel, color = :red)
scatter!(ax, Point2f.(0.6, 0.1:0.1:0.9), space = :relative, color = :green)
scatter!(ax, Point2f.(0.9, -0.9:0.2:0.9), space = :clip, color = :orange)
display(fig)
```

Screenshot from 2022-01-15 14-46-11

## Type of change

Breaking change because the `space` attribute in `text` has been renamed to `markerspace` and `space` now means something else.

## Checklist

- [x] Added an entry in NEWS.md (for new features and breaking changes)
- [x] Added or changed relevant sections in the documentation
- [x] Added unit tests for new algorithms, conversion methods, etc.
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc.
- [x] Works in GLMakie
- [x] Works in CairoMakie
- [x] Works in WGLMakie
- ~~Works in RPRMakie~~ I don't know how to make it work
- [x] Works with nested plots
- [x] Works with mixed `markerspace` in all combinations (scatter, text)
- [x] pass space through ~~everything~~ poly - I think for most recipes there isn't really a point in adding `space` as an attribute

## Related issues

- fixes #994 (somehow)
- fixes #1602 (probably because I switched away from `project`?)
- closes #1472 
- implements "display", "figure/scene/axis" and "data" coordinates from #369

PR #1632, opened 2022-02-01
move preprojection into shaders

ffreyer:
I thought #1596 might be easier to review without pulling preprojection into shaders so I did it in a separate branch. 

I checked these changes with the same examples I added as reference tests in #1596. Should be good to merge, though I want to go over this again and name things more appropriately.


GLMakie/src/drawing_primitives.jl
```diff
@@ -214,32 +214,25 @@ function draw_atomic(screen::GLScreen, scene::Scene, @nospecialize(x::Union{Scat
...
-                    transform_func_obs(x), cam.projectionview, cam.resolution
-                ) do positions, space, markerspace, tf, _, _
-
-                mat = Makie.clip_to_space(cam, markerspace) * Makie.space_to_clip(cam, space)
-                map(positions) do pos
-                    p = apply_transform(tf, pos)
-                    p4d = mat * Makie.to_ndim(Point4f, Makie.to_ndim(Point3f, p, 0), 1)
-                    p4d[SOneTo(3)] / p4d[4]
-                end
+            gl_attributes[:preprojection] = map(space, mspace, cam.projectionview) do space, mspace, pv
```
SimonDanisch, line 225: that looks a bit like this is rather the full projection matrix? Or is there some other projection matrix applied afterward?

ffreyer: It's split up to go from space -> markerspace via `preprojection` and from markerspace to clip space via `projection * view`. Basically `preprojection` does what 
https://github.com/JuliaPlots/Makie.jl/blob/85a6d145d3d931ebde190046d20482f6eb6bcc69/src/layouting/layouting.jl#L238
did without the `transform` and from a variable space to a variable markerspace (instead of just data -> pixel (or data))

SimonDanisch (approved):

ffreyer:
I did a bit of renaming and I think I'm happy with this now. Feel free to merge this into the main pr.

To document all the spaces in scatter/text:

We have two variable coordinate spaces `space` and `markerspace` and two sets of transformations, one for the position of a marker/string/character and one for the quad it is drawn on. For the position of the marker it goes:
- `position`/`pos` in the shader is the position given to text/scatter in `space`
- `preproject` projects from `space` to `markerspace`
- `marker_offset` is added to the markerspace position. This is used to calculate the individual character positions from the total string position. 
- after that `projection * view * model` is multiplied to the combined position to get the clip space origins for quads

For the quads we have:

- `quad_offset` gives the quad origin relative to the previously calculated position in markerspace
- `scale`/`markersize` gives the size of the quad in markerspace
- `projection * maybe_view * qmat(rotations) * maybe_model` transforms the quad into clip space. `maybe_model` is the model matrix if `transform_marker == true` and `maybe_view` is the view matrix if `billboard == false`.
- the clip space position and relative quad origin are then added to get the full clip space origin of the quad

ffreyer:
Ok I guess I shouldn't merge master into a branch that doesn't target master?

#1596 with this should be good to go again

ffreyer:
I merged this into #1596 manually because I was afraid it would add a bunch of people to that pr if I did it through github

PR #1633, opened 2022-02-01
fix text layouting

jkrumbiegel:
This came up when trying to build tables out of `Label`s. The alignment looked bad, which seems to be because of incorrect assumptions about glyph boundingboxes. So far, the width of a line was computed using the "ink-width" of the glyphs, but this leads to a ragged appearance when right-aligning. Also, the left origin of a line was incorrectly adjusted for bearing, which is not necessary at all, as the bearing is chosen to make the glyph look well-adjusted when left aligned.

Some test code:

```julia
f = Figure(fontsize = 30, font = "Helvetica")

words = ["Apple", "Pear", "Cherry", "Coconut", "Pineapple", "Grapes"]

for (i, word) in enumerate(words)
    Label(f[i, 1], word, halign = :left)
    Label(f[i, 2], string(i), halign = :right)
end

rowgap!(f.layout, 10)

f
```

| Before | After |
|--|--|
| grafik | grafik |

### DIFF
--- a/src/layouting/boundingbox.jl
+++ b/src/layouting/boundingbox.jl
@@ -23,12 +23,20 @@ end
 function gl_bboxes(gl::GlyphCollection)
     scales = gl.scales.sv isa Vec2f ? (gl.scales.sv for _ in gl.extents) : gl.scales.sv
     map(gl.extents, gl.fonts, scales) do ext, font, scale
-        unscaled_hi_bb = height_insensitive_boundingbox(ext, font)
+        unscaled_hi_bb = height_insensitive_boundingbox_with_advance(ext, font)
         hi_bb = Rect2f(
             Makie.origin(unscaled_hi_bb) * scale,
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Cool! I guess the test failures are all expected... But I guess the `Glyph Collections` error in the unit tests still needs fixing ;)

jkrumbiegel:
Ok the unit test is fixed. I'm not sure how to proceed given that `height_insensitive_boundingbox_with_advance` is just a Makie function now, while `height_insensitive_boundingbox` lives in FreeTypeAbstraction as if it's somehow more basic. This was more about me learning along the way how all these glyph metrics and boundingboxes work. I'm still not 100% sure what the canonical behavior should be, but it seems that taking the full hadvance into account makes some things easier and appears to be more closely to how a browser behaves when selecting text, for example.

PR #1642, opened 2022-02-04
Revert "Use julia-actions/cache"

SimonDanisch:
Reverts JuliaPlots/Makie.jl#1631 and see if this fixes our CI?
It's weird though, since that PR went through CI just fine..


t-bltg:
Same as https://github.com/JuliaPlots/Plots.jl/issues/4091.

It got fixed in https://github.com/JuliaIO/PNGFiles.jl/pull/54 (`PNGFiles@0.3.14`).

PR #1649, opened 2022-02-07
add justification for legend labels

SimonDanisch:
It wasn't possible to specify legend label justification, this PR adds it.
Justification defaults to labelhalign, if not set explicitely.


jkrumbiegel (approved):

icweaver:
Sorry if this would be better as a separate issue, but would it be possible to add this for the legend title as well? This might be lost to the slack hole now, but this is the [current workaround](https://julialang.slack.com/archives/C8RQUU2KH/p1644439666939619) for multi-line titles, but maybe there is a better way already! (copied below for convenience)

```julia
fig = Figure()
ax = Axis(fig[1, 1])

lines!(ax, rand(10), label="A looooooong label")

axislegend("A multi-line\ntitle" * " "^14, titlehalign=:left)

fig
```

index

PR #1654, opened 2022-02-08
clean up glvisualize remnants and make mesh implementation more consistent

SimonDanisch:
Long overdue cleanup...
Goal of this PR is, to get rid of some ridiculus GLVisualize constructs to ease maintenance and compilation.
- [x] get rid of `visualize`
- [x] clean up mesh to work with nan_color etc, and fetch colormap in shader
- [x] fix [#image](https://github.com/JuliaPlots/Makie.jl/issues/1230) by using mesh for plotting image


GLMakie/assets/shader/mesh.frag
```diff
@@ -29,17 +29,21 @@ vec4 get_color(Nothing color, vec2 uv, vec2 color_norm, sampler1D color_map, Not
...
+
+uniform vec4 highclip;
+uniform vec4 lowclip;
+uniform vec4 nan_color;
+
 vec4 get_color(sampler2D intensity, vec2 uv, vec2 color_norm, sampler1D color_map, Nothing matcap){
-    const float i = texture(intensity, uv).x;
+    float i = texture(intensity, uv).x;
     vec4 color = color_lookup(i, color_map, color_norm);
     if (isnan(i)) {
```
MariusDrulea, line 48: isnan shall be checked before color_lookup; unnecessary call to color lookup for nan values

SimonDanisch: Yeah makes sense... I copy and pasted this from an older shader, and it feels like I had a reason for this. I'll try it out tomorrow

PR #1658, opened 2022-02-10
Cleanup interactive interface & docs

### LINKED ISSUE #1611: Event system documentation and usability
It would be good to add a tutorial on how to use events. Perhaps a small game like Minsweeper? The general documentation could also be more extensive.

With respect to usability there are a bunch of functions that kind of got left behind when figures (and MakieLayout in general) were introduced. Lots of functions don't work with Figure, Axis, Axis3 and/or LScene, so you end up writing `ax.scene` and `fig.scene` frequently.

Some function aren't currently exported but probably should be. For example `is_mouseinside`.

Some functions could be simplified, e.g `pick(scene, pos)` is usually called with the current mouse position so why not make that the default?

ffreyer:
Generalizes a bunch of interactive function so they work with figures, axes and FigureAxisPlot as well as scenes. Also updates the documentation to be more thorough and include some more direct examples.

Closes #1611


ffreyer:
> ## Interactive Widgets
> 
> Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can read about in the \myreflink{Layoutables} section.
> 
> ## Recording Animations with Interactions
> 
> You can record a `Scene` while you're interacting with it.
> Just use the \apilink{record} function (also see the \myreflink{Animations} page) and allow interaction by `sleep`ing in the loop.
> 
> In this example, we sample from the Scene `scene` for 10 seconds, at a rate of 10 frames per second.
> 
> ```julia
> fps = 10
> record(scene, "test.mp4"; framerate = fps) do io
>     for i = 1:100
>         sleep(1/fps)
>         recordframe!(io)
>     end
> end
> ```

Do these section still make sense in the Events section?

docs/documentation/events.md
```diff
@@ -73,89 +69,283 @@ There are three mouse events one can react to:
...
+            notify(points)
         end
     end
-    # Do not consume the event
-    return Consume(false)
 end
+
+scene
+```
+\end{showhtml}
```
ffreyer, line 106: How do I make these examples work in the docs build?

ffreyer: I don't think I can so I switched to GLMakie

felixcremer (commented):
Thanks for the write up. I only looked at the documentation changes, because I can't provide feedback to the code changes.

ffreyer:
Thanks for the review. I made some changes which should address everything. Also the `pick` example now includes adding and deleting points as well as dragging

SimonDanisch (approved):

ffreyer:
There is a method ambiguity I still need to fix

SimonDanisch:
Great clean up! Is this ready to merge?

ffreyer:
Yea I think so

PR #1664, opened 2022-02-11
update Three for performance bug, add fps

### LINKED ISSUE #1663: 100% CPU utilization for WGLMakie plot
From https://github.com/JuliaPlots/Makie.jl/issues/1427#issuecomment-1035957115: 
```julia
using StaticArrays, WGLMakie, AlgebraOfGraphics

a, b = rand(SizedMatrix{2,3,SizedVector{1000}}, 2)
a = Array(Array.(a))
b = Array(Array.(b))
m = mapping(a=>"a",b => "b",
  row=collect(axes(a,1)) => nonnumeric,
  col=collect(permutedims(axes(b,2))) => nonnumeric,
)
visual(Scatter) * m |> draw
```
Leads to one core running on 100%... Which is a weird, since the CPU should do really little work while the plot is drawn on the GPU.

SimonDanisch:
The ThreeJS version we were on had a bug, that invalidated the WebGL program for every frame, leading to recompile the shader for every frame. Updating the version drops CPU times back to normal values.
This PR also throttles the framerate to 30fps, which can be configured by `activate!(; fps=30)`.
fixes #1663

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # News
 
+## Master
+
+- Fix WGLMakie performance bug and add option to set fps via `WGLMakie.activate!(fps=30)`
+
 ## v0.16
 
 #### Big Changes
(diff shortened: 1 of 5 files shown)


SimonDanisch:
Seems like 137 has a bug https://github.com/mrdoob/three.js/issues/23459, so we can only upgrade to 136 (which I think should also have the performance fix)

PR #1671, opened 2022-02-13
Fix invalid range logscale

### LINKED ISSUE #1670: `yscale=log10` not working with `lines` and small values
MWE
```julia
julia> lines(1:10, 1 .+ 1e-9 * rand(10), axis=(;yscale=log10))
ERROR: ArgumentError: range step cannot be zero
Stacktrace:
  [1] (::Colon)(start::Float32, step::Float32, stop::Float32)
    @ Base ./twiceprecision.jl:401
  [2] get_minor_tickvalues(i::IntervalsBetween, scale::typeof(log10), tickvalues::Vector{Float32}, vmin::Float32, vmax::Float32)
 ...
```

The same `MWE` was already given in #1490 but it gave a different error back then.

fatteneder:
# Description

Fixes #1670

Avoid equal axis limits and fall back to default ones.

I also modified `validate_limits_for_scale` to error if equal axis limits are encountered even if no log scale is set. Is this ok?
Previously this did not error on standard axis, e.g.
```julia
lines(1:10,1:10, axis=(; limits=(1,1,1,5))
```
worked.

Should I also add units test for this?

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/layoutables/axis.jl
+++ b/src/makielayout/layoutables/axis.jl
@@ -576,7 +576,7 @@ end
 
 function validate_limits_for_scale(lims, scale)
     all(x -> begin
-        scale in [log10, log2, log] && lims[1] == lims[2] && return false
+        lims[1] == lims[2] && return false
         x in defined_interval(scale)
     end, lims)
 end
--- a/src/makielayout/layoutables/axis.jl
+++ b/src/makielayout/layoutables/axis.jl
@@ -575,10 +575,8 @@ function validate_limits_for_scales(lims::Rect, xsc, ysc)
 end
 
 function validate_limits_for_scale(lims, scale)
-    all(x -> begin
-        lims[1] == lims[2] && return false
-        x in defined_interval(scale)
-    end, lims)
+    lims[1] == lims[2] && return false
+    all(x -> x in defined_interval(scale), lims)
 end
 
 palettesyms(cycle::Cycle) = [c[2] for c in cycle.cycle]
--- a/src/makielayout/layoutables/axis.jl
+++ b/src/makielayout/layoutables/axis.jl
@@ -773,6 +773,10 @@ function getlimits(la::Axis, dim)
 
     # otherwise start with the first box
     mini, maxi = minimum(boundingbox), maximum(boundingbox)
+
+    # equal limits not allowed
+    mini[dim] == maxi[dim] && return nothing
+
     return (mini[dim], maxi[dim])
 end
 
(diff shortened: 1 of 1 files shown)


fatteneder:
Just noted that running the `MWE` from #1670 now gives also an empty plot, because the data lies outside of the default range.
Should I also address this in this PR?

SimonDanisch:
Thank you! :) 
Looks like the text annotation is gone too in the tests:
image

fatteneder:
I looked into issue with the annotations.
Indeed, the problem is caused by the changes of this PR.
However, I am not sure whether the logic before that was even ok.

Consider this part of the code https://github.com/JuliaPlots/Makie.jl/blob/559416f27363519dde25b5faea79fd201131cbd6/src/makielayout/layoutables/axis.jl#L751-L772
and add a `display(boundingbox)` statement and rerun the test:
```julia
julia> text(". This is an annotation!", position=(300,200))
GeometryBasics.HyperRectangle{3, Float32}(Float32[300.0, 200.0, 0.0], Float32[0.0, 0.0, 0.0])
GeometryBasics.HyperRectangle{3, Float32}(Float32[300.0, 200.0, 0.0], Float32[0.0, 0.0, 0.0])
```
This suggests that the text box's `origin` is `[300,200,0]` and width is `[0,0,0]`. Because `maximum, minimum` are overloaded as `origin, origin + width`, respectively, we find equal axis limits, which this PR tries to avoid.

Q1: Is zero width for text boxes intended or a bug?
Q2: If it is a bug, should I open a new issue and resolve it there first?

fatteneder:
Just read the docs at https://makie.juliaplots.org/v0.15.1/examples/layoutables/axis/index.html and it says that 


> You can set axis limits with the functions xlims!, ylims! or limits!. The numbers are meant in the order left right for xlims!, and bottom top for ylims!. Therefore, if the second number is smaller than the first, the respective axis will reverse. 

So it seems like equal axis limits should not be ruled out or? Could somebody please confirm?

If so, then the solution for #1670 has to be different.

jkrumbiegel:
Limits where min == max should not be allowed, it doesn't make sense as all projection math in the scene breaks down in that case. It probably breaks down much earlier if we have really small distances between min and max, but I guess that is a different issue.

I have to check the logic again to make sure that the point where the equal limits are checked is the earliest-possible point, so that all logic behind that can assume non-zero-width limits. In principle, a zero width boundingbox for a plot is fine (as in the annotation case), that just means there's only one data point (the anchor of the text). But this can't proliferate to the actual axis limits. It shouldn't, as I have logic in the autolimits function that expands zero-width auto limits, I'll have to check this specific case here.

fatteneder:
Thanks for the clarification.

> ... It probably breaks down much earlier if we have really small distances between min and max, but I guess that is a different issue.

I think that is actually happening in #1670, which lead to this PR.

SimonDanisch:
Closing in favor of https://github.com/MakieOrg/Makie.jl/pull/3077

PR #1676, opened 2022-02-16
add horizontal orientation for violin plot

SirLukeSchande:
# Description

The documentation stated that violin plots could be horizontally oriented. Didn't seem to be working. I added that functionality.

## Type of change

## Checklist

### DIFF
--- a/src/stats/violin.jl
+++ b/src/stats/violin.jl
@@ -19,6 +19,7 @@ Draw a violin plot.
         boundary = automatic,
         bandwidth = automatic,
         side = :both,
+        orientation = :vertical,
         width = automatic,
         dodge = automatic,
         n_dodge = automatic,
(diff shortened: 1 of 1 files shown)


src/stats/violin.jl
```diff
@@ -47,10 +48,17 @@ end
...
     x, y = plot[1], plot[2]
     args = @extract plot (width, side, color, show_median, npoints, boundary, bandwidth,
-        datalimits, max_density, dodge, n_dodge, gap, dodge_gap)
-    signals = lift(x, y, args...) do x, y, width, vside, color, show_median, n, bound, bw, limits, max_density, dodge, n_dodge, gap, dodge_gap
+        datalimits, max_density, dodge, n_dodge, gap, dodge_gap, orientation)
+    signals = lift(x, y, args...) do x, y, width, vside, color, show_median, n, bound, bw, limits, max_density, dodge, n_dodge, gap, dodge_gap, orientation
         x̂, violinwidth = compute_x_and_width(x, width, gap, dodge, n_dodge, dodge_gap)
 
+        # for horizontal violin just flip all componentes
+        fpoint, frect = Point2f, Rectf
```
SimonDanisch, line 56: Can we revert that change? Seems unecessary and seems to degrade readability ;)

SimonDanisch:
Thank you :) @piever could you take a look too?

src/stats/violin.jl
piever: Looks like it's taken from [the crossbar recipe](https://github.com/JuliaPlots/Makie.jl/blob/7c148cd785ca5e5e9ade82c755d48263d7741125/src/stats/crossbar.jl#L67), but I agree that it decreases readability (both here and there). @SirLukeSchande, maybe something like

```julia
if orientation == :horizontal
    verts .= _flip_xy.(verts)
end
```

(and analogous for the median endpoints) would be easier to read. See eg [here](https://github.com/JuliaPlots/Makie.jl/blob/7c148cd785ca5e5e9ade82c755d48263d7741125/src/stats/boxplot.jl#L154).

SirLukeSchande: It was taken from crossbar. You are right, that the barplot implementation is more readable. Do I open a new Pull request for this?

SimonDanisch:
merged in https://github.com/JuliaPlots/Makie.jl/pull/1800

PR #1690, opened 2022-02-23
dont use Attributes type for theme

SimonDanisch:
Using the Attributes types for Themes is overly complicated, since we don't actually want every theme attribute to be converted to Observable{Any}. I think the ability to update the observables in a theme to be reflected in created plots hasn't been used a single time, and is likely also buggy.
Getting rid of Attributes in theming will also help getting rid of the Attribute type in the long run.


SimonDanisch:
Lift removal got mixed in, cleaned up PR in https://github.com/JuliaPlots/Makie.jl/pull/1694

PR #1696, opened 2022-02-24
get rid of lift

SimonDanisch:
and just use map directly


SimonDanisch:
@timholy suggested this may be worthwhile, and since it turns up as noise everywhere in the compilation profiles, it's certainly worth the simplification - also, we wanted to remove this for quite some time.
I took the chance to use some narrower types for certain observables, but I guess those changes will barely make it above the noise threshold.

SimonDanisch:
Changes in this PR:
- lift is now `const lift = map`. This mostly non breaking, main thing that this removes is `lift(f, Typ, args...)` which I don't think many people used outside Makie.
- Some tighter types for observables used that were previously using `lift(f, Any, ...)`

PR #1697, opened 2022-02-24
remove static arrays

SimonDanisch:


SimonDanisch:
Funny, didn't think this would pass :D 
This needs https://github.com/JuliaGeometry/GeometryBasics.jl/pull/169 to fully remove staticarrays, but since this seems to be non breaking, we could also merge this before tagging https://github.com/JuliaGeometry/GeometryBasics.jl/pull/169. 
Definitely a nice simplification of used types on its own.
The above plot is from running the benchmarks locally on julia 1.9 with https://github.com/JuliaGeometry/GeometryBasics.jl/pull/169 checked out.

SimonDanisch:
CI is a bit noisy and doesn't use the GeometryBasics branch, so I run the compile benchmark locally on julia master with all the right packages checked out.
Preserving this here for when CI overwrites the current plot:
image

oscardssmith:
Is this time difference changed by https://github.com/JuliaLang/julia/pull/43990?

SimonDanisch:
Both already use https://github.com/JuliaLang/julia/pull/43990?

PR #1701, opened 2022-02-25
Better cairo draw caching + code

SimonDanisch:
Looks like this is ~2s faster then the previous fastest version:
image

What's really weird though is, that every day the benchmark becomes almost exactly one second slower... 
So, while this should be, when adding up the diffs to the previously fastest version, something like 5-6s faster then the fastest version from monday, this is actually slightly slower then what I measured on monday...
Yet, when I measure the version from monday, that version is now suddenly much slower, so the gains seem real in some way. 
Since running the benchmark takes a long time, it's not that easy to pin this down, but I'll try to add more comparisons along the way to see what's going on.


timholy:
Aliens?

SimonDanisch:
Just to show what I mean, these are benchmarks from today and yesterday:
image
makie master (both julia#master) is double, one run today, one run yesteday...That's where you see the ~1s shift.
But if you compare the plots for the different versions, you see a nice speed up for each iteration.
But, no-staticarrays with julia 1.9 was 16-17s on monday, which is today at around ~22s, with every day of benchmarking it getting roughly 1s slower :-O.... 
I don't have the daily recordings for anything before yesterday, since I only started saving those yesterday, but would be pretty interesting to see...
During the day, the numbers are pretty reproducable, so benchmarking pr x against pr y shows the expected speed ups...

My best guess is, that I need to pin the manifests for the benchmarks, or that my laptop gets slower somehow ^^
I've tried the usual as deleting `.julia`, which didn't change the result of the benchmark during the day.

SimonDanisch:
My laptop is much slower on battery (at least 2x slower), so I wonder if there is some daily performance Windows 11 decides on - wouldn't be totally implausable I suppose :D

timholy:
It's Friday. Maybe it will get fast on Monday again?

Also, do you `rm -rf ~/.julia/compiled/v1.9/*` before you test? There's at least a chance that might help make it more consistent.

timholy:
I do like that trend, though, that's pretty awesome! Looks like we've already achieved a 25% savings.

Issue #521, opened 2019-08-01
Polar axis plan


asinghvi17:
- Compute datalimits (x, y) -> (r) - theta is 0-2pi no matter what
- Optimize ticks for r
- Plot circles at each tick (`lines(Circle(0f0, radius))`)
- Plot (1) linesegments going from the origin to the thetas.  Let thetas be configurable by ticks.
- All thetas in radians.
- Annotations for:
    - thetas
    - rs

- If you want different coordinates - wrap them in CoordinateTransformations.jl
- wrapper types Radii and Angles for arrays so they can be decomposed?

garrekstemo:
I was looking for a way to make a polar plot when I found this issue. I played around with the prototype @asinghvi17 posted and added some fake data. I took some ideas from Plotly, like adding a border circle outside of the data and labeling the angles just outside this border. I'm not sure how to space the r-ticks.

It would be really nice to implement this soon. I've never contributed to a project before, but is there anything I can do to help get this moving along?

```
θdata = 0:0.1:2π
rdata = 10 .* sin.(2*θdata)

rs = range(1, round(maximum(rdata)), length = 4)
θs = 0:π/4:2π
rborder = maximum(rs) * (1.10)

f = Figure()
ax = Axis(f[1, 1], aspect = 1.0)

for r in rs
    lines!(ax, Circle(Point2f(0), r), color = :lightgray)
end

lines!(ax, Circle(Point2f(0), rborder), color = :lightgray)

radiallines = zeros(Point2f, 2 * length(θs))

for (i, θ) in enumerate(θs)
    radiallines[i*2] = Point2f(rborder * cos(θ), rborder * sin(θ))
end

linesegments!(ax, radiallines, color = :lightgray)

for r in rs
    text!("$(r)", position = (r, 0), align = (:center, :bottom))
end

for θ in θs[1 : end-1]
    offset = rborder * 0.1
    text!("$(Int64(θ * 180/π))°", position = ((rborder + offset) * cos(θ), (rborder + offset) * sin(θ)), align = (:center, :center))
end

lines!(ax, rdata .* cos.(θdata), rdata .* sin.(θdata))

hidespines!(ax)
hidedecorations!(ax)

```

![plot_4](https://user-images.githubusercontent.com/8449000/155513849-87afd515-633b-460d-bdde-d06eb90632bf.png)

PR #1706, opened 2022-02-26
Add polar plot and scatter recipes

garrekstemo:
# Description

Fixes [# 521](https://github.com/JuliaPlots/Makie.jl/issues/521)

I added two recipes for creating polar plots and scatter polar plots, following the plan posted in issue #521 by @asinghvi17. I understand (from other issues) that there is some inconsistency with how Attributes are used, so I wasn't sure what to do there and mostly followed other recipes. I also would like to remove the xy axes spines and decorations, but I have seen elsewhere that this is under consideration since recipes are apparently older than the Axis framework. I wasn't sure how to handle the docs, and I put the `lines!` plot and `scatter!` plot examples in the same .md file. The plots I include in the docs are attached to this PR.

As far as setting r-ticks and theta-ticks, does my solution look reasonable? I thought this in particular would need to be reviewed.

(This is my first PR (ever!), so please guide me if there's something I'm missing)

## Type of change

## Checklist

## Images

`polar(r, theta)`

3polarplots

`polarscatter(r, theta)`
scatterplot


juliohm:
@garrekstemo any chance you could take this opportunity to also add a polar heatmap? Now that you have the concepts fresh in your mind, maybe it is easy to consider this plot as well?

ffreyer:
I think the way we would want to implement polar plots is by adding a transform function (https://github.com/JuliaPlots/Makie.jl/blob/f04ea93cde4384bad3f008bddf291a15ac35774a/src/layouting/transformation.jl) that converts polar to Cartesian coordinates. That way (r, \theta) inputs would work for any plot that uses that transform. Polar axis should probably be their own separate thing. I.e. we should have something like `PolarAxis` in analog to `Axis` so you can pick the layout you want.

garrekstemo:
OK, that makes a lot more sense and would be more generalizable. I'll close this PR and help out with transform functions where I can.

asinghvi17:
GeoMakie is now just a wrapper around Axis with some conveniences for projections, but does everything else using the Scene transformation interface, specifically `scene.transformation.transform_func` and defining `Makie.apply_transform`.

I have been meaning to make a sort of "NonlinearAxis" `Block` for a while but have been slowly iterating on how it might work and what changes would be desirable in Makie.

As of now it should be easy to simply feed a polar-to-cartesian converter to transform_func and possibly use the framework of this PR to draw gridlines.  The only question then would be the logic of where and how to draw ticks...

adigitoleo:
Thanks for the info, I should have checked more of the GeoMakie source. I'll try with transform_func, since I don't really need ticks for my case ("pole figures" of crystalline texture, [Wikipedia](https://en.wikipedia.org/wiki/Pole_figure), [rust prototype](https://github.com/MFraters/cpo_analyzer)) The contouring might get interesting. I don't expect this kind of plot in particular to be included in Makie core, but drawing the basic axis is similar to a polar plot. If NonlinearAxis is on the table, then I'll also mention  [stereographic plots](https://github.com/joferkington/mplstereonet) in general. I can try to help with all of the above.

adigitoleo:
I've started a thread [over at discourse](https://discourse.julialang.org/t/non-affine-transformations-in-makie/81951) to discuss this. I'm also interested in doing Ternary plots, and can try to help with that as well later.

adigitoleo:
#2014

PR #1707, opened 2022-02-26
Fix timeseries

### LINKED ISSUE #1669: `Invalid text boundingbox error` in `TimeSeries`
Hi.

I wanted to contribute a `timeseries` example to #1639.
But when I try to follow the docs for `timeseries` the following happens

```julia
julia> signal = Observable(1.0)
Observable{Float64} with 0 listeners. Value:
1.0

julia> scene = timeseries(signal)
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
┌ Warning: No strict ticks found
└ @ PlotUtils ~/.julia/packages/PlotUtils/rTTui/src/ticks.jl:333
ERROR: Invalid text boundingbox
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] boundingbox(glyphcollection::Makie.GlyphCollection, position::Point{3, Float32}, rotation::Quaternionf)
 ...
```

A similar error message was mentioned as a side note in #1490.
However, when trying their MWE now gives a different error (related to invalid range steps for `log10` scaling). I will open a separate issue for this.

fatteneder:
# Description

Fixes #1669

The initial scene upon creation of a `TimeSeries` plot was assumed to be a 3D one, because the data stream is filled with `NaNs`. Fixed this by adjusting `is2d` to return `true` for 3D rectangles who's width in z direction is `-Inf`.
Also overloaded `isopen` for `FigureAxisPlot` so that MWE from `TimeSeries` docs works again.

Automatic x, y limits as shown in https://github.com/JuliaPlots/Makie.jl/issues/736 don't work anymore. But that might be a different issue.

## Type of change

### DIFF
--- a/src/basic_recipes/timeseries.jl
+++ b/src/basic_recipes/timeseries.jl
@@ -40,8 +40,13 @@ function Makie.plot!(plot::TimeSeries)
     points = Observable(fill(Point2f(NaN), plot.history[]))
     buffer = copy(points[])
     lines!(plot, points)
-    start = time()
+    start = 0
+    started = false
     on(plot.signal) do x
+        if !started
+            start = time()
+            started = true
+        end
         points[][end] = signal2point(x, start)
         circshift!(buffer, points[], 1)
         buff_ref = buffer
--- a/src/basic_recipes/timeseries.jl
+++ b/src/basic_recipes/timeseries.jl
@@ -1,7 +1,10 @@
 """
-    timeseries(x::Observable{{Union{Number, Point2}}})
+    timeseries(x::Observable{{Union{Number, Point2}}}; kwargs...)
 
 Plots a sampled signal.
+
+The `history::Int=100` argument controls how many data points are tracked.
+
 Usage:
 ```julia
 signal = Observable(1.0)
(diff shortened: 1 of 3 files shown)


fatteneder:
Coming back to this after one month and looking at it, I must agree with you.
Not sure why I solved it the hacky way.

I ll give it another go perhaps at end of this week.

fatteneder:
I went through the stacktrace again and I found the real problem:
```
lines(Point2f(NaN))
```
is already not working (gives the same `Invalid boudingbox` error as in #1669).
The reason is that this code here https://github.com/JuliaPlots/Makie.jl/blob/28e5fc6f130e1f1a96989895f6ccb025e74d6999/src/figureplotting.jl#L24-L28 generates a 3D rectangle instead of a 2D one (due to `is2d`, as mentioned above).

Perhaps this recent PR #1770 could also solve the problem here?

fatteneder:
Just checked: PR #1770 will fix issue #1669, hence, this PR here is obsolete.

However, the automatic x,y limits are still disabled, but I will look into that once #1770 is merged.

PR #1716, opened 2022-03-02
Use TracePrecompiles for larger precompile coverage

SimonDanisch:


timholy:
FYI, SnoopCompile will give you the "base of the stack" and everything else that it calls (inferrably) will follow. `--trace-compiles` is a bit redundant because it will specify caller and callee (even though the callee will be compiled when the caller is). It's fine to do this, but I worry it could be a maintenance headache: just running the commands you'd like to have be fast seems more robust across Julia versions, 32/64-bit, windows/linux/mac, etc.

SimonDanisch:
Yeah this is a bit of an experiment - but it did gave me much better caching then all the other approaches...
But now, like you said, I need to figure out how to make it maintainable ;) 
I also still don't 100% understand, why running the same code that I use in the tracing during precompilation doesn't yield the same results.
I'm almost tempted to run `TracePrecompiles.trace_compiles("CairoMakie", precomp_run_file, precompile_file)` during precompilation and then just executing the result to have it run with the correct julia version etc. 
But that'd slow down precompilation quite a bit I suppose.

timholy:
That's probably the inferrability getting in the way, and I can understand why that's annoying. One option is to add manual precompiles for just the things that simpler approaches miss. For example, you could look at the flamegraph after having done the "easy" precompiles and then pick any sufficiently-fat bars for manual precompile directives.

SimonDanisch:
Yes I realize, that this is the preferred way.
But in my experience, that also comes with a heavy maintenance burden, since updating those require quite a lot of work and can't be done by anyone easily.

Ideally, I want a solution that gives me 100% coverage in a quick way 🤷 

I'll definitely investigate some alternatives, though...
Maybe there is more to gain from approaches like `precompile(::Observable)`.
I added some along that line here: https://github.com/JuliaPlots/Makie.jl/blob/b5e23fdbfaca811c13ee9965228039708383f166/CairoMakie/src/precompiles.jl#L73, which also does seem to work relatively well.

timholy:
We're all feeling our forward here together, trying to come up with best practices. I'm looking forward to learning from you as I see what you settle on!

SimonDanisch:
Closing in favor of https://github.com/JuliaPlots/Makie.jl/pull/1848

PR #1717, opened 2022-03-02
Small clean ups

SimonDanisch:
Will add some more clean ups tomorrow

### DIFF
--- a/GLMakie/src/drawing_primitives.jl
+++ b/GLMakie/src/drawing_primitives.jl
@@ -47,47 +47,36 @@ end
 make_context_current(screen::Screen) = GLFW.MakeContextCurrent(to_native(screen))
 
 function connect_camera!(gl_attributes, cam, space = gl_attributes[:space])
-    # (0, 1) x (0, 1) x (0, 1) space
-    rel = Mat4f(2, 0, 0, 0, 0, 2, 0, 0, 0, 0, 1, 0, -1, -1, 0, 1)
-    # (-1, 1) x (-1, 1) x (0, 1) space
-    id = Mat4f(I)
-
-    for key in (:pixel_space, :view, :projection, :resolution, :eyeposition, :projectionview)
-        if !haskey(gl_attributes, key)
-            gl_attributes[key] = if key in (:projection, :projectionview)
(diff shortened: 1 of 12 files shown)


src/camera/projection_math.jl
```diff
@@ -312,43 +312,44 @@ end
...
     elseif is_relative_space(space)
-        Mat4f(2, 0, 0, 0, 0, 2, 0, 0, 0, 0, 1, 0, -1, -1, 0, 1)
+        return Mat4f(2, 0, 0, 0, 0, 2, 0, 0, 0, 0, 1, 0, -1, -1, 0, 1)
     elseif is_clip_space(space)
-        Mat4f(I)
+        return Mat4f(I)
     else
         error("Space $space not recognized. Must be one of $(spaces())")
     end
 end
```
ffreyer, line 331: Relative and clip space should also return observables

GLMakie/src/drawing_primitives.jl
```diff
@@ -47,47 +47,36 @@ end
...
             return transpose(inv(v[i, i] * m[i, i]))
         end
     end
 
+    get!(gl_attributes, :projection) do
+        Makie.space_to_clip_obs(cam, space[], false)
+        return map(cam.projection, space) do _, space
+            return Makie.space_to_clip(cam, space, true)
+        end
+    end
```
ffreyer, line 70: The matrix here can be `cam.projection` or `cam.pixel_space` so it should trigger on both of those and `space`. Also seems like you didn't update this correctly

ffreyer: I think for 2d, 3d and pixel cameras reacting to `pixelspace` isn't necessary but for `RelativeCamera` it would be?

src/camera/projection_math.jl
SimonDanisch: Actually, it seems the observable version isn't needed... I'll remove it

GLMakie/src/drawing_primitives.jl
SimonDanisch: hm, for simplicity I'll just add pixel_space to it for now...
The right way to do this is hard to express with observables... I need to think about a cleaner way going forward...

ffreyer (approved):

PR #1730, opened 2022-03-06
Add mixed coordinate spaces

ffreyer:
# Description

This pr is sort of a continuation of #1596 with the goal of adding mixed coordinate spaces. We have some example applications of this in Makie already: `vlines`, `hlines`, `errorbar` and friends.

My idea for this is to allow passing a tuple of symbols as `space = (x_space, y_space[, ...])` and generate the appropriate matrices from it.  Something like axis-spanning horizontal could then be done with 

```julia
lines!(ax, [0, 1], [y, y], space = (:relative, :data))
```
(This already works in GLMakie)

## Type of change

## Checklist

- [x] GLMakie
- [x] CairoMakie
- [x] WGLMakie


SimonDanisch:
Hm, this seems to break WGLMakie volume plots...

PR #1736, opened 2022-03-10
Colorbar example for heatmap

KronosTheLate:
# Description

This PR adds documentation to the page on [`heatmap`](https://makie.juliaplots.org/stable/examples/plotting_functions/heatmap/index.html), specifically how to make a colorbar.

It also removes the previous final section [Colors](https://makie.juliaplots.org/stable/examples/plotting_functions/heatmap/index.html#colors), as it 
1) Seems to have an outdated reference, and
2) had nothing else, effectively contributing nothing but noise.

### DIFF
--- a/docs/examples/plotting_functions/heatmap.md
+++ b/docs/examples/plotting_functions/heatmap.md
@@ -128,7 +128,7 @@ ys = range(0, 2π, length=100)
 zs1 = [sin(x*y) for x in xs, y in ys]
 zs2 = [2sin(x*y) for x in xs, y in ys]
 
-my_limits = extrema(hcat(zs1, zs2))  # Get maximum and minimum for all z-values
+my_limits = (-2, 2)  # choose some good limits for sin / 2sin
 
 fig, ax1, hm1 = heatmap(xs, ys, zs1,  colorrange = my_limits)
 ax2, hm2 = heatmap(fig[1, end+1], xs, ys, zs2, colorrange = my_limits)
--- a/docs/examples/plotting_functions/heatmap.md
+++ b/docs/examples/plotting_functions/heatmap.md
@@ -81,3 +81,62 @@ zs = [1, 2, 3, 4, 5, 6, 7, 8, NaN]
 heatmap(xs, ys, zs)
 ```
 \end{examplefigure}
+
+### Colorbar for single heatmap
+
+To get a scale for what the colors represent, add a colorbar. The colorbar is 
+placed within the figure in the first argument, and the scale and colormap can be 
+conveniently set by passing the relevant heatmap to it.
+
+\begin{examplefigure}{}
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+Makie.inline!(true) # hide
+
+xs = range(0, 2π, 100)
+ys = range(0, 2π, 100)
+zs = [sin(x*y) for x in xs, y in ys]
+
+fig, ax, hm = heatmap(xs, ys, zs)
+Colorbar(fig[:, end+1], hm)
+
+fig
+```
+\end{examplefigure}
+
+### Colorbar for multiple heatmaps
+
+When there are several heatmaps in a single figure, it can be useful
+to have a single colorbar represent all of them. It is important to then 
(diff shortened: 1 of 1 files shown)


KronosTheLate:
Credit also goes to @MaPagh for pointing out this lack of documentation, so that it could be added.

KronosTheLate:
Friendly bump

docs/examples/plotting_functions/heatmap.md
```diff
@@ -82,6 +82,61 @@ heatmap(xs, ys, zs)
...
+using CairoMakie
+CairoMakie.activate!() # hide
+Makie.inline!(true) # hide
+
+xs = range(0, 2π, length=100)
+ys = range(0, 2π, length=100)
+zs1 = [sin(x*y) for x in xs, y in ys]
+zs2 = [2sin(x*y) for x in xs, y in ys]
+
+my_limits = extrema(hcat(zs1, zs2))  # Get maximum and minimum for all z-values
```
jkrumbiegel, line 131: I don't like this too much because I can see the `hcat` breaking on different datasets already, where the matrices don't match. I haven't been able to come up with a really nice and short "cumulative extrema of multiple arrays" expression though that's easy to understand. Maybe it's fine just specifying `-2, 2` here manually as we know that from the `2sin` expression.

docs/examples/plotting_functions/heatmap.md
```diff
@@ -82,6 +82,61 @@ heatmap(xs, ys, zs)
...
+
+fig
+```
+\end{examplefigure}
+
+### Colorbar for multiple heatmaps
+
+When there are several heatmaps in a single figure, it can be useful
+to have a single colorbar represent all of them. It is important to then 
+have syncronized scales and colormaps for the heatmaps and colorbar. This is done by
```
jkrumbiegel, line 112: synchronized

docs/examples/plotting_functions/heatmap.md
SimonDanisch: How about `extrema(extrema.((zs1, zs2)))`

jkrumbiegel: That makes a tuple of tuples

SimonDanisch: Ah yeah...
@KronosTheLate, should we just use `(-2, 2)`... I think that's fine here, since we're not trying to show how to get the extrema from the values ;)

jkrumbiegel: I think that's fair, it's reasonably obvious with the sine (actually it can even look better because if the limits go from -1.99123 to 1.992852 then the colorbar ticks don't have -2 and 2)

SimonDanisch: Ok, I'll edited it... @KronosTheLate I think we can merge it now, if you don't have objections

KronosTheLate: No objections - extrema of extrema is besides the point of this documentation, so manual limits, though inelegant in some ways, seems like a great solution for this documentation ^_^

docs/examples/plotting_functions/heatmap.md
```diff
@@ -128,7 +128,7 @@ ys = range(0, 2π, length=100)
 zs1 = [sin(x*y) for x in xs, y in ys]
 zs2 = [2sin(x*y) for x in xs, y in ys]
 
-my_limits = extrema(hcat(zs1, zs2))  # Get maximum and minimum for all z-values
+my_limits = (-2, 2)  # choose some good limits for sin / 2sin
```
jkrumbiegel, line 131: maybe not `my_limits` but `joint_limits` or something, and say `# here we pick the limits manually for simplicity instead of computing them` so that it's clear that that would be the normal course of action?

KronosTheLate: Good points. I agree with thoose changes.
The optimal thing would perhaps be a helper function that would find the extremas of any heatmap, contour or contour3D (I think those are the ones this would apply to?) in the given figure.

PR #1739, opened 2022-03-10
Fix `Colorbar` with all zero heatmap

t-bltg:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/issues/1738.

With pr:
hm-mid

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/layoutables/colorbar.jl
+++ b/src/makielayout/layoutables/colorbar.jl
@@ -74,12 +74,16 @@ function layoutable(::Type{<:Colorbar}, fig_or_scene; bbox = nothing, kwargs...)
         nsteps, highclip, lowclip,
         minorticksvisible, minortickalign, minorticksize, minortickwidth, minortickcolor, minorticks, scale)
 
-    limits = lift(limits, colorrange) do limits, colorrange
-        if all(!isnothing, (limits, colorrange))
-            error("Both colorrange + limits are set, please only set one, they're aliases. colorrange: $(colorrange), limits: $(limits)")
+    limits = lift(limits, colorrange) do lims, crange
+        if all(!isnothing, (lims, crange))
+            error("Both colorrange + limits are set, please only set one, they're aliases. colorrange: $(crange), limits: $(lims)")
         end
-        lo, hi = lims = something(limits, colorrange, (0, 1))
-        return lo == hi ? (zero(lo), one(hi)) : lims
+        lo, hi = real_lims = something(lims, crange, (0, 1))
+        if lo == hi
+            real_lims = oftype(lo, lo - 0.5), oftype(hi, hi + 0.5)
+            (lims !== nothing ? limits : colorrange)[] = real_lims  # update back
+        end
+        return real_lims
     end
 
     decorations = Dict{Symbol, Any}()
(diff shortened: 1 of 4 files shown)


src/makielayout/layoutables/colorbar.jl
```diff
@@ -78,7 +78,8 @@ function layoutable(::Type{<:Colorbar}, fig_or_scene; bbox = nothing, kwargs...)
         if all(!isnothing, (limits, colorrange))
             error("Both colorrange + limits are set, please only set one, they're aliases. colorrange: $(colorrange), limits: $(limits)")
         end
-        return something(limits, colorrange, (0, 1))
+        lo, hi = lims = something(limits, colorrange, (0, 1))
+        return lo == hi ? (zero(lo), one(hi)) : lims
```
SimonDanisch, line 82: shouldn't this be something like `(hi - 0.5, hi + 0.5)` ?

t-bltg: Oh, yes I forgot about centering, sorry, will fix. I also tried to fix this after in the pipeline (in `src/makielayout/lineaxis.jl`), but without success.

t-bltg: Oddly, when applying `(oftype(lo, lo - .5), oftype(hi, hi + .5))` instead, I've got this

z_rng

, which I'm afraid is obviously wrong.

jkrumbiegel: the heatmap also needs the colorrange

jkrumbiegel: what you did is build a colorbar that takes its colormap from the heatmap but sets different limits, without synching those back to the heatmap.

t-bltg: Thanks for pointing that out. Here is the new output:
hm-mid

SimonDanisch (approved):
Sounds good to me! @jkrumbiegel  any concerns?

jkrumbiegel:
Maybe the test should check what the limit values in the colorbar are? Not sure if that's accessible actually. And what happens when the heatmap observable changes so there is now a nonempty range, will the colormap adjust correctly? Maybe that should be included in the test. Also, the 0.5 offset might lead to problems for really large float32 singular values, maybe a factor off 1.5 or so could work? Unless zero

t-bltg:
> Maybe the test should check what the limit values in the colorbar are

I've added test for `colorrange` and `limits`.

> And what happens when the heatmap observable changes so there is now a nonempty range

I've also added a test changing either `hm.colorrange` or `cb.limits`.

> Also, the 0.5 offset might lead to problems for really large float32 singular values

As long as you have `hi - lo > eps(Float32)`, I think that should suffice to avoid the initial error `ERROR: ArgumentError: range step cannot be zero`.

t-bltg:
I've found a way to change the `colorrange` upstream in the pipeline, and updated the PR.
This is cleaner IMO, please comment.

jkrumbiegel (approved):

SimonDanisch:
Thank you! That looks like a fair solution!

PR #1741, opened 2022-03-11
use apply_transform directly on lims

simonbyrne:
# Description

This helps fix https://github.com/JuliaPlots/GeoMakie.jl/issues/99

As explained by @visr in https://github.com/JuliaPlots/GeoMakie.jl/issues/99#issuecomment-1020510420, the problem is that the current bounding box is computed by just using `apply_transform` on the minimum and maximum points: this doesn't work for more complicated transformations.

Instead we now call `apply_transform` directly on the `Rect`, which can then be overloaded for custom transformations (e.g. see https://github.com/JuliaPlots/GeoMakie.jl/pull/101)

## Type of change

## Checklist

### DIFF
--- a/src/layouting/transformation.jl
+++ b/src/layouting/transformation.jl
@@ -298,6 +298,14 @@ function apply_transform(f, r::Rect)
     ma_t = apply_transform(f, ma)
     Rect(Vec(mi_t), Vec(ma_t .- mi_t))
 end
+function apply_transform(f::PointTrans, r::Rect)
+    mi = minimum(r)
+    ma = maximum(r)
+    mi_t = apply_transform(f, Point(mi))
+    ma_t = apply_transform(f, Point(ma))
+    Rect(Vec(mi_t), Vec(ma_t .- mi_t))
+end
+
 # ambiguity fix
 apply_transform(f::typeof(identity), r::Rect) = r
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Thanks!

src/layouting/transformation.jl
```diff
@@ -298,6 +298,14 @@ function apply_transform(f, r::Rect)
     ma_t = apply_transform(f, ma)
     Rect(Vec(mi_t), Vec(ma_t .- mi_t))
 end
+function apply_transform(f::PointTrans, r::Rect)
+    mi = minimum(r)
+    ma = maximum(r)
+    mi_t = apply_transform(f, Point(mi))
+    ma_t = apply_transform(f, Point(ma))
+    Rect(Vec(mi_t), Vec(ma_t .- mi_t))
+end
```
simonbyrne, line 307: @SimonDanisch Would it make sense to change `apply_transform(f, r::Rect)` to operate on `Point` types as well?
https://github.com/JuliaPlots/Makie.jl/blob/284222754f4106ffb6829d39b612e5011e297db7/src/layouting/transformation.jl#L301-L307
or even change the definition of `minimum(::Rect)`/`maximum(::Rect)` to give a `Point`? I don't quite understand enough about how they're used to make the change, but if we do that then we can delete this method.

PR #1743, opened 2022-03-11
refactor and improve conversions

SimonDanisch:
...and make them type stable.
adopts breaking change from https://github.com/JuliaPlots/Makie.jl/pull/1723, which makes `to_colormap` always return `Vector{RGBAf}`.


SimonDanisch:
merged in https://github.com/JuliaPlots/Makie.jl/pull/1723

PR #1745, opened 2022-03-13
Explain axis behavior with contourf levels vector

johnbcoughlin:
# Description

Explain a subtlety with the contourf function that had me confused.

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/contourf.jl
+++ b/src/basic_recipes/contourf.jl
@@ -15,6 +15,8 @@ This can be used for example to draw bands for the upper 90% while excluding the
 In :normal mode, if you want to show a band from `-Inf` to the low edge, set `extendlow` to `:auto` for the same color as the first level, or specify a different color (default `nothing` means no extended band)
 If you want to show a band from the high edge to `Inf`, set `extendhigh` to `:auto` for the same color as the last level, or specify a different color (default `nothing` means no extended band)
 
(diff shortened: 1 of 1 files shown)


src/basic_recipes/contourf.jl
```diff
@@ -15,6 +15,8 @@ This can be used for example to draw bands for the upper 90% while excluding the
 In :normal mode, if you want to show a band from `-Inf` to the low edge, set `extendlow` to `:auto` for the same color as the first level, or specify a different color (default `nothing` means no extended band)
 If you want to show a band from the high edge to `Inf`, set `extendhigh` to `:auto` for the same color as the last level, or specify a different color (default `nothing` means no extended band)
 
+If you specify `levels` as an `AbstractVector{<:Real}`, note that the axis limits will be pulled back from the bounds of the data. This is to avoid unsightly zero margins between an irregular contour and the axis. To undo this behavior, you can use `tightlimits!(ax)` or the other axis limit functions.
```
jkrumbiegel, line 18: It's kind of the other way around, `tightlimits!` is only called when the limits are given as an integer https://github.com/JuliaPlots/Makie.jl/blob/master/src/makielayout/layoutables/axis.jl#L709-L713

jkrumbiegel:
I've gone ahead and edited your entry, thanks for the contribution

PR #1746, opened 2022-03-13
add new theme.

lazarusA:
# Description

This is a new theme, theme_makie :D, since I don't want to change the current defaults. Why theme_makie? the first 3 colours in the palette definition are close to the ones used in the logo makie. Other two, are close to the Julia Logo. 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] Added or changed relevant sections in the documentation

### DIFF
--- a/docs/documentation/theming/predefined_themes.md
+++ b/docs/documentation/theming/predefined_themes.md
@@ -118,3 +118,329 @@ with_theme(demofigure, theme_light())
 with_theme(demofigure, theme_dark())
 ```
 \end{examplefigure}
+
+## theme_makie
+
+### scatter and lines
+\begin{examplefigure}{}
+
+```julia
+with_theme(theme_makie()) do
+    x = range(0, 20; length=7)
+    fig = Figure()
+    ax = Axis(fig[1, 1])
+    for (idx, i) in enumerate(x)
+        scatterlines!(ax, x, i .* x; label="$idx")
+    end
+    axislegend("Label"; merge=true, position=:lt)
+    fig
+end
+```
+
+### Heatmaps
+
+\end{examplefigure}
+
+```julia
+with_theme(theme_makie()) do
+   x = range(0, 2π, 50)
+    fig, ax, hm = heatmap(rand(10, 10); figure=(; resolution=(1200, 900)))
+    ax1 = Axis(fig[1, 3]; xlabel="", ylabel="")
+    ax2 = Axis(fig[2, 1:3])
+    hm2 = heatmap!(ax1, rand(10, 10); colorrange=(0.2, 0.8), highclip=:black, lowclip=:grey)
+    hm3 = lines!(ax2, x, sin.(x); color=x, linewidth=5)
+    Colorbar(fig[1, 2], hm)
+    Colorbar(fig[1, 4], hm2; label="")
+    Colorbar(fig[2, 4], hm3; label="x")
+    fig
+end
+```
+
+\end{examplefigure}
+
+### On how to use Cycles with this theme for scatter and lines
+
+~~~
+<input id="hidecode" class="hidecode" type="checkbox">
+~~~
+
+```julia:demofig
+
+using CairoMakie
+CairoMakie.activate!(type = "svg") # hide
+
+function demoScatterLines()
+    x = range(0, 20; length=7)
+    fig = Figure(; resolution=(2600, 1600))
+    ax11 = Axis(fig[1, 1]; title="lines(x,y)")
+    for (idx, i) in enumerate(x)
+        lines!(ax11, x, i .* x; label="$idx")
+    end
+    axislegend("Label"; merge=true)
+
+    ax12 = Axis(fig[1, 2]; title="lines(x,y; cycle=:color)")
+    for (idx, i) in enumerate(x)
+        lines!(ax12, x, i .* x; label="$idx", cycle=:color)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax13 = Axis(fig[1, 3]; title="lines(x,y; cycle=:linestyle)")
+    for (idx, i) in enumerate(x)
+        lines!(ax13, x, i .* x; label="$idx", cycle=:linestyle)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax14 = Axis(fig[1, 4]; title="lines(x,y; linestyle=:dash,\n cycle=:color)")
+    for (idx, i) in enumerate(x)
+        lines!(ax14, x, i .* x; label="$idx", linestyle=:dash, cycle=:color)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax15 = Axis(fig[1, 5]; title="lines(x,y; color=:red,\n cycle=:linestyle)")
+    for (idx, i) in enumerate(x)
+        lines!(ax15, x, i .* x; label="$idx", color=:red, cycle=:linestyle)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax21 = Axis(fig[2, 1]; title="scatter(x,y)")
+    for (idx, i) in enumerate(x)
+        scatter!(ax21, x, i .* x; label="$idx")
+    end
+    axislegend("Label"; merge=true)
+
+    ax22 = Axis(fig[2, 2]; title="scatter(x,y; cycle=:color)")
+    for (idx, i) in enumerate(x)
+        scatter!(ax22, x, i .* x; label="$idx", cycle=:color)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax23 = Axis(fig[2, 3]; title="scatter(x,y; cycle=:marker)")
+    for (idx, i) in enumerate(x)
+        scatter!(ax23, x, i .* x; label="$idx", cycle=:marker)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax24 = Axis(fig[2, 4];
+                title="scatter(x,y; color=:transparent, \n strokewidth=2.0, \n cycle= Cycle([:strokecolor, :marker]; covary=true))")
+    for (idx, i) in enumerate(x)
+        scatter!(ax24, x, i .* x; label="$idx", color=:transparent, strokewidth=2.0,
+                 cycle=Cycle([:strokecolor, :marker]; covary=true))
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax25 = Axis(fig[2, 5]; title="scatter(x,y; color=:red,\n cycle=:marker")
+    for (idx, i) in enumerate(x)
+        scatter!(ax25, x, i .* x; label="$idx", color=:red, cycle=:marker)
+    end
+    axislegend("Label"; merge=true, position=:lt)
+
+    ax31 = Axis(fig[3, 1]; title="scatterlines(x,y)")
+    for (idx, i) in enumerate(x)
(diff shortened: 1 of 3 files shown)


lazarusA:
These are some examples added to the docs, if they work. 


demoScatterLines

demo3dmix
demoStats

jkrumbiegel:
Thank you for your work on the new theme!

After having a quick look at the docs build, here are some first suggestions, in no particular order:

- I think `theme_makie` might be confusing as it suggests it is the default when it's not. People would then need to ask, did you use the Makie theme or `theme_makie`?
- In the docs build, the scatter markers are huge
grafik
- You have changed the resolution in the theme, I think that's also confusing for a standard theme. None of the other themes change this value and it should vary a lot by use case anyway which figure size is appropriate.
- The overview plots are cool but some are kind of too huge for the website, making text look quite small etc. Also, I don't like that only this theme has extensive examples and the others don't. If you want to add docs on cycling etc, maybe on the main theming page, but the example page should have the same examples for every theme so they can be easily compared.
- Aside from the huge markers, I think it would be better if the 3d plots were less random, as that always looks less "polished" than with some more carefully chosen data like in a real plot.

lazarusA:
- theme_makie is indeed not a good name.  I'm open to suggestions. theme_makie2 ? theme_jlmke ? 
- I changed the figure size and ratio, because I'm always changing that for plots that go into papers. The default one on my end is never a good starting point.
-  "overview plots"... agree... I will rm them and just leave the same example used for the other themes.
- cycling. Those were a pain(the examples) to get them right. So, there should be more examples in the docs for them. So, probably I will moved somewhere else as suggested. 
- marker sizes.... is the usual pain between GLMakie and CairoMakie.... with GLMakie they are ok. I didn't know how to properly verify if that was being done correctly in the docs. 

thanks for your suggestions.

ffreyer:
> marker sizes.... is the usual pain between GLMakie and CairoMakie.... with GLMakie they are ok. I didn't know how to properly verify if that was being done correctly in the docs.

#1596 changed how markersizes and textsizes are handled in scatter and text plots. On master (with the default `markerspace = :pixel`) markersize should now be in pixel units regardless of 2d or 3d. It should also be consistent across backends.  So you can just drop the `markersize = 80` and you should get good results.

lazarusA:
- Ok. I rm the fig size and left the default one. 
- Changed the name to `theme_jlmke`. 
- Also, now the theme follows the convention for themes, using just the example given there. All other examples were move to theming and cycling. 

Thanks for all the suggestions, things should work now.

SimonDanisch:
Is this ready to be merged?

lazarusA:
I removed all the extra docs in order to be uniform with the other themes. Also, the theme is now called `theme_jlmke` (a mix of Julia and makie colours). This one should be now ready to merge @SimonDanisch ?

lazarusA:
bump @SimonDanisch

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,69 @@
+function theme_jlmke()
+    function jlmke_colors(alpha=1.0)
+        colors = [RGB(0.0, 0.0, 0.0), RGB(0.082, 0.643, 0.918), RGB(0.91, 0.122, 0.361),
+                  RGB(0.929, 0.773, 0.0), RGB(0.588, 0.196, 0.722), RGB(0.361, 0.722, 0.361),
+                  RGB(0.522, 0.522, 0.522)]
+        @. RGBAf(red(colors), green(colors), blue(colors), alpha)
+    end
```
asinghvi17, line 7: The function should probably live outside the main `theme_jlmke()` function, so that users can access the colors!

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,69 @@
...
+        Colorbar=(
+            label = "f(x,y)",
+            ticksize=15,
+            tickalign=0.5,
+            spinewidth=0.0,
+            minorticksvisible= true),
+        LScene = (
+            scenekw = (show_axis=true,)),
+            backgroundcolor = :white,
+    )
```
asinghvi17, line 67: I think the API on this has changed - show_axis is no longer in `scenekw`.

```suggestion
        LScene = (
            show_axis = true,
            backgroundcolor = :white
    )
```
or just remove it entirely, if we want it to default to whatever the user has set, and not interfere with that.
```suggestion
```

asinghvi17 (commented):
Looks good to me!  Couple of questions about some of the style choices though, which I've put in review comments.

src/themes/theme_jlmke.jl
lazarusA: good point.

lazarusA: could you please to a suggestion change directly to the file? please.

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,69 @@
...
+            ylabeloffset = 55,
+            zlabeloffset = 70,
+            xgridcolor = (:black, 0.07),
+            ygridcolor = (:black, 0.07),
+            zgridcolor = (:black, 0.07),
+            perspectiveness = 0.5f0,
+            azimuth = 1.275π * 1.77,
+        ),
+        Colorbar=(
+            label = "f(x,y)",
```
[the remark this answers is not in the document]
lazarusA: IMO a plot without default labels is a bad plot. Even if they don't correspond to the ones the users wants. Think of them as placeholders that the user will change later, or can do while creating the plot. Actually, you encourage the user to actually do the proper labelling of their plots.

src/themes/theme_jlmke.jl
asinghvi17, line 7:
```suggestion
function jlmke_colors(alpha=1.0)
    colors = [RGB(0.0, 0.0, 0.0), RGB(0.082, 0.643, 0.918), RGB(0.91, 0.122, 0.361),
              RGB(0.929, 0.773, 0.0), RGB(0.588, 0.196, 0.722), RGB(0.361, 0.722, 0.361),
              RGB(0.522, 0.522, 0.522)]
    @. RGBAf(red(colors), green(colors), blue(colors), alpha)
end

"""
    theme_jlmke()::Attributes

A theme by Lazaro Alonso (`@lazarusa`).
Code to set the theme to `theme_jlmke` is as follows: `Makie.set_theme!(theme_jlmke())`.

The first 3 colours in the palette definition are close to the ones used in the logo of Makie. The other two are close to those of the Julia logo.
"""
function theme_jlmke()
```

Added a small docstring and some tips, hope they make sense to have here.

src/themes/theme_jlmke.jl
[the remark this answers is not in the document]
lazarusA: and the rotations and offsets for Axis3 are there because the current default ones don't look as pleasant as they could be 😄

src/themes/theme_jlmke.jl
lazarusA: thanks.

src/themes/theme_jlmke.jl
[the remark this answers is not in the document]
asinghvi17: Hmm, I see what you're saying.  Fair enough - it just struck me as odd since I don't think any other Makie theme does that.

jkrumbiegel:
After thinking about it some more, what I overall still don't quite like is that this theme doesn't serve a specific "purpose" like the other themes that we have. At least for each of those I had a specific use case in mind, like dark, light, minimalist, general default, matching ggplot, etc. In principle it doesn't matter because we can put however many themes we like into Makie, who cares, on the other hand I'd like the main themes to be somewhat curated so the choice for the user is easier. I'm not sure what the "theme" of this theme is. For example, it would be easier to add the color scheme as a separate thing, if that defines the look. We don't have specific themes for all the other color schemes either. The other settings I could see in the changes seem more like personal preferences to me, which is fine, it's just not a purpose. Then we might as well call it `theme_lazaro` ;)

lazarusA:
> doesn't serve a specific "purpose" 

Fair enough. Then I would just simply say, let's just called it `theme_pretty` whatever that means. Everyone has his own opinion about what is pretty or not.

docs/documentation/theming/predefined_themes.md
```diff
@@ -118,3 +118,13 @@ with_theme(demofigure, theme_light())
 with_theme(demofigure, theme_dark())
 ```
 \end{examplefigure}
+
+## theme_jlmke
+
+\begin{examplefigure}{}
+```julia
+with_theme(demofigure, theme_jlmke())
```
lazarusA, line 126:
```suggestion
with_theme(demofigure, theme_ pretty())
```

src/Makie.jl
```diff
@@ -115,6 +115,7 @@ include("themes/theme_black.jl")
 include("themes/theme_minimal.jl")
 include("themes/theme_light.jl")
 include("themes/theme_dark.jl")
+include("themes/theme_jlmke.jl")
```
lazarusA, line 118:
```suggestion
include("themes/theme_ pretty.jl")
```

src/Makie.jl
```diff
@@ -202,6 +203,7 @@ export theme_black
 export theme_minimal
 export theme_light
 export theme_dark
+export theme_jlmke
```
lazarusA, line 206:
```suggestion
export theme_ pretty
```

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,78 @@
+function jlmke_colors(alpha=1.0)
+    colors = [RGB(0.0, 0.0, 0.0), RGB(0.082, 0.643, 0.918), RGB(0.91, 0.122, 0.361),
+              RGB(0.929, 0.773, 0.0), RGB(0.588, 0.196, 0.722), RGB(0.361, 0.722, 0.361),
+              RGB(0.522, 0.522, 0.522)]
+    @. RGBAf(red(colors), green(colors), blue(colors), alpha)
+end
+
+"""
+    theme_jlmke()::Attributes
```
lazarusA, line 9:
```suggestion
    theme_ pretty()::Attributes
```

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,78 @@
...
+
+A theme by Lazaro Alonso (`@lazarusa`).
+Code to set the theme to `theme_jlmke` is as follows: `Makie.set_theme!(theme_jlmke())`.
```
lazarusA, line 12:
```suggestion
Code to set the theme to `theme_ pretty` is as follows: `Makie.set_theme!(theme_ pretty())`.
```

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,78 @@
...
+
+The first 3 colours in the palette definition are close to the ones used in the logo of Makie. The other two are close to those of the Julia logo.
+"""
+function theme_jlmke()
```
lazarusA, line 16:
```suggestion
function theme_ pretty()
```

src/themes/theme_jlmke.jl
```diff
@@ -0,0 +1,78 @@
...
+    my_colors = jlmke_colors(0.8)
```
lazarusA, line 17:
```suggestion
    my_colors = pretty_colors(0.8)
```

docs/documentation/theming/predefined_themes.md
```diff
@@ -118,3 +118,13 @@ with_theme(demofigure, theme_light())
 with_theme(demofigure, theme_dark())
 ```
 \end{examplefigure}
+
+## theme_jlmke
```
lazarusA, line 122:
```suggestion
## theme_ pretty
```

PR #1750, opened 2022-03-15
use simpler to infer functions

SimonDanisch:
We wanted to stop using the fragile `split_intersections` anyways... Not sure why we haven't done so yet.
@jkrumbiegel @piever, do you remember if we actually found some edge case where the earcut triangulation doesn't work or is super slow?
Maybe violin plots?

### DIFF
--- a/src/basic_recipes/poly.jl
+++ b/src/basic_recipes/poly.jl
@@ -60,7 +60,7 @@ function plot!(plot::Poly{<: Tuple{Union{GeometryBasics.Mesh, GeometryPrimitive}
     )
     wireframe!(
         plot, plot[1],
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Violin plots are almost 10x faster with this change, so that wasn't it ^^

jkrumbiegel:
I only remember that you once said you vaguely remembered some issue with earcut, which is why you hadn't used it. But I've not encountered any issues myself.

piever:
> Violin plots are almost 10x faster with this change, so that wasn't it ^^

That's great to hear! I also don't remember issues specific to earcut.

PR #1754, opened 2022-03-15
Fix translate!/rotate!/scale! for scatter and text

### LINKED ISSUE #1753: scatter and text transformations act on the wrong space
This is probably caused by #1596.

For example
```julia
fig, ax, p = scatter(rand(10))
t = text!(ax, "Test")
```
both translate in pixel units rather than data units. More generally they probably translate in whatever units `markerspace` sets.

ffreyer:
# Description

Fixes #1753

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Fixed in GLMakie
- [x] Fix WGLMakie
- [x] Fix CairoMakie
- [x] Added refimg test

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -347,7 +347,7 @@ end
 _deref(x) = x
 _deref(x::Ref) = x[]
 
-function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation, model, space, markerspace, offsets)
+function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation, _model, space, markerspace, offsets)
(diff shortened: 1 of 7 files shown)


ffreyer:
I'm surprised we don't have any reference images for `translate!/rotate!/scale!` acting on scatter or text plots. Or maybe I'm just not seeing them?

The refimg for the test I added is:
Screenshot from 2022-03-16 18-09-58
That should be working in GLMakie, WGLMakie and CairoMakie with this pr.

SimonDanisch (approved):
Looks good, thank you :) The failing CI aren't required to pass...

PR #1758, opened 2022-03-16
CairoMakie: less specialization, better inference

timholy:
On my machine, this drops the time for
`save("/tmp/test.png", scatter(1:4))`
from 25.6s to 24.4s.  The time for CairoMakie's whole test suite drops
from 4m50s to 4m34s.


SimonDanisch:
Thank you! I fixed one problem in https://github.com/JuliaPlots/Makie.jl/pull/1769 and merged it!

PR #1764, opened 2022-03-16
probably a better README file with a quick intro

lazarusA:
# Description

A quick intro to start working as soon as possible. Namely, test and see the capabilities of Makie without going into the rabbit hole, i.e. the documentation 😄.

The poor quality of the images is due to saving png's files, nevertheless is what the end user will get, hence is fine I suppose.


README.md
```diff
@@ -39,16 +39,98 @@ BibTeX entry:
...
+
+```julia
+]st GLMakie
+```
+
+Test the package:
+
+```julia
+using GLMakie
+] test GLMakie
```
SimonDanisch, line 64: I think we should just remove that.. I personally wouldn't recommend running the tests on the local machine, since platform problems will immediately error without tests and if there are no platform specific problems, all versions should come tested by the CI.

lazarusA: agree, I never test the packages 😄, straight to using them.  Done!

SimonDanisch:
merged here: https://github.com/JuliaPlots/Makie.jl/pull/1767

PR #1771, opened 2022-03-20
Multiple Windows

### LINKED ISSUE #188: New plots should be opening new windows.
MWE:

```
plot(rand(10), rand(10))
plot!(rand(10), rand(10), color = :red)
```
```
plot(rand(10), rand(10))
plot!(rand(10), rand(10), color = :red)
```

If one runs these 4 lines of code, a single makie window will be opened. The result of the first 2 lines of code are totally discarded, erased. Only the final two lines take effect.

This is a bug. Instead each call to a plotting function that does not end with `!` must open a new window instead of discarding all previous ones.

ffreyer:
# Description

This is another attempt at getting multiple windows working in GLMakie. 

closes #1287 
closes #188

## Type of change

## TODO / Checklist

- [x] multiple window open and render
- [x] multiple windows process events independently
- Fix rendering issues:
    - [x] scatter markers disappear when opening a window for `ax.scene` and translating them off screen to the left
    - [x] resizing affects multiple screens 
- [x] Add an entry in NEWS.md (for new features and breaking changes)
- ~~Add or change relevant sections in the documentation~~ There isn't really anything to document. Trigger `display` and you get a new window...
- [x] Add unit tests for new algorithms, conversion methods, etc.
- [x] Stop new windows from popping up on `save`, `record` and when `fig` is called at the end of a jupyter notebook cell
- [x] Cleanup the event cleanup
- [x] Restrict to one window per scene tree

## Future TODO

To keep this pr from taking forever I'm leaving a bunch of nonessential TODOs for the future:

- processing events from different windows acting on the same scene tree (this requires a partial rework of how Events)
- Moving globals like window settings, enable_SSAO etc to scene attributes
- reducing the number of context swaps, specifically by pooling updates of gpu buffers rather than executing them immediately after Observable updates. This should also help with general rendering performance and is a step towards on demand rendering.
- `record` opens a new window - should it?
- When both a `fig` and `ax.scene` have an open window, adding a plot to `fig.scene` will create a renderobject in both screens. I assume this is because every scene has a full copy of `current_screns`
- When displaying `ax.scene` the (initial) window size matches the axis size, but the drawing area is offset by `minimum(scene.px_area[])`


ffreyer:
Getting something like
```julia
fig, ax, p = scatter(10)
ax.scene
```
to work with overlapping interactions seems like big rabbit hole to me. Mouse positions from `axis.scene` will be based on a different window size and therefore incompatible with those from `fig.scene`. So either we need to adjust the position in `axis.scene` to be based on the size of `fig.scene`, or we need to adjust all child scenes to have position based on their size. In the latter case we also need to check if that breaks any of the current interaction and we will likely break some assumption people have made in using events.

So for now I stopped connecting events to scenes that already have a window attached. (This seems to be shared from the top down.)

Another issue I've noticed with the setup above is that `ax.scene` stops displaying some markers when the plot moves outside the axis area to the left. (The lack of axis lines etc is expected.) The axis outline also fails to resize correctly sometimes, leaving gaps.
Screenshot from 2022-03-20 18-20-01
Screenshot from 2022-03-20 18-20-03

With
```julia
fig, ax, p = scatter(rand(10))
fig2, ax, p = scatter(rand(10))
```
there is still some event duplication going on with resizing. This seems to cause gaps like in the example above and other graphical errors (half updated ticks, grids, etc).

GLMakie/src/screen.jl
```diff
@@ -387,34 +391,14 @@ function Screen(;
...
-        # resize failed. This may happen when screen was previously
-        # enlarged to fill screen. WE NEED TO DESTROY!! (I think)
-        destroy!(screen)
-        # try again
-        return global_gl_screen(resolution, visibility, 2)
-    end
-    # show loading image on fresh screen
-    isold || display_loading_image(screen)
-    screen
-end
```
ffreyer, line 412: Is there anything important here that should be added to `Screen()`?

GLMakie/src/GLAbstraction/AbstractGPUArray.jl
```diff
@@ -61,12 +61,13 @@ function setindex!(A::GPUArray{T, N}, value::Array{T, N}, ranges::UnitRange...)
     nothing
 end
 
+ShaderAbstractions.switch_context!(A::GPUArray) = switch_context!(A.context)
 function update!(A::GPUArray{T, N}, value::AbstractArray{T2, N}) where {T, N, T2}
     update!(A, convert(Array{T, N}, value))
 end
 function update!(A::GPUArray{T, N}, value::AbstractArray{T, N}) where {T, N}
     @sync begin
```
SimonDanisch, line 69: Does this actually do anything?

>help?> @sync
  Wait until all lexically-enclosed uses of @async, @spawn, @spawnat and @distributed are complete.

I think that should only wait for anything, if you have an `@async` in the same scope

SimonDanisch: Yeah, I guess we need to make sure that this doesn't happen..
I'm actually not sure if can assure this besides erroring when we switch the context, and then making sure that all our code is written in a way that it doesn't switch...
Ideally, we'd also use multiple threads and then have one context per thread...But that'd be a more advanced refactor 😅

ffreyer: I removed all the `@sync`s and tested with 4 windows (2 figures each displaying fig and ax.scene) and it seemed to work fine...

SimonDanisch: I already tried to make sure there are no yields in the rendering code, since that also would crash without multi windows...so maybe that's enough

ffreyer:
Just having `fig` at the end of a cell in a jupyter notebook opens a window and generates an image with this. I'm not sure if that's a bad thing but it's different from master...

GLMakie/src/screen.jl
fatteneder: The if block in this function caused GLMakie tests to fail when you have a space-filling window manager (e.g. `i3` or `bspwm`) installed, cf. https://github.com/JuliaPlots/Makie.jl/issues/1443

Will this be avoided with your rework?

SimonDanisch: Hm, I remember coming up with this horrible code when trying to fix some horrible windows / linux bugs ^^ 
I'd love to get rid of this, but we should make sure it works on all platforms

SimonDanisch:
Damn, I can't reproduce the failure locally... 
Will need to come back after the current release.

ffreyer:
Maybe singleton_screen or scene2image needs a context switch?

ffreyer:
I can't reproduce it either :man_shrugging:

jkrumbiegel:
I think this one needs some docs, for example in the GLMakie backend section maybe, or in the docstring of `display`.

ffreyer:
I wonder if we should even have this `display(GLMakie.Screen(), fig_or_scene)` complication to display another screen. The changes made here basically all take effect whether or not multiple screens are used. The only thing you remove are context switches, which you can also do by just not having multiple screens open.
Personally I think we should just stop one scene from opening multiple windows but allow multiple windows otherwise. (I.e. every unique Scene tree can open a new window without any extra code)

src/interaction/events.jl
```diff
@@ -12,19 +11,65 @@ dropped_files(scene, native_window) = not_implemented_for(native_window)
...
+
+    disconnect!(screen, dropped_files)
+    # disconnect!(e.dropped_files)
+
+    disconnect!(screen, hasfocus)
+    # disconnect!(e.hasfocus)
+
+    disconnect!(screen, entered_window)
+    # disconnect!(e.entered_window)
 end
```
ffreyer, line 73: @SimonDanisch What were the `disconnect!(event_obs)` methods supposed to do?

ffreyer:
After some event cleanup fixes

```julia
fig, ax, p = scatter(rand(10));
s1 = display(fig)
s2 = display(GLMakie.Screen(), fig)
```

now results in two windows, `s1` without interactivity and `s2` with interactivity. If `s2` is closed, `s1` remain unresponsive. Also, neither window/screen is present in `scene.current_screens`. (This is generally the case when using `display(GLMakie.Screen(), ...)`)
We should decide on how this should behave and make some more adjustments before merging:
- Should you be able to open multiple windows for one Scene tree?
- If so, should every window be in `current_screens`?
- If so, should an old window regain control when another is closed? Alternatively, should interactivity be based on which window is hovered/focused?

ffreyer:
> Should you be able to open multiple windows for one Scene tree?

We'll go with no for now.

I found another issue with the display system in here: Repeatedly calling `display(fig)` adds more and more copies of the same screen to current_screens. I tried clearing them in `connect_screen`, but that triggers a segfault in `GLFW.SetWindowCloseCallback(window, nothing)` on the second `display(fig)`. I tried adding a context switch right before it but that doesn't seem to help. Calling `display(fig)` multiple times also breaks interactivity.

jkrumbiegel:
Do I understand correctly from your previous comment that you can only have interactivity in the scene that was opened last? That seems like a big caveat, unresponsive windows aren't of much use

ffreyer:
> Do I understand correctly from your previous comment that you can only have interactivity in the scene that was opened last? That seems like a big caveat, unresponsive windows aren't of much use

I was talking about the when you open the same figure multiple times. If you open independent figures you get interactions for each

jkrumbiegel:
Aha then I'd be all for closing the old window of a Scene. That seems to me to only have potential for confusion. Or rather, displaying a Scene that is already open should just focus that window, no real need to open it again right?

ffreyer:
I restricted things to one window per scene tree now and added some tests for window/screen bookkeeping.

> I found another issue with the display system in here: Repeatedly calling display(fig) adds more and more copies of the same screen to current_screens. I tried clearing them in connect_screen, but that triggers a segfault in GLFW.SetWindowCloseCallback(window, nothing) on the second display(fig). I tried adding a context switch right before it but that doesn't seem to help. Calling display(fig) multiple times also breaks interactivity.

The segfault happened when a closed window was used. I wrote the same general code a couple of times, now it works. I must've closed the wrong screen or gotten something about the order wrong with earlier versions because it works fine now.

>  Cleanup the event cleanup

While I still allowed multiple windows per scene tree I got duplicate mouse and window area updates because `screen.render_tick.listeners` didn't get cleared. I fixed that here

src/interaction/events.jl
SimonDanisch: Hm, good question. I think I planned to remove any listener, but `connect!` shouldn't really register listeners when I think about it...

ffreyer:
The changes boil down to:
- adjust window management to allow more than one window
- make shaders per screen
- add lots of context switched
- add `display(GLMakie.Screen(), scene)` (?)
- fix a few cleanup issues

I think all of those are either internals or non-breaking/additions (display stuff). I don't think we need to wait for a breaking release for this

SimonDanisch:
That regression looks real:
image

Since it isn't that much, and we haven't had the chance to really optimize display times for GLMakie yet anyways, I'll just go ahead and merge this. We should try to make this faster in a separate PR when we have time for it.

goretkin:
I'm very happy to see that Makie should support multiple windows.

I want to be sure that 
> I restricted things to one window per scene tree now and added some tests for window/screen bookkeeping.

means that it's not possible to have two windows showing different camera views of the same scene, as is common in [3D modeling software](https://docs.blender.org/manual/en/latest/editors/3dview/navigate/views.html#quad-view).

image
(If you wanted each of the quadrants to be different OS windows.)

ffreyer:
Yea that's right. The main issue is that if two windows show the same scene tree you end up with things like `events.mouseposition` getting triggered from two sources with conflicting position. This could be solved by only allowing position updates from a focused window, but that would deny some interactivity if done in the backend. It probably also still causes jumps

PR #1777, opened 2022-03-21
ylims!() as vector

daniel-frisch-kit:
# Description
`ylims!(ax, [0,1])` raised an error; only `xlims!()` had the correct treatment for vectors. Only `ylims!(ax, (0,1))` and `ylims!(ax, 0,1)` was possible. 

Conversion from vectors to tuples is now added for `ylims!()` and `zlims!()` as well. Now it works as documented. 

## Type of change

## Checklist


SimonDanisch:
Actually, I'm on a quest to make sure we use fewer types + signatures for all functions, so that there is a clear preferred way, and we don't compile as much...
What would you think about restricting `ylims!` and friends to `ylims!(ax, min, max)` ?

cc: @jkrumbiegel

jkrumbiegel:
Yeah I'm also against an additional vector arg, probably the simple two arg function is the only one we really need.

daniel-frisch-kit:
For me it was quite convenient to make prior calculations like `lim=[-1,1]*size`. But then I could do `ylims!(ax, lims...)` which is also perfect. And I would LOVE reduced compile time. But please change the doc then; there it is [sold](https://makie.juliaplots.org/v0.15.2/examples/layoutables/axis/) as a feature to freely choose between two-arg, tuple, and vector.

jkrumbiegel:
Ah I probably didn't read correctly, ok I see now that not adding `ylims!` was just an oversight. These small methods will also not add anything to compile time by themselves, this is more about following a general design pattern not to add too many convenience overloads for everything. So I'd still think removing the vector option for xlims would be reasonable. Tuples or Rects are debatable I think.

SimonDanisch:
Ah I also didn't get, that this mainly fixes an oversight for ylims.
Ok, let's merge this, to make it consistent and without breaking existing behavior.
If we want to clean up accepted  arguments, we should do it in a separate PR

PR #1790, opened 2022-03-30
Fix glyphlayout not being an Observable for LaTeXStrings

### LINKED ISSUE #1417: Latex text observables not work
As per the docs, changing  "a text object which has been instantiated with a normal String input to use a LaTeXString" can't be done.
But actually, changing a text object which has been instantiated with a LaTeXString input to use another LaTeXString can't be done, either.
For example,
```julia
using GLMakie
fig = Figure()
ax1 = Axis(fig[1, 1], title = L"\frac{x}{y}")
ax2 = Axis(fig[1, 2], title = "x/y")
ax1.title = L"x^2" #does not change the title
ax2.title = "x^2" #does change
fig
```

ffreyer:
# Description

I noticed that using latex strings as axis labels in `Axis3` results in wrong, static rotations so I did some digging. It turns out that glyphcollection decides it no longer is an Observable after `notify(plot.position)` (it is before) but instead is the glyphcollection in the first `lift`. Removing the name collision fixes this.

Fixes #1417 as well.

## Type of change

### DIFF
--- a/src/basic_recipes/text.jl
+++ b/src/basic_recipes/text.jl
@@ -112,7 +112,7 @@ function plot!(plot::Text{<:Tuple{<:Union{LaTeXString, AbstractVector{<:LaTeXStr
             end
             tex_elements, glyphcollections, offsets
         else
-            tex_elements, glyphcollection, offset = texelems_and_glyph_collection(latexstring, ts,
+            tex_elements, glyphcollections, offsets = texelems_and_glyph_collection(latexstring, ts,
                 al[1], al[2], rot, color, scolor, swidth)
         end
     end


SimonDanisch (approved):
Seems about right although now it conflicts with the `glyphcollections` variable.
But I think that's how it should be?

jkrumbiegel:
Hm yeah I think we don't need the output tuple variable names at all, I just put them there to show the relationship to the output above it. The difference is that the lower function returns just one glyphlayout, so the missing `s` was intentional. Below, the code checks whether we're dealing with one or multiple glyphlayouts. It's not the cleanest logic to be sure

Issue #1566, opened 2022-01-09
I want to contribute (open for suggestions)


kimkroener:
Hi everyone, 

I'm a Masters's student at the University of Stuttgart and participating in the course [Simulation Software Engineering](https://simulation-software-engineering.github.io/homepage/). As part of this course we are asked to contribute to an open source project to get familiar with the development workflow of such projects.  For this challenge I chose Makie and I'm looking for suggestions :) 

The contribution should be small but meaningful and there are no limitations for this contribution. It could be a new tutorial, bug fix, documentation, or any other maintenance task. However it should be something that is a good first issue as I'm a beginner Julia (and Makie) user (I do have programming experience in Matlab and Python) and I'm looking forward to learn by contributing.

ffreyer:
One suggestion I have is working on patterns/hatching. A probably very simple thing you could fix is that the [show method](https://github.com/JuliaPlots/Makie.jl/blob/c453051af7e1cccdf63b8465f4d805b0cdebabab/src/patterns.jl#L11) for `Makie.ImagePattern` errors.
Currently patterns are only supported by `mesh` plots in GLMakie and are therefore still an unofficial feature. I think the most important step towards officially supporting patterns is a CairoMakie implementation. This means figuring out how Cairo handles hatching, adjusting `LinePattern` to be compatible with it and adding it to the [CairoMakie mesh method](https://github.com/JuliaPlots/Makie.jl/blob/c453051af7e1cccdf63b8465f4d805b0cdebabab/CairoMakie/src/primitives.jl#L644).  (And if `LinePattern` is a terrible representation it should be fine to replace it too. There just needs to be a `to_image` method so that GLMakie know what image texture to draw.)

The original prs implementing patterns are https://github.com/JuliaPlots/AbstractPlotting.jl/pull/457 and https://github.com/JuliaPlots/GLMakie.jl/pull/108. There is also an issue requesting this: https://github.com/JuliaPlots/Makie.jl/issues/1385

kimkroener:
Thank you for your suggestions and sorry for my late reply. Pattern support sounds super interesting. 
I started looking into it and I realized I don't feel comfortable enough with my Julia (and Makie) skills yet to work on a feature implementation. 

[...]

PR #1794, opened 2022-03-31
fix show AbstractPattern

harisorgn:
# Description

Fixed the error in the show function for AbstractPatterns 

Based on @ffreyer's suggestion on issue #1566, I fixed the show function for AbstractPatterns. I tested this with both a LinePattern and an ImagePattern. The show function converts the pattern to an image and prints it as a Scene object.

### DIFF
--- a/src/patterns.jl
+++ b/src/patterns.jl
@@ -9,9 +9,14 @@ function, which must return a `Matrix{<: AbstractRGB}`.
 abstract type AbstractPattern{T} <: AbstractArray{T, 2} end
 
 function Base.show(io::IO, ::MIME"text/plain", p::AbstractPattern)
+    
     println(io, typeof(p))
-    show(io, mime, image(p))
+    print(io, p)
 end
+
+Base.show(io::IO, p::AbstractPattern) = print(io, image(to_image(p)))
+
+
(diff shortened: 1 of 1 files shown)


src/patterns.jl
```diff
@@ -9,9 +9,14 @@ function, which must return a `Matrix{<: AbstractRGB}`.
 abstract type AbstractPattern{T} <: AbstractArray{T, 2} end
 
 function Base.show(io::IO, ::MIME"text/plain", p::AbstractPattern)
+    
     println(io, typeof(p))
-    show(io, mime, image(p))
+    print(io, p)
 end
+
+Base.show(io::IO, p::AbstractPattern) = print(io, image(to_image(p)))
```
ffreyer, line 17: This should not have `image`, just `to_image`. The output should either be an actual image like Colors.jl has or a matrix style print. Right now it creates a full plot and then prints the Scene structure.

Also the printing for `ImagePattern` still fails.

SimonDanisch:
I'm going to close this, since it changes behavior without actually fixing all problems.

harisorgn:
> I'm going to close this, since it changes behavior without actually fixing all problems.

Sorry about the delay! Misunderstood the original task, will give it another go when time allows and open a new PR.

PR #1800, opened 2022-04-04
violin horizontal

SimonDanisch:
Just a rename for #1676, so we can merge it.


SimonDanisch:
Cloeses https://github.com/JuliaPlots/Makie.jl/issues/1312

PR #1803, opened 2022-04-05
GLBuffer-related bug fixes

maleadt:
A small set of fixes for issues I ran into when using GLBuffer inputs directly.
Note that I first fixed `iterate(::GLBuffer)` before replacing the implementation (so that we can always go back to the old implementation, but in a working state), which means it may be better to merge this instead of squashing the commits.


SimonDanisch:
Thanks :) I think squash merge is fine (the others are disallowed and I don't think it's too bad to still squash)...

PR #1810, opened 2022-04-07
[breaking] clean up conversion pipeline

SimonDanisch:
- [x] ~~rely on `NoConversion` return type to recursively apply conversions in a well defined order, and stop using `MethodError` as recursion anchor~~ ... Actually, recursive application was getting to complex and error prone. I left a simpler version, and mostly just got rid of `MethodError` and cleaned up the implementation a bit.
- [ ] ~~clean up conversion application in `interfaces.jl`~~. I'll leave this for another PR
- [x] add `@convert_target`, which specifies what a plot type expect `convert_arguments` to convert to, which allows us to throw better errors.


MakieCore/src/MakieCore.jl
```diff
@@ -11,4 +11,5 @@ include("recipes.jl")
 include("basic_plots.jl")
 include("conversion.jl")
 
+
```
jkrumbiegel, line 14: just white space

MakieCore/src/basic_plots.jl
```diff
@@ -107,6 +107,8 @@ Plots a surface, where `(x, y)`  define a grid whose heights are the entries in
     )
 end
 
+
+
```
jkrumbiegel, line 111: just white space

MakieCore/src/conversion.jl
```diff
@@ -30,3 +30,103 @@ conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
 
 struct VolumeLike <: ConversionTrait end
 conversion_trait(::Type{<: Volume}) = VolumeLike()
+
+
+"""
+    convert_arguments_typed(::Type{<: AbstractPlot}, args...)::NamedTuple
+
+Converts the arguments to their correct type for the Plot type.
+Throws apropriate errors if it can't convert!
```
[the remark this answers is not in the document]
jkrumbiegel: and I think doc strings don't need exclamation marks! :)

MakieCore/src/conversion.jl
```diff
@@ -30,3 +30,103 @@ conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
...
+"""
+function convert_arguments_typed end
+
+
+function convert_arguments_typed(P::Type{<: AbstractPlot}, @nospecialize(args...))
+    convert_arguments_typed(conversion_trait(P), args...)
+end
+
+function convert_arguments_typed(::CT, @nospecialize(args...)) where {CT <: ConversionTrait}
```
jkrumbiegel, line 48: why this redirect? the functions should either take CT or CT(), not both

MakieCore/src/conversion.jl
```diff
@@ -30,3 +30,103 @@ conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
...
+    return convert_arguments_typed(CT, args...)
+end
+
+function convert_arguments_typed(ct::Type{<: ConversionTrait}, @nospecialize(args...))
+    # We currently just fall back to not doing anything if there isn't a convert_arguments_typed defined for certain plot types.
+    # This makes `@convert_target` an optional feature right now.
```
jkrumbiegel, line 54: Explain here what the consequence is of args just being returned as is. Where will that lead to an error?

MakieCore/src/conversion.jl
```diff
@@ -30,3 +30,103 @@ conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
...
+
+"""
+    @convert_target(expr)
+
+Allows to define a conversion target for a plot type, so that convert_arguments can be checked properly, if it converts to the correct types.
+
+Usage:
+```Julia
+@convert_target struct PointBased{N} # Can be the Plot type or a conversion Trait
+    positions::AbstractVector{Point{N, Float32}}
```
jkrumbiegel, line 69: it's not clear what this does from the docstring. can this have multiple fields? are the fields just args of some function or are we making a struct here? or is the struct just a way to list the fields. that would be kind of weird

src/figureplotting.jl
```diff
@@ -18,17 +18,11 @@ function is_plot_3d(p::PlotFunc, args...)
...
-            conv = non_obs
-        else
-            rethrow(e)
-        end
-    end
-
-    Typ, args_conv = apply_convert!(p, Attributes(), conv)
+    RealP = plottype(p, non_obs...)
+    result = is_plot_type_3d(RealP)
+    isnothing(result) || return result
```
jkrumbiegel, line 23: I think 
```julia
if result !== nothing
    return result
end
```
is clearer, this kind of confuses me

src/utilities/utilities.jl
```diff
@@ -299,3 +299,26 @@ function attribute_names(PlotType)
...
 end
+
+
+"""
+    regularly_spaced_array_to_range(arr)
+If possible, converts `arr` to a range.
+If not, returns array unchanged.
+"""
+function regularly_spaced_array_to_range(arr)
+    diffs = unique!(sort!(diff(arr)))
```
jkrumbiegel, line 310: this seems  a bit weird, but maybe it's floating point error avoiding magic ;) I think something like this would be simpler:

```julia
distance = last(arr) - first(arr)
step = distance / (length(arr) - 1)
# iterate through pairs and compare each diff to step, if they all match return range from first to last element
```

I don't know why extrema are being taken and vectors full of diffs being allocated, shouldn't any valid array be already sorted? so we don't need to do the weird min, max switch either

jkrumbiegel (changes requested):

MakieCore/src/conversion.jl
SimonDanisch: The conversion pipeline has been using `::Type{<:AbstractPlot}` and `::ConversionTrait` for dispatch, but in the macro I can't differentiate between those, so I overload `::Type{<:T}` for both.
I guess I could also use `typeof(conversion_trait(P))`, instead of redirecting.

MakieCore/src/conversion.jl
SimonDanisch: Well, the consequence is, that it's optional, because you don't need to define it for every plot type this way.
The alternative would be erroring here, and demanding to add an `@convert_target` for the plot type.

src/figureplotting.jl
```diff
@@ -18,17 +18,11 @@ function is_plot_3d(p::PlotFunc, args...)
...
+    conv = convert_arguments(RealP, non_obs...)
+    Typ, args_conv = apply_convert!(RealP, Attributes(), conv)
```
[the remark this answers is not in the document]
SimonDanisch: The other overloads are mutating attributes...

src/utilities/utilities.jl
SimonDanisch: This function is just getting moved in this PR and wasn't written by me. If you think it should be written differently, we should make a new PR ;)

src/figureplotting.jl
SimonDanisch: I feel like the boat has sailed for this one :D 
I've actually started to use `isnothing` consistently and removed all `=== nothing` instead.
`|| return x`  is debatable, i got confused by it as well in the beginning but if you use it more it's kind of nice.
If we remove it here, we should remove it in the whole code base...Maybe at some point we should introduce some style guidelines?

MakieCore/src/conversion.jl
```diff
@@ -30,3 +30,103 @@ conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
...
+            # This seems to work well, since Julia drops any unecessary where clause/ type parameter, while lowering.
+            # TODO figure out a way to drop the where close... Eval is the only thing I found, but shouldn't be used here.
+
+            expr = quote
+                # Unions etc don't work well with `convert(T, x)`, so we try not to convert if already the right type!
+                if $name isa $TargetType
+                    $conv_name = $name
+                else
+                    $conv_name = try
+                        convert($TargetType, $name)
```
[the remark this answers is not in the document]
SimonDanisch: I want to come back here later and use e.g. `convert_attribute` here, or some of the newer versions we're using in `Block` or the new `Attribute` type.
For now, I decided to stick with `convert` for two reasons:
* the `convert` here is kind of optional, since `convert_arguments` should already convert to the right types
* it's mainly a softer version of just asserting the type, and therefore `convert` does the job well in throwing the appropriate errors even for union types

PR #1816, opened 2022-04-12
Allow weights in statistical recipes

greimel:
Add the keyword `weights` to `boxplot`, `density`, `hist`, `violin`.

## `density`

image

```julia
let
	N = 100_000
	x = rand(Uniform(-2, 2), N)
	
	w = pdf.(Normal(), x)

	fig = Figure()
	density(fig[1,1], x)
	density(fig[1,2], x, weights = w)

	fig
end
```

## `violin`
image

```julia
let
	N = 100_000
	x = rand(1:3, N)
	y = rand(Uniform(-1, 5), N)
	
	w = pdf.(Normal(), x .- y)

	fig = Figure()

	violin(fig[1,1], x, y)
	violin(fig[1,2], x, y, weights = w)

	fig
end
```

## `boxplot`
image

```julia
let
	N = 100_000
	x = rand(1:3, N)
	y = rand(Uniform(-1, 5), N)
	
	w = pdf.(Normal(), x .- y)

	fig = Figure()

	boxplot(fig[1,1], x, y)
	boxplot(fig[1,2], x, y, weights = w)

	fig
end
```

## `hist`
image
```julia
let
	N = 100_000
	x = rand(Uniform(-5, 5), N)
	
	w = pdf.(Normal(), x)

	fig = Figure()
	hist(fig[1,1], x)
	hist(fig[1,2], x, weights = w)

	fig
end
```


greimel:
Friendly bump 😃 

The added functionality generalizes what `AlgebraOfGraphics.density()` and `AlgebraOfGraphics.histogram()` can already do. (e.g. `data(df) * mapping(:y, weights = :wgts) * density() |> draw`)

@piever, since you have written most of the code that this PR touches – would you mind reviewing it if you find some time?

src/stats/violin.jl
```diff
@@ -5,6 +5,7 @@ Draw a violin plot.
 - `x`: positions of the categories
 - `y`: variables whose density is computed
 # Keywords
+- `weights`: vector of statistical weights (length of data). By default, each observation has weight 1.
```
piever, line 8:
```suggestion
- `weights`: vector of statistical weights (length of data). By default, each observation has weight `1`.
```

src/stats/violin.jl
```diff
@@ -71,6 +73,7 @@ function plot!(plot::Violin)
                 npoints = n,
                 (bound === automatic ? NamedTuple() : (boundary = bound,))...,
                 (bw === automatic ? NamedTuple() : (bandwidth = bw,))...,
+                (w === automatic ? NamedTuple() : (weights = StatsBase.weights(view(w, idxs)),))...
```
piever, line 76: We should probably add a utility function to filter out `automatic` keywords, but that should be done separately.

src/stats/boxplot.jl
```diff
@@ -17,6 +17,7 @@ The boxplot has 3 components:
 - `x`: positions of the categories
 - `y`: variables within the boxes
 # Keywords
+- `weights`: vector of statistical weights (length of data). By default, each observation has weight 1.
```
piever, line 20:
```suggestion
- `weights`: vector of statistical weights (length of data). By default, each observation has weight `1`.
```

piever (changes requested):

piever:
Nice! Looks good to me (just left two tiny formatting comments).

I have a minor doubt as to whether we should call `StatsBase.weights` indiscriminately, or instead not call it when the user passes a `StatsBase.AbstractWeight` object. I have no idea whether the various subtypes of `AbstractWeight` make any difference here. I suspect not, so probably it's fine to leave as is.

greimel:
I've applied your suggestions, thanks for the review!

piever (approved):

SimonDanisch:
Great!
There is still an error in the docs:
```
┌ Franklin Warning: in <examples/plotting_functions/violin.md>
│ There was an error of type 'ArgumentError' when running a code block.
│ Checking the output files '/home/runner/work/Makie.jl/Makie.jl/docs/__site/assets/examples/plotting_functions/violin/code/output/example_figure.(out|res)'
│ might be helpful to understand and solve the issue.
│ 
│ Relevant pointers:
│ - evaluation of Julia code blocks: https://franklinjl.org/code/
│ 
│ Details:
│ ArgumentError: Package Distributions not found in current path:
│ - Run `import Pkg; Pkg.add("Distributions")` to install the Distributions package.
```
Can you add that to the docs project?

greimel:
Done! I hadn't realized that the docs are actually built for non-member contributors these days. That's great news!

greimel:
I would like to use this for a lecture in about a week.

If it's not too much of a hassle for you, I'd appreciate this being merged and tagged before May 8.

If you don't plan a new tag soon, please let me know and I'll adjust my material accordingly

SimonDanisch:
I'll try to release a new version!

SimonDanisch:
Would a breaking version work for you as well?

greimel:
that's fine, thanks for asking!

Issue #1595, opened 2022-01-14
How to clear a figure?


jarbus:
I'm making a UI for data exploration, and want to have a button that I can click to switch between different views of the data, kind of like tabs. For example, in one view, I have one big axis that takes up most of the figure, and in another view, I have two smaller axes that take up half of the figure each. In each view, I don't want to see the axes from the other view. 

Crucially, I don't want to have to switch between two figures, as that will launch another window. I need to use the same window, with the ability to change a figure's entire layout table on the fly.

Is there a way to clear, and then reconstruct, a Figure's layout table?

Mo8it:
One can do a lot with Observables. Take a look at this page with the video:
https://makie.juliaplots.org/stable/documentation/nodes/index.html

But I would also appreciate an easy method to clear a figure. I did not find any.

jarbus:
I got it, I need to delete entries in `Figure.content`

fatihyaman661:
> I got it, I need to delete entries in `Figure.content`

And how do you delete the entries exactly😅

PR #1818, opened 2022-04-14
Add `empty!(fig)` and cleanup `empty!(scene)`

ffreyer:
# Description

This fixes an error in `empty!(scene)`, makes it delete all plots in all child scenes and stops it from disconnecting the screen. 

It also adds `empty!(fig)` clearing all attached scenes and the layout. (Following https://github.com/jkrumbiegel/GridLayoutBase.jl/pull/2 here, though `delete!` seems unnecessary to me?)

Related: #1595

## Type of change

- New feature & bug fix

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 - **Breaking** Added `space` as a generic attribute to switch between data, pixel, relative and clip space for positions. `space` in text has been renamed to `markerspace` because of this. `Pixel` and `SceneSpace` are no longer valid inputs for `space` or `markerspace`.
 - **Breaking** Deprecated `mouse_selection(scene)` for `pick(scene)`.
 - **Breaking** Bumped `GridLayoutBase` version to `v0.7`, which introduced offset layouts. Now, indexing into row 0 doesn't create a new row 1, but a new row 0, so that all previous content positions stay the same. This makes building complex layouts order-independent [#1704](https://github.com/JuliaPlots/Makie.jl/pull/1704).
+- Added `empty!(fig)` and changed `empty!(scene)` to remove all child plots without detaching windows
 
 ##  v0.16.4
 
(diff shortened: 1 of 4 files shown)


ffreyer:
Triggering `empty!` and plotting off of an event rapidly sometimes results in plots being left behind

GLMakie/src/screen.jl
```diff
@@ -58,6 +58,47 @@ function Makie.insertplots!(screen::GLScreen, scene::Scene)
     foreach(s-> insertplots!(screen, s), scene.children)
 end
 
+function Base.delete!(screen::Screen, scene::Scene)
+    if !isempty(scene.children)
+        delete!.((screen,), scene.children)
```
SimonDanisch, line 63: I think I'd prefer a for loop here for compilation, and since broadcasting creates a result vector.

SimonDanisch (approved):

src/types.jl
```diff
@@ -172,22 +172,22 @@ end
 # Compat only
 function Base.getproperty(e::Events, field::Symbol)
     if field === :mousebuttons
-        error(
-            "`events.mousebuttons` is deprecated. Use `events.mousebutton` to " *
-            "react to `MouseButtonEvent`s instead and ``."
-        )
+        error("""
```
ffreyer, line 175: I think this way of writing the error strings needs backslashes to avoid linebreaks, which is a rather new feature iirc.

SimonDanisch: Ah... But do we want to avoid them? I have the rule of thumb, that if it's nice in the code to have line breaks, it's probably nice in the print out as well.

ffreyer: I think the line breaks here were just to stay in an 80 characters/line limit or something similar. 

Don't log messages also handle line breaks themselves?

SimonDanisch: I made the linebreak a bit nicer.. I think it's good to break the line like that for the error message

SimonDanisch: nevermind :D I looked at the actual error some more, and it does look better in one line.
If someone doesn't like such long lines in the code, my general philosophy has been to enable word wrap in the editor and just keep semi long lines in code.

Issue #757, opened 2020-11-18
Prevent overlapping text


jtrakk:
Sometimes text overlaps other objects, which can be ugly or hard to read. Some ways to avoid this include:

- raise an error
- jitter them so they are offset from one another
- wrap the text
- shrink the font (but not too small)
- grow the container

Examples:

### Overlapping title

```julia
outer_padding = 30
scene, layout = layoutscene(
    outer_padding, 
    resolution = (1200, 700),
    backgroundcolor = RGBf0(0.98, 0.98, 0.98),
)

ax1 = layout[1, 1] = LAxis(scene, title = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt")
ax2 = layout[1, 2] = LAxis(scene, title = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt")
scene
```
![image](https://user-images.githubusercontent.com/43392409/99481185-6ffc0380-290e-11eb-94b1-99a8d726f36b.png)

### Overlapping ticks

```julia
outer_padding = 30
scene, layout = layoutscene(
    outer_padding, 
    resolution = (1200, 700),
    backgroundcolor = RGBf0(0.98, 0.98, 0.98),
)

for i in 1:6
layout[1, i] = LAxis(scene)
end
scene
```
![image](https://user-images.githubusercontent.com/43392409/99481357-c8cb9c00-290e-11eb-81bf-578681db8704.png)

### Overlapping labels

![image](https://user-images.githubusercontent.com/43392409/99481395-e436a700-290e-11eb-8916-f3e4d5c6f693.png)

jkrumbiegel:
This is a real constraint of the layout system, but for good reason. For example, removing a tick label because it overlaps a neighboring one could decrease the width needed for the tick labels, which could expand the axis, which could then give more space for tick labels, which could re-add a tick label, which could reduce the width available, which would shrink the axis, which would overlap the tick labels again.

Basically there is only a limited amount of things that can reasonably be included in the layout calculations, and sideways expansion of ticks is not really one of them. It's fortunately simple enough to change the number of ticks or introduce a line break in an axis title to tweak these things to taste. It is very very hard if not impossible to write good rules to govern this behavior automatically, without things going crazy in an interactive setting.

[...]

PR #1819, opened 2022-04-14
Add basic word wrapping

ffreyer:
# Description

This adds `word_wrap_width` as an attribute for text, which sets a limit for the size of a line. If a word (currently anything between two spaces with no newline in between) overflows this limit, a newline is inserted before it to wrap it to the next line. 

Example (which is also the test I added):
Screenshot from 2022-04-14 18-08-51

## Type of change

## Checklist


ffreyer:
I think this could be nice with `Label`. Instead of having it resize to "fit" long text, the text could be wrapped to a region fitting the Label instead. Could also be helpful for Axis titles, x and y labels, etc. Related: #757

ffreyer:
Tests only failed because there is no refimage for new test

ffreyer:
I added `word_wrap::Bool` to `Label` now. Setting it to false (default) gives the old behavior. With `word_wrap = true` the Label uses the suggested width as its width and wraps text to fit (height is derived). The refimage test looks like this:

Screenshot from 2022-04-15 12-15-13

ffreyer:
Moved some stuff around to avoid creating a second GlyphCollection and got this to work for latex strings. (At least strings with mostly text.) New first refimage:
Screenshot from 2022-04-15 16-23-39

fatteneder:
Hi,

I have implemented the same feature here (although in a hacky way) https://github.com/fatteneder/MakieSlides.jl
This is just a proof-of-principle implementation to see whether we can use `Makie.jl` and its layouting functionality as the building blocks for a new presentation/slideshow engine with Markdown syntax (a la `reveal.js` and `Marpit` etc, but without the `html/css` madness :) The initial idea was mentioned on Zulip.
In case you wanna take a look please check the `test` folder for examples, the ones in `examples` are broken right now.

The reference image you added already goes into the same direction with what I want to do. So I very much support this PR for `Makie` native line wrapping support. Please let me know if I can be of any help here.

fatteneder:
That's funny. If you want we can join forces :)

ffreyer:
For the next 2 weeks I want to keep my package (more or less) as is because I have a presentarion planned with it and I don't want to break it. After that I'd be happy to merge them.

ffreyer:
All test errors are because of the missing refimg for the new test, so this is ready to be merged again

PR #1823, opened 2022-04-16
Fix rotation with Irrational and Uniful's degree

blegat:
`rotate!` doesn't work with irrational angle like `π` because it's not an `AbstractFloat` so the user has to use `1.0π` for instance as a workaround.
Moreover, Unitful's degree implement `sin` and `cos` but they are subtypes of `Number` and not `AbstractFloat` so by using `Number` we allow both the use of Unitful's degree and irrational.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -965,8 +965,8 @@ function to_rotation(s::VecTypes{N}) where N
     end
 end
 
-to_rotation(s::Tuple{VecTypes, AbstractFloat}) = qrotation(to_ndim(Vec3f, s[1], 0.0), s[2])
-to_rotation(angle::AbstractFloat) = qrotation(Vec3f(0, 0, 1), Float32(angle))
+to_rotation(s::Tuple{VecTypes, Number}) = qrotation(to_ndim(Vec3f, s[1], 0.0), s[2])
+to_rotation(angle::Number) = qrotation(Vec3f(0, 0, 1), angle)
 to_rotation(r::AbstractVector) = to_rotation.(r)
 to_rotation(r::AbstractVector{<: Quaternionf}) = r
 
--- a/test/quaternions.jl
+++ b/test/quaternions.jl
@@ -1,3 +1,10 @@
+struct Degree{T} <: Number
+    θ::T
+end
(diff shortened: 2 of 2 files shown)


ffreyer (approved):

test/quaternions.jl
```diff
@@ -1,3 +1,10 @@
+struct Degree{T} <: Number
```
asinghvi17, line 1:
```suggestion
struct Degree{T} <: Real
```

src/conversions.jl
```diff
@@ -965,8 +965,8 @@ function to_rotation(s::VecTypes{N}) where N
     end
 end
 
-to_rotation(s::Tuple{VecTypes, AbstractFloat}) = qrotation(to_ndim(Vec3f, s[1], 0.0), s[2])
-to_rotation(angle::AbstractFloat) = qrotation(Vec3f(0, 0, 1), Float32(angle))
+to_rotation(s::Tuple{VecTypes, Number}) = qrotation(to_ndim(Vec3f, s[1], 0.0), s[2])
+to_rotation(angle::Number) = qrotation(Vec3f(0, 0, 1), angle)
```
asinghvi17, line 969:
```suggestion
to_rotation(s::Tuple{VecTypes, Real}) = qrotation(to_ndim(Vec3f, s[1], 0.0), s[2])
to_rotation(angle::Real) = qrotation(Vec3f(0, 0, 1), angle)
```
We don't want people accidentally passing complex values to rotation, so better to head It off here.

asinghvi17 (commented):
This looks good to me!  Quick comment - I think we should use Real instead of Number as the constraint, since we don't want someone to pass a Complex by mistake.

test/quaternions.jl
SimonDanisch: I'm not sure about adding this type to Makie directly... would it be possible to leave it to another package and make things work via ducktyping?

SimonDanisch: Oh, didn't see this is just for testing purposes... Excellent, so I guess it's already like I requested

SimonDanisch:
Thank you!

PR #1827, opened 2022-04-18
initial implementation

jkrumbiegel:


Datseris:
wow this is so good and so much better than my version LOLZ

Datseris:
> The options to style this are potentially infinite, and it's not going to be practical to implement them all with overlapping keywords etc

Completely agree, and seems to me that what you have already coded is already enough!

Datseris:
The example figure you posted in https://github.com/JuliaPlots/Makie.jl/issues/1581 does not have tick labels in the x-axis, even though the x axis are not shared. My argument would be to hide tick labels only if sharing x is requested

SimonDanisch:
Do we want to merge this?

jkrumbiegel:
I'm not sure if this helps or not. On the one hand, it can handle the spanned x and y labels that are a little bit annoying to implement by hand, as well as hiding the correct axis decorations and linking the right x / y limits.

On the other hand, a Block is supposed to offer interactive editing of attributes, but this doesn't really have any of those. The number of rows / columns is essentially defined at creation and can't be changed afterwards. Adding more things to the internal grid layout could mess up internal state, so users wouldn't be encouraged to do that.

So currently I'm thinking that maybe @Datseris's initial proposal to use a simple function wasn't so bad. The disadvantage is that there's not one big object that can be manipulated and placed as one. But I already foresee anyway that it might not be practical to force treating the whole grid as one rectangular object. That would prohibit placing other elements in relation to specific rows or columns. Essentially, it takes away some of the flexibility and malleability of the GridLayout.

I have to think about a good api for a replacement function. I'm going to close this until that point and we can continue the discussion in an issue if necessary. Otherwise I'll link any new PR against this one.

Datseris:
Sounds, good, I have updated the code here to have better axis linkage: https://github.com/JuliaPlots/Makie.jl/issues/1581#issuecomment-1010439260

PR #1828, opened 2022-04-19
Fix CairoMakie heatmap filtering

jkrumbiegel:
# Description

Heatmap filtering was broken again because it used Cairo.FILTER_BEST, which does nothing in bitmap backends. This PR fixes that and reorders the logic a bit, so it should be easier to follow why the surface or rectangle path is chosen.

## Type of change

## Checklist

- [x] Added unit tests


CairoMakie/src/primitives.jl
```diff
@@ -458,34 +458,6 @@ end
...
 
-"""
-    interpolation_flag(is_vector, interp, wpx, hpx, w, h)
-
-* is_vector: if we're using vector backend
-* interp: does the user want to interpolate?
-* wpx, hpx: projected size of the image in pixels, so the actual width in pixels on screen
-* w, h: size of image in pixels
-"""
-function interpolation_flag(is_vector, interp, wpx, hpx, w, h)
```
SimonDanisch, line 469: Why not leave this in a function?

jkrumbiegel: Because the flag logic didn't work anyway, there's no point in switching between the three, there's only two basically. So the main logic isn't about picking the flag but deciding whether to take the fast path or not.

PR #1830, opened 2022-04-21
Fix segfault during streamplot

NauticalDrop:
Somehow, when I used streamplot on my machine it segfaults without this.  These are my fixes for it.

* Specify the length of `Point` when creating `x0`
* Remove broadcasted addition of arrays in projection

# Description

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/streamplot.jl
+++ b/src/basic_recipes/streamplot.jl
@@ -104,7 +104,7 @@ function streamplot_impl(CallType, f, limits::Rect{N, T}, resolutionND, stepsize
         end)
         ind += 1
         if mask[c]
-            x0 = Point(ntuple(N) do i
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Oh wow! What Julia version, OS and CPU was this?

NauticalDrop:
I'm not sure whether that was the exact cause - the stacktrace pointed explicitly to the line which created the Point.  Could be a different bug, but this was Julia v1.7.2 on Mac M1.

asinghvi17 (approved):

SimonDanisch:
Ok, merging this, although this very likely just masks a Julia M1 bug.
If you can find the time, it would be helpful to try to distill an MWE from this and open an issue with Julia;)

PR #1835, opened 2022-04-23
Improve Axis3 interactivity performance

ffreyer:
# Description

### Change in text

I noticed that `Axis3` can be rather sluggish at times and one reason for it seem to be text updates. A large part of the time spent on updating the angles in `Axis3` was in the update of `camera.projectionview`. From there I intercepted the updates with timed updates with this:

```julia
using GLMakie, LaTeXStrings
fig = Figure()
ax = Axis3(fig[1, 1], xlabel = L"Test $\frac{1}{1+x}$")

function print_time(f, i)
    x -> begin
        t = time()
        val = f(x)
        dt = time() - t
        if dt > 1e-3
            printstyled("\nTime: $dt\n", color = :red)
            println("@ index $i")
            flush(stdout)
        end
        return val
    end
end

src = ax.scene.camera.projectionview
for i in eachindex(src.listeners)
    src.listeners[i] = print_time(src.listeners[i], i)
end

fig
```

and found that most of the time was spend in `add_ticks_and_ticklabels!`, in (and after) `label_pos_rot_valign`. (You can find that out by doing autocompletes on the functions held by the observable.)

Well, long story shot - I removed the `position` trigger from text layouting and that reduced this particular timing from `~0.06s` to `~0.006s`.  I also changed the method with `(string, pos)` pairs to only update when necessary to avoid rebuilding glyphcollections too much.

### In Axis3

For this I tested with `src = ax.azimuth`. After the text changes the first 3 listeners took ~0.006, ~0.006 and ~0.012s. The first two are `mi1` and `mi2` and the last is the calculation of `projectionview`. I changed `mi1`, `mi2` and `mi3` to only update on changes which practically eliminates the first two calls.

I also split up some observables for the axis label to avoid triggering text layouting unnecessarily. That (or the changes `text!` for latexstrings) reduced the last timing to ~0.006s for me.

## Type of change

- Performance


ReferenceTests/src/tests/text.jl
```diff
@@ -298,7 +298,7 @@ end
 
     Makie.step!(st)
     ## change lengths
-    textnode.val = push!(textnode[], L"\int_0^5x^2+2ab")
```
ffreyer, line 301: With the changes in the pr this doesn't work anymore for CairoMakie, but updating the nodes in either order normally works

SimonDanisch: what does that mean?

ffreyer: I think I made the `textnode.val = ...` -> `textnode[] = ...` explicitly for CairoMakie tests to pass

SimonDanisch: ah that way around...

SimonDanisch (approved):
As far as I can tell, this works pretty well :) tested a few latex + text updates and Axis3 interaction

ffreyer:
It would probably be worth testing updates to attributes like `xticklabelalign` as well. I think those are the most likely to stop working from this

jkrumbiegel (approved):
Looks good to me although I didn't have time to run it.

SimonDanisch:
Hm,  
```julia
f, ax, pl = scatter(1:4; axis=(xticklabelalign=(:left, :bottom),))
ax.xticklabelalign = (:right, :top)
```
Doesn't update, but also doesn't update on last tagged version..Also, the block refactor actually broke setting it completely, but I'll make a PR to fix that.

ffreyer:
https://github.com/JuliaPlots/Makie.jl/blob/b3b0c6daaff5a7c5d5ed36310e5a65fc11832b67/GLMakie/src/drawing_primitives.jl#L313-L327

This is what makes updates of textsize, align, font and probably everything else that isn't position, offset or transfunc fail with `Vector{GlyphCollection}`. Why is it there?

ffreyer:
Good old synchronization issues is why it's there...

ffreyer:
Didn't mean to link the second issue here lol. If tests pass the issue with updates to text attributes not propagating properly with vector positions should be fixed.

SimonDanisch:
>Somehow this changed the colors in "Dynamically adjusting number of particles in a meshscatter".

ColorTypes.jl changed their rand implementation, which lead to this... I rebased against master and updated the ref images, so hopefully should turn green now!

PR #1838, opened 2022-04-24
Better colormap handling in CairoMakie

asinghvi17:
Define a method `to_colormap(::Nothing) = nothing`, which is required for CairoMakie to plot polygons as meshes 

## Type of change


SimonDanisch:
I'm trying to be more correct with the conversions these days. `nothing` is not a valid colormap, so that shouldn't leak into the conversion pipeline. We'll need to change CairoMakie's code in this case.

asinghvi17:
ping @SimonDanisch - this should be ready to merge now.

PR #1845, opened 2022-04-28
Test on latest 1.x release

greimel:
... as seen here https://github.com/JuliaData/DataFrames.jl/blob/ca0fb4d7614786fb08b1feb4021b52d6aba9f244/.github/workflows/ci.yml#L17

### DIFF
--- a/.github/workflows/cairomakie.yaml
+++ b/.github/workflows/cairomakie.yaml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         version:
           - '1.3'
-          - '1.6'
+          - '1' # automatically expands to the latest stable 1.x release of Julia
         os:
           - ubuntu-latest
         arch:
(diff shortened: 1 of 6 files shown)


SimonDanisch:
I don't really like `1`, since it will change the version suddenly one day to another.
Also, what we really want to test is the supported range of julia versions, so it shouldn't get shifted by a new julia release, but instead by dropping/adding support for specific julia versions.

greimel:
Right, on the other hand julia 1.7 has been out for half a year and Makie hasn't been tested on it so far.

If something breaks wouldn't it be better to catch it here on CI, rather then waiting until somebody notices in the wild?

SimonDanisch:
Ok, I thought about it a bit more and am fine with merging this..Thank you :)

PR #1846, opened 2022-04-28
Specify size of matrices in `rotatedrect`

asinghvi17:
# Description

Fixes an issue with MakieTeX layouting.

There was a method ambiguity when constructing `Mat(args...)`, which I fixed by explicitly providing sizes (`Mat{n, m}(...)`).  I guess this method wasn't used so the error was never caught.

## Type of change

## Checklist


SimonDanisch:
Thanks :)

PR #1847, opened 2022-04-29
Add vertical text centering methods

jkrumbiegel:
- add tex gyre heros for ascender testing
- add new centering modes
- add centering mode tests
- use em_center and em_top for tick label defaults

# Description

I tried to change the default font to TeX Gyre Heros (a free Helvetica / Arial clone that is
also used by ProPlot, for example) which I think looks more clean and professional than Dejavu Sans.
I noticed that the y tick labels were off center, and investigation showed that this is because
TeX Gyre Heros has a large ascender value. When text is centered from descender to ascender,
it looks like it's too low. Therefore this PR adds different centering modes. The vertical align modes are now:

- center: From lowest decender to highest ascender
- em_center: From lowest baseline to highest em (height of M character)
- ex_center: From lowest baseline to highest ex (height of x character)
- baseline: At lowest baseline
- bottom: At lowest descender
- top: At highest ascender
- em_top: At highest em (height of M character)

## Type of change

- New feature

## Checklist


SimonDanisch:
Closing for now in favor of https://github.com/JuliaPlots/Makie.jl/pull/1897.
There may still be useful changes in here, but since it's heavily conflicting right now, we likely will need to start anew anyways.

PR #1848, opened 2022-04-29
add trace compile script

SimonDanisch:
Cleaned up version of #1716


SimonDanisch:
I only need this as a branch for now, since it's not clear how/when we want to actually merge this

PR #1851, opened 2022-04-29
Add a tutorial: creating a marginal histogram

asinghvi17:
This PR adds a short tutorial on creating a marginal histogram.  It's mainly aimed at showing another use-case which involves gridlayouts within figures, and as a reference in case anyone else wants to do the same thing.

It would also be a good starting point to create a `Block` MarginalHist recipe.

## Type of change

Addition to documentation

## Checklist
- [x] Added or changed relevant sections in the documentation.


asinghvi17:
I could also rework this to use protrusions but thought that the gridlayout approach might be more flexible.  We could of course show both.

jkrumbiegel:
Where do you think this would fall in the new 4 sector docs approach we'd like to move towards? Is it simple and non-explanatory enough for a tutorial? I wanted to rework the others under that principle.

asinghvi17:
Yeah, I'd definitely call it a tutorial!

asinghvi17:
This probably needs to be updated quite a bit for the Vitepress docs system, I'll get on that today.

asinghvi17:
Do we need a changelog for this?

ffreyer:
For docs only if it's very big/significant

jkrumbiegel:
Yes there should be tutorials for layouting for sure, the one I have is too complex for learning the concepts I think. An example like yours could fit that well. I'm going to close this for now, then

PR #1872, opened 2022-05-03
Add single-plot rasterization capability to CairoMakie

asinghvi17:
# Description

Allows the user to tell CairoMakie to rasterize individual Plots in a Scene, by setting `plt.rasterize = scale::Int || true`.  

CairoMakie will rasterize that specific plot to an image with `scale^2` pixels per Scene pixel (1 px per scene px if scale=1 or scale=true) and draw it to the main context.  The drawing order also preserves the z-order of the scene.

This has the potential to drastically decrease file sizes, which is extremely useful for publication plots and large meshes.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/infrastructure.jl
+++ b/CairoMakie/src/infrastructure.jl
@@ -255,21 +255,9 @@ end
 #   instead of the whole Scene
 # - Recognize when a screen is an image surface, and set scale to render the plot
 #   at the scale of the device pixel
-function draw_plot_as_image(scene::Scene, screen::CairoScreen, primitive::Combined, scale = 1)
-    # you can provide `p.rasterize = scale::Int` or `p.rasterize = true`,
-    @assert scale isa Int || scale isa Bool
+function draw_plot_as_image(scene::Scene, screen::CairoScreen, primitive::Combined, scale::Number = 1)
+    # you can provide `p.rasterize = scale::Int` or `p.rasterize = true`, both of which are numbers
 
-    # If plt.rasterize = true, then we want to rasterize to the exact scale necessary.
-    # So we find the device scaling factor so that we can scale our own rendered
-    # image to that precise scale.
-    if scale == true
-        xscale = Ref(0e0)
-        yscale = Ref(0e0)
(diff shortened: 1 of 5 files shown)


greimel:
That's a great feature! This will be handy for plotting maps.

It would be great to have that as a global option as well. Pluto displays `png`s by default because to many `svg`s slow down the browser. Probably this problem could be solved by rasterizing all plots, while keeping axis decorations and text as vector graphics.

asinghvi17:
I think that this could actually be themable as it is now - if you set `rasterize = true` in the default themes of Heatmap and Surface, for example, that should "just work".  The only potential issue right now is having too many rasterized plots - since each one gets its own image, that would cause a lot of size bloat.  Fixing that would be a bit complicated, though.

CairoMakie/src/infrastructure.jl
```diff
@@ -240,6 +250,43 @@ function draw_plot(scene::Scene, screen::CairoScreen, primitive::Combined)
...
+    Cairo.rectangle(screen.context, 0, 0, w, h)
+    Cairo.save(screen.context)
+    Cairo.translate(screen.context, 0, 0)
+    # Cairo.scale(screen.context, w / scr.surface.width, h / scr.surface.height)
+    Cairo.set_source_surface(screen.context, scr.surface, 0, 0)
+    p = Cairo.get_source(scr.context)
+    # this is needed to avoid blurry edges
+    Cairo.pattern_set_extend(p, Cairo.EXTEND_PAD)
+    # Set filter doesn't work!?
+    Cairo.pattern_set_filter(p, Cairo.FILTER_BEST)
```
SimonDanisch, line 283: Vector backends don't support that filter, see: https://github.com/JuliaPlots/Makie.jl/pull/1828
So, we should remove it?

asinghvi17:
The CI failure on Makie/Julia 1 does not appear to be related - seems related to mouse click detection.

asinghvi17:
@Datseris this may be useful if you are saving GeoMakie plots to pdf - just pass `rasterize = 5` or so as a keyword argument to `surface` or `heatmap`!

PR #1873, opened 2022-05-03
fix manual cycling of attrs

SimonDanisch:


jkrumbiegel (approved):
Made some small changes, rest is good to go.

PR #1876, opened 2022-05-04
Add a path clip recipe and implement it in CairoMakie

asinghvi17:
# Description

This PR defines a `clip` recipe which takes a path as `Vector{Point2f}`, and clips all plots in the Scene to be within that path.  
It also implements the recipe in CairoMakie.
This is very useful for nonlinear transforms, such as geographic projections and relativistic plots.

I found interesting links which discuss how one might do this in OpenGL [here](https://stackoverflow.com/questions/9374622/how-to-clip-with-circles-in-opengl) and [here](https://stackoverflow.com/questions/36225906/lwjgl-3-clipping-mask), but unfortunately I don't know much about OpenGL as of now and can't really implement it there.

iTerm2 atHzFl
iTerm2 oIk8MP

Fixes https://github.com/JuliaPlots/GeoMakie.jl/issues/105

## Type of change

## Checklist

- [ x Added or changed relevant sections in the documentation


jkrumbiegel:
It seems to me that clip should be an attribute that a plot or a scene can have, no? For example, if you want to plot a circular cutout of a heatmap.

asinghvi17:
Good point...that would make things somewhat easier on the backend.  I can definitely implement that on the Cairo end, but someone would probably have to implement it for GLMakie.  

I would lean towards only accepting `clip` as a Scene attribute, and having that clip in addition to what we currently have (clip at the Scene's pixel boundary).  Currently, one can pass a Vector of Point2f to the recipe - would that be sufficient to pass to the Scene?  I could just include the clip code in `prepare_for_scene`.

jkrumbiegel:
Another question is what the space should be in which the clip is given. At least a 3d clip doesn't really make so much sense? On the other hand, our 3d plots currently stick out of Axis3 for example, if the limits are chosen manually. I'm not sure how other plotting softwares handle this problem. To me this kind of 3d clipping seems different than 2d cutouts, and Cairo can only do 2d anyway.

asinghvi17:
Definitely!  For Scenes, I would argue that the clip should, by default, be in the same coordinate space as the input.  We can definitely have both `clip` and `clipspace` as attributes, so the user could theoretically clip in the associated Scene's pixel space  or (-1, 1) space if required.

PR #1882, opened 2022-05-05
fix (x/y)ticklabelaling

SimonDanisch:


asinghvi17:
Oops...for some reason I read labelalign when it should have been ticklabelalign.  My bad, sorry about that :D

PR #1896, opened 2022-05-07
close #1893

### LINKED ISSUE #1893: downstream CI for GraphMakie?
would a PR adding downstream CI for GraphMakie like https://github.com/SciML/ModelingToolkit.jl/blob/master/.github/workflows/Downstream.yml be accepted?

I ask because GraphMakie tests don't seem to pass with 0.17, maybe it'd be nice to know for a subset of dependents when things are breaking for that package. AoG may be good too

anandijain:


.github/workflows/downstream.yaml
```diff
@@ -0,0 +1,50 @@
...
+            Pkg.update()
+            Pkg.test()  # resolver may fail with test time deps
+          catch err
+            err isa Pkg.Resolve.ResolverError || rethrow()
+            # If we can't resolve that means this is incompatible by SemVer and this is fine
+            # It means we marked this as a breaking change, so we don't need to worry about
+            # Mistakenly introducing a breaking change, as we have intentionally made one
+            @info "Not compatible with this release. No problem." exception=err
+            exit(0)  # Exit immediately, as a success
+          end
```
asinghvi17, line 50:
```suggestion
          julia -e '
          using Pkg
          try
            # force it to use this PR's version of the package
            Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
            Pkg.update()
            Pkg.test()  # resolver may fail with test time deps
          catch err
            err isa Pkg.Resolve.ResolverError || rethrow()
            # If we can't resolve that means this is incompatible by SemVer and this is fine
            # It means we marked this as a breaking change, so we don't need to worry about
            # Mistakenly introducing a breaking change, as we have intentionally made one
            @info "Not compatible with this release. No problem." exception=err
            exit(0)  # Exit immediately, as a success
          end 
          '

**Comment by asinghvi17** [comment]
From what I can tell, it doesn't look like GraphMakie was actually tested here, but that seems to be the whole purpose of this PR?  Maybe we could edit the Project.toml of the downstream packages (using TOML.jl) to be compatible with the latest version of Makie during CI, so we can see if there are any issues.

**Comment by anandijain** [comment]
See https://github.com/JuliaPlots/Makie.jl/runs/6335691155?check_suite_focus=true#step:6:21. 
It's intentional that it didn't run since 0.17 was breaking and GraphMakie hasn't updated their compat bounds

**Comment by anandijain** [comment]
You make a good point though. I don't know of an example where they actually handle the breaking case and test with the new version.

File: .github/workflows/downstream.yaml [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+      - name: Clone Downstream
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          path: downstream
+      - name: Load this and run the downstream tests
+        shell: julia --color=yes --project=downstream {0}
+        run: |
+          using Pkg
+          try
```
asinghvi17, line 38:
```suggestion
           makie_version = TOML.parsefile("./Project.toml")["version"]
           downstream_project = TOML.parsefile("downstream/Project.toml")
           downstream_project["compat"]["Makie"] = makie_version

           write("downstream/Project.toml") do io
               TOML.print(io, downstream_project)
           end

           try
```

asinghvi17: We could try something like this to actually edit the downstream's Project.toml to be compatible with Makie's current master version.

SimonDanisch:
Sorry, I forgot to comment on the issue, but I'm not a big fan of downstream CI tests. 
Shouldn't this be handled by compathelper, whenever there is a breaking version release?
I mean, what would a CI here do, besides getting ignored? It's usually pretty clear that breaking changes will break tightly integrated packages like AoG, and it will need to get addressed in the PR upping the compat version.

jkrumbiegel:
I think this might be more about non-breaking changes or internal changes that break downstream packages. But that poses a different problem, should the maintenance burden of depending on internal behavior be passed to Makie? I don't really want to have to go through downstream packages or consider their internals when updating Makie internals that are not API.

SimonDanisch:
I think we've been pretty fine with releasing patch release that aren't breaking downstream packages these days.
If it happens again, we should add more tests to the test suite.

SimonDanisch:
I'm going to close this for the above reasons. Sorry for the work you already put into this.
If you really want something like this, feel free to still convince us ;)

anandijain:
Sorry, just getting to it. It was no work at all, just wanted to get a sense.
I think it makes sense to not have it since it's really only because there was a breaking release that things are sort of hectic.

.github/workflows/downstream.yaml
anandijain: thanks, this is still really useful

SimonDanisch:
> it's really only because there was a breaking release that things are sort of hectic.

what do you mean by this? What is, and what's hectic?

SimonDanisch:
My main question is: what is the workflow you imagine happening?
If I make a Makie PR that breaks e.g. AoG, and that will fail the CI, I will neither start fixing AoG nor will it stop me from merging that PR and releasing a breaking Makie version.
Nobody is forced to upgrade to a new version, and we've been trying to make better release notes to help with upgrades to breaking versions (admittedly, we can still improve even more on that front).

PR #1901, opened 2022-05-09
properly deprecated to_colormap(cm, n)

SimonDanisch:
`to_colormap(cm, n)` got removed without properly deprecating it, and without actually testing & exporting the new way of doing things (`categorical_colors`).
This adds an descreptive error to `to_colormap(cm, n)` and tests, exports and fixes `categorical_colors`


SimonDanisch:
The original PR: https://github.com/JuliaPlots/Makie.jl/pull/1723

PR #1916, opened 2022-05-11
Minor fixes and improvements

asinghvi17:
# Description

Adds some minor fixes and improvements.

## Type of change

## Checklist


SimonDanisch (approved):
Looks good to me. @jkrumbiegel, are those `GridLayoutBase` overloads desired?

jkrumbiegel:
Hmm I had thought about doing something similar before. In one way, `f.layout` is not super Julian, we don't like to access fields. On the other hand, this will basically never change. At least I don't foresee it. So I don't see `rowsize!(f.layout, ...)` as such a burden, you need to know anyway this applies to layouts in case you do nested ones.

jkrumbiegel:
I'd rather see `resize!(f, ...)`

asinghvi17:
@jkrumbiegel how do you feel about the docstrings in https://github.com/JuliaPlots/Makie.jl/pull/1916/commits/c5f26cd29a9f8a5245f69ba2cb80693a1af5e58c?  I'd like to upstream them to GridLayoutBase but am not sure whether they're 100% correct or not.

jkrumbiegel:
They look ok on first glance, yes better to directly PR them into GridLayoutBase

asinghvi17:
This looks good to merge from my end.

In regards to the row/col method forwarding, I agree it's not particularly required but it is a pretty good quality-of-life factor.  When writing complicated layouts it somewhat helps to keep things short, IMO.  What the methods do is also pretty clear in the docstrings, so that should help clear up confusion on that front.

src/scenes.jl
```diff
@@ -559,3 +559,51 @@ struct FigureAxisPlot
...
+are already present; it only changes the figure's physical size.
+
+See also [resize_to_layout!(fig::Figure)](@ref).
+"""
+Makie.resize!(figure::Figure, args...) = resize!(figure.scene, args...)
+Makie.resize!(figure_axis_plot::FigureAxisPlot, args...) = resize!(figure_axis_plot.figure.scene, args...)
+
+
+# Overload some GridLayoutBase methods for Figures
+# Boasically, just forward `f.layout` in place of `f`.
```
asinghvi17, line 585:
```suggestion
# Basically, just forward `f.layout` in place of `f`.

jkrumbiegel:
Ok I have come to the conclusion that I don't want layout forwards on the Figure at this point. There are tons of functions from GridLayoutBase that shoud then in principle also work (why introduce a separation between supported and non supported) and I don't think that's a good way to go right now. With `resize!` this feels a bit different because the normal user should not be too concerned that there's a Scene below the Figure, but they should know about the GridLayout. So to me it currently makes sense to support one but not the other. I might rather add a `layout(f::Figure)` function at some point to remove the dot access if that looks too un-Julian.

Also, could you please remove the unrelated changes and call the PR `add resize! for Figure` or something along those lines? That makes it easier to follow what's going on.

asinghvi17:
Fair enough, makes sense.  I'll just open a new PR, should be simpler.

PR #1924, opened 2022-05-12
bump versions

piever:
# Description

Bump versions for release


CairoMakie/Project.toml
```diff
@@ -22,7 +22,7 @@ FFTW = "1"
 FileIO = "1.1"
 FreeType = "3, 4.0"
 GeometryBasics = "0.4.1"
-Makie = "=0.17"
+Makie = "=0.17.1"
```
asinghvi17, line 25: Is this necessary?  As I understand it, forcing Makie to `"^0.17.1" in AlgebraOfGraphics should be sufficient?  It seems a bit strict to me, especially since the new version of CairoMakie should be compatible with Makie 0.17.

piever: I think that's just how it's being done since the monorepo: the backends pick a specific version of Makie, see eg [here](https://github.com/JuliaPlots/Makie.jl/blob/v0.16.6/CairoMakie/Project.toml).

asinghvi17: Makes sense.  LGTM!

SimonDanisch:
We need a new release for MakieCore as well!

piever:
Ah, fair point, I've also bumped the compatibility with MakieCore (I imagine the latest one is needed for the word wrapping changes).

PR #1944, opened 2022-05-15
also consider newline chars for line wrapping

fatteneder:
# Description

In the course of adapting the text line wrapping feature I noticed that newline chars are not considered when looking for line wrap positions. Consequently, visual box width overflows can appear.

Unfortunately, I could not come up with a MWE that shows this with Makie only features.
Instead, I have two gifs here https://github.com/fatteneder/MakieSlides.jl/pull/12 that illustrate the issue.

## Type of change

- [x] Bug fix

## Checklist

- [ ] Also ignore whitespace and newline chars at which we wrap when determining the full width.
```
                      currently line width includes this trailing whitespace/newline, but it shouldn't
                            ↓                          
word{space}word{space}word{space}
```
@ffreyer Could this be related to this TODO here: https://github.com/JuliaPlots/Makie.jl/blob/07f0763bcb8e990274aec5335deaa247bcf60229/src/layouting/layouting.jl#L127

### DIFF
--- a/src/layouting/layouting.jl
+++ b/src/layouting/layouting.jl
@@ -114,7 +114,8 @@ function glyph_collection(
             push!(xs[end], x)
             x += charinfos[i].hadvance
             
-            if 0 < word_wrap_width < x && (ci.char == ' ' || i == length(charinfos)) && 
+            if 0 < word_wrap_width < x &&
+                    ((ci.char == ' ' || ci.char == '\n') || i == length(charinfos)) &&
                     last_space_local_idx != 0
 
                 newline_offset = xs[end][last_space_local_idx + 1]
@@ -133,6 +134,14 @@ function glyph_collection(
                 
                 if i == length(charinfos)
                     push!(lineinfos, view(charinfos, last_line_start:i))
+                end
+
+                if ci.char == '\n'
+                    push!(xs, Float32[])
+                    push!(lineinfos, view(charinfos, last_line_start:i))
+                    last_space_local_idx = 0
+                    last_line_start = i+1
+                    x = 0f0
                 end
 
             elseif ci.char == '\n' || i == length(charinfos)


src/layouting/layouting.jl
```diff
@@ -135,6 +136,14 @@ function glyph_collection(
                     push!(lineinfos, view(charinfos, last_line_start:i))
                 end
 
+                if ci.char == '\n'
+                    push!(xs, Float32[])
+                    push!(lineinfos, view(charinfos, last_line_start:i))
+                    last_space_local_idx = 0
+                    last_line_start = i+1
+                    x = 0f0
+                end
```
ffreyer, line 145: I think we could just make the `elseif ci.char == '\n' || i == length(charinfos)` an `if ci.char == '\n' || i == length(charinfos)` instead of repeating this

fatteneder:
I ticked off the only TODO because I think it does not apply.
Ready for merge then.

ffreyer:
```julia
fig = Figure()
str = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
lbl = Label(fig[1, 1], str, word_wrap = true)
Box(fig[1, 1], color = (:red, 0.3))
fig
```
This errored for me immediately with the changes I suggested. I tried to fix that and also a lack of wrapping for the final word. I checked the above and the above with a couple newlines in there. I think it's working now, could you verify?

On a side note I noticed that the Label doesn't update immediately/correctly. We should look into that at some point too.

fatteneder:
Can confirm that your MWE did work with my initial changes, but not with the 'polished fix'. With the latest version it works again. And it also works with https://github.com/fatteneder/MakieSlides.jl/pull/12.

> On a side note I noticed that the Label doesn't update immediately/correctly. We should look into that at some point too.

Can confirm too it is not working perfectly. It appears that sometimes it is a bit off vertically. Here is a gif: 
delayed-box

fatteneder:
The wrong placement seems to happen if there is only one word in the last row.

ffreyer:
The labels should update correctly now

ffreyer:
I looked into this some more. If you use
```julia
fig = Figure()
lbl = Label(fig[1, 1], lorem_ipsum, word_wrap = true)
mesh!(fig.scene, lbl.layoutobservables.computedbbox, color = (:yellow, 0.3))
fig
```
there are no problems with the bounding boxes. It's also fine if you set up the `Box` first:
```julia
fig = Figure()
Box(fig[1, 1], color = (:red, 0.3))
lbl = Label(fig[1, 1], lorem_ipsum, word_wrap = true)
fig
```
But if you use the `Box` after the label its bbox can become missaligned. 
```julia
fig = Figure()
lbl = Label(fig[1, 1], lorem_ipsum, word_wrap = true)
Box(fig[1, 1], color = (:red, 0.3))
fig
```
As far as I can tell the recursive updates in `Label` are messing up `Box` here. In this configuration the bbox of `Box` triggers after the update to `word_wrap_width` which updates the Label bbox to the correct size. So `Box` gets the correct size from lower levels of the recursion, but then gets set back to the old size in the outer level.
I guess one way to fix this is to do the bbox update of `Box` at higher priority/first, but I don't know what that involves.

fatteneder:
Oh, I see. I thought I had solved this in `MakieSlides` by providing an exit from the recursion which triggered when the box's height is approximately the text's auto height. However, adjusting your MWE to use `FormattedLabel` now behaves exactly the same ...

ffreyer (approved):

fatteneder:
@ffreyer thank you for taking care of this and getting it merged!

PR #1947, opened 2022-05-16
Change docs to use `Theme` instead of `Attributes`

### LINKED ISSUE #332: What is the relationship between theme and attributes?
Hi,

I was wondering what is the relationship between theme and attributes. For example, if I update one of the Attributes in attribute do the attributes in theme get update as well? 

Best,

kunzaatko:
- edit(docs): recipes docs to use Theme (#332)

Fixes #332


SimonDanisch:
Not sure what happened here... 
But this PR is a bit doomed, since I rather want to deprecate Theme and use Attributes consistently throughout. 
I'm actually currently working on figuring that out, so this PR will be difficult to combine with my changes.
Sorry for the confusion!

kunzaatko:
It was actually suggested to me that `Theme` is the way to go. So maybe instead there should be a PR that prefers `Attributes` to `Theme`?

SimonDanisch:
>It was actually suggested to me that Theme is the way to go.

Yeah sorry about that, I didn't reply fast enough to clear up the misunderstanding.

>So maybe instead there should be a PR that prefers Attributes to Theme?

As I said, I'm actively working on refactoring `Attributes` so, there is a high chance that whatever you do will be in conflict with my PR.

kunzaatko:
OK. Sure. Sorry for the misunderstanding and thank you for clearing things up.

PR #1952, opened 2022-05-18
Kolaru compat with mathtexengine

SimonDanisch:
Looked at https://github.com/JuliaPlots/Makie.jl/pull/1785 today with @jkrumbiegel and using `height_insensitive_boundingbox` for chars seemed rather important for the rest of the text pipeline.

It doesn't seem to solve it though, since the M and O isn't centered correctly with `height_insensitive_boundingbox`:
image

It's centered correctly with `inkboundingbox`, but then the (:left, :bottom) is wrong, since the M shouldn't change positions depending on whether the g is present:
image

test code:
```julia
begin
    f = Figure()
    for (i, str) in enumerate(["M", "gM", "O"])
        for (j, align) in enumerate([(:left, :baseline), (:left, :bottom), (:center, :center)])
            ax, pl = scatter(f[i, j], Point2f(0, 0), axis=(title="$align",))
            text!(str, position = Point2f(0, 0), textsize = 50, align = align)
            hidedecorations!(ax)
        end
    end
    f
end
```


src/types.jl
```diff
@@ -319,7 +319,7 @@ end
 
 function GlyphExtent(font, char)
     extent = get_extent(font, char)
-    ink_bb = FreeTypeAbstraction.inkboundingbox(extent)
+    ink_bb = FreeTypeAbstraction.height_insensitive_boundingbox(extent, font)
```
ffreyer, line 322: I think this makes bounding boxes less accurate. Why not undo my changes instead? (use character height instead of font height)

jkrumbiegel: it is that way on purpose. text should not be bounding-boxed tightly around the glyphs, but given the font's ascenders and descenders. unless there is a specific purpose for having a tighltly bounded glyph, but normal text is not that

jkrumbiegel: the question is just if LaTeX should use something else because text is positioned all over the place there. I don't know how real LaTeX handles equation bounding boxes

jkrumbiegel: ah ok I guess I misunderstood, the ink_bb should of course be the inkboundingbox, but that doesn't mean we have to use it to compute text boundingboxes. it's still useful to have in that struct

ffreyer: Another direction I thought about going with my changes was to use something like `overallheight = max(font_height, max_height_of_line)` for latex. I guess that would be the way if we want to keep text aligned based on the largest character/font rather than individual characters.

Maybe we should also just have both styles for alignments and bounding boxes. I.e. one based on characters and one based on the (if necessary padded) font (height)?

jkrumbiegel: ideally the font ascender / descenders shouldn't be that far off from normal strings that have the occasional f or g. Just some fonts define those metrics in a peculiar way (case in point, the need for TeX Gyre Heros Makie). So only if you were to do something where absolute precision of boundingboxes of strings was necessary would you need those other styles. I think in all other situations the non-jitteryness is preferable and seems to be what other text-based tools or browsers etc are doing.

jkrumbiegel:
With this I get 
grafik

Kolaru:
Thanks everyone who stepped up to finish that! I was on vacation for the last two weeks and couldn't contribute.

PR #1961, opened 2022-05-20
latency improvements

SimonDanisch:
Trying to make lineaxis more inferable, and maybe finally get the unnecessary updates under control.


SimonDanisch:
Somehow the merges became really messy, and this PR should also be rather based on master.
Continued in: https://github.com/JuliaPlots/Makie.jl/pull/1968

PR #1963, opened 2022-05-21
Update makie_logo_transparent.svg

cormullion:
# Description

The text "Makie" is invisible if you use Github in Dark Mode. The simplest fix is to add a thin white stroke. Better would be to add a slight border on the original. Or to make a non-transparent image that has some cool background elements!


jkrumbiegel:
Ah yes that makes sense, I don't think Github allows to have css-dependent images in a Readme so maybe the banner-like solution could also work.

jkrumbiegel:
I'll merge this for a quick fix and we can think about a different solution when we have time :) thanks!

cormullion:
Haha, I'm stupid - I edited the wrong one...I should have changed the "canvas" one : 

Screenshot 2022-05-24 at 15 18 22

Never mind, it will keep for another day! :)

jkrumbiegel:
Yeah I noticed that too after merging ;)

PR #1964, opened 2022-05-21
make broadcast_foreach generated for better performance

jkrumbiegel:
This got my attention while looking at CairoMakie plotting functions. It seemed that the `args...` gave the implementation trouble when `map`ing or broadcasting over it, so I changed to a `@generated` implementation that can correctly unroll these expressions.

This is approximately 100x faster and has no allocations on its own. Imagine that the below is a scatter plot of 1 million points, then at least 7 seconds of that would have been just this function, without any drawing.

```julia
julia> f = ((args...,) -> nothing)
#19 (generic function with 1 method)

julia> function test()
           for i in 1:1000000
               Makie.broadcast_foreach(f, 1:3, 1.0:3, 'a':'c')
           end
       end
test (generic function with 1 method)

# before

julia> @time test()
  6.923584 seconds (22.00 M allocations: 1.222 GiB, 3.12% gc time, 0.00% compilation time)

# after

julia> @time test()
  0.069359 seconds
```

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -181,23 +181,25 @@ An example would be a collection of scatter markers that have different sizes bu
 The length of an attribute is determined with `attr_broadcast_length` and elements are accessed with
 `attr_broadcast_getindex`.
 """
-function broadcast_foreach(f, args...)
-    lengths = attr_broadcast_length.(args)
-    maxlen = maximum(lengths)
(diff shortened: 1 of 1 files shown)


asinghvi17:
Nice!  Looking forward to seeing how this changes the previously long mesh/surface drawing times in CairoMakie.

PR #1968, opened 2022-05-23
Improve inferability of lineaxis and axis

SimonDanisch:
- ~~use `@max_methods` for performance insensitive modules~~ moved to https://github.com/JuliaPlots/Makie.jl/pull/2000
- try to clean up line axis code and make it more inferable and remove more double updates

### DIFF
--- a/CairoMakie/src/CairoMakie.jl
+++ b/CairoMakie/src/CairoMakie.jl
@@ -14,6 +14,10 @@ using Makie: inline!
 using Makie.Observables
 using Makie: spaces, is_data_space, is_pixel_space, is_relative_space, is_clip_space
 
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@max_methods"))
+    @eval Base.Experimental.@max_methods 1
+end
+
 const OneOrVec{T} = Union{
     T,
     Vec{N1, T} where N1,
--- a/CairoMakie/src/CairoMakie.jl
+++ b/CairoMakie/src/CairoMakie.jl
@@ -14,10 +14,6 @@ using Makie: inline!
 using Makie.Observables
 using Makie: spaces, is_data_space, is_pixel_space, is_relative_space, is_clip_space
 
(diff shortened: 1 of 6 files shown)


SimonDanisch:
@IanButterworth, you may want to check this PR out, as I turned quite a lot of observables into "change observables" while at it...
I don't have a benchmark setup right now, so not 100% sure if I hit the right observables and how much it makes a difference.

ffreyer:
Does `max_methods` just shift compile time to run time? It kinda seems that way from the benchmarks

SimonDanisch:
> It kinda seems that way from the benchmarks

What benchmarks? We don't have run time benchmarks and both using + ttfp times are an improvement

>Does max_methods just shift compile time to run time

Depends on the code, fully typed + inferable code shouldn't change. Only code which calls functions, where the arguments aren't fully inferable may get slower runtime dispatch.

ffreyer:
Iirc the first compile time benchmark I saw here had around -15% on using and maybe 1% more or less on ttfp. With all the other changes now I don't know what's actually improving things. If the ttfp improvements are mostly from change observables it might not be worth having the experimental stuff. Especially since "experimental" hints that it may break soon.

SimonDanisch:
>specially since "experimental" hints that it may break soon.

It was regarded as a bad move to name the module `Experimental`, because in hindsight, it wasn't really meant like that... So I think we can safely us it for now.

Some more detailed benchmarks:

branch | using | ttfp | runtime
 --- | ---   | --- | ---
this | 4.8 | 14 | 19.8, 3.5mb
this, no max_method | 5.6 | 14 | 19.8ms, 3.5mb
master | 5.7  | 15.2 | 21ms, 3.68 MiB

I think this shows nicely, that `max_methods` gives us significant less using times, without influencing ttfp or runtime, while the net improvement of the other changes are pretty nice on top.

ffreyer:
Oh I thought ttfp would include using time as well. So then this is really just a straight up improvement

PR #1979, opened 2022-05-25
Respect scale and nonlinear values in PlotUtils cgrads

### LINKED ISSUE #1977: Unexpected behavior of colormaps (cgrad)
This code:

```julia
sing CairoMakie

cmap = cgrad([:black,:white,:orange],[0,0.2,1])
f = Figure(resolution=(400,200))
ax = Axis(f[1,1])
x = range(0,1,50)
scatter!(f[1,1],Point2.(x,fill(0.,50)),color=x,colormap=cmap)
hidedecorations!(ax)
Colorbar(f[2,1],vertical=false,colormap=cmap)
```

creates this image: temp1

The Colorbar is rendered as expected, but the points are not. The behavior I expected was what you get from calling

```julia
scatter!(f[1,1],Point2.(x,fill(0.,50)),color=get(cmap,x))
```

asinghvi17:
# Description

Fixes #1977 

Essentially, reverts a change from #1723.

I'd like to release a v0.16.7 with this change as well. since it is somewhat serious.

### Before
iTerm2 pNXQmN
(note how they are all the same)
### After
iTerm2 NH1H6Q

## Type of change

## Checklist


asinghvi17:
CI errors are because of a missing reference image for the new test.

SimonDanisch:
Alright, I took a deep dive to make sure this is correct now... I guess I misunderstood `cgrad` a bit when I made that change.
I added some more tests to make sure we do the right thing now

asinghvi17:
Thanks for taking this forward!

On a side note - I wonder if returning a `Sampler` might be a more efficient solution?  Colormaps with intentionally tiny variations, like those with e.g. `scale = x -> log(100, x)`,  might then be somewhat misrepresented.

I'm not sure whether [W]GLMakie will accept Sampler (from colorsampler.jl) as a native type but I think something like that might be a somewhat more elegant solution to the problem.

SimonDanisch:
Yeah we need a refactor of the whole colormap pipeline and then make sure that wherever colormaps are used, they work exactly the same without hacks like we have here

PR #1989, opened 2022-05-27
Ensure that scatter respects low/high clip

### LINKED ISSUE #1980: scatter ignores highclip
```julia
let
	fig = Figure(resolution=(400, 300))
	ax = Axis(fig[1, 1];)
	ys = 1:10
	sc = scatter!(ax, ys; color=ys, colorrange=(1, 6), highclip=:red)
	fig
end
```
image
I'd expect points from 7 to 10 to be red ?

`CairoMakie v0.8.3`

asinghvi17:
# Description

Fixes #1980 

Adds checking code for highclip and lowclip to `numbers_to_colors`.

## Type of change

### DIFF
--- a/CairoMakie/src/utils.jl
+++ b/CairoMakie/src/utils.jl
@@ -125,8 +125,8 @@ function numbers_to_colors(numbers::AbstractArray{<:Number}, primitive)
         Float64.(numbers), # ints don't work in interpolated_getindex
         Ref(colorrange))
 
-    !isnothing(primitive.lowclip[])  && (raw_colors[numbers .< colorrange[1]] .= Makie.to_color(primitive.lowclip[]))
-    !isnothing(primitive.highclip[]) && (raw_colors[numbers .< colorrange[2]] .= Makie.to_color(primitive.highclip[]))
+    !isnothing(to_value(get(primitive, :lowclip,  nothing))) && (raw_colors[numbers .< colorrange[1]] .= Makie.to_color(primitive.lowclip[]))
+    !isnothing(to_value(get(primitive, :highclip, nothing))) && (raw_colors[numbers .< colorrange[2]] .= Makie.to_color(primitive.highclip[]))
 
 end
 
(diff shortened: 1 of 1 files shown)


ffreyer:
I wonder if this would be better implemented in Makie. I could see it being useful occasionally if a plot type doesn't implement highclip and lowclip (and nancolor) or if you want to tweak something very specific. In those cases you could generate the colors manually, make your adjustments and then pass them in the color attribute.

asinghvi17:
Yeah, it would definitely be cleaner if nothing else!  Ideally, what we would do (and I think we had been working towards this with `Sampler`) is use PlotUtils cgrads as an intermediate step, and have the backends get their colors off of `Sampler` objects - which could then have well-defined behaviour for high/lowclip, nancolor, etc.

That would also make the colormap conversion a lot cleaner - right now it doesn't even work on master for scaled colormaps, and in #1973 I'm bringing back the old brute force solution.

CairoMakie/src/utils.jl
```diff
@@ -120,10 +120,14 @@ function numbers_to_colors(numbers::AbstractArray{<:Number}, primitive)
...
     end
 
-    Makie.interpolated_getindex.(
+    raw_colors = Makie.interpolated_getindex.(
         Ref(colormap),
         Float64.(numbers), # ints don't work in interpolated_getindex
         Ref(colorrange))
+
+    !isnothing(to_value(get(primitive, :lowclip,  nothing))) && (raw_colors[numbers .< colorrange[1]] .= Makie.to_color(primitive.lowclip[]))
+    !isnothing(to_value(get(primitive, :highclip, nothing))) && (raw_colors[numbers .< colorrange[2]] .= Makie.to_color(primitive.highclip[]))
```
ffreyer, line 129: This should be `>`, right?

asinghvi17: whoops, thanks!  guess this is what I get for coding on a plane :D

asinghvi17:
```suggestion
    !isnothing(to_value(get(primitive, :highclip, nothing))) && (raw_colors[numbers .> colorrange[2]] .= Makie.to_color(primitive.highclip[]))

lazarusA:
Oh, I see that this still some work. Hopefully it will propagate also to meshscatter.... which also is currently not working.

asinghvi17:
This is probably a bad idea, would rather go at this from the sampling perspective (so that highclip, lowclip, etc. get merged into one sampler object).

PR #1990, opened 2022-05-27
mouseposition -> mouseposition_px in interactiveapi

### LINKED ISSUE #1978: difference between `pick(ax)` and `pick(fig)`
Hello :) Makie 0.17 deprecated `mouse_selection(ax)` in favor of `pick(ax)`. However, it seems there is a difference between `pick(ax)` and `pick(fig)`. Based on the [interaction example](https://makie.juliaplots.org/dev/documentation/events/#point_picking) in the docs here is a MWE
```julia
using GLMakie

positions = Observable(rand(Point2f, 5))

fig, ax, p = scatter(positions, markersize=50)

on(events(fig).mousebutton, priority = 2) do event
    if event.button == Mouse.left && event.action == Mouse.press
        # Delete marker
        plt, i = pick(ax)   # <- thats the culprit here
        if plt == p
            deleteat!(positions[], i)
            notify(positions)
            return Consume(true)
        end
    end
    return Consume(false)
end

fig
```
If one uses `pick(fig)` the function works as expected. If one uses `pick(ax)` one has to click right above the marker in order to delete it, it seems like there is some miscalculation between figure and axis pixel coordinate system or something like that.

xref https://github.com/JuliaPlots/GraphMakie.jl/issues/82

asinghvi17:
# Description

Fixes #1978

On @ffreyer's advice, this PR replaces all instances of `mouseposition` in the interactive API functions with `mouseposition_px`, which takes the Screen's pixel area into account.

## Type of change

## Checklist

### DIFF
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -261,7 +261,7 @@ function select_rectangle(scene; blocking = false, priority = 2, strokewidth = 3
     on(events(scene).mousebutton, priority=priority) do event
         if event.button == key
             if event.action == Mouse.press && is_mouseinside(scene)
-                mp = mouseposition(scene)
+                mp = mouseposition_px(scene)
                 waspressed[] = true
                 plotted_rect[:visible] = true # start displaying
                 rect[] = Rectf(mp, 0.0, 0.0)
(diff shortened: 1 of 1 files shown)


asinghvi17:
Ah.  I guess I did rush into it based on the issue I ran into in GeoMakie.  Will look into the issue more closely later and try again - thanks!

PR #2008, opened 2022-05-31
Remove MakieLayout module

SimonDanisch:
Pretty much the same as https://github.com/JuliaPlots/Makie.jl/pull/2007 just for MakieLayout.
The extra module has just been leading to repeating code without real benefits, and we have been acting like Makie is the sole API exporter anyways.
This also removes the extra `__init__` and combines the precompiles and adds some new ones.


jkrumbiegel (commented):
I think you didn't check the docs for MakieLayout occurrences right? I know there's a couple mentions on the Axis page at least, so it would be good to also remove those

jkrumbiegel (approved):

PR #2012, opened 2022-05-31
Force colorbuffer to use the size of the screen, not the scene

asinghvi17:
# Description

This also changes the VideoStream constructor similarly, allowing the user to use patterns like this:
```julia
CairoMakie.activate!(px_per_unit=3)

record(fig, "streamplot_anim_keener.mp4") do io
    ...
end
```

This works with GLMakie and CairoMakie, but will probably break WGLMakie since there seems to be no `GeometryBasics.width` definitions for either WGLDisplay or ThreeDisplay, and I'm not sure how to go about implementing those since the display structs don't contain their Scenes directly.

I built this off my other PR which accelerates recording speed in the presence of rasterized plots, but will rebase once that is merged.

## Type of change

## Checklist


src/display.jl
```diff
@@ -333,9 +333,11 @@ function VideoStream(fig::FigureLike; framerate::Integer=24, visible=false, conn
     #codec = `-codec:v libvpx -quality good -cpu-used 0 -b:v 500k -qmin 10 -qmax 42 -maxrate 500k -bufsize 1000k -threads 8`
     dir = mktempdir()
     path = joinpath(dir, "$(gensym(:video)).mkv")
-    scene = get_scene(fig)
     screen = backend_display(scene; start_renderloop=false, visible=visible, connect=connect)
```
asinghvi17, line 336:
```suggestion
    scene = get_scene(fig)
    screen = backend_display(scene; start_renderloop=false, visible=visible, connect=connect)

File: WGLMakie/src/three_plot.jl [hunk 2]
```diff
@@ -4,6 +4,14 @@ end
 
 JSServe.session(td::ThreeDisplay) = td.session
 
+function GeometryBasics.widths(screen::ThreeDisplay)
+    # look at d.qs().clientWidth for displayed width
+    width = Int(WGLMakie.JSServe.evaljs_value(screen.session, WGLMakie.JSServe.js"document.querySelector('canvas').width"; time_out=100))
```
SimonDanisch, line 9: Every evaljs_value call is super expensive, can we make this one call? Have you tried if this works?

asinghvi17: Will try that - was unsure how the structure of jsserve worked but will try to return an array or something.

WGLMakie/src/three_plot.jl
```diff
@@ -4,6 +4,13 @@ end
 
 JSServe.session(td::ThreeDisplay) = td.session
 
+function GeometryBasics.widths(screen::ThreeDisplay)
+    # look at d.qs().clientWidth for displayed width
+    width, height = Int(WGLMakie.JSServe.evaljs_value(screen.session, WGLMakie.JSServe.js"[document.querySelector('canvas').width, document.querySelector('canvas').height]"; time_out=100))
```
asinghvi17, line 9:
```suggestion
    width, height = round.(Int, WGLMakie.JSServe.evaljs_value(screen.session, WGLMakie.JSServe.js"[document.querySelector('canvas').width, document.querySelector('canvas').height]"; time_out=100))

File: WGLMakie/src/three_plot.jl (hunk 2 above)
Comment (asinghvi17) [reply WGLMakie/src/three_plot.jl]: Looked like it worked - thanks for the suggestion!  Should be good to go from my end.

**Comment by jkrumbiegel** [comment]
I don't really understand where in the code the `px_per_unit` attribute is picked up by `colorbuffer`, is it possible to pass it to `record` as well, just like you can do it for `save`? That's what I'd want, I mainly see the hidden state as a convenience while plotting, but it should always be directly override-able.

**Comment by asinghvi17** [comment]
Hmm, that's a good idea - might require a bit of refactor but will do that.

The px_per_unit is in this case read from the backend in `backend_show`, and in order to pass it on I'll refactor `backend_display` to pass some kwargs on to `backend_show` through an IOContext.

File: CairoMakie/src/infrastructure.jl [hunk 4]
```diff
@@ -313,7 +315,7 @@ end
 
 function Makie.backend_display(x::CairoBackend, scene::Scene; kw...)
     return open(x.path, "w") do io
-        Makie.backend_show(x, io, to_mime(x), scene)
+        Makie.backend_show(x, IOContext(io, kw), to_mime(x), scene)
```
asinghvi17, line 318:
```suggestion
        Makie.backend_show(x, IOContext(io, (; kw...)), to_mime(x), scene)

File: CairoMakie/src/infrastructure.jl [hunk 5]
```diff
@@ -313,7 +315,7 @@ end
 
 function Makie.backend_display(x::CairoBackend, scene::Scene; kw...)
     return open(x.path, "w") do io
-        Makie.backend_show(x, io, to_mime(x), scene)
+        Makie.backend_show(x, IOContext(io, (; kw...)), to_mime(x), scene)
```
asinghvi17, line 318:
```suggestion
        Makie.backend_show(x, IOContext(io, collect(kw)...), to_mime(x), scene)

File: NEWS.md [hunk 6]
```diff
@@ -12,9 +12,10 @@
   Added single image marker support to WGLMakie [#979](https://github.com/MakieOrg/Makie.jl/pull/979).
 - Allow `CairoMakie` to render `scatter` with images as markers [#2080](https://github.com/MakieOrg/Makie.jl/pull/2080).
 - Reworked text drawing and added ability to draw special characters via glyph indices in order to draw more LaTeX math characters with MathTeXEngine v0.5 [#2139](https://github.com/MakieOrg/Makie.jl/pull/2139).
-- Recording and colorbuffers use screen size instead of scene size, allowing e.g. the `px_per_unit` argument in CairoMakie to take effect.  
-  It is sufficient to set the global `px_per_unit` setting by calling `CairoMakie.activate!(px_per_unit=3)` or whichever scaling factor is desired.  
+- Recording and colorbuffers use screen size instead of scene size, allowing e.g. the `px_per_unit` argument in CairoMakie to take effect.
+  It is sufficient to set the global `px_per_unit` setting by calling `CairoMakie.activate!(px_per_unit=3)` or whichever scaling factor is desired.
```
jkrumbiegel, line 16: Is that a change in behavior? I'm not sure if the `CairoMakie.activate!(...` settings were proliferated to `save` before. Do we want them to or not?

SimonDanisch:
closed in favor of #2306

PR #2017, opened 2022-06-02
add linestyle and default to extrema(z) for contour

SimonDanisch:
So, fillrange has been pretty much ignored and been implemented completely incorrectly via some obscure GLMakie backend feature, which has been recently removed.
Also, contour didn't allow to change the linestyle for no good reason.
This implements fillrange correctly now via contourf.
Changes:
* ~~fillrange is now implemented via `contourf`~~
* linestyle can be changed for `contour`
* `colorrange` for `contour` uses now the extrema of z, and not of `calculated_levels` anymore (previously, overlaying `contour` over `heatmap` would result in differently scaled colors)


jkrumbiegel:
I'm not sure how I feel about the duplication with contourf. I think a natural thing to do is to use `contourf` and then `contour` on top, and the thing we can do to improve that situation is to document better how to achieve correct overlaps of those two plot types. Or if we think that's too difficult for users, remove contourf entirely. I find `fillrange` a very obscure way to plot a contourf to be honest. As it didn't work before anyway, and wasn't really documented, we don't have to bring it back.

I additionally don't like it because it's a non-reversible attribute, so setting it false doesn't remove the contourf after the fact.

SimonDanisch:
Yeah I agree... Just noticed, that we can easily implement it correctly with contourf when I looked at the code, but I think you're right that we shouldn't.
I removed it as an attribute.

PR #2020, opened 2022-06-04
Change `text` signature to `positions, text = text`

jkrumbiegel:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/issues/1468, https://github.com/JuliaPlots/Makie.jl/issues/1138

The `text` function has an unusual signature compared to the other plotting functions, it takes the string(s) as its positional arg and the positions as a keyword arg. That means positions have to be passed in a specific way (usually as Point or Vector{Point}) because the conversion pipeline that usually applies for PointBased does not apply to attributes. This additionally causes problems for AlgebraOfGraphics, as the expected `mapping(x, y, text = text)` doesn't work, instead one has to construct a vector of points manually.

Another side effect is that a text plot initialized with a String cannot be changed to a LaTeXString later, as the signature of `Text` is fixed to `String` at that point and any subsequent `LaTeXString` will be converted to `"$content$"`.

This PR changes the default args for `text` such that it's called with `text(positions, text = strings)`. It has the conversion trait `PointBased` now so all expected signatures for passing x and y work. All the old signatures still exist and reroute to that new definition. The last `Text` in the chain is still the one receiving a vector of `GlyphCollection`s, so the backends didn't need to change.

TODO

- [x] Add new signature with `PointBased` conversion trait
- [x] Reroute old signatures
- [ ] Check that all `Block`s use the new format internally (only Axis so far)
- [ ] Rewrite docs
- [ ] Check that interactive updating still works
- [ ] Line elements from MathTeXEngine are currently broken because they were implemented with a separate LineSegments plot object and it's not clear how that should work when switching back and forth between String and LaTeXString args. An option could be to render lines as stretched line chars instead, which would mean that no separate LineSegments shenanigans would be needed. (cc @Kolaru)

## Type of change

- [x] New feature (non-breaking change which adds functionality) It should hopefully be non-breaking as the old signatures reroute.

## Checklist

### DIFF
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -233,12 +233,12 @@ Plots a mesh for each element in `(x, y, z)`, `(x, y)`, or `positions` (similar 
 end
 
 """
-    text(string)
+    text(positions)
 
 Plots a text.
 
 """
-@recipe(Text, text) do scene
+@recipe(Text, positions) do scene
     Attributes(;
         default_theme(scene)...,
         color = theme(scene, :textcolor),
--- a/MakieCore/src/conversion.jl
+++ b/MakieCore/src/conversion.jl
@@ -19,6 +19,7 @@ convert_arguments(::NoConversion, args...) = args
 
 struct PointBased <: ConversionTrait end
 conversion_trait(::Type{<: XYBased}) = PointBased()
+conversion_trait(::Type{<: Text}) = PointBased()
 
 abstract type SurfaceLike <: ConversionTrait end
 
--- a/src/basic_recipes/text.jl
+++ b/src/basic_recipes/text.jl
@@ -1,6 +1,9 @@
 function plot!(plot::Text)
+    positions = plot[1]
     # attach a function to any text that calculates the glyph layout and stores it
-    glyphcollection = lift(plot[1], plot.textsize, plot.font, plot.align,
+    glyphcollections = Observable(GlyphCollection[])
+    
+    onany(plot.text, plot.textsize, plot.font, plot.align,
             plot.rotation, plot.justification, plot.lineheight, plot.color, 
             plot.strokecolor, plot.strokewidth, plot.word_wrap_width) do str,
                 ts, f, al, rot, jus, lh, col, scol, swi, www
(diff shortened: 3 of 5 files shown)


Kolaru:
> * An option could be to render lines as stretched line chars instead, which would mean that no separate LineSegments shenanigans would be needed. (cc @Kolaru)

Wouldn't a string that mix stretched and unstreched characters cause the same problem (i.e. mixing stuff that were not planned to be mixed)? If not I can think about how to do that properly.

ffreyer:
We have scale, rotation and translation as part of the per character interface so stretching is fine. The bad part is that it's currently using two plot types.

Iirc unicode also has a section of characters that are undefined and meant to be overwritten for stuff like this. I think with how GLMakie works we can even reproduce lines one to one as a character.

jkrumbiegel:
The way it works right now is that a string including stylistic attributes is translated to a GlyphCollection and that is then rendered. So it's in principle fine to put a long underscore character in there instead of the separate line elements you're currently using. That would just simplify the implementation but I can imagine that the long underscore could look blurry at the edges in glmakie, due to the bitmap

ffreyer:
> Iirc unicode also has a section of characters that are undefined and meant to be overwritten for stuff like this. I think with how GLMakie works we can even reproduce lines one to one as a character.

I played around a bit with this now. The characters I was thinking of are private use character: https://www.unicode.org/faq/private_use.html. I guess it's not generally safe to use them like this, but we can (and should anyway) restrict ourselves to a font that doesn't use them.

My basic idea is to manually create signed distance fields for vline and hline and insert them into the texture atlas linked against characters in the private use section. During latex layouting we can then replace VLine and HLine with the corresponding characters, set up the metrics and be more or less done with it on the GLMakie side. 

Without thinking carefully if all my scales make sense I get this:

```julia
scene = Scene()
text!(scene, L"\frac{1}{2}", textsize = 200, align = (:center, :center))
```

Screenshot from 2022-06-06 20-54-48

The bit of fuzz on the left is texture bleed I believe. I think that just needs a pixel of padding to be fixed. The size of the sdf in the atlas is `(2 * GLYPH_PADDING[] + 1)^2`, so as little as possible.

jkrumbiegel:
I tried out just using the underscore character for this, calculating how to stretch it so that it can replace the HLine. Works pretty well and would remove the need for LineSegments. The implementation is very ad-hoc right now because I just wanted to get an example quickly.

grafik

SimonDanisch (approved):

jkrumbiegel:
Yes they didn't render well in GLMakie at small sizes..

PR #2024, opened 2022-06-06
add nightly test to CI

Moelf:


SimonDanisch:
I kind of want to keep CI runs to a minimum... And I also don't want to fix Julia nightly bugs in Makie ;) 
Is there any really important reason besides that to test on nightly, that would justify this?

Moelf:
For one 1.8 is about to release and AFAIC Makie doesn't work on 1.8?

We need CI so Makie is ready day 1 of new releases

SimonDanisch:
>We need CI so Makie is ready day 1 of new releases

Well, it won't be magically ready just by having a CI ;)

SimonDanisch:
Also, I got convinced to use `1` now for CI, which should switch us to 1.8 once it's released.

Moelf:
Yeah but when it is released, Makie will be broken for a long time right?

SimonDanisch:
Well, but I don't think anyone here will debug nightly bugs... Especially since Julia is not supposed to break packages with releases < 2.0, so they almost by definition shouldn't be anything that's fixable in Makie itself.

greimel:
This proposal will not help, because `nightly` would test on Julia 1.9-dev, not 1.8-rc.

What you could do is `"^1.8.0-0"`. This though might be too cumbersome because CI scripts need to be updated regularly.

Hopefully it will become easier to test on pre-releases in the future. See the proposal here https://github.com/julia-actions/setup-julia/issues/84 to add `latest-prerelease` analogous to `nightly` for usage in CI.

jkrumbiegel:
Don't they run tests across the ecosystem anyway to screen for changes that might break packages?

Moelf:
But before 1.8 is frozen, we would have fixed the bugs, once it's frozen, it's not gonna break more stuff usually

Moelf:
Actually if we add 1.8, it would work apparently, using rc, we should at least test rc right?

Moelf:
> Especially since Julia is not supposed to break packages with releases < 2.0

assuming none of Makie or any recursively dependent upon packages use any non-public API

giordano:
> Don't they run tests across the ecosystem anyway to screen for changes that might break packages?

You may be hooking into internals of julia which have been intentionally broken, and no one will tell you (I think they used to, but not anymore).  So you're better off actually running tests on nightly.

jkrumbiegel:
I think this might be more of a psychological thing. We don't want the CI to return failure status for normal PRs if the only reason is that there is a yet unfixed incompatibility with the Julia release candidate. As we can't really communicate well what the exact reason is that something failed and rely for these quick cues while sifting through all the notifications, it would not help us that much. Practically speaking, people try Makie with unreleased versions often enough so that we get a heads up anyway if something fails.

SimonDanisch:
Yeah, what @jkrumbiegel says ;) 
I just don't want to deal with the CI failing on unrelated PRs, just because nightly/pre-release changed something - it's hard enough already to review CI failures with the current volume of Makie PRs.
If someone critically relies on Makie to be immediately ready once 1.8 gets released, they should invest the work and run the tests etc locally.

Moelf:
>We don't want the CI to return failure status for normal PRs if the only reason is that there is a yet unfixed incompatibility with the Julia release candidate

you can still merge it citing "CI 1.8 fails for unrelated reason"

SimonDanisch:
So... what's the point then? Also, that will be shown in the status overview for CI, which I quite heavily rely on, to make sure everything is working.

Moelf:
>So... what's the point then?

so we have a place to see and talk about potential breaking on 1.8???

SimonDanisch:
So an unrelated PR is a good place to talk about why nightly is failing?
I think you don't realize how tough it is right now to review all the PRs, and every little problem with CI is really compounding into making it super hard to get the huge volume of PRs reviewed.
As long as nobody as paying me or @jkrumbiegel or anyone else to deal with this, I'm not going to take on that extra work.

Moelf:
just to find some examples here (sorted by stars):
https://github.com/FluxML/Flux.jl/blob/f86b3562c2f171ceaed61ec3a4d37295b4344d9f/.github/workflows/ci.yml#L22
https://github.com/GenieFramework/Genie.jl/blob/8c3788a773e5d9ca9c63531e7e498b7d534d980e/.github/workflows/ci.yml#L15
https://github.com/jump-dev/JuMP.jl/blob/master/.github/workflows/nightly_ci.yml

also, I am literally paying you LOL

Moelf:
you can use `allow-fail` for nightly if you rely on green tick visual aids btw

PR #2027, opened 2022-06-06
fix minor ticks

SimonDanisch:
https://github.com/JuliaPlots/Makie.jl/pull/1968 broke minor ticks + grids:
image
This should fix it and add a reference image test for it.


jkrumbiegel (approved):

jkrumbiegel:
I still have the feeling that there are double updates in there but I'm not motivated right now to go in again and check. Such a time sink

PR #2035, opened 2022-06-08
fix label shorthands

### LINKED ISSUE #1808: xlabel! and ylabel! do not behave as expected
When using `xlabel!` or `ylabel!` to set axis labels according to the docstring, Makie throws a method error.

```julia
using CairoMakie

lines(0:0.1:2π, sin)
xlabel!("My X Label")
```

``` julia
MethodError: no method matching xlabel!(::String)

Closest candidates are:

xlabel!(::Any, !Matched::AbstractString)
```

In the source, `xlabel!` requires a `scene` positional argument. This conflicts with the docstring which specifies it as optional.

https://github.com/JuliaPlots/Makie.jl/blob/84a026f5a59ee71df50f2486be89da27a41af82d/src/shorthands.jl#L7-L18

From the [documentation](https://makie.juliaplots.org/v0.15.2/documentation/scenes/), it seems like `scene`s where phased out in favor of `Figure`s. Also this [discourse post](https://discourse.julialang.org/t/makie-xlabel/53905) further describes the same issue.

If this is a valid issue, I don't mind creating a draft pull request simply implementing a method for `xlabel!(xlabel::AbstractString)`.

czimm79:
# Description
First pull request, please be kind, I'd like to learn. :)

Fixes #1808 

I fixed the `xlabel!`, `ylabel!` and `zlabel` shorthands as described in #1808. However, tests fail on precompilation `LoadError: UndefVarError: Axis not defined`. This makes sense, but I still would like to dispatch on the `Axis` and `Figure` Makie types. How can I improve this?

## Type of change

## Checklist

### DIFF
--- a/src/shorthands.jl
+++ b/src/shorthands.jl
@@ -5,50 +5,47 @@ function ylims! end
 function zlims! end
 
 """
-    xlabel!([scene,] xlabel)
+    xlabel!([axis,] xlabel)
 
-Set the x-axis label for the given Scene.
-Defaults to using the current Scene.
+Set the x-axis label for the given Axis.
+Defaults to using the current Axis.
 """
-function xlabel!(scene, xlabel::AbstractString)
-    axis = scene[OldAxis]
-    @assert !isnothing(axis) "The Scene does not have an axis!"
-    scene[OldAxis][:names][:axisnames][] = (xlabel, scene[OldAxis][:names][:axisnames][][2:end]...)
-    nothing
+function xlabel!(axis::Axis, xlabel::AbstractString)
+    axis.xlabel = xlabel
 end
 
-"""
-    ylabel!([scene,] ylabel)
+xlabel(figure::Figure, xlabel::AbstractString) = xlabel!(current_axis(figure), xlabel)
+xlabel(xlabel::AbstractString) = xlabel!(current_axis(), xlabel)
 
-Set the y-axis label for the given Scene.
-Defaults to using the current Scene.
 """
-function ylabel!(scene, ylabel::AbstractString)
-    axis = scene[OldAxis]
-    @assert !isnothing(axis) "The Scene does not have an axis!"
-    if axis isa Axis3D
-        scene[OldAxis][:names][:axisnames][] = (scene[OldAxis][:names][:axisnames][][1], ylabel, scene[OldAxis][:names][:axisnames][][3])
-    else
-        @error("Unknown axis type $(typeof(axis)).")
-    end
-    nothing
+    ylabel!([axis,] ylabel)
+
+Set the y-axis label for the given Axis.
+Defaults to using the current Axis.
+"""
+function ylabel!(axis::Axis, ylabel::AbstractString)
+    axis.ylabel = ylabel
 end
 
-"""
-    zlabel!([scene,] zlabel)
+ylabel(figure::Figure, ylabel::AbstractString) = ylabel!(current_axis(figure), ylabel)
+ylabel(ylabel::AbstractString) = ylabel!(current_axis(), ylabel)
 
-Set the z-axis label for the given Scene.
-Defaults to using the current Scene.
+"""
+    zlabel!([axis,] zlabel)
+
+Set the z-axis label for the given Axis.
+Defaults to using the current Axis.
 !!! warning
     The Scene must have an Axis3D.  If not, then this function will error.
 """
-function zlabel!(scene, zlabel::AbstractString)
-    axis = scene[OldAxis]
-    @assert !isnothing(axis) "The Scene does not have an axis!"
-    @assert axis isa Axis3D "The scene does not have a z-axis"
-    scene[OldAxis][:names][:axisnames][] = (scene[OldAxis][:names][:axisnames][][1], scene[OldAxis][:names][:axisnames][][2], zlabel)
+function zlabel!(axis::Axis, zlabel::AbstractString)
+    @assert axis isa Axis3D "The scene does not have a z-axis!"
+    axis.zlabel = zlabel
     return
 end
+
+zlabel(figure::Figure, zlabel::AbstractString) = zlabel!(current_axis(figure), zlabel)
+zlabel(zlabel::AbstractString) = zlabel!(current_axis(), zlabel)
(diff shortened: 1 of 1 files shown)


src/shorthands.jl
```diff
@@ -5,51 +5,48 @@ function ylims! end
...
-    axis = scene[OldAxis]
-    @assert !isnothing(axis) "The Scene does not have an axis!"
-    scene[OldAxis][:names][:axisnames][] = (xlabel, scene[OldAxis][:names][:axisnames][][2:end]...)
-    nothing
+function xlabel!(axis::Axis, xlabel::AbstractString)
+    axis.xlabel = xlabel
 end
 
+xlabel(figure::Figure, xlabel::AbstractString) = xlabel!(current_axis(figure), xlabel)
+xlabel(xlabel::AbstractString) = xlabel!(current_axis(), xlabel)
```
asinghvi17, line 18: IMO these functions should be `xlabel!` also (instead of having `xlabel` and `xlabel!` both).  Same for the y and z equivalents.

src/shorthands.jl
```diff
@@ -5,51 +5,48 @@ function ylims! end
...
-    xlabel!([scene,] xlabel)
+    xlabel!([axis,] xlabel)
 
-Set the x-axis label for the given Scene.
-Defaults to using the current Scene.
+Set the x-axis label for the given Axis.
+Defaults to using the current Axis.
 """
-function xlabel!(scene, xlabel::AbstractString)
-    axis = scene[OldAxis]
```
asinghvi17, line 14: We might want to keep this implementation for people who are still using the old axes.  Perhaps you could have this implementation dispatch on the signature `xlabel!(scene::Scene, xlabel::AbstractString)`?

asinghvi17 (changes requested):
Looks like a good starting point!  I left some comments on the code.

Regarding the precompile issue: this is likely because the `Axis` type is not yet defined in the code.  In order to fix this, you would be best served by defining overloads in each individual layoutable's file.  For example,  you could keep the Scene method in `shorthands.jl`, and move the Axis method to `makielayout/blocks/axis.jl`.

BTW, your dispatch on Axis3 won't work, since Axis3 is not a subtype of Axis.  I would suggest that you define the following signatures instead:

```julia
xlabel!(ax::Union{Axis, Axis3}, xlabel::AbstractString)
ylabel!(ax::Union{Axis, Axis3}, ylabel::AbstractString)
zlabel!(ax::Axis3, zlabel::AbstractString)
```
This way the user just gets a method error.  If you want to be more specific you could define a `zlabel!` method for Axis with a more descriptive error message.


To clarify some more: back in the really old days when we didn't have layouting, there were innate axis types in Makie - Axis2D and Axis3D.  These were actually implemented as plot recipes instead of layoutable `Block`s.

With the coming of layouts in MakieLayout, we have `Axis` (2D) and `Axis3` (3D), both of which are layoutables in their own right which contain their own scenes, et cetera. 

PS - thanks for the PR!

src/shorthands.jl
czimm79: I see - I will keep the scene implementation in `shorthands.jl`. 

I'm a bit confused though - why can I refer to dispatching `::Scene` in `shorthands.jl` but not `::Axis`? Neither is defined in the scope of this file. What am I missing here?

asinghvi17: The order in which the files are loaded also comes into play here - scenes.jl is loaded before shorthands.jl whereas makielayout is loaded after it, IIRC.

jkrumbiegel:
I'm still wondering if we need these functions. They exist because `Scene` didn't have `xlabel` etc attributes, so they needed to use setter functions. However, Axis and Axis3 do have such attributes so having the setters is a duplication of API. However, as long as the methods for Scene exist, users might be confused why they don't work for Axis and Axis3. I don't know the best way around it, but I do feel that it's also confusing if some arbitrary attributes have setter functions that do nothing special and the rest don't.

czimm79:
> I'm still wondering if we need these functions. They exist because `Scene` didn't have `xlabel` etc attributes, so they needed to use setter functions. However, Axis and Axis3 do have such attributes so having the setters is a duplication of API. 

So some options would be:
1. Remove all setter functions that modify accessible axis attributes like `xlims!`, and replace them with the lengthier notation `ax = current_axis(); ax.limits.val= (0, 10, ax.limits[][2])` notation. `xlabel!` would be a bit easier as `ax = current_axis(); ax.xlabel = "string"`.
2. Add setter functions for the majority of the axis attributes to keep the interface constant. This would duplicate some API but would not require user knowledge about the inner attribute structure of Makie. From a non-developer point of view, it is completely arbitrary why `ax.xlabel = ` should work while `ax.xlims = ` shouldn't.

src/shorthands.jl
czimm79: Ah! Makes sense, thanks.

jkrumbiegel:
Yes but xlims and hidedecorations to not have direct correspondences to attributes. That's the main difference, why should all the other attributes not have functions to set them?

czimm79:
I mean its really up to you @jkrumbiegel. If you don't like Makie having multiple ways to accomplish the same task, here's another idea.

1. Scrap this PR. Open a new PR where we:
2. Add a note in the docstring for `xlabel!` that the function is depreciated, and to use `ax.xlabel = ` instead.
3. Throw a warning when a user attempts to use `ax.xlims`, saying to use `xlims!` instead because the attribute does not exist.
4. (optional) Change the plot limits from being held in `ax.limits` to being held in `ax.xlims`, `ax.ylims`, `ax.zlims`. That way `xlims` has a direct correspondence to an attribute.

ffreyer:
> why should all the other attributes not have functions to set them?

I'd say they should, at least the common ones. And I think grouping them like this would be good to avoid having a million setter functions.

```julia
xlabel!(ax[, str][; visible, font, fontsize, color, padding, text = str])
```

jkrumbiegel:
Making more useful setters that do multiple things at once is a different thing I'd say, could be useful.

jkrumbiegel:
That's kind of in the same vein as hidedecorations

PR #2059, opened 2022-06-16
Improve basic tutorial

jkrumbiegel:


jkrumbiegel:
fixes https://github.com/JuliaPlots/Makie.jl/issues/2053

PR #2060, opened 2022-06-16
Better PointTrans interface

asinghvi17:
- Define PointTrans for VecTypes, not only Point
- Add transformation methods for Rects, so that PointTrans just works

The ultimate goal is that `scene.transformation.transform_func[] = PointTrans{2}(f)` where `f(::VecTypes{N})::VecTypes{N}` just works.

Here it's assumed that the method returns the input type, i.e., will return Point on an input of Point, Vec on an input of Vec.

## Type of change

## Checklist


asinghvi17:
We could definitely do that. It's even possible, that transforming lines could be a purely backend operation, and plot objects could be made more or less unaffected by the transform func in base Makie.  This seems consistent with the model matrix approach we use now.

SimonDanisch:
>lines etc could be a purely backend operation

Well, it is right now, since the backends are applying any transformation ... Or what do you mean?

asinghvi17:
I meant densifying / transforming lines, rects, heatmaps, etc.. so that a line with 2 points becomes a curve.

SimonDanisch:
Yeah, so if we find a way to let `apply_transformation` guide this, we could do it quite elegantly in the backend, no?

asinghvi17:
I think so - might want to make it another method in the interface though, and let apply_transform be a low-level method.  I'll whip something up for this in a couple of weeks, but this might be best for a larger pull request.  Does this look OK to merge as is?

SimonDanisch (changes requested):
I'd remove the global, either by:
a) Putting it into PointTrans
b) just have it in `apply_transform` with a TODO

Or do you expect this to be set to different values often enough?

asinghvi17:
I'll put it into `apply_transform` for now, then spin it out into PointTrans later.  That makes sense - a global was the quick and dirty way to allow the transform to be configured but it does make sense to keep it as either a field of PointTrans or as a method (e.g. `Makie.default_transform_density(f::T) where T`).

src/layouting/transformation.jl
```diff
@@ -247,6 +247,63 @@ function apply_transform(f::PointTrans{N1}, point::Point{N2}) where {N1, N2}
...
+This controls the density of the grid within the default implementation of
+`apply_transform(f::PointTrans{2}, rect::Rect{2, T})`.  Set it by:
+```julia
+Makie._DEFAULT_RECT_TRANSFORM_DENSITY[] = new_transform_density::Int
+```
+"""
+const _DEFAULT_RECT_TRANSFORM_DENSITY = Ref{Int}(21)
+
+function apply_transform(f::PointTrans{2}, rect::Rect{2, T}) where T
+    N = _DEFAULT_RECT_TRANSFORM_DENSITY[]
```
asinghvi17, line 267:
```suggestion
function apply_transform(f::PointTrans{2}, rect::Rect{2, T}) where T
    N = 21 # TODO: spin this into PointTrans or out into a method

**Comment by asinghvi17** [comment]
This should be ready for merge now

**Comment by jkrumbiegel** [comment]
Not sure if a method for rects makes sense because the Rect around a set of points in linear space does not necessarily envelop those points in the transformed space right?

File: src/layouting/transformation.jl [hunk 2]
```diff
@@ -242,11 +242,11 @@ end
 # PointTrans{N}(func::F) where {N, F} = PointTrans{N, F}(func)
 Base.broadcastable(x::PointTrans) = (x,)
 
-function apply_transform(f::PointTrans{N}, point::Point{N}) where N
+function apply_transform(f::PointTrans{N}, point::VecTypes{N}) where N
     return f.f(point)
 end
 
-function apply_transform(f::PointTrans{N1}, point::Point{N2}) where {N1, N2}
+function apply_transform(f::PointTrans{N1}, point::VecTypes{N2}) where {N1, N2}
```
SimonDanisch, line 249: Do we actually need this anywhere? Since we've been fighting a lot with compile times and clarity of what types a function accepts, I started restricting the types more whenever possible.
This way, this method is compiled less and anyone calling `apply_transform` is forced to use less types.
Which I think is a good trade off for an internal function as this

src/layouting/transformation.jl
```diff
@@ -257,6 +257,52 @@ function apply_transform(f::PointTrans{N1}, point::Point{N2}) where {N1, N2}
...
+                umaxs[i] = max(umaxs[i], u[i])
+            end
+        end
+    end
+
+    return Rect(Vec2(umins), Vec2(umaxs .- umins))
+end
+
+function apply_transform(f::PointTrans{3}, rect::Rect{3, T}) where T
+    N = _DEFAULT_RECT_TRANSFORM_DENSITY[]
```
SimonDanisch, line 285: I think this global is gone? And we should add tests, since CI is green even though this should error!

src/layouting/transformation.jl
asinghvi17: I believe it gets used for `Arrows` and some other things, there was definitely a reason I had this here although I don't remember exactly what it was now.

src/layouting/transformation.jl
asinghvi17: Oops...did not notice that.  I'll write some tests up tomorrow then.  Thanks for the review!

asinghvi17:
@jkrumbiegel that's why I implemented that method, it basically takes a grid across the Rect (21x21 in this case), transforms each of those points, and finds the extrema, to get the min/max in transformed space.  It's the same method which Proj (geospatial projection library) uses as a general fallback.

src/layouting/transformation.jl
asinghvi17: I think that was the problem - anything with Vec caused a methoderror when it was being transformed.

SimonDanisch:
I think this will need further thoughts about the whole transformation pipeline and a proper refactor!

PR #2069, opened 2022-06-17
Fix version of MakieCore in Makie

greimel:
This is to avoid breakage within the short timespan after the release has been made and all packages are available from the package server.


SimonDanisch:
I think we can close this after the new versions got tagged...

PR #2073, opened 2022-06-18
Update DataInspector to work with text changes

ffreyer:
# Description

Fixes https://discourse.julialang.org/t/problem-with-datainspector/82958

Also makes some changes/additions to `boundingbox(::Text)`:
- `gl_bboxes` now treats '\n' as 0 width and height
- added `boundingbox(::GlyphCollection, rotations)`
- transform text.position[] to pixelspace to avoid mixing different spaces
- added methods that work top level text plots (rather than just the final child version)

## Type of change

### DIFF
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -181,30 +181,6 @@ function ncoords(poly::Polygon)
     N
 end
 
-
-### Text bounding box
-########################################
-
-function Bbox_from_glyphcollection(text, gc)
-    bbox = Rect2f(0, 0, 0, 0)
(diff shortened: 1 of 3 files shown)


ffreyer:
Fixes #2076

garrekstemo:
I also notice a couple of issues with DataInspector in 0.17.8. I also encounter @tecosaur's issue with the inspector and BoundingBox. Also, I am unable to change inspector attributes anymore, like `indicator_color` (it's always red, when two or three updates ago I was able to change it) and `indicator_linewidth`.

Screen Shot 2022-06-23 at 13 05 59

PR #2074, opened 2022-06-19
Menu improvements

jkrumbiegel:
# Description

Mostly fixes https://github.com/JuliaPlots/Makie.jl/issues/2013

Menu was using Label and Box internally which was already quite heavy in terms of the amount of callback functions and plot objects before the Block refactor, but with the Block refactor also the number of Scenes used rose, so much so that you can easily hit the current limit of 255 scenes. In this PR, we just use two texts and two polys, with one additional scene for the list so it can be visually cut off.

As a bonus, the list can also be scrolled now if it's too long, and it's easier to set a default value with the `default` keyword.

## Type of change

## Checklist


tecosaur:
A little feedback on the scrolling: It's great to have, but I think it would be good to scroll more than one pixel at a time :slightly_smiling_face:. Scrolling to a new option currently takes as much scroll wheel motion as 1-2 pages of content in anything else.

jkrumbiegel:
Ok it works fine on my macbook but that probably outputs quite different scrolling values than a mouse. Not sure how to get around that, maybe there's something in glfw that tells us what device we should expect?

tecosaur:
I have no idea on device detection, but could a special case when delta=1 make sense? I'm wondering if a trackpad would tend to emit delta>1 events.

src/makielayout/blocks/menu.jl
```diff
@@ -96,12 +102,136 @@ function initialize_block!(m::Menu)
         end
     end
 
-    selectiontext = Label(scene, selected_text, tellwidth = false, halign = :left,
-        padding = m.textpadding, textsize = m.textsize, color = m.textcolor)
+    selectionarea = Observable(Rect2f(0, 0, 0, 0))
+
+    selectionpoly = poly!(blockscene, selectionarea, color = m.selection_cell_color_inactive[])
```
ffreyer, line 107: The way `DataInspector` is set up it just works with any plot that doesn't explicitly exclude itself. So plot calls in blocks should have `inspectable = false` to be hidden

SimonDanisch: I guess that makes sense, but shouldn't DataInspector be axis specific?

ffreyer: I didn't want it to be coupled to Axis directly when I made it. I'd rather have it work with anything.

Maybe we could make it look for a certain parent scene. I.e. effectively check `parent_scene(hovered_plot) == ax.blockscene.childrenb[1]` after initialising with `DataInspector(ax)`. That would probably need to be a recursion to still allow it to work with any plot after `DataInspector(fig)` though.

SimonDanisch: Well, seems kind of important, because you can't just expect anyone creating a recipe/whatnot foresee if someone might want to inspect it at some point... or not ;)

tecosaur: I have to add, just from an end-user's perspective, if I want to make one plot inspectable it's a bit of a hassle then having to add `inspectable = false` to *every other component of the figure*.

garrekstemo:
The improvements to Menu in the last couple of updates have been really great. I'm seeing much better performance with multiple menus, each with 10+ options, whereas before I could really only have one menu with 6-7 items.

I notice that sometimes a menu does not register a click or the click is registered behind the menu. For example, if the opened menu covers a plot, clicking the menu will cause the plot to zoom. Another example is when the opened menu covers a button, sometimes a click will register on the button instead of the menu selection.

garrekstemo:
Here is an example where a menu click selects the button behind the drop-down. The bug works without fail (both on my Intel MacBook and M1 iMac). If you try to click on an option that covers the button "Press Here", the plot line width will get larger, but the menu item will also be selected.

The issue of zooming in on the plot behind menu is a little harder to reproduce, but when I switch to my mouse (from trackpad) it very often tries to zoom in on the plot when selecting a menu item. Maybe my "clicks" are a little longer on the mouse. I'm not sure what's going on there.

```
fig = Figure()
display(fig)

menu1 = Menu(fig, options = ["A", "B", "C"])
menu2 = Menu(fig[2, 2], options = ["D", "E", "F", "G", "H", "I"])
button = Button(fig, label = "Press Here")

fig[1, 1] = vgrid!(
    menu1,
    button;
    tellheight = false, width = 200
)

ax = Axis(fig[1, 2])

lw = Observable(1)

x = 0:0.3:4π
l = lines!(x, sin.(x), linewidth = lw)

on(button.clicks) do _
    lw[] += 1
end
```

jkrumbiegel:
zooming is a click drag operation so maybe that only happens if you slightly shift the mouse (a pixel would be enough I guess)

PR #2080, opened 2022-06-20
Support for rendering image markers with CairoMakie

fatteneder:
# Description

Overload `CairoMakie.draw_marker` for `Matrix{<:Colorant}` types which allows to render images as markers.

## Type of change

## Checklist


fatteneder:
Tests are failing because of a failure in the conversion pipeline for the markers.

I retraced it to this PR https://github.com/JuliaPlots/Makie.jl/pull/1981 which introduced the `_marker_convert` method that is now erroring:
```julia
# an array of markers is converted to string by itself, which is inconvenient for the iteration logic
_marker_convert(markers::AbstractArray) = map(m -> convert_attribute(m, key"marker"(), key"scatter"()), markers)
_marker_convert(marker) = convert_attribute(marker, key"marker"(), key"scatter"())
```

The issue is that when scattering with image markers you pass a marker of type `Matrix{<:Colorant} <: AbstractArray` for which `_marker_convert` tries to convert all elements as markers. I think adjusting the overload here should do the trick.

CairoMakie/src/primitives.jl
```diff
@@ -204,10 +204,13 @@ end
 # an array of markers is converted to string by itself, which is inconvenient for the iteration logic
 _marker_convert(markers::AbstractArray) = map(m -> convert_attribute(m, key"marker"(), key"scatter"()), markers)
 _marker_convert(marker) = convert_attribute(marker, key"marker"(), key"scatter"())
+# image arrays need to be converted as a whole
+_marker_convert(marker::AbstractArray{<:Colorant}) = convert_attribute(marker, key"marker"(), key"scatter"())
```
asinghvi17, line 208: Maybe make this AbstractMatrix?  Might be too general like this.

CairoMakie/src/primitives.jl
```diff
@@ -204,10 +204,13 @@ end
...
 
 function draw_atomic_scatter(scene, ctx, transfunc, colors, markersize, strokecolor, strokewidth, marker, marker_offset, rotations, model, positions, size_model, font, markerspace, space)
     broadcast_foreach(positions, colors, markersize, strokecolor,
-        strokewidth, marker, marker_offset, remove_billboard(rotations)) do point, col,
+            strokewidth, marker isa AbstractArray{<:Colorant} ? [marker] : marker,
```
asinghvi17, line 212: This should probably be AbstractMatrix.  Also, it might be a little cleaner if we set this outside the `broadcast_foreach` IMO.

asinghvi17 (approved):
Looks good, thanks!  Couple of minor changes requested.

CairoMakie/src/primitives.jl
fatteneder: Indeed, thanks for spotting.

asinghvi17:
Oops...I guess this breaks WGLMakie.  Maybe we could exclude this test from there?

CairoMakie/src/primitives.jl
fatteneder: This doesn't work, because if `marker::AbstractMatrix{<:Colorant}` then it will again iterate over each pixel. I have a better fix for this, pushing it now ...

fatteneder:
Can't we fix this right now? Although I would not know how :sweat_smile:

SimonDanisch:
Is this really just for PDF, or for CairoMakie? Looks to me like the latter, no?

fatteneder:
Aeeehhh, could be :sweat_smile:. I only tested with PDFs. I guess @asinghvi17's :+1: to your answer indicates that it also works for `CairoMakie`. I'll update the news then.

Honestly, I am don't know yet which parts of Makie are responsible for saving scenes to files? And what file formats it can handle?
So far I only used CairoMakie to generate PDFs. Does it also work with SVGs and PNGs? Perhaps I should read the docs more carefully :laughing:

asinghvi17:
Basically, you can think of it this way - Makie specifies the points to render, and the backend packages actually render them.

For example, a complex recipe will be degraded to atomic plots (lines, scatter, heatmap/image, surface, mesh, meshscatter, volume) by Makie.  The backend's basic task is to render these atomic plots in the correct positions.  Of course, you can then hook into higher level recipes to render those directly if you want, as CairoMakie does with poly.  GLMakie renders only atomic level plots to OpenGL, WGLMakie renders them to browser.

asinghvi17:
GLMakie CI actually passes, but since this post is from an external PR I guess the bot can't comment that there is a missing refimage.  Otherwise this is good to go.

asinghvi17 (commented):
Looks like there are some bugs here - shouldn't be too difficult to fix though.

Left is reference image, right is recorded image - this is from the `Image scatter different sizes` example.
Screen Shot 2022-07-12 at 6 25 59 PM

The scale of the bottom marker looks weird and the top marker seems (a) misaligned and (b) somehow more compressed?

Screen Shot 2022-07-12 at 6 25 37 PM

More detail on the top marker weirdness

fatteneder:
I pushed a fix for the wrong scaling, at least now the Makie logo looks ok (the problem were swapped dimensions due to the `permutedims` call).

However, the logo and the doggy now both appear smaller than in the ref image (as you already noted above).
I don't know how to fix that given only the size of the picture and the scale factor.
Adding an artificial factor of 3/2 to the scaling (and offset) would give roughly the right sizes.
Any idea where this difference could origin from?

CairoMakie/src/primitives.jl
```diff
@@ -315,6 +317,24 @@ function draw_marker(ctx, marker::Rect, pos, scale, strokecolor, strokewidth, ma
...
+    # convert marker to Cairo compatible image data
+    argb32_marker = convert.(ARGB32, marker)
+    argb32_marker = permutedims(argb32_marker, (2,1)) # swap x-y for Cairo
+    marker_surf   = Cairo.CairoImageSurface(argb32_marker)
+
+    Cairo.translate(ctx, pos[1]+marker_offset[1], pos[2]+marker_offset[2])
+    Cairo.rotate(ctx, to_2d_rotation(rotation))
+    px_scale = scale ./ size(argb32_marker)
+    Cairo.scale(ctx, px_scale[1], px_scale[2])
+    Cairo.set_source_surface(ctx, marker_surf, 0, 0)
```
asinghvi17, line 333: I think you might want to set something like:
```julia
    p = Cairo.get_source(ctx)
    Cairo.pattern_set_extend(p, Cairo.EXTEND_NONE)
    filt =  Cairo.FILTER_BILINEAR # or Cairo.FILTER_NEAREST, whichever seems best
    Cairo.pattern_set_filter(p, filt)
```

in order to get the sharpest image.

asinghvi17:
> Any idea where this difference could origin from?

Could you try rendering with different `px_per_unit`s and see what happens?  My guess is that somehow you are drawing in device space, not Cairo's user space.  Especially since the reference images seem to be the same.

CairoMakie/src/primitives.jl
fatteneder: Just tried this but it did not have any effect.
I also tried with adding
```julia
Cairo.fill(ctx)
Cairo.restore(ctx)
```
because this is also present in `draw_atomic` for `Heatmap, Image` where your code snippet is also used, but with no effect.

Also noticed that the doggy now has weirdly colored boundaries around its face: 
image

Is that supposed to be fixed by the filter? Or is this a different issue?

fatteneder:
> Could you try rendering with different `px_per_unit`s and see what happens?

I already tried with different scaling. E.g. adding an artificial factor of 3/2 gives roughly the same size as the ref image.


> My guess is that somehow you are drawing in device space, not Cairo's user space. Especially since the reference images seem to be the same.

Perhaps it is the difference between `markerspace = :data, :pixel` and the test needs to be updated?
Using `markerspace=:data` seems to work fine in terms of sizing and positioning:
```julia
using FileIO
using CairoMakie
CairoMakie.activate!()
using Makie

img = Makie.logo()
img2 = load(Makie.assetpath("doge.png"))
images = [img, img2]
# markersize = map(img-> Vec2f(reverse(size(img) ./ 10)), images)
markersize = [ ( 1.0,0.5 ), ( 1.0,1.0 ) ]
s = scatter(1:2, [0.75,2], marker = images, markersize=markersize, axis=(limits=(0.5
            markerspace=:data)

save("scatter.pdf", s)
```

image

fatteneder:
> 

    Could you try rendering with different px_per_units and see what happens?

I already tried with different scaling. E.g. adding an artificial factor of 3/2 gives roughly the same size as the ref image.

Sorry, did not notice that there is an actual `px_per_unit` attribute for `save`.
Changing it does not change anything in the output though.

CairoMakie/src/primitives.jl
fatteneder: Neverminde, the weird colors are due to me saving the figure as `.pdf`.
The edges disappear with `.png`.

asinghvi17: Huh...i guess those are there because of (a) weird rendering or (b) color interpolation.  Does the same thing happen if you open it in Google Chrome?

fatteneder:
I spent more time on testing and finally compared the `png` outputs generate by `GLMakie` and `CairoMakie`:
```julia
using FileIO
using Makie

function test_scatter(filename)
  img = Makie.logo()
  img2 = load(Makie.assetpath("doge.png"))
  images = [img, img2]
  markersize = map(img-> Vec2f(reverse(size(img) ./ 10)), images)
  s = scatter(1:2, 1:2, marker = images, markersize=markersize, axis=(limits=(0.5, 2.5, 0.5, 2.5),))
  save(joinpath(@__DIR__, filename), s)
end

using CairoMakie
CairoMakie.activate!()
test_scatter("cairo_scatter.png")

using GLMakie
GLMakie.activate!()
test_scatter("gl_scatter.png")

```

- `GLMakie` output: 
gl_scatter

- `Cairo` output:
cairo_scatter

Booth pictures are identically, but they differ from the test output of `test CairoMakie`:
Image Scatter different sizes

I guess the difference is in different settings in the test, but I could not figure out what those are. Where can I find them?

But since the tests now pass I guess the sizing is not the issue anymore. Only the colored edges in `pdf` output remain.

CairoMakie/src/primitives.jl
fatteneder: It appears that the colored edges issue is not specific to the `scatter` implementation, since it also occurs when using an `image`.

I think we have to track this in a separate issue.

fatteneder: Adding `fxaa=true` does not help either.

fatteneder:
I opened an issue for the colored edges problem since it also appears when saving it as `pdf` using an `image` plot.

This PR should now be good to go.

SimonDanisch:
Thank you all! :)

SimonDanisch:
Just noticed, that the added test never worked in CairoMakie:
image-scatter
Was this working at some point in this PR?

fatteneder:
Might be that I forgot to check the `CairoMakie` result, my bad.
I will take a look.

PR #2094, opened 2022-06-24
Menu cleanup (DataInspector, events)

ffreyer:
# Description

Fixes [DataInspector triggering on Menu](https://github.com/JuliaPlots/Makie.jl/pull/2073#issuecomment-1163169654) and [other elements below Menu triggering](https://github.com/JuliaPlots/Makie.jl/pull/2074#issuecomment-1164332288)

## Type of change


jkrumbiegel (approved):
There's one commented out return statement left in there that should be removed before merging

PR #2096, opened 2022-06-25
Cleanup stroke artifacts in scatter and text

ffreyer:
# Description

Using `strokwidth > 0` in GLMakie with a non transparent (and non-background) `color` creates a thin outline around characters or makers in text and scatter.

With text, white stroke and glow:
Screenshot from 2022-06-25 16-19-00
With scatter, stroke only:
Screenshot from 2022-06-25 16-23-57

The reason for this is that the blending uses a transparent version of `color` as the blend target for stroke. So the rgb component of the `strokecolor` and the `color` get mixed at the edge, resulting in the artifacts seen above. The pr changes the reference color to be a transparent version of `strokecolor` at the outer edge, fixing this problem.

After:

Screenshot from 2022-06-25 16-19-13
Screenshot from 2022-06-25 16-24-10

## Type of change

## Checklist


asinghvi17:
Looks good to me, but could you add the example you posted in the PR as a test for this situation?

ffreyer:
I made a new version of the marker vs rotation test with stroke and glow. I also noticed that this sometimes creates some at the edges: (y, o, ...)
Screenshot from 2022-07-02 01-11-15

The reason for this is texture bleed. 

Characters are encoded with signed distance fields, so the values in the texture represent the distance from the edge of a character, with positive numbers being outside. Stroke and glow basically draw between 0 distance and strokewidth/glowwidth. If opengl ends up interpolating between a positive value and a smaller value at the edge of the glyph texture, that region can become relevant for stroke and glow again. And that is what happens here because the texture is initialized to 0. Changing that to `floatmax(Float16)` fixes this issue, though at large strokewidths/glowwidths it results in a sharp cutoff.

Screenshot from 2022-07-02 01-30-01

ffreyer:
I reduced the background value to `0.5PIXELSIZE_IN_ATLAS[] + GLYPH_PADDING[]` because that should be the maximum reachable distance from the center of a glyph/character. That makes the cutoff problem a bit less of an issue

Is there a way to force the texture atlas to get regenerated with the next tagged version?

ffreyer:
I also tried an over operator for blending now, doesn't seem better to me
glmakie_stroke_over

PR #2097, opened 2022-06-27
interpolate between values instead of interpolating between colors

SimonDanisch:
We've been interpolating between colors in the mesh shader...
TODO: We also still interpolate between colors in cairo at least for mesh.
Bug report: https://discourse.julialang.org/t/how-to-create-a-patch-shape-with-a-color-gradient-in-a-simple-2d-plot/62607/7?u=sdanisch

Before:
```julia

using GLMakie
vertices = Point2[
    (0.0, -1.0)
    (1.0, 0.0)
    (0.6, 1.0)
]
faces = [1, 2, 3]
colors = [1, 2, 3]
f, ax, p = mesh(vertices, faces, color = colors, shading = false, colormap = Reverse(:viridis))
Colorbar(f[1, 2], p)
f
```
image

With this PR:
image

### DIFF
--- a/GLMakie/assets/shader/mesh.frag
+++ b/GLMakie/assets/shader/mesh.frag
@@ -25,9 +25,6 @@ flat in uvec2 o_id;
 vec4 get_color(Nothing image, vec2 uv, Nothing color_norm, Nothing color_map, Nothing matcap){
     return o_color;
 }
-vec4 get_color(Nothing color, vec2 uv, vec2 color_norm, sampler1D color_map, Nothing matcap){
-    return o_color;
-}
 
 vec4 get_color(sampler2D color, vec2 uv, Nothing color_norm, Nothing color_map, Nothing matcap){
     return texture(color, uv);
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Cairo can't do any other interpolation than between color. So to really support it we would need to software-render our own meshes to bitmaps directly.

SimonDanisch:
Yeah, I've been wanting to do that... It's not even that hard, just need to find some time for it!

asinghvi17:
Hmm...wouldn't that essentially lose the "vector" nature of the mesh (i.e., zoomability without pixelation)?  Or are you proposing subdividing the mesh and interpolating along color gradients of smaller magnitude?  Either way, it would probably be nice to be able to turn it on and off by some switch in CairoMakie.

jkrumbiegel:
Mesh is already rasterized (at least in svgs). Could be that it's not in pdfs, not sure, but the mesh operator hasnt been added to the svg spec yet.

NEWS.md
```diff
@@ -8,6 +8,7 @@ role as `datalimits` in `violin` [#2137](https://github.com/JuliaPlots/Makie.jl/
 - Improve performance around Axis(3) limits [#2115](https://github.com/JuliaPlots/Makie.jl/pull/2115).
 - Cleanup stroke artifacts in scatter and text [#2096](https://github.com/JuliaPlots/Makie.jl/pull/2096).
 - compile time improvements [#2153](https://github.com/JuliaPlots/Makie.jl/pull/2153).
+- interpolate between values instead of interpolating between colors for WGLMakie + GLMakie [#2097](https://github.com/JuliaPlots/Makie.jl/pull/2097)
```
jkrumbiegel, line 11: Capitalization here and above :)
It's not that good to paste the PR titles as those are usually not even complete sentences.

PR #2105, opened 2022-07-02
Mirrored Axis ticks

jkrumbiegel:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/issues/2091

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -369,17 +369,17 @@ function initialize_block!(ax::Axis; palette = nothing)
         end
     end
 
-    _xticksmirrored = lift(mirror_ticks, xaxis.tickpositions, ax.xticksize, ax.xtickalign, Ref(scene.px_area), :x, ax.xaxisposition[])
-    linesegments!(topscene, _xticksmirrored, visible = ax.xticksmirrored,
+    xticksmirrored = lift(mirror_ticks, xaxis.tickpositions, ax.xticksize, ax.xtickalign, Ref(scene.px_area), :x, ax.xaxisposition[])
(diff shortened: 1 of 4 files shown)


Moelf (approved):

jkrumbiegel:
@SimonDanisch seems to be slightly slower, but I guess that's expected because it's doing more? I'll merge if you approve

src/makielayout/blocks/axis.jl
```diff
@@ -369,6 +369,19 @@ function initialize_block!(ax::Axis; palette = nothing)
         end
     end
 
+    _xticksmirrored = lift(mirror_ticks, xaxis.tickpositions, ax.xticksize, ax.xtickalign, Ref(scene.px_area), :x, ax.xaxisposition[])
```
SimonDanisch, line 372: could we try to make this one linesegment call, by just creating one point array for the linesegments?
Also, whats the meaning of `_` as a prefix in this context?

SimonDanisch: I'd like to avoid using `_` for scenarios where no better name was found... But it's just a nitpick, not a deal breaker ^^

SimonDanisch (approved):

src/makielayout/blocks/axis.jl
jkrumbiegel: I can change it now, I only used keyword args when developing to avoid a struct redefinition.

Moelf (approved):

src/makielayout/blocks/axis.jl
jkrumbiegel: I've changed the underscore names now, the combining of linesegments would be nice but I'm not motivated right now

PR #2106, opened 2022-07-02
support `Makie.AbstractPattern` as color in `CairoMakie` for non-mesh plots

### LINKED ISSUE #1385: Feature request: Patterned bar plots in CairoMakie
Following up on the discussion from slack, it would be great if similar functionality to this example in GLMakie from [100daysOfMakie](https://github.com/lazarusA/100daysOfMakie#31-stripped-patterns) could be brought to CairoMakie as well

```julia
using Makie, GLMakie, Random
Random.seed!(3)
# patterns
# `'/'`, `'\\'`, `'-'`, `'|'`, `'x'`, and `'+'`
directions = [Vec2f0(1), Vec2f0(1, -1), Vec2f0(1, 0), Vec2f0(0, 1),
    [Vec2f0(1), Vec2f0(1, -1)], [Vec2f0(1, 0), Vec2f0(0, 1)]]
colors = [:white, :red, (:green, 0.5), :white, (:navy, 0.85),:black]
patterns = [Makie.LinePattern(direction= hatch; width = 5, tilesize=(20,20),
    linecolor = colors[indx], background_color = colors[end-indx+1])
    for (indx, hatch) in enumerate(directions)]
fig = Figure(resolution = (1200,800), fontsize = 32)
ax = Axis(fig[1,1])
for (idx, pattern) in enumerate(patterns)
    barplot!(ax, [idx], [idx*(2rand()+1)], color = pattern, strokewidth = 2)
end
ax.xticks = (1:6, ["/", "\\", "-", "|", "x", "+"])
save("strippedPatterns.png", fig)
fig
```

strippedPatterns

Moelf:
## Demo
```julia
julia> p = Pattern('/');

julia> save("/tmp/f.png", poly(Point2f[(0, 0), (2, 0), (3, 1), (1, 1)], color = p, strokecolor=:red, strokewidth=2))

```
```julia
using CairoMakie, Random
Random.seed!(3)
const Vec2f0 = CairoMakie.Vec2f0
# patterns
# `'/'`, `'\\'`, `'-'`, `'|'`, `'x'`, and `'+'`
directions = [Vec2f0(1), Vec2f0(1, -1), Vec2f0(1, 0), Vec2f0(0, 1),
    [Vec2f0(1), Vec2f0(1, -1)], [Vec2f0(1, 0), Vec2f0(0, 1)]]
colors = [:white, :red, (:green, 0.5), :white, (:navy, 0.85),:black]
patterns = [Makie.LinePattern(direction= hatch; width = 5, tilesize=(20,20),
    linecolor = colors[indx], background_colo
r = colors[end-indx+1])
    for (indx, hatch) in enumerate(directions)]
fig = Figure(resolution = (1200,800), fontsize = 32)

ax = Axis(fig[1,1])
for (idx, pattern) in enumerate(patterns)
    barplot!(ax, [idx], [idx*(2rand()+1)], color = pattern, strokewidth = 2)
end
ax.xticks = (1:6, ["/", "\\", "-", "|", "x", "+"])
save("/tmp/strippedPatterns.png", fig)
```

| first | second |
| -----  | ----- |
| image | strippedPatterns |

# Description

fix  #1385

## Type of change

## Checklist

- [x] Added or changed relevant sections in the documentation [NO NEED]
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc. [ALREADY PRESENT]

### DIFF
--- a/CairoMakie/src/overrides.jl
+++ b/CairoMakie/src/overrides.jl
@@ -35,7 +35,9 @@ function draw_poly(scene::Scene, screen::CairoScreen, poly, points::Vector{<:Poi
     draw_poly(scene, screen, poly, points, poly.color[], poly.model[], poly.strokecolor[], poly.strokewidth[])
 end
 
-function draw_poly(scene::Scene, screen::CairoScreen, poly, points::Vector{<:Point2}, color, model, strokecolor, strokewidth)
+# when color is a Makie.AbstractPattern, we don't need to go to Mesh
+function draw_poly(scene::Scene, screen::CairoScreen, poly, points::Vector{<:Point2}, color::Union{Symbol, Colorant, Makie.AbstractPattern},
+        model, strokecolor, strokewidth)
     space = to_value(get(poly, :space, :data))
     points = project_position.(Ref(scene), space, points, Ref(model))
     Cairo.move_to(screen.context, points[1]...)
@@ -43,7 +45,14 @@ function draw_poly(scene::Scene, screen::CairoScreen, poly, points::Vector{<:Poi
         Cairo.line_to(screen.context, p...)
     end
     Cairo.close_path(screen.context)
-    Cairo.set_source_rgba(screen.context, rgbatuple(to_color(color))...)
+    if color isa Makie.AbstractPattern
+        cairopattern = Cairo.CairoPattern(color)
+        Cairo.pattern_set_extend(cairopattern, Cairo.EXTEND_REPEAT);
+        Cairo.set_source(screen.context, cairopattern)
+    else
+        Cairo.set_source_rgba(screen.context, rgbatuple(to_color(color))...)
+    end
+
     Cairo.fill_preserve(screen.context)
     Cairo.set_source_rgba(screen.context, rgbatuple(to_color(strokecolor))...)
     Cairo.set_line_width(screen.context, strokewidth)
@@ -72,6 +81,9 @@ function draw_poly(scene::Scene, screen::CairoScreen, poly, rects::Vector{<:Rect
     elseif color isa String
         # string is erroneously broadcasted as chars otherwise
         color = to_color(color)
+    elseif color isa Makie.AbstractPattern
+        cairopattern = Cairo.CairoPattern(color)
+        Cairo.pattern_set_extend(cairopattern, Cairo.EXTEND_REPEAT);
     end
     strokecolor = poly.strokecolor[]
     if strokecolor isa AbstractArray{<:Number}
@@ -83,7 +95,11 @@ function draw_poly(scene::Scene, screen::CairoScreen, poly, rects::Vector{<:Rect
 
     broadcast_foreach(projected_rects, color, strokecolor, poly.strokewidth[]) do r, c, sc, sw
         Cairo.rectangle(screen.context, origin(r)..., widths(r)...)
-        Cairo.set_source_rgba(screen.context, rgbatuple(to_color(c))...)
+        if c isa Makie.AbstractPattern
+            Cairo.set_source(screen.context, cairopattern)
+        else
+            Cairo.set_source_rgba(screen.context, rgbatuple(to_color(c))...)
+        end
         Cairo.fill_preserve(screen.context)
         Cairo.set_source_rgba(screen.context, rgbatuple(to_color(sc))...)
         Cairo.set_line_width(screen.context, sw)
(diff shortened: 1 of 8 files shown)


CairoMakie/src/utils.jl
```diff
@@ -258,3 +261,12 @@ end
 function mesh_pattern_set_corner_color(pattern, id, c::Colorant)
     Cairo.mesh_pattern_set_corner_color_rgba(pattern, id, rgbatuple(c)...)
 end
+
+# not piracy
+function Cairo.CairoPattern(color::Makie.AbstractPattern)
+    # the Cairo left and right are fliped
```
asinghvi17, line 267: in this case it is the y coordinate which is flipped, so top and bottom

CairoMakie/src/overrides.jl
```diff
@@ -35,15 +35,24 @@ function draw_poly(scene::Scene, screen::CairoScreen, poly, points::Vector{<:Poi
...
     points = project_position.(Ref(scene), space, points, Ref(model))
     Cairo.move_to(screen.context, points[1]...)
     for p in points[2:end]
         Cairo.line_to(screen.context, p...)
     end
     Cairo.close_path(screen.context)
-    Cairo.set_source_rgba(screen.context, rgbatuple(to_color(color))...)
+    if color isa Makie.AbstractPattern
+        cairopattern = Cairo.CairoPattern(color)
+        Cairo.pattern_set_extend(cairopattern, Cairo.EXTEND_REPEAT);
```
asinghvi17, line 50: You should also set the filter so that Cairo doesn't interpolate...
```suggestion
        Cairo.pattern_set_extend(cairopattern, Cairo.EXTEND_REPEAT);
        Cairo.pattern_set_filter(cairopattern, Cairo.FILTER_NEAREST); # or something like this anyway

File: NEWS.md [hunk 3]
```diff
@@ -2,8 +2,10 @@
 
 ## master
 
+- `Makie.AbstractPattern` is now supported by `CairoMakie` in poly plot that doesn't involve mesh, such as `bar, poly` [#2106](https://github.com/JuliaPlots/Makie.jl/pull/2106/).
```
asinghvi17, line 5:
```suggestion
- Patterns (`Makie.AbstractPattern`) are now supported by `CairoMakie` in `poly` plots that don't involve `mesh`, such as `bar` and `poly` [#2106](https://github.com/JuliaPlots/Makie.jl/pull/2106/).

File: CairoMakie/src/utils.jl [hunk 4]
```diff
@@ -258,3 +261,12 @@ end
 function mesh_pattern_set_corner_color(pattern, id, c::Colorant)
     Cairo.mesh_pattern_set_corner_color_rgba(pattern, id, rgbatuple(c)...)
 end
+
+# not piracy
+function Cairo.CairoPattern(color::Makie.AbstractPattern)
```
asinghvi17, line 266: This should probably dispatch on ImagePattern instead, since if we add a specifically vector pattern it may be too broad.

asinghvi17 (commented):
I will try this out soon but it looks good so far!  Thanks for the PR :)

CairoMakie/src/overrides.jl
jkrumbiegel: not sure, why no interpolation? should make diagonal line patterns look better no

CairoMakie/src/utils.jl
Moelf: we can't
```julia
julia> p = Pattern('/');

julia> typeof(p)
Makie.LinePattern

julia> typeof(p) |> supertype
Makie.AbstractPattern{ColorTypes.RGBA{Float32}}
```

jkrumbiegel: if we add the other thing we can add a dispatch site for that, I don't think it matters much here

Moelf: the logic right now is we always convert it to a image because that's what `CairoImageSurface` consumes

CairoMakie/src/overrides.jl
asinghvi17: Makes sense - was looking at it from the perspective of ImagePattern, where you may not want the pattern to be interpolated.  Maybe we could make this user-configurable somehow, perhaps by checking the `interpolate` keyword...

asinghvi17 (approved):
Makes sense to me, and looks good so far!

This looks mergeable to me.  

At a later point I would like to have a vectorized version of LinePattern in CairoMakie, but that sounds like something for a different PR!

As for the Cairo recording surface: basically, one can draw on it using `cairo_move_to` or `cairo_line_to`, and then "replay" it onto a surface.  It would be something like drawing lines across the surface in a pattern and the rest of the code would likely remain the same.

PR #2115, opened 2022-07-05
Improve performance around Axis(3) limits

jkrumbiegel:
Seems like performance slows quite a bit if many plots are added in sequence to an Axis. Because each plot triggers a `resize_limits!`, first of all the limits have to be recomputed for all previous plots when a new one is added, and this also triggers tick changes, which trigger layout changes, etc. So all in all this can be quite slow. This PR adds an option to disable the `reset_limits!` call, which should mainly be interesting for workflows where all plots are added in one go, without a person looking at the progress. So for example this should probably be used in AlgebraOfGraphics @piever .

Also, I noticed dynamic dispatch in the code path that computes the limits in the first place and tried to fix that. I'm not sure if the logic is perfectly right with regards to NaNs, that behavior can be a bit tricky to pin down. For example, what if there's a NaN only in one dimension of a point, do we skip that whole point or use the remaining components. This PR does the latter.

Here's some examples how the PR improves performance:

## Many plots in sequence
Before: 4.999897 seconds (206.07 M allocations: 6.235 GiB, 14.02% gc time)
After: 0.255500 seconds (2.88 M allocations: 151.533 MiB)
```julia
f = Figure()
ax = Axis(f[1, 1])
ax.reset_limits_on_plot = false # new feature
@time for i in 1:500
    lines!(ax, randn(1000) .+ 5i)
end
ax.reset_limits_on_plot = true # new feature
reset_limits!(ax)
f
```

## Single plot
Before:  0.176131 seconds (10.00 M allocations: 305.179 MiB)
After: 0.012864 seconds (92 allocations: 4.203 KiB)
```julia
f = Figure()
ax = Axis(f[1, 1])
lines!(cumsum(randn(1_000_000)))
@time reset_limits!(ax)
f
```

## Type of change

## Checklist


SimonDanisch:
I still think we should only update limits on display, before we add the workaround of disabling limit calculation to all kind of 3rd party libraries... This could be quite the easy change, no?

jkrumbiegel:
How does the Axis know it's being displayed?

SimonDanisch:
Adding something like
```julia
update_state_before_display!(figlike) # or some better name
```
Here: https://github.com/JuliaPlots/Makie.jl/blob/master/src/display.jl#L62=
And overload that function appropriately for figure, scene, axis etc, so that it gets called for all axes in a figure.

Moelf:
I think compute on display is reasonable, `display` is already non-pure (I/O).

One problem I can think of is: do people use reflection (to read axis limits) before display in order to do something? I hope the answer is no

jkrumbiegel:
In that case they could always call `reset_limits!` themselves I guess?

ffreyer:
Wouldn't update on display be an issue when plotting in the repl? I.e. starting with `fig = Figure()` in the REPL you'd display immediately and need another display or `reset_limit!` at the end to update limits.

Another option to throttle these kind of updates would be to generalize `screen.render_tick` from GLMakie to all backends and use it as a trigger. That would reduce things to one update per frame but it would require some bookkeeping to decide what needs to be updates.

jkrumbiegel:
Could we do it such that in the `plot!` call we check if any parent `Scene` of the `Axis` is currently displayed? If yes, we reset limits. If not, we don't (most programmatic cases).

And additionally the `update_state_before_display!` thing that Simon suggested.

jkrumbiegel:
I've changed the implementation to `update_state_before_display!` and reset if the Axis scene or any of its parents are open.

However, there's still the problem of all the functions that can save a scene. `save`, `record`, `colorbuffer`, `VideoStream`, are there more? Do all of them need a separate `update_state_before_display!` inserted?

SimonDanisch:
Most of them should use backend_display / backend_show under the hood... I can take a look if you want

jkrumbiegel:
Problem is that that's already Scene, and we don't have the back link to Axis

jkrumbiegel:
Tests pass now, question is if all the places to call `update_state_before_display!` make sense.

SimonDanisch (approved):

SimonDanisch:
Guess it would be good to check the docs before merging this ;)

PR #2120, opened 2022-07-08
expand rectzoom to y or x if rectangle is very narrow

jkrumbiegel:
# Description

Changes the rectangle zoom behavior in Axis such that at very narrow rectangle selections, the narrow side is expanded to full width. This makes it much easier to narrow the field of view in one direction and is a common behavior in comparable tools.

## Type of change

- [x] New feature (non-breaking change which adds functionality.


jkrumbiegel:
Anyone with a non-retina screen wants to test whether the threshold of 25 px feels ok on their monitor? @ffreyer or @SimonDanisch? As long as we can't rescale for hidpi monitors I can't really test user experience for normal sizes :)

ffreyer:
I don't like this as an always-on feature. I've had cases where I want to zoom into narrow regions, which would get more annoying with this. (For example a large data set where I'm interested in convergence on a large scale and variance/periodicity on a small scale. Or data that falls into groups, where the difference between groups is much larger than within.)

jkrumbiegel:
Will close for now, might return later.

PR #2139, opened 2022-07-14
jk/glyph indices

jkrumbiegel:
- switch from chars to Culong glyph indices
- save glyph indices in glyphcollection
- draw glyphs by index in cairomakie
- remove new path
- add stub for MathTeXEngine


Kolaru:
@jkrumbiegel It doesn't seem to work, after a quick (and terribly hacky) test. Despite passing the correct font and the correct index in the stub you provide, nothing appear when I use the `text` function.

I tested it by using glyph index 5930 in the NewCMMath-Regular font (from the NewComputerModern family), which should contain an integral symbol, as confirmed by http://fontdrop.info :
image

Unfortunately I couldn't quite understand how to select font in Cairo directly, so I couldn't test if it works in pure Cairo.jl. I'll have another try tomorrow.

CairoMakie/src/primitives.jl
```diff
@@ -453,6 +449,25 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
...
+    index::Culong
+    x::Cdouble
+    y::Cdouble
+end
+
+function show_glyph(ctx, glyph::Culong, x, y)
+    cg = Ref(CairoGlyph(glyph, x, y))
+    ccall((:cairo_show_glyphs, Cairo.libcairo),
+            Nothing, (Ptr{Nothing}, Ptr{CairoGlyph}, Cint),
+            ctx.ptr, cg, 1)
```
asinghvi17, line 462: In this case you should use [cairo_show_text_glyphs](https://www.cairographics.org/manual/cairo-text.html#cairo-show-text-glyphs), which allow text to be highlighted.  We had glyph support earlier but AFAIK this is why we switched away from it.

jkrumbiegel: At this point I don't have the utf 8 anymore and it's not guaranteed that there is any, as for Mathtexengine

jkrumbiegel:
For me it worked when I used the computer modern font and just spliced in the value `5390` for the glyph in the CairoMakie drawing code. Just like that:
```julia
function show_glyph(ctx, glyph::Culong, x, y)
    cg = Ref(CairoGlyph(5930, x, y))
    ccall((:cairo_show_glyphs, Cairo.libcairo),
            Nothing, (Ptr{Nothing}, Ptr{CairoGlyph}, Cint),
            ctx.ptr, cg, 1)
end
```

```julia
cm = Makie.FTFont("~/.julia/packages/MathTeXEngine/XFmQJ/assets/fonts/NewComputerModern/NewCMMath-Regular.otf")

s = Scene(camera = campixel!)
text!(s, 100, 100, text = "hi", font = cm)
s
```
image

So it's definitely possible accessing the glyph. However, what I first did didn't work, returning 5930 from the place where I introduced the error for you. But I assume now that this was because MathTeXEngine actually selected the wrong font for the expression that I did, which was just `L"x+y"`.

jkrumbiegel:
Great, I think it should be sufficient that TeXChar carries the Culong and then I can build it in from Makie's side. Would need to be a minor version change anyway right? So no compat problem until Makie is ready to integrate

Kolaru:
Yes, on MathTeXEngine side it is breaking, since the font will slightly change as well as the API of TeXChar (slightly).

jkrumbiegel:
I'm making some changes to FreeTypeAbstraction in this PR, you can probably just use the functions there for getting glyph indices instead of duplicating that on your side https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/pull/72

CairoMakie/src/primitives.jl
```diff
@@ -453,6 +450,25 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
...
 end
 
+struct CairoGlyph
+    index::Culong
+    x::Cdouble
+    y::Cdouble
+end
+
+function show_glyph(ctx, glyph::Culong, x, y)
+    cg = Ref(CairoGlyph(5930, x, y))
```
Kolaru, line 460:
```suggestion
function show_glyph(ctx, glyph, x, y)
    cg = Ref(CairoGlyph(glyph, x, y))
```
The type info is not needed directly, it gets converted anyway by `ccall` (and for some reason I sometime get an `Int64` reaching this function).

src/basic_recipes/text.jl
```diff
@@ -162,7 +162,7 @@ function texelems_and_glyph_collection(str::LaTeXString, fontscale_px, halign, v
     scales_2d = [Vec2f(x[3] * Vec2f(fs)) for x in els]
 
     texchars = [x[1] for x in els]
-    chars = [texchar.char for texchar in texchars]
+    glyphindices = [glyph_index(texchar) for texchar in texchars]
```
Kolaru, line 165:
```suggestion
    glyphindices = [MathTeXEngine.glyph_index(texchar) for texchar in texchars]
```

Kolaru (commented):
With these two changes, it is compatible with https://github.com/Kolaru/MathTeXEngine.jl/pull/67

The MathTeXEngine PR will require a bit more work to make sure everything work as it should, but the API shouldn't change.

src/basic_recipes/text.jl
jkrumbiegel: Should this just be `FreeTypeAbstraction.glyph_index` from the PR, which you can extend in MathTeXEngine?

jkrumbiegel:
Ok @Kolaru I could use your branch locally, the integral for example looks like this:

image

Seems like there's some wrong boundingbox information on that though?

Anyway, you can further test your changes with this branch, then when you tag MathTeXEngine it should be picked up by this PR. I've already bumped compat.

jkrumbiegel:
Hm currently the only thing still problematic is that the word wrapping checks for `\n` but I'm only storing glyph indices now. Any opinions @ffreyer what would be a good reimplementation? Maybe the characters should still be stored side by side, but that's not really possible or meaningful for MathTeXEngine

jkrumbiegel:
Actually I'm not sure if word wrapping makes much sense for MathTeXEngine at all..

ffreyer:
> Actually I'm not sure if word wrapping makes much sense for MathTeXEngine at all..

I think it does. If you have some formulas you may want other text as latex strings for visual cohesion.

> Hm currently the only thing still problematic is that the word wrapping checks for `\n` but I'm only storing glyph indices now. Any opinions @ffreyer what would be a good reimplementation? Maybe the characters should still be stored side by side, but that's not really possible or meaningful for MathTeXEngine

Can't you just compare to `glyph_index('\n', font)` instead?

jkrumbiegel:
> Can't you just compare to glyph_index('\n', font) instead?

True, didn't think this way around.

Hm yeah maybe, I haven't used latexstrings much and especially not at sizes that require word wrapping.

PR #2151, opened 2022-07-19
Fix DataInspector hover over NaN values

MariusDrulea:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/issues/1714

## Type of change

### DIFF
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -707,13 +707,9 @@ function show_imagelike(inspector, plot, name, edge_based)
     # in case we hover over NaN values
     if isnan(z)
         a._visible[] = false
-        # return false as this plot is not processed (due to NaN) 
-        # and we have to process other plots from the figure
-        # see @show_data_recursion
-        return false
+        return true
     end
 
-    # from this line down, z is not a NaN
     a._color[] = if z isa AbstractFloat
         interpolated_getindex(
             to_colormap(plot.colormap[]), z,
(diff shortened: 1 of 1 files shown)


MariusDrulea:
I have tested the change in the PR with the following sample code. Works as expected.
```
using GLMakie

##
x = rand(100, 100)
x[20:60, :] .= NaN
x2 = 100 * rand(100, 100)
x2[40:end, :] .= NaN


##
fig, ax, plot_obj = image(rotr90(x), interpolate=false)
image!(rotr90(x2), interpolate=false, colormap=(:jet1, 0.5))
DataInspector(fig)


##
screen = display(fig)                  
wait(screen)

##
```

src/interaction/inspector.jl
```diff
@@ -704,6 +704,16 @@ function show_imagelike(inspector, plot, name, edge_based)
         _pixelated_getindex(plot[1][], plot[2][], plot[3][], mpos, edge_based)
     end
 
+    # in case we hover over NaN values
+    if isnan(z)
+        a._visible[] = false
+        # return false as this plot is not processed (due to NaN) 
+        # and we have to process other plots from the figure
+        # see @show_data_recursion
+        return false
```
ffreyer, line 713: This should still return true. In this case getting processed means that there is a `show_data` method that decided what to do with a hovered plot.

ffreyer (changes requested):
Thanks for the fix. We can merge this after fixing the comment/return value.

ffreyer (approved):

PR #2152, opened 2022-07-19
trace compile in build step

SimonDanisch:
Lets see how stable that is!


SimonDanisch:
Guess the last commit killed it ^^ Just for reference (because I won't fix it, since this PR is obsolete):
image

Luckily, @timholy just came around the corner with [SnoopPrecompile](https://github.com/timholy/SnoopCompile.jl/pull/282) which makes this approach obsolete!
Tried out in https://github.com/JuliaPlots/Makie.jl/pull/2153, which indicates similar gains :)

PR #2153, opened 2022-07-19
use the new SnoopPrecompile

SimonDanisch:
Since it's not registered yet and actually a pretty small macro, I've just inlined it into Makie for now...


SimonDanisch:
The using times slow down are expected, since this now caches precompiles for much more than just `scatter`....
The speed up and one slow down of run time must be noise, although the GLMakie one seems to be real, since it has been showing up pretty consistently.
Must be the change in screen.jl (maybe ` closeall(screen_ref)`?).
Pretty happy about the improvement in general (thanks @timholy, this macro is really amazing and has been on my bucket list for quite some time :) ).

PR #2167, opened 2022-07-22
Fix stackoverflow in show

SimonDanisch:
Adding that method messed with `showable` as now Makie claimed to be able to show every mime...


SimonDanisch:
Fixes https://github.com/JuliaPlots/Makie.jl/issues/2166

SimonDanisch:
Ahhh come on:
image

99% Sure, that it's not due to this PR and rather a CDN problem or so... Going to merge, and investigate later, since we should tag this quickly...

PR #2173, opened 2022-07-23
DataInspector shall show img[row, col] = value

MariusDrulea:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/issues/2172

## Type of change

This is how it looks now:


MariusDrulea:
potential issue: 5 commits from my previous PR still appear here; they were merged into Makie/master as a unified commit; after merging Makie/master into MariusDrulea/Makie/master, these individual commits still appear here, although the files are obviously the same

ffreyer:
I think this is more of a misunderstanding. What DataInspector tells you here is how you need to index your data to get the value you are hovering over. That's not the same as telling you the axis coordinates.

```julia
img = [0 1; 0 0 ]
fig, ax, p = heatmap(img)
DataInspector(fig)
fig
```
for example shows the `1` in the top left corner and DataInspector correctly reports `H[1, 2] = 1.000` for it.

MariusDrulea:
See the following example. In the top right corner the correct way do get the value is `x[1, 20]`. Still the DataInspector shows `img[20, 1]`.
Note I had to transpose `x'` and reverse the y-axis.

```
using GLMakie

x = Float64.(repeat(1:10, 1, 20))

fig, ax, plot_obj = image(x', interpolate=false)
ax.yreversed = true
DataInspector(fig)
```

image

x looks like:
image

MariusDrulea:
At the end I think we shall edit the default behaviour for images and heatmaps. We shall only call `image(x)` and get the expected behavior. So no more transposition, no rotr90 and no more reverse of the yaxis.

MariusDrulea:
**Update:**
@ffreyer After a few more tests I realized DataInspector changes have to be merged. It has nothing to do with the transposition, rotr90 etc.
DataInspector simply shows the value at a specific location, _w.r.t. the horizontal and vertical axes_. 
The problem is: DataInspector displays the vertical coordinate first and this is counter intuitive. The horizontal axis has to go first. See the before and after pictures bellow.

**Before:**


**After:**

PR #2056, opened 2022-06-16
use lightweight GeometryBasics

SimonDanisch:

PR #2175, opened 2022-07-25
feat: add lowclip and highclip keywords to poly

### LINKED ISSUE #2156: [Feature Request] Add highclip, lowclip and nan_color to poly
Hi,

thanks for this great package! Would you consider adding `highclip`, `lowclip` and `nan_color` to the `poly()` function? Reading the code [here](https://github.com/JuliaPlots/Makie.jl/blob/05413d3206565a1dcdd1697106e5dc39da08da48/src/basic_recipes/poly.jl#L49-L64) it seems that it should be straightforward, as these parameters only need to be passed to `mesh()`.

Cheers

cristobaltapia:
# Description

Fixes #2156 

Adds support for the keywords `lowclip` and `highclip` to the function `poly`.
I still would need to add documentation, but first I wanted to ask, whether the implementation and the code style was ok. Specifically, I would like to know whether the handling of the colors is correctly done with `parse(RGBA, lowclip)`.

Cheers,

Cristóbal

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/poly.jl
+++ b/src/basic_recipes/poly.jl
@@ -25,6 +25,9 @@ $(ATTRIBUTES)
         strokecolor = theme(scene, :patchstrokecolor),
         colormap = theme(scene, :colormap),
         colorrange = automatic,
+        lowclip = automatic,
+        highclip = automatic,
+        nan_color = :transparent,
         strokewidth = theme(scene, :patchstrokewidth),
         shading = false,
         fxaa = true,
@@ -49,9 +52,17 @@ convert_arguments(::Type{<: Poly}, m::GeometryBasics.Mesh) = (m,)
 function plot!(plot::Poly{<: Tuple{Union{GeometryBasics.Mesh, GeometryPrimitive}}})
     mesh!(
         plot, plot[1],
-        color = plot[:color], colormap = plot[:colormap], colorrange = plot[:colorrange],
-        shading = plot[:shading], visible = plot[:visible], overdraw = plot[:overdraw],
-        inspectable = plot[:inspectable], transparency = plot[:transparency],
+        color = plot[:color],
+        colormap = plot[:colormap],
+        colorrange = plot[:colorrange],
+        lowclip = plot[:lowclip],
+        highclip = plot[:highclip],
+        nan_color = plot[:nan_color],
+        shading = plot[:shading],
+        visible = plot[:visible],
+        overdraw = plot[:overdraw],
+        inspectable = plot[:inspectable],
+        transparency = plot[:transparency],
         space = plot[:space]
     )
     wireframe!(
@@ -122,6 +133,9 @@ function plot!(plot::Poly{<: Tuple{<: Union{Polygon, AbstractVector{<: PolyEleme
         color = plot.color,
         colormap = plot.colormap,
         colorrange = plot.colorrange,
+        lowclip = plot.lowclip,
+        highclip = plot.highclip,
+        nan_color = plot.nan_color,
         overdraw = plot.overdraw,
         fxaa = plot.fxaa,
         transparency = plot.transparency,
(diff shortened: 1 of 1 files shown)


cristobaltapia:
This PR can be tested with the following code:

```jl
using GLMakie

f = Figure();
Axis(f[1, 1]);
poly!(
    [
        Point2f[(2, 0), (4, 0), (4, 1), (2, 1)],
        Point2f[(0, 0), (2, 0), (2, 1), (0, 1)],
        Point2f[(2, 1), (4, 1), (4, 2), (2, 2)],
        Point2f[(0, 1), (2, 1), (2, 2), (0, 2)],
    ];
    color=[1, 2, 3, 4],
    colormap=:viridis,
    colorrange=(2, 3),
    lowclip=:white,
    highclip=:black,
    strokewidth=2,
)
display(f)

```

src/basic_recipes/poly.jl
```diff
@@ -163,16 +177,21 @@ function plot!(plot::Mesh{<: Tuple{<: AbstractVector{P}}}) where P <: Union{Abst
...
+        lift(meshes, color_node, attributes.colormap, attributes.colorrange, attributes.lowclip, attributes.highclip) do meshes, colors, cmap, crange, lclip, hclip
             # Color are reals, so we need to transform it to colors first
             single_colors = if colors isa AbstractVector{<:Number}
                 interpolated_getindex.((to_colormap(cmap),), colors, (crange,))
             else
                 to_color.(colors)
             end
+            # Consider low- and highclips
+            single_colors[colors .< crange[1]] .= parse(RGBA, lclip)
+            single_colors[colors .> crange[2]] .= parse(RGBA, hclip)
```
asinghvi17, line 194:
```suggestion
            single_colors[colors .< crange[1]] .= to_color(lclip)
            single_colors[colors .> crange[2]] .= to_color(hclip)
```

This handles symbols as well as RGB, HSV, and other color types.

cristobaltapia: Thanks for the hint! I changed that accordingly.

cristobaltapia:
I also had to check for `colorrange` to be a tuple, otherwise I would get an error when precompiling. This was changed in the last commit.

src/basic_recipes/poly.jl
```diff
@@ -122,6 +133,9 @@ function plot!(plot::Poly{<: Tuple{<: Union{Polygon, AbstractVector{<: PolyEleme
         color = plot.color,
         colormap = plot.colormap,
         colorrange = plot.colorrange,
+        lowclip = plot.lowclip,
```
SimonDanisch, line 136: wireframe doesn't use those ;)

src/basic_recipes/poly.jl
cristobaltapia: I tried that, but that does not work with the example I gave before. I could not really understand why, but doing these changes solved the problem. Do you have any idea?

src/basic_recipes/poly.jl
cristobaltapia: But that function is calling mesh. Or is it doing something in the background to also call wireframe?

SimonDanisch: Oh sorry, I guess you're right... the diff view made it look that way:
image

cristobaltapia: Oh, that is missleading :P haha

src/basic_recipes/poly.jl
cristobaltapia: I think that the problem is that the colors are passed to `mesh()` already as colors and not as values (see [these lines](https://github.com/JuliaPlots/Makie.jl/blob/master/src/basic_recipes/poly.jl#L167-L186)). Therefore, it is not possible to determine whether a certain color should be clipped or not. This is done like this, because we want to have one color per polygon, which is different to the behavior of mesh, where one color per node is defined.

asinghvi17: Yeah, makes sense in that case.

SimonDanisch: Ah damn, I was already in the PR https://github.com/JuliaPlots/Makie.jl/pull/2056, where I cleaned up the multi mesh recipe, to pass through numbers directly to the mesh call...
I'll open a PR with your changes and the changes from #2056, to do this more cleanly ;)

cristobaltapia: That sounds good to me. I was thinking that such a solution would be cleaner, but then was struggling with passing numbers to mesh directly.

SimonDanisch: https://github.com/JuliaPlots/Makie.jl/pull/2183

SimonDanisch:
closed in favor of https://github.com/JuliaPlots/Makie.jl/pull/2183

PR #2183, opened 2022-07-26
Cristobaltapia feat 2156 additional kws to poly

SimonDanisch:
Continuation of https://github.com/JuliaPlots/Makie.jl/pull/2175
with:

1) improvements from #2056, so that high/low clip doesn't need to be handled in recipe
2) overload `convert_arguments` for single geometries, to not dispatch them to the multi mesh recipes (it seems like that was the case for all these years, which should have made `poly(Rect(...))` unnecessarily slow)
3) added a test for low/high clip


jkrumbiegel:
Can you check if this also fixes contourf high/low clip? That was broken by some refactor using integers for colors a while ago

jkrumbiegel:
And if so, a test for that would ideally be added as well. There's an example in the docs for contourf which you can use, it's currently broken

PR #2198, opened 2022-08-04
up versions

SimonDanisch:
... And fix MakieCore version to be an exact match!


jkrumbiegel (approved):
Maybe add news first?

Issue #368, opened 2019-06-25
Plots.jl shorthands


asinghvi17:
Plots.jl has a lot of small, easy-to-implement convenience functions and recipes that would be nice to have in AbstractPlotting, or StatsMakie.

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L70-L77

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L79-L86

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L139-L151

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L139-L151

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L166-L178

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L180-L192

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L224-L239

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L368-L380

- [ ] https://github.com/JuliaPlots/Plots.jl/blob/e061d156bc0a826db1d8a53d89cd8a2852f806e2/src/shorthands.jl#L382-L384

asinghvi17:
Implementation for sticks (needs to be turned into a recipe but you get the picture):

```julia
function sticks!(scene, x, y)
for (i, j) in zip(x, y)
    linesegments!(scene, (Point2f0(i, min(j, 0)), Point2f0(i, max(j, 0)))
end
end
```
![sticks-makie](https://user-images.githubusercontent.com/32143268/60091389-e7e7ed80-975d-11e9-9a65-7a0d9f7d3db7.png)

asinghvi17:
[...]

PR #2201, opened 2022-08-05
Hexbin recipe

TabeaW:
# Description

Adds the recipe for hexbin plots (#368) with documentation.

## Type of change

## Checklist


TabeaW:
As this is my first work in Makie, it would be great if someone experienced can have a look at this :) I am not quite sure, if I had done reasonable work with the axis :D

jkrumbiegel:
Very cool, thank you, I'll try to take a look soon when I have some spare time.

jkrumbiegel:
For some inputs I get an empty line (usually second to topmost it seems?)

```julia
x = [-0.8229475043895498, -0.46621767925243984, -2.0720758196284805, 0.524687403763896, 0.9653685757558091, 1.7528735168451055, -0.33237984203429993, 0.3506865765367531, -0.8944929017899972, 0.2532684038441482]
y = [-1.7463655960215445, 1.8957111366590151, 0.5550468594781066, 1.2254393653690707, 1.371174685353981, -1.0206975175484125, 0.3360700428234833, 2.195537222554637, 0.9952661972027315, 0.28852172975641416]
hexbin(x, y)
```

image

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+   
+    polys = Observable(Polygon[])
+    count_hex=Observable(Float64[])
+    function calculate_grid(x,y,grid_size,mincnt,scale)
+        empty!(polys[])
+        empty!(count_hex[])
+        if (length(x)==0)|(length(y)==0)
+            return
+        end
+        axis = current_axis()
```
jkrumbiegel, line 35: What exactly do you need the axis for? It looks like you're trying to access its limits? First of all, `current_axis()` isn't necessarily the axis you're currently plotting to, so you can't use that. Second, you're not supposed to access any axis stuff in a recipe, it lives a level below that in a simple Scene. (At least how recipes currently work). And third, you are only doing your calculation once, so if the limits changed the plot wouldn't update. You can access limits via the projection matrix of the Scene if you really need to (check out vlines etc for how it's done there) but it's kind of rare that we do this.

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+    colorrange=Makie.automatic,
+    gridsize = 20,
+    mincnt = 0,
+    scale=identity)
+end
+
+function Makie.plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Any}, <:AbstractVector{<:Any}}})
+    x,y =hb[1:2]
+   
+    polys = Observable(Polygon[])
```
jkrumbiegel, line 27: Polygon is an abstract type so this doesn't really help in terms of type stability.

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+        xe = sin.(0:pi/3:5/3*pi)
+        ys = cos.(0:pi/3:5/3*pi)
+        hex_points=Point2f[]
+        for (a,b) in zip(xe,ys)
+            push!(hex_points, Point2f(a*r*1/scaling_x,b*r*1/scaling_y))
+        end
+        hex = [Polygon(hex_points.+p*Point2f(1/scaling_x,1/scaling_y)) for p in grid_points]
+
+        for b in hex
+            push!(polys[],b)
```
jkrumbiegel, line 116: looks like you could also just `append!(polys[], hex)`? But anyway this would accumulate if you updated, so I'm not sure you want that

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+        for a in amount_hex
+            push!(count_hex[],a)
+        end
+
+    end
+    onany(calculate_grid,x,y,hb.gridsize, hb.mincnt,hb.scale)
+    calculate_grid(x[],y[],hb.gridsize[],hb.mincnt[],hb.scale[])
+
+    replace_automatic!(hb,:colorrange)do 
+        if isempty(count_hex[])
```
jkrumbiegel, line 127: not sure but this probably wouldn't update if you update the plot data

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+            (0,1)
+        else
+            (minimum(count_hex[]),maximum(count_hex[]) )
+        end
+    end 
+    if ! isempty(polys[])
```
jkrumbiegel, line 133: This is also written for a static plot. In Makie, we always try to make plots update-able if possible. So you would just plot `polys` one way or another, and if the data were updated such that there are now polygons, the plot would update as well.

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+        ind,dist = nn(tree,values)
+        amount_hex = zeros(size(grid_points))
+        for i in ind
+            amount_hex[i]+=1
+        end
+        grid_points=grid_points[amount_hex.>=mincnt]
+        amount_hex=amount_hex[amount_hex.>=mincnt]      
+        amount_hex = scale.(amount_hex)
+
+        xe = sin.(0:pi/3:5/3*pi)
```
jkrumbiegel, line 107: you don't need to allocate the vectors here, you can loop over the range below and calculate x,y on the fly

src/stats/hexbin.jl
jkrumbiegel: ah I see now you empty the vector above

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+        for b in hex
+            push!(polys[],b)
+        end
+        for a in amount_hex
+            push!(count_hex[],a)
+        end
+
+    end
+    onany(calculate_grid,x,y,hb.gridsize, hb.mincnt,hb.scale)
+    calculate_grid(x[],y[],hb.gridsize[],hb.mincnt[],hb.scale[])
```
jkrumbiegel, line 124: this is done a little more easily by just saying `notify(hb.gridsize)` or whatever

src/stats/hexbin.jl
```diff
@@ -0,0 +1,136 @@
...
+        for (a,b) in zip(xe,ys)
+            push!(hex_points, Point2f(a*r*1/scaling_x,b*r*1/scaling_y))
+        end
+        hex = [Polygon(hex_points.+p*Point2f(1/scaling_x,1/scaling_y)) for p in grid_points]
+
+        for b in hex
+            push!(polys[],b)
+        end
+        for a in amount_hex
+            push!(count_hex[],a)
```
jkrumbiegel, line 119: so you're updating the vectors but never notify the observables again, so you wouldn't see an update in a live plot

src/stats/hexbin.jl
jkrumbiegel: you could construct a fake poly and do `typeof(fakepoly)[]` if the type is hard to write manually (sometimes the case with GeometryBasics types I think)

jkrumbiegel (changes requested):

jkrumbiegel:
Very good first pr, a lot of good things in there already! I made some initial comments, maybe you can clarify what the logic is supposed to be that you need limits for, I didn't quite understand from reading.

TabeaW:
Thanks a lot for the feedback!
I used the axis with its limits to cut the data. For example, if you have data points from 0:100, but you want to show only those up to 50. As, the size of the hex is calculated through the difference of min and max of axis or data. Maybe I could change the gridsize parameter, so that one can give the size and not the number of hex.
I thought that the hexbin is also updatable (I have already created a video in any case), but I will check your suggestions!

jkrumbiegel:
Yes I think it's better to decouple this from the axis limits, and have the user couple those if desired :)

SimonDanisch:
Thank you @TabeaW, we're trying to merge this in https://github.com/MakieOrg/Makie.jl/pull/2236. If you have any further improvements, you should be able to PR against that PR :)

PR #2202, opened 2022-08-05
Fix scroll lock in Menu

ffreyer:
# Description

Currently the scrolling interaction of menu scenes always triggers and always blocks lower priority/later scroll events (as long as it isn't blocked by something else). This means that creating a `Menu` before an `Axis` will make you unable to zoom in that axis.

This is fixed here by only running code and returning `Consume(true)` if `is_mouseinside(menuscene)` is true. 

## Type of change

### DIFF
--- a/src/makielayout/blocks/menu.jl
+++ b/src/makielayout/blocks/menu.jl
@@ -242,10 +242,14 @@ function initialize_block!(m::Menu; default = nothing)
     end
 
     on(menuscene.events.scroll) do (x, y)
-        t = translation(menuscene)[]
-        new_y = max(min(t[2] - y, 0), height(menuscene.px_area[]) - listheight[])
-        translate!(menuscene, t[1], new_y, t[3])
-        return Consume(true)
+        if is_mouseinside(menuscene)
+            t = translation(menuscene)[]
+            new_y = max(min(t[2] - y, 0), height(menuscene.px_area[]) - listheight[])
+            translate!(menuscene, t[1], new_y, t[3])
+            return Consume(true)
+        else
+            return Consume(false)
+        end
     end
 
     on(m.options) do options
--- a/src/makielayout/blocks/menu.jl
+++ b/src/makielayout/blocks/menu.jl
@@ -241,7 +241,7 @@ function initialize_block!(m::Menu; default = nothing)
         return Consume(false)
     end
 
-    on(menuscene.events.scroll) do (x, y)
+    on(menuscene.events.scroll, priority=61) do (x, y)
         if is_mouseinside(menuscene)
             t = translation(menuscene)[]
             new_y = max(min(t[2] - y, 0), height(menuscene.px_area[]) - listheight[])


jkrumbiegel:
I honestly don't understand how the priority system works in such cases so I will probably introduce similar bugs in the future. Is it explained somewhere which elements get which event at what time and how the priorities/consume settings affect that? Or are we currently doing it in an ad-hoc "this is blocked so let's increase its priority" way

ffreyer:
The rules are:
1. If two listeners have different priority the higher one executes first
2. If two listeners have the same priority the one added first (i.e. first in the list of listeners) executes first. (Like without priority)

And if a listener returns `Consume(true)` the callback loop exits. So it will deny any listener that has lower priority or is added later. 

So basically priority is a way to set execution order, with listeners at equal priority executing more or less in a random order (because it often depends on outside factors, i.e. creation order) and `Consumne` is a way to shut out lower priority events.

In this case you probably want a high priority so the menu scrolling is considered first, and consume the event only if the menu is opened and hovered.

jkrumbiegel:
Yes the basically random execution order otherwise is what's confusing me. Should it really be that random or should we have a clearer plan how events move through the scene graph?

ffreyer:
I tried a set up something similar to axis interactions with priority and scene ordering in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/586. It ended up too slow imo. (Maybe because the approach with recursive functions was bad. But I think triggering observables in each scene child-first would be awkward and slow too.) 

Iirc I was also not sure on whether priority or scene order should come first. Scene order first makes more sense to me right now, but that might not always be what you want. For example you could set up some sort of pen widget that draws on top of everything on the root scene. With scene order first this would get blocked all the time. `DataInspector` would too, as it is now.

Going without priority entirely is problematic too. Any kind of optional interaction that supersedes an existing one needs it. Without, you run into the same ordering issue as now. And with priorities you still need to pick the right ones.

jkrumbiegel:
Then maybe we should come up with an enum of priority levels or so? I face kind of the same problem with z values, for example the legend is shifted forward. But by how much is kind of arbitrary, I just picked some values that worked, there's no "system"

ffreyer:
> Then maybe we should come up with an enum of priority levels or so?

I don't think restricting it makes sense, but we could define `const LOW_PRIORITY = 0` etc. Not sure how useful that is but I guess it's a bit more readable?

> I face kind of the same problem with z values, for example the legend is shifted forward. But by how much is kind of arbitrary, I just picked some values that worked, there's no "system"

For that (at least in GLMakie) `p = projection * view * model * point` must be in a (-1, -1, 0) to (1, 1, 1) range to be drawn. You can go back from there. 

Generally `near` and `far` describe the range in data space, after `view * model`:
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/camera2d.jl#L85
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/camera2d.jl#L323
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/camera2d.jl#L343
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/camera3d.jl#L144-L145

For `space = :pixel` it's currently hardcoded to -10_000 .. 10_000 here:
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/camera.jl#L71-L76

`space = :clip` and `space = :relative` are hardcoded to 0 .. 1 here: (The matrices constructed here usually end up replacing `projection * view`)
https://github.com/JuliaPlots/Makie.jl/blob/0252621cbd4e9479e5c0db8ea4b528d800bf3e22/src/camera/projection_math.jl#L314-L326

ffreyer:
Can we merge this?

jkrumbiegel:
I think so!

PR #2224, opened 2022-08-16
improve cycling documentation

Datseris:
Closes https://github.com/JuliaPlots/Makie.jl/issues/2215

### DIFF
--- a/docs/documentation/theming.md
+++ b/docs/documentation/theming.md
@@ -148,16 +148,17 @@ cannot cycle markers). This is exemplified in the following code blocks.
 
 \begin{examplefigure}{}
 ```julia
-set_theme!(
-    palette = (color = [:red, :blue], marker = [:circle, :xcross]), 
-    Scatter = (cycle = [:color, :marker],)
-)
-scatter(fill(1, 10))
-scatter!(fill(2, 10))
-scatter!(fill(3, 10))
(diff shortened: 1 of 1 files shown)


docs/documentation/theming.md
```diff
@@ -142,9 +142,16 @@ cycle = [[:linecolor, :markercolor] => :color, :marker]
...
 ```
 
+Notice that cycles must be given as attributes to a plot object, not the top-level theme
+(because different plot objects can cycle different attributes, e.g., a density plot
+cannot cycle markers). This is exemplified in the following code blocks.
+
 \begin{examplefigure}{}
 ```julia
-set_theme!(palette = (color = [:red, :blue], marker = [:circle, :xcross]), Scatter = (cycle = [:color, :marker],))
+set_theme!(
```
jkrumbiegel, line 151: Could you please change all `set_theme!`s to `with_theme` blocks? It happens again and again that the global theme is modified and not reset accidentally and then some random plots somewhere else start looking wonky :) So the `with_theme` guards against that

jkrumbiegel:
looks like some `)` or so is missing, the docs build failed

jkrumbiegel:
Have you run all the code on the page? There's still an error

jkrumbiegel (approved):

PR #2226, opened 2022-08-17
Tricontourf plot

jkrumbiegel:
# Description

Adds a tricontourf plot type.

```julia
using CairoMakie
using Random

Random.seed!(1234)

x = randn(100)
y = randn(100)
z = x .* y

Makie.tricontourf(x, y, z, colormap = :heat)
scatter!(x, y, color = z, colormap = :heat, strokewidth = 1, strokecolor = :black)
current_figure()
```

image

## Type of change

## Checklist


kimauth:
This looks super cool! :purple_heart: 

It seems like it always computes the connectivity pattern between the points (`trianglelist`) by a Delaunay triangulation right now. Coming from finite elements, it is common in my field to have that pattern available. Are there any plans to allow handing it over to `tricontourf`? And would knowing the triangulation actually be enough to do contour plots on non-convex domains, too?

jkrumbiegel:
Oh sure that could definitely be an option. I don't really know much about tricontourf plots and just implemented what I thought was the most common usage pattern.  You could look in the source where I pass the list of triangles to the contour generating function and just pass your triangles in there and see if it works? We could just make the triangulation mode an attribute and maybe the default would be `:delauney` but one could pass a list of triangles instead.

jkrumbiegel:
I have no idea if the contour generation function needs a specific form of triangulation though.

jkrumbiegel:
I just added the option to pass a matrix as triangles (probably should add a vector of triangle faces from GeometryBasics as well, and it seems fine:

```julia
using CairoMakie
x = [0.0, 0.5, 1.0, 0.25, 0.75, 0.5]
y = sqrt(3)/2*[0.0, 0.0, 0.0, 0.5, 0.5, 1.0]
z = [0.0, 0.5, 0.0, 1.0, 1.0, 0.0]
f, ax, _ = tricontourf(x, y, z, triangles = [
    1 2 4
    2 3 5
    4 5 6
]')
scatter!(x, y, color = z, strokewidth = 1, strokecolor = :black)
tricontourf(f[1, 2], x, y, z, triangles = Makie.DelaunayTriangulation())
scatter!(x, y, color = z, strokewidth = 1, strokecolor = :black)
f
```

image

kimauth:
That's exactly what I needed, thanks for putting it in so quickly!

I gave it a trial run with some of my real data, it worked great:
test_tricontourf
(In the past I used to put a fine regular grid over my data to obtain this type of plot, now it is only a few lines from the `Ferrite.jl` data structures to the plot.)

jkrumbiegel:
Cool, that's good to know! I'll continue with that path then. Maybe I'll call the parameter `triangulation` instead of `triangles`

src/basic_recipes/tricontourf.jl
```diff
@@ -0,0 +1,198 @@
...
+                    Point2f(tupl...)
+                end
+            end
+            if isempty(pointvecs)
+                continue
+            end
+            
+            p = Makie.Polygon(pointvecs[1], pointvecs[2:end])
+            push!(polys[], p)
+            push!(colors[], lc)
```
kimauth, line 143:
```suggestion
            
            for pointvec in pointvecs
                p = Makie.Polygon(pointvec)
                push!(polys[], p)
                push!(colors[], lc)
            end
```
seems to fix the GLMakie issue for me. Not sure if this is the correct approach in general though.

src/basic_recipes/tricontourf.jl
```diff
@@ -0,0 +1,198 @@
...
+    xs, ys, zs = c[1:3]
+
+    c.attributes[:_computed_levels] = lift(zs, c.levels, c.mode) do zs, levels, mode
+        _get_isoband_levels(Val(mode), levels, vec(zs))
+    end
+
+    colorrange = lift(c._computed_levels) do levels
+        minimum(levels), maximum(levels)
+    end
+    computed_colormap = lift(c._computed_levels, c.colormap, c.extendlow,
```
SimonDanisch, line 63: should we share this with contourf?

src/basic_recipes/tricontourf.jl
```diff
@@ -0,0 +1,198 @@
...
+        highs = levels[2:end]
+
+        trianglelist = compute_triangulation(triangulation, xs, ys)
+        filledcontours = filled_tricontours(xs, ys, zs, trianglelist, levels)
+
+        levelcenters = (highs .+ lows) ./ 2
+
+        for (fc, lc) in zip(filledcontours, levelcenters)
+            pointvecs = map(fc.polylines) do vecs
+                map(vecs) do tupl
```
SimonDanisch, line 133: I think `map(Point2f, vecs)` should work?

src/basic_recipes/tricontourf.jl
jkrumbiegel: Yeah that's copied, should be factored out

src/basic_recipes/tricontourf.jl
jkrumbiegel: yeah I don't remember where the previous version comes from :)

jkrumbiegel:
This needs some refactoring so that code duplication with contourf is reduced. Also, currently each triangle generates its own polygon for each band, so we get edge artifacts in CairoMakie, but not GLMakie:

image

Even though the bands run continuously across several triangles, the triangle borders are always visible with white artifacts.

jkrumbiegel:
Should be ok now

SimonDanisch (approved):

thomvet:
Just wanted to say thank you for creating this! Have a year ago I posted a hacky and ugly coded version of it on discourse (and got some help making it work!).

And now it's a full blown recipe inside Makie! Greatly appreciated!

PR #2228, opened 2022-08-17
Add option to update when releasing mouse button from slide

lwhitefox:
# Description

Fixes # (issue)

Added the option `drag_update` to Slider.   If true, updates selected index while dragging; otherwise only does update on mouse release.  Thus if true, graphics will update while dragging the slider in common usage.  If false,  graphics update only on mouse release, which may be useful if updates rely on comparatively slow calculations for example.

In addition a couple lines of code that appear to have no function in current implementation were identified and commented out with a note added to allow them to be found easily. 

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] Added or changed relevant sections in the documentation


jkrumbiegel:
Thank you! Looks good so far code-wise, but I was thinking more about this and maybe it's not quite the right solution.

The problem as I see it is that this way, if you have a slider and a text field for example, the text field would also not update while you drag the slider, which is potentially confusing or unhelpful because the user might want to stop dragging knowing they're at the right position.

So I thought, maybe it's easier to just add a second output observable, with a name like `static_value` or so (maybe you can come up with something better). And then one could use that if one desires updates only at drag-stop while still updating other ui like labels dynamically.

The other option I thought about was to have an observable called `dragging` which could be `true` or `false` and on `false` you could pick up `value` for the same behavior. But if `set_close_to!(slider, value)` is called, this observable wouldn't update because no drag happened, so the `static_value` solution is probably better.

lwhitefox:
> Thank you! Looks good so far code-wise, but I was thinking more about this and maybe it's not quite the right solution.
> 
> The problem as I see it is that this way, if you have a slider and a text field for example, the text field would also not update while you drag the slider, which is potentially confusing or unhelpful because the user might want to stop dragging knowing they're at the right position.
> 
> So I thought, maybe it's easier to just add a second output observable, with a name like `static_value` or so (maybe you can come up with something better). And then one could use that if one desires updates only at drag-stop while still updating other ui like labels dynamically.
> 
> The other option I thought about was to have an observable called `dragging` which could be `true` or `false` and on `false` you could pick up `value` for the same behavior. But if `set_close_to!(slider, value)` is called, this observable wouldn't update because no drag happened, so the `static_value` solution is probably better.

Makes sense in general - of course my approach fit my immediate needs and wants :-) 

However, you have a good example with the textfield views.  I'll have to think a little about implementing something along the lines of what you propose.  A lot of effort the first effort was trying to get the general code structure and operation, so now time to dig deeper

jkrumbiegel:
The approach would be almost exactly what you have, just remove the input observable you added, and add an output observable there just like `value`, only that this one is updated on dragstop only (and when using set_close_to!)

lwhitefox:
I've now pushed code that should have the approach discussed above implemented.  I added a new example to the docs to help demo the behavior as well.

docs/examples/blocks/slider.md
```diff
@@ -9,6 +9,12 @@ The currently selected value is in the attribute `value`.
 Don't change this value manually, but use the function `set_close_to!(slider, value)`.
 This is necessary to ensure the value is actually present in the `range` attribute.
 
+If the slider value is used for a relatively slow task, it may be more effective to
+use the attribute `value_dragstop` instead of `value`.  The `value_dragstop` is only
+updated when the mouse is released to conclude the slider drag operation.  It is 
+is synchronized with `value` when calling `set_cloe_to!(slider, value)` or one a mouse
```
jkrumbiegel, line 15: set_cloe_to

jkrumbiegel: this section should go before the example I think, with an appropriate subheading

src/makielayout/blocks/slider.jl
```diff
@@ -93,19 +96,18 @@ function initialize_block!(sl::Slider)
...
     onmouseleftdrag(mouseevents) do event
         dragging[] = true
-        dif = event.px - event.prev_px
-        fraction = clamp(if sl.horizontal[]
-            (event.px[1] - endpoints[][1][1]) / (endpoints[][2][1] - endpoints[][1][1])
-        else
-            (event.px[2] - endpoints[][1][2]) / (endpoints[][2][2] - endpoints[][1][2])
-        end, 0, 1)
+        # dif is never used?
+        # dif = event.px - event.prev_px
```
jkrumbiegel, line 100: these comments can be removed then

jkrumbiegel (changes requested):

jkrumbiegel:
The example should also have some small explanatory sentence probably, something like "In the following example one slider updates when ... and the other ...

SimonDanisch:
This is actually a really nice PR! @lwhitefox, do still want to finish it or do you require some help?

lwhitefox:
> This is actually a really nice PR! @lwhitefox, do still want to finish it or do you require some help?

I had some issues with email notifications for a while and missed updates from you and @jkrumbiegel - I'll get back to this in the next couple of days, it is good to have the reminder!

lwhitefox:
@SimonDanisch I've finished up the tasks for this PR; I did have to wrestle with git and GitHub a bit to get synchronized again so hopefully the updated code and .md file are pushed here successfully
If something is missing let me know!

f-ij:
This would be a very welcome addition!

jkrumbiegel:
I'm closing this in favor of https://github.com/MakieOrg/Makie.jl/pull/3231. Thanks for your work @lwhitefox, if you're still interested you can check out the other PR and see if it satisfies your needs as well. The implementation is a bit different as I decided I found one output observable simpler and a bit more flexible, too.

PR #2231, opened 2022-08-20
Implemented creation of videos (via `VideoStream` and `record`) directly in the output formats without first passing through mkv

rben01:
# Description

As is, when using `VideoStream` or `record` to create a video, Makie first encodes to an mkv, then transcodes to the desired output format (or returns the video as is if the output is `mkv`).

This PR instead encodes directly to desired output format without first passing through mkv. For non-mkv outputs, this should both lower the time to produce them and increase their quality.

No existing Makie APIs were modified nor were any new APIs were introduced . Rather, this change will become evident through the aforementioned improvements.

## Type of change

## Checklist

- [x] (N/A) Added or changed relevant sections in the documentation. Just updated the documentation of the functions modified. 
    - Nope, I gave up. Adding a test that requires `display`ing something is beyond my grasp — it requires making one of the backend packages available, but you need to use the version that lives in the repo to avoid mismatches with Makie's master branch. I don't know if this is possible in a way that's compatible with the rest of Makie. How do other Makie tests test their integration with one of the backends?
    - That said, I have pretty extensively tested this on my own (by just changing the file extension of the path passed to `record`); everything works fine. But it would be great to have some real tests of this.
- [x] (N/A) Added reference image tests for new plotting functions, recipes, visual options, etc.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,11 @@
 
 ## master
 
+- Updated `VideoStream` and `record` to encode directly to their output format instead
+  of first passing through `mkv`. This should have no user-facing consequences except
+  that `record` and `save(path, video_stream)` become faster and produce outputs of
+  higher quality.
+
 ## v0.17.13
 
 - Fix boundingboxes [#2184](https://github.com/JuliaPlots/Makie.jl/pull/2184).
@@ -100,7 +105,7 @@ role as `datalimits` in `violin` [#2137](https://github.com/JuliaPlots/Makie.jl/
 - Fixed manual cycling of plot attributes [#1873](https://github.com/JuliaPlots/Makie.jl/pull/1873).
 - Fixed type constraints in ticklabelalign attributes [#1882](https://github.com/JuliaPlots/Makie.jl/pull/1882).
 
-##  v0.16.4
+## v0.16.4
 
 - Fixed WGLMakie performance bug and added option to set fps via `WGLMakie.activate!(fps=30)`.
 - Implemented `nan_color`, `lowclip`, `highclip` for `image(::Matrix{Float})` in shader.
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Thank you for the PR! Could you please remove unrelated formatting changes? They make it hard to see what is relevant in the diff. Formatting changes are better done in separate PRs.

rben01:
> Thank you for the PR! Could you please remove unrelated formatting changes? They make it hard to see what is relevant in the diff. Formatting changes are better done in separate PRs.

Done! Sorry about the changes to NEWS.md. Kind of a bummer that .JuliaFormatter.toml is committed but is seemingly not actually used to format files — it means I have to disable auto formatting in my editor altogether. Would a formatting-only PR be welcome, or that too much trouble to be worth it at this point?

jkrumbiegel:
I think in general we do want auto formatting, it's just that we would need additional CI to ensure that formatting is still correct before merging, and we'd need some push to reformat everything once. It's just more difficult right now to work with PRs that touch lots of unrelated lines so until that other infrastructure is in place I'd still avoid the autoformatters. You can also disable that just for the Makie workspace.

rben01:
Got it, thanks. Also I figured out how to unit test using the repo's version of CairoMakie (required mucking with `LOAD_PATH`), so I added unit tests as well. The tests are very simple, just checking for no thrown exceptions and that the expected warnings are thrown, but better than nothing.

rben01:
Also, I realized that this is a breaking change as `save(path, video_stream)` no longer takes keyword arguments. I don't see a good way to work around this — there's no way for kwargs to propagate from `save` to the `VideoStream` because `VideoStream` now gets all its kwargs when it's constructed, so it's too late to do anything with `save`'s kwargs when `save` is called. Even if I permitted `save` to take kwargs to avoid hard breakage, it would no longer treat those keyword arguments the same as it does now — they'd be ignored — which would probably worse (silent change of behavior of existing code). This will not affect anyone using the `record` function, so I imagine impact will be low. Thoughts?

PR #2236, opened 2022-08-25
Hexbin recipe (continuation of 2201)

jkrumbiegel:
continuation of https://github.com/JuliaPlots/Makie.jl/pull/2201/

poly updates still fail, probably related to a similar problem with contourf


jkrumbiegel:
The thing that I'm still unsure about is what `bins` and `binsize` should mean. Because of the inherent asymmetry in x and y, it's not so easy to choose. Currently, one "bin" means one step from center to center for me (so at least two hexagons) because when you want to cover a rectangular area with hexagons, you need to cut half of the hexagons in half and they'll stick out. So it seems sensible to me to go from center left to center right in steps of adjacent centers. If you wante

jkrumbiegel:
This only fails on WGLMakie because `highclip` and `lowclip` are `nothing` by default (forwarded to Scatter) and WGLMakie doesn't seem to like that. I thought scatter supported those already? But if it was supposed to, it was not covered by tests.

SimonDanisch:
We should probably document this better, but things that need to be calculated somewhere (e.g. backend) , should be set to automatic instead of nothing!

src/stats/hexbin.jl
```diff
@@ -0,0 +1,206 @@
...
+                (
+                    fld(dvx, 2),
+                    iseven(dvy) ? dvy+1 : dvy,
+                )
+            end
+
+            d[id] = get(d, id, 0) + 1
+        end
+
+        # this iteration scheme misses points at the edges and I don't understand why
```
SimonDanisch, line 138: outdated?

src/stats/hexbin.jl
```diff
@@ -0,0 +1,206 @@
...
+                for ix in 0:_nx-1
+                    _x = xoff + 2 * ix * xspacing + (isodd(iy) * xspacing)
+                    _y = yoff + iy * yspacing
+                    c = get(d, (ix, iy), 0)
+                    push!(points[], Point2f(_x, _y))
+                    push!(count_hex[], scale(c))
+                end
+            end
+        else
+            # we only need to iterate dict values if we don't plot zero cells
```
SimonDanisch, line 153:
```suggestion
            # If we don't plot zero cells, we only have to iterate the sparse entries in the dict
```

src/stats/hexbin.jl
jkrumbiegel: yes!

PR #2255, opened 2022-09-05
clean up stairs + use linesegments

### LINKED ISSUE #1129: Rendering of stairs plot is distorted when zoomed out
It's great that you've added a stairs plot -- it's super useful for time series.

I've noticed that the rendering seems to be a bit off unless I zoom in.  The horizontal lines seem to be be drawn with a minimum length that is too wide, so they end up sticking out.  For example, it's very noticeable in the following plot (especially near the mouse pointer):
s

As a side note, I think `step=:post` may be a more common use case to use as the default.

SimonDanisch:
Fixes #1129 ... Linesegments should be a better fit for stairs anyways!

### DIFF
--- a/src/basic_recipes/stairs.jl
+++ b/src/basic_recipes/stairs.jl
@@ -22,48 +22,35 @@ end
 
 conversion_trait(::Type{<:Stairs}) = PointBased()
 
-function plot!(p::Stairs{<:Tuple{<:AbstractVector{<:Point2}}})
-    points = p[1]
-
-    steppoints = lift(points, p.step) do points, step
-        if step == :pre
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
But isn't it still a lines rendering bug in GLMakie that needs to be fixed? Linesegments don't have nice joins, that could be a visible deterioration

SimonDanisch:
Yeah sure...but I can't fix it right now :D And since steps are perfect for linesegments, this should be a nice work around for now!

jkrumbiegel:
yeah as I said I'd like to see how the step joins look with linesegments, usually you have these cutout corners where the lines overlap, when they're not joined. at least for the single-color case, `lines` avoids that

PR #2262, opened 2022-09-09
Fix Latex in tick labels

fatteneder:
# Description

Fixes an issue mentioned on Julia Discourse: https://discourse.julialang.org/t/makie-latex-in-tick-labels/86980

Problem was that `LaTeXString`s were implicitly converted into `Strings` in the pipeline.

## Type of change


src/makielayout/types.jl
```diff
@@ -126,7 +126,7 @@ mutable struct LineAxis
     elements::Dict{Symbol, Any}
     tickpositions::Observable{Vector{Point2f}}
     tickvalues::Observable{Vector{Float32}}
-    ticklabels::Observable{Vector{String}}
+    ticklabels::Observable{Vector{AbstractString}}
```
fatteneder, line 129: This change is not needed for the fix.
I thought I include it because otherwise `LaTeXStrings` decay here too.

jkrumbiegel: I guess that's fair. Just goes to show how difficult it is to type these fields statically

src/basic_recipes/text.jl
```diff
@@ -125,7 +125,10 @@ end
 function plot!(plot::Text{<:Tuple{<:AbstractArray{<:Tuple{<:AbstractString, <:Point}}}})    
     strings_and_positions = plot[1]
 
-    strings = Observable(first.(strings_and_positions[]))
+    strings = Observable(
```
jkrumbiegel, line 128: I would do `Observable{Vector{AbstractString}}(...)`

src/makielayout/lineaxis.jl
```diff
@@ -632,7 +632,7 @@ end
 
 Gets tick labels by applying `Showoff.showoff` to `values`.
 """
-get_ticklabels(::Automatic, values) = Showoff.showoff(values)
+get_ticklabels(::Automatic, values) = convert(Vector{AbstractString}, Showoff.showoff(values))
```
jkrumbiegel, line 635: hm this looks weird to me, usually you only have to do the type widening wherever you store the result, so wherever an observable is created holding the results.

fatteneder: Yeah, I don't really understand it either. Issue is that `Showoff.showoff` itself converts `Vector{LaTexStrings}` to `Vector{Strings}`, e.g.
```julia
julia> using Makie.Showoff

julia> s = [ L"x" , L"y" ]
2-element Vector{LaTeXStrings.LaTeXString}:
 L"$x$"
 L"$y$"

julia> cs1 = Showoff.showoff(s)
2-element Vector{String}:
 "L\"\$x\$\""
 "L\"\$y\$\""

julia> cs2 = convert(Vector{AbstractString}, Showoff.showoff(s))
2-element Vector{AbstractString}:
 "L\"\$x\$\""
 "L\"\$y\$\""
```

What is even more weird is that in the last one the `LaTeXString`s also decayed, e.g.
```julia
julia> cs2[1] |> typeof
String
```

But without the `convert` the fix does not work.

jkrumbiegel: Well yeah showoff is not for printing a vector of LatexStrings, so it should not hit that method at all

jkrumbiegel: I might have to take a look directly, hard to understand just from reading the diff what's going on exactly

jkrumbiegel:
I'll check this tomorrow

jkrumbiegel:
continued in #2322, I fixed the incorrectly narrow observable type there instead of doing a conversion in a kind of unrelated place

PR #2263, opened 2022-09-11
Fix image markers in CairoMakie

fatteneder:
# Description

Fixes https://github.com/JuliaPlots/Makie.jl/pull/2080#issuecomment-1242751992

## Type of change


fatteneder:
There is at least one more issue. Consider:
```
using CairoMakie, Colors, FileIO
CairoMakie.activate!()

pixel_types = [ RGBA, RGBAf, RGBA{Float16}, ARGB, ARGB{Float16}, RGB, RGBf, RGB{Float16} ]
rotations = [ 2pi/3 * (i-1) for i = 1:length(pixel_types) ]
w, h = 100+100*length(pixel_types), 400
s = Scene(resolution = (w, h), camera = campixel!)
pngname = "logo.png"
# pngname = "icon_transparent.png"
filename = Makie.assetpath(pngname)
marker_image = FileIO.load(filename)
f = Figure()
for (i, (rot, pxtype)) in enumerate(zip(rotations, pixel_types))
    marker = convert.(pxtype, marker_image)
    p = Point2f((i-1) * 100 + 100, 200)
    scale!(s, 2, 1)
    scatter!(s,
        p,
        marker = marker,
        markersize = 75,
        rotations = rot,
        transform_marker=true
    )
end

save(pngname, s)
```

Outputs:
- `CairoMakie`:
logo
- `GLMakie`:
ref_logo

fatteneder:
I suspect the problem has to with this part here, because its the place where the `model` and `space` come together and where `scale` for `draw_marker` is computed. https://github.com/JuliaPlots/Makie.jl/blob/0fc6e78d9d2bb583db2d2c0aee9a2e88b17f258e/CairoMakie/src/utils.jl#L35-L47

However, in the above MWE `is_data_space(space)` is false. Not sure why though ...

ffreyer:
For reference, what you get in `draw_marker` is:
- `pos` is an input position from scatter transformed from `space` to pixel space. This is usually the center of the marker.
- `scale` is `markersize` transformed from `markerspace` to pixel space. This transformation is dodgy in 3D because depth can affect scale there.
- `marker_offset` is `markeroffset` transformed from `markerspace` to pixel space. This is usually just `-0.5 * markersize` (centered marker) and `pos + markeroffset` should always be the lower left corner of the marker (or upper depending on how the coordinate system is set up).

> However, in the above MWE `is_data_space(space)` is false. Not sure why though ...

This makes sense because `project_scale` uses `markerspace` which is `:pixel` by default.

I think what's wrong here is the order of transformations, `scale` should apply to the width and height of the marker, but if you rotate first it applies to some mixture of x and y direction of the marker. (E.g. with a 90° rotation the x direction of the marker/image is the y direction in the canvas, so the y scale actually scales x.)

fatteneder:
> This makes sense because project_scale uses markerspace which is :pixel by default.

Ok.
From the docs of `scatter`:
> transform_marker::Bool = false controls whether the model matrix
       (without translation) applies to the marker itself, rather than just the
       positions. (If this is true, scale! and rotate! will affect the marker.)

As it is now `transform_marker=true` does not have an effect on `scale`. This is confirmed by running the above example where it is always `(75,75)` either the option is enabled or disabled. I would have expected `(150,75` with a scale of `(2,1)`.

>  I think what's wrong here is the order of transformations, scale should apply to the width and height of the marker, but if you rotate first it applies to some mixture of x and y direction of the marker. (E.g. with a 90° rotation the x direction of the marker/image is the y direction in the canvas, so the y scale actually scales x.)

This makes sense and I now observe some stretching of the markers too. But it is still wrong because of `scale` is not being modified.

I think I am still missing something. I can take a closer look again tonight.

fatteneder:
The `scale` issue is indeed independent from image markers, because it also happens if I replace the `scatter` call above with
```julia
scatter!(s,
        p,
        marker = :star,
        markersize = 75,
        rotations = rot,
        transform_marker=true
    )
```

Here are the outputs:
- `CairoMakie.jl`
logo

- `GLMakie.jl`
ref_logo

fatteneder:
The recorded and reference image now agree.
I opened a separate issue for the `transform_marker` problem.
This should be good to go now.

fatteneder:
> I suspect the problem has to do with this part here, because its the place where the `model` and `space` come together and where `scale` for `draw_marker` is computed.
> 
> https://github.com/JuliaPlots/Makie.jl/blob/0fc6e78d9d2bb583db2d2c0aee9a2e88b17f258e/CairoMakie/src/utils.jl#L35-L47
> 
> However, in the above MWE `is_data_space(space)` is false. Not sure why though ...

@ffreyer Merging the `is_data_space` and `is_pixel_space` branches would fix the issue, also in 3d. But then those two things would be the same. This does not feel right ...

ffreyer:
No, that's only correct if the scene has a pixel camera. Rather than that `model` should probably apply to `to_ndim(Vec4, s, 0)` in every branch. I'm not sure why it doesn't, maybe that was an oversight... I could see that breaking some other stuff though, but that should be easy to fix by passing an identity instead of `model`.

For scatter and text it might also be necessary to strip the translation part of model since GLMakie does that. For offsets:
https://github.com/MakieOrg/Makie.jl/blob/d3fc50373bb7106d7b5f58f4de0c797f95aaedb5/GLMakie/assets/shader/sprites.vert#L99
Not sure if it's happening for markersize as well.

ffreyer (approved):
Apparently scale going before rotate was wrong. I reverted that and adjusted `project_scale` to always include `model`. With that the scaling issue is fixed, and `translate!(s, ...)` also works. `rotate!(s, ...)` however does not. I think for that to work we need to rewrite all the scatter draw code to work with transformation matrices like the text draw code. But that doesn't need to happen here. The original issue is fixed.

fatteneder:
The failing CairoMakie tests might be due to this PR fixing the regression in the ref image.
Could somebody with write privileges update the `scatter with images.png`?

SimonDanisch:
Cool, thanks everyone :)

PR #2272, opened 2022-09-14
More precise and short wording and signature

KronosTheLate:
The original wording was imprecise, so I made it more precise and shorter.


src/interaction/inspector.jl
```diff
@@ -363,12 +363,12 @@ disable!(inspector::DataInspector) = inspector.plot.enabled[] = false
 
 """
     DataInspector(figure; kwargs...)
-    DataInspector()
+    DataInspector( ; kwargs...)
```
jkrumbiegel, line 366: you can remove the space before `;`

KronosTheLate: Done

Issue #1703, opened 2022-02-25
`Legend` fails on a 3D plot ?


BambOoxX:
On a basic example, it seems `Legend` does not work on 3D axes, is it normal ?

```julia
f,ax,p = meshscatter(rand(10),rand(10),rand(10))
Legend(f[1,2],p,"test")
```

fatteneder:
Hi,

I looked into this and I found the following:
1. an overload of `legendelements` for `MeshScatter` is indeed missing inside ``src/makielayout/layoutables/legend.jl``, e.g.
```julia
function legendelements(plot::MakieCore.ScenePlot, legend)
    LegendElement[LineElement()]
end
```
From what I can see overloads are also missing for `Surface, Heatmap`.
2. Instead of the plot `p` we have to pass the axis `ax` to `Legend`. This also happens e.g. with `lines`. Perhaps a better error message could help here.
3. We also have to pass a `label` to the actual plot command.

The following works now
```
f,ax,p = meshscatter(rand(10),rand(10),rand(10), label="some data")
Legend(f[1,2],p,"test")
```
and gives the output
![image](https://user-images.githubusercontent.com/26469701/155844498-09107e8f-8e1b-4e49-ad2e-4324f08ebbcb.png)

BambOoxX:
@fatteneder Thanks for your answer. I have to say I just started using Makie, so I wouldn't know how to correct this issue. 
Would you mind opening a PR for this since you seem to have found the source ?

fatteneder:
Yes, I can.

But first @SimonDanisch should confirm whether it is really the overload that is missing or something else. The fix seems 'hacky' to me, because `ScenePlot` is not exported from `MakieCore`.

PR #2276, opened 2022-09-14
Add hide/show interactions to `Legend`

fatteneder:
# Description

Adds the ability to toggle plot elements by clicking on their label inside a `Legend`.
Mimics `gnuplot`s toggling behavior.

demo_hideshow

The functionality relies on `<:AbstractPlot` objects having a `visible` attribute. If there isn't one, we still put a shade above the legend entry.

## Type of change

## Checklist

      I think this not needed.
      Should we add a ref image that uses `pick` and generates a plot with missing elements? And can I even add one myself?

### DIFF
--- a/src/makielayout/blocks/legend.jl
+++ b/src/makielayout/blocks/legend.jl
@@ -214,9 +214,9 @@ function initialize_block!(leg::Legend,
                 justification = map(leg.labeljustification, e.labelhalign) do lj, lha
                     return lj isa Automatic ? lha : lj
                 end
-                label = Label(scene; text=e.label, textsize=e.labelsize, font=e.labelfont, justification=justification,
-                              color=e.labelcolor, halign=e.labelhalign, valign=e.labelvalign)
-                push!(etexts, label)
+                push!(etexts,
+                      Label(scene; text=e.label, textsize=e.labelsize, font=e.labelfont, justification=justification,
+                            color=e.labelcolor, halign=e.labelhalign, valign=e.labelvalign))
 
                 # create the patch rectangle
                 rect = Box(scene; color=e.patchcolor, strokecolor=e.patchstrokecolor, strokewidth=e.patchstrokewidth,
@@ -238,7 +238,7 @@ function initialize_block!(leg::Legend,
                 push!(eshades, shade)
 
                 # add mouseevent to hide/show elements
-                events = addmouseevents!(label.blockscene, label.layoutobservables.computedbbox)
+                events = addmouseevents!(blockscene, shade.layoutobservables.computedbbox)
                 onmouseleftdown(events) do event
                     for el in e.elements
                         el.plot.visible[] = !el.plot.visible[]
--- a/src/makielayout/blocks/legend.jl
+++ b/src/makielayout/blocks/legend.jl
@@ -241,7 +241,10 @@ function initialize_block!(leg::Legend,
                 events = addmouseevents!(blockscene, shade.layoutobservables.computedbbox)
                 onmouseleftdown(events) do event
                     for el in e.elements
-                        el.plot.visible[] = !el.plot.visible[]
+                        isnothing(el) && continue
+                        if hasproperty(el.plot, :visible)
+                            el.plot.visible[] = !el.plot.visible[]
+                        end
                     end
                     shade.visible[] = !shade.visible[]
                     return Consume(true)
--- a/src/makielayout/blocks/legend.jl
+++ b/src/makielayout/blocks/legend.jl
@@ -55,6 +55,8 @@ function initialize_block!(leg::Legend,
     entrytexts = [Label[]]
     entryplots = [[AbstractPlot[]]]
     entryrects = [Box[]]
+    entryevents = [MouseEventHandle[]]
+    entryshades = [Box[]]
 
     function relayout()
         manipulating_grid[] = true
@@ -70,6 +72,7 @@ function initialize_block!(leg::Legend,
             title = titletexts[g]
             etexts = entrytexts[g]
             erects = entryrects[g]
+            eshades = entryshades[g]
 
             subgl = if leg.orientation[] == :vertical
                 if leg.titleposition[] == :left
@@ -89,10 +92,11 @@ function initialize_block!(leg::Legend,
                 end
             end
 
-            for (n, (et, er)) in enumerate(zip(etexts, erects))
+            for (n, (et, er, es)) in enumerate(zip(etexts, erects, eshades))
                 i, j = leg.orientation[] == :vertical ? rowcol(n) : reverse(rowcol(n))
                 subgl[i, 2j-1] = er
                 subgl[i, 2j] = et
+                subgl[i, 2j-1:2j] = es
             end
 
             rowgap!(subgl, leg.rowgap[])
@@ -174,6 +178,14 @@ function initialize_block!(leg::Legend,
         end
         empty!(entryplots)
 
+        for events in entryevents
+            clear!.(events)
+        end
+        empty!(entryevents)
+
+        [delete!.(eshades) for eshades in entryshades]
+        empty!(entryshades)
+
         # the attributes for legend entries that the legend itself carries
         # these serve as defaults unless the legendentry gets its own value set
         # TODO: Fix
@@ -192,6 +204,8 @@ function initialize_block!(leg::Legend,
             etexts = []
             erects = []
             eplots = []
+            eevents = []
+            eshades = []
             for (i, e) in enumerate(entries)
                 # fill missing entry attributes with those carried by the legend
                 merge!(e.attributes, preset_attrs)
@@ -200,9 +214,9 @@ function initialize_block!(leg::Legend,
                 justification = map(leg.labeljustification, e.labelhalign) do lj, lha
                     return lj isa Automatic ? lha : lj
                 end
-                push!(etexts,
-                      Label(scene; text=e.label, textsize=e.labelsize, font=e.labelfont, justification=justification,
-                            color=e.labelcolor, halign=e.labelhalign, valign=e.labelvalign))
+                label = Label(scene; text=e.label, textsize=e.labelsize, font=e.labelfont, justification=justification,
+                              color=e.labelcolor, halign=e.labelhalign, valign=e.labelvalign)
+                push!(etexts, label)
 
                 # create the patch rectangle
                 rect = Box(scene; color=e.patchcolor, strokecolor=e.patchstrokecolor, strokewidth=e.patchstrokewidth,
@@ -216,16 +230,32 @@ function initialize_block!(leg::Legend,
                     append!(symbolplots,
                             legendelement_plots!(scene, element, rect.layoutobservables.computedbbox, e.attributes))
                 end
-
                 push!(eplots, symbolplots)
+
+                # create a shade above label and marker to indicate hidden plots
+                shade = Box(scene; color=RGBAf(0.9,0.9,0.9,0.6), visible=false,
+                            strokewidth=0)
+                push!(eshades, shade)
+
+                # add mouseevent to hide/show elements
+                events = addmouseevents!(label.blockscene, label.layoutobservables.computedbbox)
+                onmouseleftdown(events) do event
+                    for el in e.elements
+                        el.plot.visible[] = !el.plot.visible[]
+                    end
+                    shade.visible[] = !shade.visible[]
+                    return Consume(true)
+                end
+                push!(eevents, events)
             end
             push!(entrytexts, etexts)
             push!(entryrects, erects)
             push!(entryplots, eplots)
+            push!(entryevents, eevents)
+            push!(entryshades, eshades)
         end
         relayout()
     end
-
 
     # trigger suggestedbbox
     notify(leg.layoutobservables.suggestedbbox)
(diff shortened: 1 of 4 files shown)


fatteneder:
We should also make it so that either clicking the title or right clicking anywhere in the legend box toogles all entries.

jkrumbiegel:
That's cool! Also it's probably a behavior many people expect from a legend nowadays.

I'm wondering how this can be made to work with multiple axes that share a legend. Usually the legend items are constructed separately in that case, or one plot object is passed to the constructor as a proxy for all others. Which would mean that no or only one facet would have its plot hidden with the current technique.

fatteneder:
I think I have not used `Legend` enough to understand what you are describing.
Could you maybe provide a MWE of the setup you have in mind?
Or is the example with `merge & unique` in the docs here close to what you want to do? https://docs.makie.org/v0.17.13/examples/blocks/legend/index.html

jkrumbiegel:
I just mean that semantically, one legend entry often describes multiple plots across facets. So you'd expect all of them to turn off, but the legend might have been created referencing only one (or even none) of them. So this method needs to be able to accomodate that scenario, where a list of related plots is passed by the user somehow.

fatteneder:
Still not sure if I understand correctly.
Are you maybe referring to this comment:
> We should also make it so that either clicking the title or right clicking anywhere in the legend box toogles all entries.

Now I could say what the problem would be here.

Nevertheless I could gather some more things from your comment that we should take care of:

- If there is no plot connected to a label, then we should disable the toggle. This can be done easily, I think.
- If one entry in a legend refers to multiple plot elements (which can be spread across more than one Axis), we want that clicking the label hides/shows all of them simultaneously. I did not test yet, but I thought the current implementation might do that already, because the mouse event goes through all the `<:LegendEntry`s that belong to a label's `element` connection, cf. this https://github.com/MakieOrg/Makie.jl/pull/2276/files#diff-93b8d91868f7576e0971bdc75440cc6f9bfa0562952fbd630d52cd2a84ccce36R243-R247
- Some doc examples show the use of `LineElement` etc. As it is now you can only add a single handle to that using the `plot` kwarg. We should allow to pass a vector of handles so that we can connect multiple plot elements to one legend entry, which (I think) is automatically done if you only use `Legend`.

Again, I am really sorry that I don't understand. I usually I am not that slow, I promise :smile:

jkrumbiegel:
Yeah maybe your last bullet point covers what I mean. I was thinking of facet plots where you have the same line plot let's say 12 times across 12 axes, but you haven't referenced all of them at once when creating the legend. I didn't check your implementation yet but if you added a plots kwarg where one could pass a vector that could take care of it.

fatteneder:
I added two three more things.
I illustrate them with the following MWE:
```julia
using GLMakie, Random
GLMakie.activate!()

Random.seed!(145)

f = Figure()

ax  = Axis(f[1,1])
lb = lines!(ax, 0:4, 0:4, color=:blue, label="lines 1")
lo = lines!(ax, 0:4, -4:0, color=:orange, label="lines 2")
sb = scatter!(ax, rand(3), rand(3), color=:blue, label="scatter 1")
so = scatter!(ax, rand(3), rand(3), color=:orange, label="scatter 2")
x = LinRange(0, 4, 100)
slb = band!(ax, x, cos.(x) .- 1, cos.(x) .- 2, color=:blue, label="band 1")
slo = band!(ax, x, sin.(x) .- 1, sin.(x) .- 2, color=:orange, label="band 2")

ax2 = Axis(f[2,1])
x, y, yerr = 1:2:20, 5 * rand(10), 0.4 * abs.(randn(10))
bb = barplot!(ax2, x, y; strokewidth = 1, color = :blue, label="barplot 1")
bo = barplot!(ax2, x, -y; strokewidth = 1, color = :orange, label="barplot 2")

Legend(f[1,2], ax)
Legend(f[2,2],
       [ PolyElement(plots = [lb, sb, slb, bb], color = :blue),
         PolyElement(plots = [lo, so, slo, bo], color = :orange) ],
       [ "blue", "orange" ], "Colors" )

f
```

1. Construct `<:LegenEntry`s from a vector handles. Clicking the entry then toggles all the plot elements in this group. If you watch close you can spot an issue where plot elements are toggled but not any shades in other legends. I think this is what Julius mentioned above. 
demo_custom_legend

2. A right click anywhere in a legend toggles all the entries in that legend. 
demo_toggle_all

3. Middle click anywhere in a legend brings shows all elements again and following middle clicks just switch to toggeling. 
demo_sync



Side note: `scatterlines` does not provide a `visible` attribute.

fatteneder:
I reworked the toggling strategy again so that it addresses the problem of partly hidden elements. We now use a hatched box if not all plots of an element are either visible or hidden. If all are hidden then we put a full box.

demo_new_toggeling

There is also a question about what to do with plot elements that are already hidden when the legend is created. Consider the following
```julia
using GLMakie, Random
GLMakie.activate!()

Random.seed!(145)

f = Figure()

ax  = Axis(f[1,1])
lb = lines!(ax, 0:4, 0:4, color=:blue,    label="lines 1", visible=true)
lo = lines!(ax, 0:4, -4:0, color=:orange, label="lines 2", visible=false)
Legend(f[1,2], ax)

f
```
The plot would start with only `lb` shown in the axis but with two legend entries which both appear to be shown. Clicking the second entry would show `lo` but put a shade on its legend entry. I think we have to live with that, otherwise it will be difficult to decide when to put a shade when there are multiple plot elements associated to a legend entry. The above example also appears artificial to me, because when I plot stuff I usually want to see it too (unless I tinker around with `Block` elements or so).



Otherwise, I think this would be ready for a review.

fatteneder:
Here is a list of plot elements for which toggling does not work:
- `spy`
- `timeseries` (is that still broken? I could not see any lines)

I could not test for the following ones because they cannot be to added to `Legend`s yet. IIUC the problem here is that there is no default implementation for `legendelements`, cf. #1703.
- `heatmap`
- `image`
- `annotations` and `text` (might not even make sense)
- `mesh`
- `meshscatter`
- `volumeslices`
- `volume`
- `surface`
- `wireframe`

Lastly: The legend symbol for `arrows` could be improved. Right now it works like that of a `scatterlines`, but we could make it so that the arrow head sets at one of the ends of the symbol (or at both).

fatteneder:
The `WGLMakie` failure seems to be due to a problem with `LinePattern`.
However, I cannot test locally because `test WGLMakie` fails for a different reason on my machine.

asinghvi17:
```julia

│ Details:
│ MethodError: no method matching iterate(::Nothing)
│ Closest candidates are:
│   iterate(!Matched::Union{LinRange, StepRangeLen}) at range.jl:872
│   iterate(!Matched::Union{LinRange, StepRangeLen}, !Matched::Integer) at range.jl:872
│   iterate(!Matched::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at dict.jl:712
│   ...
│ Stacktrace:
│   [1] get_plot_visibilities(entry::LegendEntry)
│     @ Makie ~/work/Makie.jl/Makie.jl/src/makielayout/blocks/legend.jl:739
│   [2] (::Makie.var"#1701#1726"{Legend, Makie.LinePattern, ColorTypes.RGBA{Float32}, Vector{Vector{Box}}, Vector{Vector{Box}}, Vector{Vector{Union{Nothing, Makie.MouseEventHandle}}}, Vector{Vector{Box}}, Vector{Vector{Vector{AbstractPlot}}}, Vector{Vector{Label}}, Vector{Union{Nothing, Label}}, Scene, Scene})(entry_groups::Vector{Tuple{Union{Nothing, AbstractString}, Vector{LegendEntry}}})
│     @ Makie ~/work/Makie.jl/Makie.jl/src/makielayout/blocks/legend.jl:249
│   [3] #invokelatest#2
│     @ ./essentials.jl:729 [inlined]
│   [4] invokelatest
│     @ ./essentials.jl:726 [inlined]
│   [5] notify
│     @ ~/.julia/packages/Observables/PHGQ8/src/Observables.jl:169 [inlined]
│   [6] setindex!(observable::Observable, val::Any)
│     @ Observables ~/.julia/packages/Observables/PHGQ8/src/Observables.jl:86
│   [7] Legend(fig_or_scene::GridPosition, contents::Vector{PolyElement}, labels::Vector{String}, title::String; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
│     @ Makie ~/work/Makie.jl/Makie.jl/src/makielayout/blocks/legend.jl:574
│   [8] Legend(fig_or_scene::GridPosition, contents::Vector{PolyElement}, labels::Vector{String}, title::String)
│     @ Makie ~/work/Makie.jl/Makie.jl/src/makielayout/blocks/legend.jl:561
└
```
was the error message in the docs, everything else seems to pass tests.

mgyoo86:
@fatteneder Hello, I'm new to Makie and was looking for this show/hide feature. This pull request seems to be exactly what I'm looking for. Is this pull request going to be merged into the master soon or is there a serious problem? 
If anyone can let me know if there is a workaround to accomplish something similar, I would greatly appreciate it.

fatteneder:
The last problem was that `legendelements` has no default implementation (would be fixed by https://github.com/MakieOrg/Makie.jl/pull/2348).
Without that this feature here would only be available for selected plot types.

Regarding 2348: That PR would be ready as is.
There were objections regarding the use of blank markers for legend entries.
I agree that we could do better with the markers with Makie.
However, I did not progress with that, because there is no a stable API for `Legend` yet and, hence, this is why it stalled.
Cf. https://github.com/MakieOrg/Makie.jl/issues/2727


> Is this pull request going to be merged into the master soon or is there a serious problem?

I don't plan to work on 2727 myself soon. I personally would be ok with merging 2348 and then this here.

artemsolod:
> @fatteneder Hello, I'm new to Makie and was looking for this show/hide feature. This pull request seems to be exactly what I'm looking for. Is this pull request going to be merged into the master soon or is there a serious problem? If anyone can let me know if there is a workaround to accomplish something similar, I would greatly appreciate it.

As a workaround for now I use toggles to show/hide some traces, maybe it will suite your task
https://docs.makie.org/stable/reference/blocks/toggle/#toggle

mgyoo86:
Thanks for the kind explanation and heads up for the workaround.
As the description/demonstration above is really cool,
I hope this show/hide feature can be merged to master sometime in the future.
Meanwhile, I will try the toggle feature and also take a look at this "fa/hideshow" branch to see if it works on top of the current master version.
Thanks again for the advice!

matthewgcooper:
This feature would be great! I've been waiting for something like this since I started using Makie a few years ago.

ffreyer:
I merged master and 
- refactored event handling to avoid creating multiple mouse state machines (this is rather slow, especially with WGLMakie I believe)
- fixed and simplified right click (dropped checks to avoid visible triggers - these should be pretty cheap)
- did bit of general cleanup (rename some variables, avoid some temporary arrays, clean up visible listeners, fix cleanup of entryrects)

TODO:
- Right click doesn't "toggle" but makes everything visible.
- Middle mouse button interaction is gone
- Safeguards against legend entries that do not keep track of plots (deprecate old struct and warn?)
- fix test errors
- add new tests/refimages

> If you watch closely you can spot an issue where plot elements are toggled but their shades in other legends are not.

Need to think about whether this is fixable without too much effort

ffreyer:
> > If you watch closely you can spot an issue where plot elements are toggled but their shades in other legends are not.
> 
> Need to think about whether this is fixable without too much effort

Ah this is not the case anymore, 

I think I added/updated everything that's necessary to merge this pr now.

docs/src/reference/blocks/legend.md
```diff
@@ -327,6 +327,60 @@ Label(f[3, 1:2, Top()], "titleposition = :left, orientation = :horizontal\nnbank
 f
 ```
 
+## Legend Interactivity
+
+The Legend can be interacted with to show or hide connected plots.
+A left click on a specific element will toggle the visibility of the connected plots.
+A right click anywhere in the Legend will toggle the visibility of all plots associated with every element.
+A middle click will toggle all plots if they have the same visibility state or make all visible if they do not.
+Note that if you construct elements (MarkerElement, PolyElement, etc) yourself, you need to also pass a plot or vector of plots as the first argument to enable these interactions.
```
Mo-Gul, line 336:
```suggestion
Note that if you construct elements (MarkerElement, PolyElement, etc.) yourself, you need to also pass a plot or vector of plots as the first argument to enable these interactions.
```

src/makielayout/blocks/legend.jl
```diff
@@ -794,6 +945,44 @@ function legend_position_to_aligns(t::Tuple{Any, Any})
...
+    end
+    return
+end
+
+function get_plot_visibilities(entry::LegendEntry)
+    visibilities = Observable{Bool}[]
+    for element in entry.elements
+        if (element !== nothing) && hasfield(typeof(element), :plots) && (element.plots !== nothing)
+            for plot in element.plots
+                !hasproperty(plot, :visible) && continue
```
SimonDanisch, line 965: I feel like we should just guarantee that every plot has a visibility field.

ffreyer: We should, but I see that as wasted effort before #4630.

SimonDanisch:
Closing in favor of: https://github.com/MakieOrg/Makie.jl/pull/4927

PR #2281, opened 2022-09-18
Add support for pasting text into a textbox

### LINKED ISSUE #2266: Pasting text into a textbox?
Pasting text (with `ctrl+v`) into the textbox does not seem to work - at least on Windows.

```julia
using GLMakie

f = Figure()

tb = Textbox(f[1, 1], placeholder = "Try copy/pasting text into me")

gl_screen = display(f)
wait(gl_screen)
```

Relevant dump from status:

```julia
pkg> status GLMakie
Status `C:\[project path]\Project.toml`
  [e9467ef8] GLMakie v0.6.13
```

ConnectedSystems:
Support pasting text into a textbox.

Fixes #2266 

## Type of change

## Checklist


ffreyer:
I was a bit worried this would paste multiple times even on press but it seems fine to me. It behaves pretty much the same as the textbox on github does for me, i.e. paste - pause - repeat pastes with ctrl+v held down. It might be good to check on windows and mac as well though. 

For testing we could do something similar to the camera and mouse state machine tests in https://github.com/MakieOrg/Makie.jl/blob/master/test/events.jl. I.e. create a figure with a textbox, trigger some events to select it, fill the clipboard with something (you can pass content to it) trigger down+release of `ctrl+v` and check if the contents end up in textbox.

ConnectedSystems:
Hi @ffreyer 

For the test I'm using a mouse click to simulate the textbox being selected.
Seems I have to display a window for the click to occur correctly. Is this okay, given none of the other tests seem to display a window?

Another issue is that I clear the figure and recreate the scene to ensure the right control + v combination works. But the test fails for this attempt, and I don't know why: it always evaluates the `stored_string` in the second textbox to be `nothing`.

Would you mind taking a look?

ffreyer:
I don't think display should be necessary. Maybe you need remove either the `tb.focused = true` or the mouse selection bit?

ConnectedSystems:
Yes, the `focused` flag could be removed but without the mouse selection, the textbox will not be selected.

Without `display()`, the window size will be `(0,0)`, and so the textbox selection (via mouse selection) will not work either.

Is there a more straightforward way to selecting widgets that I'm not aware of?

ffreyer:
I cleaned up the tests, they should be working now.

> Without display(), the window size will be (0,0), and so the textbox selection (via mouse selection) will not work either, as it relies on clicking the textbox location.

There are multiple sources for the window size which can be used. `events.window_area` is what a backend reports when/after it creates a window. I.e. if there is no window this will report 0 size. There is also `scene.px_area` which initially reports what size a scene wants to be and then gets resized to match the window, layouting etc. This is what is actually used for most mouse events in makielayout. (Technically they can still depend on point picking but I removed most of these cases a while ago for better performance.)

ffreyer:
> no clipboard command found, please install xsel or xclip

Is it worth setting one of those up to test this?

ConnectedSystems:
> > no clipboard command found, please install xsel or xclip
> 
> Is it worth setting one of those up to test this?

I guess it would be good for completeness, but I'll have to leave this up to you as I don't run *nix at the moment. Sorry!

src/makielayout/blocks/textbox.jl
```diff
@@ -210,6 +213,26 @@ function initialize_block!(tbox::Textbox)
...
+                local content::String = ""
+                local no_err::Bool = true
+                try
+                    content = clipboard()
+                finally
+                    for char in content
+                        if is_allowed(char, tbox.restriction[])
+                            insertchar!(char, cursorindex[] + 1)
+                        else
+                            no_err = false
```
SimonDanisch, line 227: This doesn't really seem like an error case, so I'd rather call this `all_chars_allowed` or something like that.
Also, we should add a catch case and rethrow, no? Or at least warn...

SimonDanisch (changes requested):

src/makielayout/blocks/textbox.jl
ConnectedSystems: Excuse the clumsy implementation.

My thinking here was that it would be obvious that the copy/paste stopped.
I take your point, though, it would make more sense if I check for disallowed characters in the string first, and just abort the paste if any is found. Will make changes in the morning

SimonDanisch: No worries, this is great :) Maybe just use:
```julia
if all(char-> is_allowed(char, tbox.restriction[]), content)
   foreach(char-> insertchar!(char, cursorindex[] + 1), content)
   return Consume(true)
else
    return Consume(true)
end
```

ConnectedSystems: Thanks! 

That second `Consume` returns `false` instead though, correct? But yes, I get the idea.

SimonDanisch (approved):

src/makielayout/blocks/textbox.jl
```diff
@@ -210,6 +213,24 @@ function initialize_block!(tbox::Textbox)
...
+                catch err
+                    @warn "Pasting failed: $err"
+                    return Consume(false)
+                end
+
+                if all(char -> is_allowed(char, tbox.restriction[]), content)
+                    foreach(char -> insertchar!(char, cursorindex[] + 1), content)
+                    return Consume(true)
+                else
+                    return Consume(false)
```
jkrumbiegel, line 230: Shouldn't there be some sort of informative warning if pasting fails because some chars fall under the restriction? Such as it is, nothing would happen, right?

src/makielayout/blocks/textbox.jl
```diff
@@ -210,6 +213,24 @@ function initialize_block!(tbox::Textbox)
 
     on(events(scene).keyboardbutton, priority = 60) do event
         if tbox.focused[]
+            ctrl_v = (Keyboard.left_control | Keyboard.right_control) & Keyboard.v
+            if ispressed(scene, ctrl_v)
+                local content::String = ""
+                try
+                    content = clipboard()
+                catch err
+                    @warn "Pasting failed: $err"
```
jkrumbiegel, line 222: I would rather say "Accessing the clipboard failed"

src/makielayout/blocks/textbox.jl
SimonDanisch: I thought about this... My conclusion was, that this is a user restriction, so the user should chose, right?
And if the user wants to warn other users, they could include a warning in their `restriction` function...

jkrumbiegel: it's mostly for stuff like "only allow numbers" but of course the validation function should notify about that anyway. maybe it doesn't matter

jkrumbiegel: haven't used restriction in practice yet

SimonDanisch:
Thanks :)

PR #2282, opened 2022-09-18
Make GLMakie relocatable

### LINKED ISSUE #1480: GLMakie, relocatable with PackageCompiler ?
I created an interactive simulation/animation using GLMakie of the Vicsek Model.  Everything runs fast and without any issues. Based on that,  i wanted to make a standalone app with the create_app() function of the package PackageCompiler. This also works well on my machine, but when i try to run it on another machine (which is the entire purpose of an standalone app), i get the error: 
Vicsek_Error - Copy
I belive, this is due to the fact GLMakie works with absolute paths, which are only valid on the machine that created the app. 
There is an package called "RelocatableFolders" which provides a workaround to that problem. The @path macro can be put infront of every @__dir__ macro to handle absolute paths and provide an alternative local copy of the contents of the path. 
I belive the function loadshader(name) = normpath(joinpath(@__DIR__, "..", "assets", "shader", name)) in the GLMakie.jl file is the root of  all evil. So i tried to use the @path macro there. normpath(@path joinpath(@__DIR__, "..", "assets", "shader", name)). It still works on my machine but does not create a local copy of the objects in the given path, therefore it gives me an error on a different machine aswell: 
ArgumentError: not a path: `C:\Users\ank10ki\.julia\packages\GLMakie\pFGSp\assets\shader\postprocessing\fullscreen.vert`not a path: `C:\Users\ank10ki\.julia\packages\GLMakie\pFGSp\assets\shader\postprocessing\fullscreen.vert`
Therefore, i have a few questions at hand: 
1. Am i doing something horribly wrong here ? 
2. If not, is GLMakie relocatable at all, due to its apparent reliance on absolute paths  ? 
3.  Is this the right place to ask this question, or should i aks the guys from PackageCompiler or RelocatableFolders ?

ConnectedSystems:
Changes to allow GLMakie to be compiled into an app.

Fixes #1480

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/GLMakie.jl
+++ b/GLMakie/src/GLMakie.jl
@@ -17,6 +17,7 @@ using Makie: ClosedInterval, (..)
 using Makie: inline!, to_native
 using Makie: spaces, is_data_space, is_pixel_space, is_relative_space, is_clip_space
 import Makie: to_font, glyph_uv_width!, el32convert, Shape, CIRCLE, RECTANGLE, ROUNDED_RECTANGLE, DISTANCEFIELD, TRIANGLE
+import Makie: RelocatableFolders
 
 using ShaderAbstractions
 using FreeTypeAbstraction
@@ -43,7 +44,9 @@ export Sampler, Buffer
 struct GLBackend <: Makie.AbstractBackend
 end
 
-loadshader(name) = normpath(joinpath(@__DIR__, "..", "assets", "shader", name))
+const GL_ASSET_DIR = RelocatableFolders.@path joinpath(@__DIR__, "../assets")
+const SHADER_DIR = RelocatableFolders.@path joinpath(GL_ASSET_DIR, "shader")
+loadshader(name) = joinpath(SHADER_DIR, name)
 
 # don't put this into try catch, to not mess with normal errors
 include("gl_backend.jl")
--- a/GLMakie/src/GLAbstraction/GLShader.jl
+++ b/GLMakie/src/GLAbstraction/GLShader.jl
@@ -220,7 +220,7 @@ function gl_convert(cache::ShaderCache, lazyshader::AbstractLazyShader, data)
     # Tuple(Source, ShaderType)
     if all(paths) do x
             isa(x, Tuple) && length(x) == 2 &&
-            isa(first(x), String) &&
+            isa(first(x), AbstractString) &&
             isa(last(x), GLenum)
         end
         # we don't cache view & templates for shader strings!
@@ -231,7 +231,7 @@ function gl_convert(cache::ShaderCache, lazyshader::AbstractLazyShader, data)
         end
         return compile_program([shaders...], fragdatalocation)
     end
-    if !all(x-> isa(x, String), paths)
+    if !all(x -> isa(x, AbstractString), paths)
         error("Please supply only paths or tuples of (source, typ) for Lazy Shader
             Found: $paths"
         )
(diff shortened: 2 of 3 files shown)


SimonDanisch:
Cool, thank you :) 
I've tried this, but it hangs indefinitely :( 
image
This has been running for more than 20minutes now...
Did you do anything special? I made a simple new package with:
```julia
module MakieApp

using GLMakie

function julia_main()::Cint
    screen = display(scatter(1:4))
    # wait(screen) commented out to test if this blocks anything, but didn't change anything
    return 0 # if things finished successfully
end

end # module MakieApp

```
and then run:
```julia
using Pkg; Pkg.activate() # just the default project
using PackageCompiler
create_app("dev/MakieApp", "MakieExe")
```
Run this with Julia 1.8 and `PackageCompiler@2.0.9`

ConnectedSystems:
I wonder if you've encountered a variation of this issue I tripped over: https://github.com/JuliaLang/PackageCompiler.jl/issues/721

The compile command I usually use is: `create_app(".", "MakieExe", incremental=true)`
Assumes you're inside the project root of course.

I can confirm it hangs for me too when `incremental=false` is used instead.

ConnectedSystems:
Does GLMakie pull anything in from Makie itself during precompilation?

I notice there's a fair few places which rely on `@__DIR__` in the Makie project, maybe that's what is causing it to hang?

SimonDanisch:
Ah, incremental=true works indeed :) 

>Does GLMakie pull anything in from Makie itself during precompilation?

Hm, shouldn't really... Maybe the waiting logo.

Btw, did you get around the App installing MKL at runtime? That's very annoying, also it fails to unzip it for some reason -.-

SimonDanisch:
Ended up with: 
```julia
create_app(".", "makieapp"; force=true, incremental=true, include_transitive_dependencies=false)
```

Maybe we can add this to the docs?

GLMakie/src/GLMakie.jl
```diff
@@ -43,7 +44,9 @@ export Sampler, Buffer
 struct GLBackend <: Makie.AbstractBackend
 end
 
-loadshader(name) = normpath(joinpath(@__DIR__, "..", "assets", "shader", name))
+const GL_ASSET_DIR = RelocatableFolders.@path joinpath(@__DIR__, "../assets")
```
greimel, line 47:
```suggestion
const GL_ASSET_DIR = RelocatableFolders.@path joinpath(@__DIR__, "..", "assets")
```

ConnectedSystems:
> Btw, did you get around the App installing MKL at runtime? That's very annoying, also it fails to unzip it for some reason -.-

No sorry. The app I'm working on doesn't do the download. The chatter online indicates to me that MKL is only downloaded if it is required, but given this example app is just Makie by itself, I don't know why my app does not download it... unless it is and I've somehow hidden the download progress bar.

I would have thought adding MKL.jl to the dependencies would include the required binaries during compile, but I just tried it out and it did not - still triggered the download at runtime.

> Maybe we can add this to the docs?

Sure, I can do this. Any preference on where though?

SimonDanisch:
Huh, I got this one time, then i inserted a print statement to debug the path and then it was gone...I guess the print statement didn't fix it but forced a recompile of Makie which fixed it...or something!? :(

ConnectedSystems:
That's weird, I did a `] precompile` as well.

Although it finished pretty much instantly so I should have thought something was weird.
I'll try again later.

Issue #2204, opened 2022-08-07
updating mesh in WGLMakie doesn't work as expected.


lazarusA:
There two issues with the following:
- the first one is that the initial plot corresponds to the last slider index, it should be the first one? 
- the second one is that update the mesh doesn't work as expected. There are some artefacts.

```Julia
using WGLMakie, JSServe, Observables
using GeometryBasics
using JSServe: Slider
using Random, Colors
function do_viz(t)
    spheret = @lift(normal_mesh(Tesselation(Sphere(Point3f(0), 1.0f0), $t)))
    wireframe(spheret; color = :white)
    mesh!(spheret; color = @lift($t*rand(RGBf)/32), transparency=true)
    return current_figure()
end

WGLMakie.activate!()
page = Page(offline=true, exportable=true)
app = JSServe.App() do session::Session
    index_slider1 = Slider(8:8:32, style="width: 350px;")
    slider1 = DOM.div("tess: ", index_slider1, index_slider1.value)
    fig  = do_viz(index_slider1.value)
    rpage = JSServe.record_states(session, DOM.div(DOM.div(slider1), fig))
    return rpage
end
open("app_tsphere.html", "w") do io
    println(io, """
    <body>
      <div id="container"></div>
        """)
      show(io, MIME"text/html"(), page)
      show(io, MIME"text/html"(), app)
      println(io, """
    </body>
    </html>
    """)
end
```

<img width="414" alt="Screenshot 2022-08-07 at 11 37 04" src="https://user-images.githubusercontent.com/19525261/183284735-3e8f849c-b63d-4f58-b82d-707d795f9d0c.png">

<img width="417" alt="Screenshot 2022-08-07 at 11 32 06" src="https://user-images.githubusercontent.com/19525261/183284675-04576ae1-1ef5-41b0-a6da-30c9056c615f.png">

PR #2297, opened 2022-09-23
fix changing input types

SimonDanisch:
Incredibly enough, nobody seems to have run into this in a long time...
Seems like changing input types is pretty unpopular.
Changing types _after_ `convert_arguments` isn't expected to work, since that usually changes plot type, but changing the input type to `convert_arguments` should have been working. Might have regressed in some of the observable refactors.


lazarusA:
This looks like https://github.com/MakieOrg/Makie.jl/issues/2204, although is for WGLMakie, so it might be a completely different issue.

SimonDanisch:
It's quite the different issue ;)

PR #2299, opened 2022-09-24
better performance for Menus and fix clicks on items

SimonDanisch:
`addmouseevents!` seems to ignore clicks and I couldn't really get to the bottom of the problem :( 
I started using `scene.events` directly to debug things, and since it fixed the issue, should be faster, uses less listeners and does all the work in one function, i sticked with that solution ;) 

I'm not 100% sure though, if it's correct in regards to Consume behaviour, but it seems to work pretty well.
We should also think about adding a z-index value, since in some cases the z-value seems too low.
Maybe, we should also think about having `overdraw=true` for scenes, so that we can better create pop-up scenes without worrying about any z-index.

This also defaults to the first item in the option list instead of `nothing`, which at least to me seems more practical. Willing to remove that change, though ;)

### DIFF
--- a/src/makielayout/blocks/menu.jl
+++ b/src/makielayout/blocks/menu.jl
@@ -83,7 +83,7 @@ function initialize_block!(m::Menu; default = 1)
                 d == :down ? bottom(bbox) : min(top(bbox) + h, top(blockscene.px_area[]))))
     end
 
-    menuscene = Scene(blockscene, scenearea, camera = campixel!, clear=true, backgroundcolor=:black)
+    menuscene = Scene(blockscene, scenearea, camera = campixel!, clear=true)
     translate!(menuscene, 0, 0, 200)
 
     onany(scenearea, listheight) do area, listheight
--- a/src/makielayout/blocks/menu.jl
+++ b/src/makielayout/blocks/menu.jl
@@ -182,10 +182,16 @@ function initialize_block!(m::Menu; default = 1)
 
     was_inside_options = false
     was_inside_button = false
-    was_pressed_options = Ref(false)
-    was_pressed_button = Ref(false)
+
     e = menuscene.events
 
+    # Up events are notoriusly hard,
+    # especially if we want to react only to presses that went down inside an element & went up inside
+    # was pressed needs to be tracked per item, and also needs to be invalidated outside `mouse_up`
+    # which makes the state handling especially annoying
+    # TODO, move this back to mousestatemachine, which does exactly this
+    was_pressed_options = Ref(false)
+    was_pressed_button = Ref(false)
     function mouse_up(butt, was_pressed)
         if butt.button == Mouse.left
             if butt.action == Mouse.press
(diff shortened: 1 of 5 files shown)


src/makielayout/blocks/menu.jl
```diff
@@ -202,41 +179,58 @@ function initialize_block!(m::Menu; default = nothing)
...
+    was_inside_button = false
+    e = menuscene.events
+    onany(e.mouseposition, e.mousebutton, priority=typemax(Int)) do position, butt
+        mp = screen_relative(menuscene, position)
+        if Makie.is_mouseinside(menuscene)
+            is_over = mouseover(menuscene, optionpolys, optiontexts)
+            pressed = ispressed(menuscene, Mouse.left)
+            if is_over
+                was_inside_options = true
+                if pressed
```
ffreyer, line 192: I don't think this matters/happens in practice but `ispressed` could technically trigger multiple times because it's true as long as the left mouse button is held down. It's also a bit different from previous behavior which reacted to left-mouse-up on an option instead of mouse down (or technically any action while left-mouse-down) if I'm not mistaken.

`if butt.button == Mouse.left && butt.action = Mouse.up` should match master behavior a bit better. Might also be slightly faster, though irrelevant compared to picking.

SimonDanisch: Ah good point, I think it does actually matter, since I noticed that sometimes it seemed to press 2 times, so it closed the menu again!

ffreyer: Yea I think that's happening in the one below

src/makielayout/blocks/menu.jl
```diff
@@ -202,41 +179,58 @@ function initialize_block!(m::Menu; default = nothing)
...
+                end
+            elseif was_inside_options
+                was_inside_options = false
+                optionpolycolors[] = map(eachindex(optionstrings[])) do i
+                    i == m.i_selected[] ? m.cell_color_active[] :
+                    iseven(i) ? to_color(m.cell_color_inactive_even[]) :
+                        to_color(m.cell_color_inactive_odd[])
+                end
+            end
+        elseif Makie.is_mouseinside(blockscene)
```
ffreyer, line 214: This can just be `else` since `is_mouseinside` just checks area and `blockscene` covers the full window.

ffreyer: Actually maybe not since the mouse can trigger outside the window as well... Nvm

ffreyer: Hmm, maybe it should trigger outside as well. If you move the mouse quickly you might otherwise suppress the `was_inside_button` true -> false switch.

ffreyer:
Some more interaction issues/differences:

- When moving outside the expanded selection menu, the last hovered element remains highlighted.
- After selecting a new element and opening the menu again, the selected element is highlighted as hovered and the previous element is highlighted as selected. This updates when the mouse hovers one of the options
- Clicking outside the expanded selection menu does not close the menu anymore.

SimonDanisch:
Thanks for the thorough review :) I should have fixed all problems now - and found a new appreciation for mousestatemachine :D 
But, maybe we could refactor it to be without observables and just use a stateful struct? Then we can do everything in one function as well:
```julia

mousestate1= mousestatemachine(scene, plots...)
mousestate2 = mousestatemachine(scene, other_plots...)
on(events.mouse) do mouse 
   if ishover(mousestate1, mouse) 
       .... 
    end
    if ishover(mousestate2, mouse)  
        ....
    end
end
```
Well, that's for another PR and I also still need to take a closer look at mousestatemachine. Maybe it's not as bad to have one observable and multiple functions registered...But then one needs to fight for the event via consume, which seems easier if the whole state of something complex as a menu is handled in one function 🤷

jkrumbiegel:
If mouseevents ignores clicks it's probably because they are registered on multiple objects or scenes at once I think. Unless the mouse down and mouse up events are already not received. But if they are, then some other condition must not be met, which could be a bug in the implementation but also in our event system.

ffreyer:
Back when I reworked the event system (i.e. introduced priority) I also noticed some differences between how mouse state machine works and how my OS reacts. I reworked it back then but didn't push it because I didn't want to add more complexity to the pr. I noted some of the oddities in the tests but I never got back to making a pr for it...
https://github.com/MakieOrg/Makie.jl/blob/087dfbdd22acead0c14eb5d4c507ea63cf418885/test/events.jl#L361-L435

jkrumbiegel:
ah good that you collected some of those, it's much better to test them in this way, although I only found some of the edge cases by trying to break the code and acting weird with the gui. Some of these cases are not well defined but I think if the big OSes differ in behavior we should use that as reference. Like with the mentioned case where left and right click happen in a nested fashion

SimonDanisch:
I say we merge this and leave further refactors & tests of the event system / mousestate machine for another day?

SimonDanisch:
Are you sure you have all the new changes? That sounds exactly like all the issues I fixed...

ffreyer:
Oh nvm, I didn't pull correctly. Looks good to me :laughing:

Issue #2162, opened 2022-07-21
Artificial colored edges in CairoMakie when saving an image to PDF.


fatteneder:
### MWE

```julia
using CairoMakie
CairoMakie.activate!()
using FileIO
using Makie

img = load(Makie.assetpath("doge.png"))
i = image(permutedims(img)[:,end:-1:1])
save(joinpath(@__DIR__, "mwe.pdf"), i)
```

Output: 
![image](https://user-images.githubusercontent.com/26469701/180289323-3a7d98ae-59f1-463f-9fcf-f4528bdb76ea.png)

Adding `fxaa=true` to `image` does not fix it either.
Saving as `png` does work though.

Related: #2080

fatteneder:
Could this be an issue with a transparent background in the `png` file?

jkrumbiegel:
Have you looked at this with multiple different pdf viewers? Just to make sure we're not looking at an implementation-dependent artifact here. Or is the artifact really baked into a bitmap?

fatteneder:
I checked with
- Evince
- Okular
- Firefox
- Chromium
and all show the same artifacts.

Here is the PDF file if you want to check yourself: [mwe.pdf](https://github.com/JuliaPlots/Makie.jl/files/9162295/mwe.pdf)

fatteneder:
Fixed by #2304

PR #2304, opened 2022-09-27
use premultiplied alpha for cairo argb surfaces

jkrumbiegel:
# Description

Cairo needs premultiplied alpha for image ARGB32 image surfaces, while everywhere else they don't. So we never noticed, which has been leading to all kinds of artifacts with transparent bitmaps. This should fix all of those hopefully.

Before:

```julia
CairoMakie.activate!()
x = collect(0:2.0)
y = collect(0:2.0)
z = x .* y'
f = Figure()
ax = Axis(f[1, 1], title="CairoMakie", backgroundcolor = :white)
heatmap!(ax, x, y, z, colormap=(:viridis, 0.5))
f
```

grafik

After:

grafik

## Type of change

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -557,6 +557,14 @@ end
 
 regularly_spaced_array_to_range(arr::AbstractRange) = arr
 
+function premultiplied_rgba(a::AbstractArray{<:ColorAlpha})
+    map(premultiplied_rgba, a)
+end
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Should fix #2162 as well but I can't select that in the link menu for some reason

SimonDanisch:
Yay finally :) 

Before:
before
With this PR:
after

SimonDanisch (approved):

PR #2306, opened 2022-09-27
Refactor display code to be Screen centric

SimonDanisch:
Loosely a continuation of #2012, but really Rabbit holed into a refactor of the display code.
Screen is now the holder of all display related information like `device_scaling_factor`, desired output type and things like visibility

This PR adds:

* saving/recording with backend keyword
* better and consistent screen constructors between backends, with a clear purpose of what they do
* cleaned up display code and more consistent use of display
* removed backend_display + inline! since they're somewhat redundant at this point
* CairoMakie opens plots in OS image viewer/browser 

..And some more docs and clean ups to come


CairoMakie/src/display.jl
```diff
@@ -0,0 +1,171 @@
...
+
+function Makie.backend_showable(::Type{Screen}, ::MIME{SYM}) where SYM
+    supported_mimes = Base.setdiff(SUPPORTED_MIMES, DISABLED_MIMES)
+    return string(SYM) in supported_mimes
+end
+
+"""
+    to_mime_string(mime::Union{String, Symbol, MIME})
+
+Converts anything like `"png", :png, "image/png", MIME"image/png"()` to `"image/png`.
```
jkrumbiegel, line 120:
```
to `"image/png"`.
```

jkrumbiegel (approved):

GLMakie/src/screen.jl
```diff
@@ -3,13 +3,141 @@ const ZIndex = Int
...
+* `decorated = true`: shows the window decorations or not.
+* `title::String = "Makie"`: Sets the window title.
+* `fullscreen = false`: Starts the window in fullscreen.
+* `debugging = false`: Starts the GLFW.Window/OpenGL context with debug output.
+* `monitor::Union{Nothing, GLFW.Monitor} = nothing`: Sets the monitor on which the Window should be opened.
+
+## Preproccessor
+* `oit = true`: Enles order independent transparency for the window.
+* `fxaa = true`: Enables fxaa (anti-aliasing) for the window.
+* `ssao = true`: Enables screen space occlusion, which gives 3D meshes a soft shadow towards their edges.
```
ffreyer, line 34: We have ssao off by default on master because a bunch of people had issues with the extra vram it uses

ffreyer: Also shouldn't these be post-processors since they run after standard rendering?

jkrumbiegel: and it's screen space _ambient_ occlusion, which simulates natural shadowing at inner edges and crevices

jkrumbiegel: towards the edges sounds to me like every edge gets like a gradient effect :)

SimonDanisch:
Final summary of changes: 

## GLMakie Bug fixes

* Fix Heisenbug for multiple windows, where shaders got compiled for the wrong window (OpenGL context), which works fine on most nvidia hardware but may crash other OpenGL implementations.
* Due to the above fix, it became clear that the scene clean up after displaying on a screen wasn't happening correctly, which got fixed
* SSAO/FXAA/transparency_weight_scale were a global, which was fine until we got multi window support. Now, ssao/fxaa/oit are screen options, which you can set via `GLMakie.activate!(ssa=true)` or per screen via `GLMakie.Screen(ssa=true)`.

## Saving + record

* new backend argument, allowing one to save a file with a specific backend (`save("plot.png", figure; backend=GLMakie)`). Applies also to `record` and `display`.
* `record` and `save` now support forwarding all `Backend.Screen` options, which enables you to use `record(...; px_per_unit=2.0)` for higher resolution videos with CairoMakie, or e.g. `save(...; backend=GLMakie, ssao=true)`.
* All backend screen types have been renamed to `Backend.Screen`, which was previously not the case. This shouldn't break too much code, since the only backend where one was interacting with `Screen` is GLMakie, which was already named `Screen`. In the future, it will be more common to interact with `Screen` directly for all backends. Use `?Backend.Screen` to see all screen options.

## Speed and allocation improvements for record

We gained some substantial speed improvements for `record`, by reusing more screen state and directly writing out to a file via ffmpeg [$2231](https://github.com/MakieOrg/Makie.jl/pull/2231).

### GLMakie
```
PR:     171.387 ms (57775 allocations: 4.93 MiB)
master: 322.244 ms (50714 allocations: 32.58 MiB)
```

### CairoMakie
```
PR:     647.093 ms (3076666 allocations: 174.11 MiB)
master: 807.175 ms (3070484 allocations: 335.48 MiB)
```

### WGLMakie
```
PR:     1.450 s (2200125 allocations: 189.61 MiB)
master: 1.628 s (2192751 allocations: 279.25 MiB)
```


## Other

* adding more tests for display / record
* CairoMakie now opens a browser/os-image-viewer when plotpane is disabled or `display(fig)` is called explicitely.
* deprecation of `Makie.inline!(false/true)` which was only disabling show, which isn't very helpful (can just be done by `fig;` or `return nothing`) and was instead making trouble accidantely disabling show.
* deprecating `GLMakie.set_window_config!(; screen_config...)` in favor of `GLMakie.activate!(; screen_config...)`.
* deprecating internal function `backend_display` in favor of `display`.
* cleaning up `colorbuffer(figlike)` and `colorbuffer(screen)`, making it faster and more consistent across backends.
* clean up of unused code and making code more consistent between backends
* backend is now registered as module not it's own `GLMakieBackend` struct. Deprecated internal global `current_backend` in favor of `CURRENT_BACKEND` and renaming `register_backend!` to `set_active_backend!`. One can now query the current backend via `Makie.current_backend()::Module`
* `Backend.activate!(; screen_config...)` now always accepts attributes for the backends screen config, and writes it to the theme. Alternatively, one can use `set_theme!(Backend=(screen_config...,))`
* better warnings for wrong ffmpeg flags in `record`/`Record`

PR #2314, opened 2022-10-02
Add `stack=:dodge` and `stack=:waterfall` to `barplot` to allow easy waterfall charts creation

daschw:
# Description

This PR implements the options

```julia
barplot(...; stack=:waterfall, ...)
barplot(...; stack=:dodge, ...)
```

I frequently use what some people call waterfall charts ([plotly](https://plotly.com/python/waterfall-charts/), [python](https://github.com/chrispaulca/waterfall)) to visualize individual positive and negative components that add up to a net result.

Suppose I have a vector `y` of such components and colors `c` to indicate the sign.

```julia
using CairoMakie
green, red = Makie.wong_colors()[[3, 6]]
y = [3, 2, 1, -4, 1, -3, -2, 4, -1, 3]
c = map(v -> v < 0 ? red : green, y) # color bars by sign
```

To achieve something similar with Makie I can recalculate *stacked* `y` values and corresponding `fillto` values.

```julia
y_waterfall = cumsum(y)
fill_waterfall = [0; y_waterfall[1:end-1]]
barplot(y_waterfall, color=c, fillto=fill_waterfall)
```
waterfall

With this PR I can do the same thing with just:

```julia
barplot(y, color=c, stack=:waterfall)
```

Now suppose I want to compare multiple net results grouped by `x` with individual components grouped by `dodge`.

```julia
x = repeat(1:2, inner=5)
dodge = repeat(1:5, outer=2)
```

Here I have to preprocess my data in some way like this:

```julia
y_dodge = similar(y)
fill_dodge = similar(y)
for i in unique(x)
    inds = findall(==(i), x)
    y_dodge[inds] .= cumsum(y[inds])
    fill_dodge[inds] .= [0; y_dodge[inds][1:end-1]]
end
barplot(x, y_dodge, color=c, dodge=dodge, fillto=fill_dodge)
```
stackdodge

It would be even more cumbersome if `x` and `dodge` were not sorted properly. This PR enables me to just use:

```julia
barplot(x, y, color=c, dodge=dodge, stack=:dodge)
```

I'm not really sure about the name choices `:waterfall` and `:dodge` though and open to better suggestions.
Something similar was also asked for in https://discourse.julialang.org/t/waterfall-plot-makie-jl/79988.

## Type of change

## Checklist

### DIFF
--- a/docs/examples/plotting_functions/barplot.md
+++ b/docs/examples/plotting_functions/barplot.md
@@ -82,6 +82,30 @@ barplot(tbl.x, tbl.height,
 \end{examplefigure}
 
 \begin{examplefigure}{}
+
+using CairoMakie
+CairoMakie.activate!() # hide
+Makie.inline!(true) # hide
+green, red = Makie.wong_colors()[[3, 6]]
+
+y = [3, 2, 1, -4, 1, -3, -2, 4, -1, 3]
+c = map(v -> v < 0 ? red : green, y) # color bars by sign
+
+fig = Figure()
+waterfall_ax = Axis(fig[1, 1], title="stack=:waterfall")
+barplot!(waterfall_ax, y, color=c, stack=:waterfall)
+
+x = repeat(1:2, inner=5)
+dodge = repeat(1:5, outer=2)
+
+dodge_ax = Axis(fig[2, 1], title="stack=:dodge")
+barplot!(dodge_ax, x, y, color=c, dodge=dodge, stack=:dodge)
+
+fig
+
+\end{examplefigure}
+
+\begin{examplefigure}{}
 ```julia
 colors = Makie.wong_colors()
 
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
Thanks for the PR! My recurring feeling with `barplot` is that it kind of handles too much, all the stacking, dodging, labelling. Wouldn't a `waterfall` recipe work as well? Or is that not good because you then need to implement dodging again?

daschw:
> My recurring feeling with barplot is that it kind of handles too much, all the stacking, dodging, labelling.

Yeah, I was expecting these concerns with this PR and I totally understand them.

> Wouldn't a waterfall recipe work as well? Or is that not good because you then need to implement dodging again?

Sure, this would work as well. I guess I wouldn't even have to reimplement dodging, because I could use `barplot` in the `waterfall` recipe and just calculate corresponding `y` and `fillto` values, right?. I'm not sure if you are suggesting a waterfall recipe for my personal use or as addition to Makie. In the second case, I could rewrite this PR. I was already thinking of this alternative, but I decided to hook into barplot, because it seemed like the smaller change.

jkrumbiegel:
Yes as a recipe in Makie on top of barplot. I thought the up down coloring scheme from the waterfall might need its own logic anyway? Or is that not the typical mode of using it

daschw:
closed by #2416

Issue #2311, opened 2022-09-30
Allow `object.visible[] = false` for all objects (or something similar)


rafaqz:
I'm building some interfaces in GLMakie that have a few different plot windows, but largely keep the same widgets besides a few that need to change.

I can show and hide plots simply with `plot.visible[] = false`, but showing and hiding the few controls specific to each window appears to be difficult. 

My first attempt involved simply rebuilding the whole interface, but it took a few seconds, which makes the interface feel very sluggish. Just hiding plots is very snappy in comparison. The problem is buttons, sliders, toggles etc. Once they are in a `GridLayout()` it's not so easy to hide them/swap them for a different widget for a new context, keeping all of their observable connections intact.

What I propose is to extend the `visible` fields on plots to all objects - sliders, buttons and whole GridLayouts, so we can more easily control what is shown in the figure at any time. This does not need to affect the layout - they can still be accounted for in spacing, just not actually visualised. I'm not sure how easy this is to do internally.

jkrumbiegel:
This would mostly be about making `scene.visible = false` work first, as then you could easily set that on a `Block`'s container `Scene`. `GridLayout`s on the other hand are not display objects, they just control positioning, so you can't "hide" them. But one can traverse a layout's children and hide those, which is what you mean anyway I guess.

PR #2317, opened 2022-10-05
fix invisible scenes

SimonDanisch:
Fix for e.g. https://github.com/MakieOrg/Makie.jl/issues/2311


rafaqz:
This is amazing, thanks for adding it so quickly.

PR #2320, opened 2022-10-07
Add description for how to fix dynamic linking issue with GLFW

KronosTheLate:
This PR adds the fix that worked for me under the troubleshooting for GLMakie, to avoid the user having to go searching the thread of the linked issue for fixes.


fatteneder:
(I think) This PR https://github.com/JuliaLang/julia/pull/46976 tries to solve the problem. If it will make it into `v1.9` then hopefully the problem is gone for all times.

KronosTheLate:
That would be awesome! However, I see no reason not to help people solve the problem themselves as much as possible in the meantime.

SimonDanisch:
thanks :)

PR #2321, opened 2022-10-08
Rich text

jkrumbiegel:
# Description

Adds a rich text data structure which allows to plot strings with colors, subscripts, superscripts, different fonts and fontsizes specified internally. Also adds the ability to specify fonts by symbols such as `:regular` and passing a top-level dictionary that resolves these symbols.

This PR is breaking in two ways:
- Symbols are not resolved equivalently to strings anymore. This should mostly not matter because all our examples used strings and the symbol behavior was not advertised.
- Doing `Figure(..., font = xyz)` will not have the expected effect anymore, because all the `Block`s like `Axis` which previously inherited their fonts from this field do not do this anymore. Instead, they are set to `:regular` or `:bold`. The new way to change fonts should therefore be `Figure(..., fonts = (; regular = xyz, bold = abc))`. The benefit of this is obviously that you can theme more than one font category, before you would have had to change Axis3 title, Axis title, Legend title all separately if you didn't want the bold font used there.

Examples:

```julia
using Random
Random.seed!(123)

scatter(rand(100, 2),
    axis = (
        yscale = log10,
        xscale = log,
        title = rich(
            "H",
            subscript("2"),
            "O is a ",
            rich("great", font = :bold_italic, color = :dodgerblue1),
            " way to hydrate"
        )
    ),
)
```

image

Using rich superscripts also fixes the mismatching exponents because TeX Gyre Heros only had superscripts 1 to 3:

Before:

image

After:

image

## Checklist


jkrumbiegel:
I think this is at a state where it can be reviewed. The most breaking part of this PR is that `font = ...` doesn't have the effect of changing most fonts anymore (it would still change that of `text` plots I think). But I also think the mechanism to supply a dict of fonts basically and call them semantic names like `:regular` and `:bold` is much better and more themeable. I decided against supporting a boolean bold and italic flag in `rich` because choosing the font directly is just that much more precise and less "do what I mean", while setting `font = :bold` seems very convenient and clear to me as well.

ffreyer:
There should be some printing for rich text so it doesn't throw a display error (e.g. `rstr = rich("H", subscript("2"), "O")` does.)

The attribute pass-through also needs a bit more work. `rotation` seems be ignored entirely for example, and `textsize` can't be changed inside `rich`. I think it would be good if all the per-character attributes worked, just in case they become useful later.

jkrumbiegel:
Yes textsize was still missing, rotation I'm not sure about because of how it would affect text layouting.. What I did also want to add was manual kerning but I can do that later, it can be useful for mixtures of italic and normal fonts.

And yes rich should get some printing, it fails because it's an AbstractString (to fit inside the existing infrastructure) but doesn't really overload any of the methods.

src/makielayout/lineaxis.jl
```diff
@@ -575,6 +575,11 @@ function get_ticks(::Automatic, scale::Union{typeof(log10), typeof(log2), typeof
     get_ticks(LogTicks(WilkinsonTicks(5, k_min = 3)), scale, any_formatter, vmin, vmax)
 end
 
+# the hyphen which is usually used to store negative number strings
+# is shorter than the dedicated minus in most fonts, the minus glyph
+# looks more balanced with numbers, especially in superscripts or subscripts
+const MINUS_SIGN = "−"
```
SimonDanisch, line 581: I actually think the hyphen looks more pleasing in the screenshots... But not a hill to die on for me ;)

src/makielayout/lineaxis.jl
```diff
@@ -674,9 +679,13 @@ function get_ticks(m::MultiplesTicks, any_scale, ::Automatic, vmin, vmax)
     dvmax = vmax / m.multiple
     multiples = Makie.get_tickvalues(LinearTicks(m.n_ideal), dvmin, dvmax)
 
-    multiples .* m.multiple, Showoff.showoff(multiples) .* m.suffix
+    multiples .* m.multiple, showoff_minus(multiples) .* m.suffix
 end
 
+# Replaces hyphens in negative numbers with the unicode MINUS_SIGN
+function showoff_minus(x::AbstractVector)
```
SimonDanisch, line 686: Seems like quite the effort as well, especially considering that the tick code is pretty performance sensitive?
Do we memoise this somehow?

SimonDanisch:
Looks really great :) 
I guess this still needs some new reference tests...

Not sure if the minus sign change is really worth it. I think hyphen looks better and also doesn't steal as much screen estate:
hyphens-vs-minus

jkrumbiegel:
> I think hyphen looks better and also doesn't steal as much screen estate
Oh really? I thought the minus looked better (it makes sense to me because it's meant to be the character for that kind of thing), also the hyphen gets quite hard to read in the smaller exponents. Matplotlib for example also appears to use the longer minus, I just checked some example pictures. And in terms of screen real estate it's only a couple of pixels :)

src/makielayout/lineaxis.jl
jkrumbiegel: yeah this one was a rather hacky way to get it working, as I couldn't access `showoff`'s logic directly..

SimonDanisch:
As I said, it's not a hill to die on for me ;) But seeing that we need to regex replace things out of the tick strings and having less pixels in the plot just doesn't seem worth it... Maybe we can keep it for a separate PR and discuss it with a wider audience? Seems to be quite subjective after all...

jkrumbiegel:
Yeah sure we can do that in a separate one as well. Just as I said, the main motivation was that the hyphen looks kind of cramped in the superscripts.

jkrumbiegel:
Ok I removed the minus, added printing, added some more docs, ability to change fontsize and offsets, fixed rotation, added a reference test. Any other points?

Moelf:
2 cents: nobody will complain if it's a minus for `minus`, but people WILL complain why this looks weird/like shit/ etc etc if you use `hyphen`, at some point. (especially if this is carried over to other fonts, where hyphen looks very different from minus, somehow?)

jkrumbiegel:
We can try that in a separate PR with input from more users.

ffreyer:
I was thinking something like `text(rich("Test", textsize = 10:5:25))` should work. Right now it's getting ignored for me.

Also mixed constant and per-character attributes should combine cleanly. This for example throws an error for me:
```julia
scatter(rand(100, 2),
    axis = (
        yscale = log10,
        xscale = log,
        title = rich(
            "H",
            subscript("2", color = :dodgerblue1),
            "O is a ",
            rich("great", font = :bold_italic, color = [RGBf(0.2, b, 0.4) for b in 0.6:0.05:0.8],
            " way to hydrate"
        )
    ),
)
```

jkrumbiegel:
Ah yeah I actually don't like per character attributes that much, and I think the per-section logic is going to be 99% of the real use cases. Like making one word italic or red or so.

ffreyer:
`fig, ax, p = text(rich("Test", textsize = 30), textsize = 10)` also remains at 10 for me.

jkrumbiegel:
Ah that's because it's fontsize... I've talked to Simon about the overall textsize / fontsize discrepancy and we're going to go with fontsize only in the next breaking release with rich text. We've waited too long with that already but now is as good a time as any, as long as we include clear error messages it should cause no further trouble for users than doing a search and replace.

docs/examples/plotting_functions/text.md
```diff
@@ -199,3 +199,131 @@ Legend(f[1, 2], ax)
...
+f = Figure(fontsize = 30)
+Label(
+    f[1, 1],
+    rich(
+        "H", subscript("2"), "O is the formula for ",
+        rich("water", color = :cornflowerblue, font = :italic)
+    )
+)
+
+str = "A beautiful rainbow"
```
ffreyer, line 227: Maybe this should be all caps to make it easier to see the changes in font size?

jkrumbiegel: Good idea, you can change it if you want :)

asinghvi17:
Can I suggest that instead of specifying bold etc fonts directly, we allow the user to simply pass a `FontFamily` object, which could contain preresolved fonts?  This would allow people to use arbitrary styles (small caps, semibold, etc).

What I thought of was something like this:

```julia
...
font = "TeX Gyre Heros Makie"
# or
texgyre_family = FontFamily("TeX Gyre Heros Makie"; semibold = "Calibri") # for example, can set any font
...
font = texgyre_family
```

This way, one could simply specify via the x-label formatter if the font should be bold etc.

jkrumbiegel:
> Can I suggest that instead of specifying bold etc fonts directly, we allow the user to simply pass a `FontFamily` object, which could contain preresolved fonts? This would allow people to use arbitrary styles (small caps, semibold, etc).
> 
> What I thought of was something like this:
> 
> ```julia
> ...
> font = "TeX Gyre Heros Makie"
> # or
> texgyre_family = FontFamily("TeX Gyre Heros Makie"; semibold = "Calibri") # for example, can set any font
> ...
> ```

That's kind of how it works in this PR, no?

```julia
fonts = Dict(
    :regular => "Makie Regular",
    :bold => "Makie Bold",
    :semibold => "Calibri" # whatever
)

set_theme!(fonts = fonts)

text(1, 2, text = "hi", font = :semibold)
```

asinghvi17:
Hmm, that does make sense.  Basically what I wanted to make sure of was that (a) people can still set `axis.font="...."`, and that (b) the user should be able to set set `fonts = "Calibri"`, and then `axis.font = :bold` would resolve to Calibri Bold automatically, i.e., the user should deal in font families, not fonts - which is what most layman users would think in.  One doesn't set the font to "Calibri Bold" in Word, after all :D

For (a) it looks like you can already do this, but (b) would basically just require an easier way to set the font family.  One way we could do this is by somehow auto-converting when a String is given to `fonts`, into a similar dictionary...MakieSlides had some utilities for this as I recall.  Have to step away for a bit but will post where that is later, it's basically just searching through FreeType for a bold, italic, etc version of the font.

jkrumbiegel:
> One doesn't set the font to "Calibri Bold" in Word, after all

I personally really dislike choosing fonts not by name but by abstract attributes. I know that there are some fonts that only have the regular, italic, bold, bold italic combination, but I tend to want to use fonts that have like 30 variants. That's why I wanted to make it easy to pick exactly the fonts you want (by name) and not make it convenient to do something that is often correct but bites you in all other cases. That's matplotlib's system to me, for example. I once dove down all the way into their font handling code just to understand why my font theme didn't resolve correctly, and it was all heuristics about font weights and stuff. The family plus style name is actually really easy to do find out and use in my opinion, so I stuck with that in Makie.

asinghvi17:
So I was looking into this, and ended up diving a bit into Fontconfig/Freetype interaction.  Ended up writing a basic integration which allows us to use Fontconfig.  Will make that a PR to this one in the next couple of days.

Good news: figured out how to use Fontconfig with Makie, meaning that we can use that as a somewhat better search engine...

This raises another edge case which should be solved before this gets merged - how should we handle font names?  Some fonts have them labelled as oblique, instead of italic, which is fair enough but liable to get users confused.  Additionally, most style names in fonts come with spaces - meaning that we would either have to force the users to use the exact syntax of the font, or create some sort of lookup engine for the most common ones (possibly both!)

P.S. With this, we can actually use single-file fonts with multiple faces!  If you tried "Helvetica Bold" as a font now, it wouldn't work, since we currently have no way of distinguishing between multiple fonts in a single file in Makie.

jkrumbiegel:
I did already think a lot about these problems years back when I rewrote the font finding in FreeTypeAbstraction, maybe have a look at the closed PRs there, there might still be bits of discussion. Anyway, I settled on using font names with a shortest match policy, and that seems to have worked well so far. I wouldn't want you to spend tons of time on getting fontconfig to work if that doesn't clearly improve on a problem we currently have. And so far I haven't seen that problem, so if you want to go forward, probably best to collect problematic use cases first and present them here for further discussion :)

asinghvi17:
Well, the problem as I see it is the following:

`to_font("Helvetica Bold")` returns `FTFont (family = Helvetica, style = Medium)`, which is not what was requested - and on assigning this font, there's no change in the plot.  This is despite the fact that there is a bold font in the ttf file, but the font-finding algorithm we have now cannot find it, since it can only access the 0th index of the file.

Similar things apply to basically any font which a user downloads as a single file, which is somewhat prevalent on common font sites.

I just put up #2400 which implements this interface - after looking at the Fontconfig docs, it was surprisingly easy.  Do give it a look!

SimonDanisch (approved):
Really great to have this!! :) 
From what I can tell, this is not breaking, right?
If it is, we should tag current master as a patch/bugfix release first.
Apart from this, I think this is ready to be merged! 
We should not forget to add the new refimage, though!

jkrumbiegel:
The description notes two ways in which the PR is breaking, so it should go into a minor release. I was thinking to bundle it with the  fontsize refactor.

stevengj:
It would be nice to open a separate PR that contains the hyphen/minus change, as suggested above.

[Matplotlib does this](https://matplotlib.org/stable/gallery/text_labels_and_annotations/unicode_minus.html), and it seems like all professional typesetting (e.g. LaTeX) uses a different (wider) glyph for the minus sign than for a hyphen.   It's quite painful to ask users to do this manually.

Issue #2264, opened 2022-09-12
Png output in HTML-capable environments


jkrumbiegel:
I'm wondering if it would make sense to output png with an `<img>` wrapper tag instead of as raw image data when the environment can display HTML. The problem is that if I set `px_per_unit = 2` to have a crisp output from CairoMakie, the image can become very large but still kind of blurry in something like Pluto or Quarto. I'm using a retina screen Macbook, so what happens is that one image pixel is displayed with four screen pixels by default. I guess in order to override this, one would have to pass half the image size as `<img width=...` etc, hence the idea to output HTML when png is activated. Not sure how the scale factor should be determined, though. Maybe one would just divide by `px_per_unit`.

jkrumbiegel:
This seems to do it if called directly, question is if we should use this if `CairoMakie.activate!(type = "png")` was used. I think so, because then the image is the same size no matter the `px_per_unit` value.

```julia
function Base.show(io::IO, ::MIME"text/html", f::Figure)
    w, h = widths(f.scene.px_area[]) ./ 0.75
    iob = IOBuffer()
    Base.show(iob, MIME"image/png"(), f)
    b64 = Base64.base64encode(String(take!(iob)))
    print(io, "<img width=$w height=$h src=\"data:image/png;base64, $(b64)\"/>")
end
```

ffreyer:
https://github.com/MakieOrg/Makie.jl/blob/904e7f6370101fed2fad2580d980b2686ec705f4/src/display.jl#L488-L496

Looks like this got added

PR #2333, opened 2022-10-13
Default to `px_per_unit=2` in CairoMakie

ericphanson:
# Description

1 is way too blurry in my opinion. I generally use `px_per_unit=8` but the file sizes get a bit large, so I figured 2 was less controversial.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/CairoMakie.jl
+++ b/CairoMakie/src/CairoMakie.jl
@@ -31,7 +31,7 @@ include("primitives.jl")
 include("overrides.jl")
 
 function __init__()
-    activate!()
+    activate!(; px_per_unit=2)
 end
 
 include("precompiles.jl")


ffreyer:
You can also set a resolution directly in `Figure`. With `px_per_unit = 2` the final image would be twice the size you set, which seems like a weird default to me. It would also make the size different between backends

ericphanson:
Ah ok, maybe this approach is not right then. Perhaps the default resolution should be higher instead? I'm not really sure, I just know "blurry plots [when viewed on high-resolution screens] by default" doesn't seem ideal.

Moelf:
What background? CairoMakie can produce svg which shouldn't be blur?

ericphanson:
it can; CairoMakie produces PNGs by default, which I think is fine; high density SVGs can be a big perf cliff for IDEs, saving to disk etc (every single point gets serialized so if you have a massive scatter plot, you have a massive file). So I think PNGs by default is a reasonable choice. PNGs are also supported by more software, so it makes it more portable/usable.

One can do `CairoMakie.activate!(; px_per_unit=8)` (or whatever value they want) to change it globally, or set the resolution in the `Figure`, or set the scaling at `save` time, or use SVGs. So there's definitely *options* to improve things. 

So my gripe here is just about the default settings.

Moelf:
svg is best for publication, and I think having massive amount of scattered points is, rare? why not use 2d histogram...

ericphanson:
sure, which is why folks should use SVG for publications? That doesn't mean it should be the default.

A massive amount of points is somewhat rare, and a scatter plot of them is not optimal. The problem is that if you do plot a scatter w a massive number of points (e.g. due to a typo or not realizing) it can crash or hang your julia session, browser, or ide, depending on what goes wrong. That's a bad enough thing that it shouldn't be the default, IMO. Also, apparently it had other quirks, discussed in the PR which changed the default to PNG: https://github.com/JuliaPlots/CairoMakie.jl/pull/147.

jkrumbiegel:
I also always increase the px_per_unit when I use CairoMakie, so I'm in principle open to a higher value there. The problem is that you cannot set pixel density metadata on an image in a way that would be respected by IDEs / webapps like Pluto.

When you display a plot with the `image/png` mime type, the png image data is written out as a blob and pasted in some `` tag in the output cell / plot pane. But, and that's the reason for the default setting of 1, if the Figure had resolution 800x600 and px_per_unit is 2, then the output image will take 1600x1200 css pixels of space (if it's available). So on retina displays it will still be blurry, as it still uses 4 native pixels per image pixel / css pixel.

At some point the viewer space will run out and the image will not get larger anymore, only at that point do you get pixel density improvement or less effective blurriness on your screen. But this only works if you don't have enough space, which may or may not be the case.

So my preferred solution for the future, which isn't perfectly implemented yet, is to use the html mime type for output. Then we can write `` so the image will always have the correct base size, just using a higher pixel density.

jkrumbiegel:
Compare with https://github.com/MakieOrg/Makie.jl/issues/2264

Moelf:
do the viewer in e.g. VS Code respect this:
https://www.w3.org/TR/2003/REC-PNG-20031110/#11pHYs

?

jkrumbiegel:
No, basically nothing respects density metadata, aside from maybe Photoshop and similar tools.

Moelf:
since we (Julia) writes the viewer panel, can we not support it?

jkrumbiegel:
it's not a scalable solution if it only works in vscode :)

Moelf:
fine, blurry then.

jkrumbiegel:
With the html thing I'm proposing it wouldn't be blurry

PR #2334, opened 2022-10-13
move poly to MakieCore

jw3126:
https://github.com/JuliaGeo/GeoInterface.jl/pull/71#discussion_r994977229


jw3126:
Bump! Is there some arguments against this? Would be really helpful for us to have lightweight access to poly.

SimonDanisch:
Yes, thank you, I think this is a good idea, although it's not really clear on where to draw the line... But lets just merge this for now, if it helps your use case! :)

jw3126:
@SimonDanisch thanks for merging! What are the downsides of declaring more polytypes in MakieCore?

jkrumbiegel:
it's just that makiecore is supposed to be lightweight, and the more we add the less it is that

SimonDanisch:
Also we really want to start having a "real" solution, where you can just address any plot/recipe without taking on the dependency ;)

jw3126:
> Also we really want to start having a "real" solution, where you can just address any plot/recipe without taking on the dependency ;)

Sounds awesome!

jw3126:
> it's just that makiecore is supposed to be lightweight, and the more we add the less it is that

But do `@recipe` calls make the package more heavy weight? I am not talking about moving any implementations to MakieCore. Just adding more `"some docs" @recipe(...)` to MakieCore.

SimonDanisch:
I don't really think it matters in this case, and won't for a couple more, as long as we don't need further package dependencies. It'll just a bit random, which recipes make it into the exclusive MakieCore club ;)

jw3126:
Can we have a MakieCore release? Should I bump the version?

Issue #1683, opened 2022-02-18
REPL can be extremely slow after plotting with GLMakie


rafaqz:
Its hard to make an MWE for this, as it doesn't happen the first few times, or completely predictably after that.

But after displaying a large, detailed 1000*1000 heatmap in GLMakie.jl, showing another object in the REPL can take 20s. It doesn't seem to happen (or as much) with a lower resolution heatmap.

I could profile this with ProfileView.jl, producing this output from a function that plot and array with makie then displays the array in the repl:
![2022-02-18-152554_1920x1080_scrot](https://user-images.githubusercontent.com/2534009/154703519-9fdc587d-de4f-4b14-993c-c66822279f46.png)

Nearly all of the time is in glfw3.jl, SwapBuffers line 696. The GLMakie.jl window was mostly sitting idle during this time, displaying the heatmap.

rafaqz:
The weird thing is that this is happening _after_ the plot has rendered. Are there constant updates happening even with no changes? 

I'm wondering if it has anything to do with my opengl terminal (alacritty). It doesn't seem to affect the vscode REPL.

SimonDanisch:
I see...Maybe the renderloop doesn't quit correctly.
Can you try:
```julia
screen = display(figure)
...
GLMakie.destroy!(screen)
```

rafaqz:
What's slow is displaying the array after showing the heatmap:

```julia
using Makie, GLMakie
A = rand(1000, 1200)
Makie.heatmap(A)
display(A)
```
But this takes essentially no CPU time at all. Its either all GPU or some task waiting problem. For me it takes about 30 seconds to display the heatmap in the repl with the GLMakie window open, and shows all in glfw3.jl in ProfileView.

My suspicion is that each number string being streamed to the opengl terminal are during `show` is triggering a refresh of some sort in GLFW, although I'm not sure if that makes sense.

rafaqz:
Running that code with 4000 * 2000 matrix pretty much kills my laptop.

SimonDanisch:
What platform is this? What terminal?

SimonDanisch:
And does the MWE reproduce consistently for you?

SimonDanisch:
For me `display(A)` is instant

rafaqz:
The heatmap renders nearly instantly, so I don't think this is #1427. 

[...]

Issue #1427, opened 2021-11-03
GLMakie 100% gpu utilization to display a 2000x2000 image


MariusDrulea:
Displaying a 2000x2000 image with GLMakie leads to 100% GPU utilization, almost blocking the GPU. I am curious why is this happening, 'cause a GPU can render a lot more and for a lot less utilization.  I have Nvidia Quadro M2000M.

```
x = Observable(rand(2000, 2000))
fig2, ax2, plotobj2 = plot(x)
```

![image](https://user-images.githubusercontent.com/27725572/140184719-19566b7b-ba1d-46b3-9909-5b5c10b909b7.png)

SimonDanisch:
Ah, I think this is a bug I recently discovered and haven't fixed yet... I think this currently creates a mesh for each pixel 😅

MariusDrulea:
A small update:
```
xl = rand(2000 * 2000) # big 1D array (so no image)
plot(xl)
```
This too leads to a GPU usage of nearly 100%.

SimonDanisch:
I haven't gotten around to fixing the performance but yet, so it still creates a mesh for every pixel...

SimonDanisch:
It's a bit hairy to bring back the optiomization to heatmap, where everything is just rendered as one big quad.
The easiest work around for now should be to use `image(...)` which should behave in most cases pretty similar to heatmap.

MariusDrulea:
@SimonDanisch : `image(rand(2000, 2000))` only eats 8% of the GPU, awesome. It's just that `image` does not support NaN for transparency https://github.com/JuliaPlots/Makie.jl/issues/1230, but I guess this is easy to fix right? Or is there any other way to mask out some values when displaying the image? https://discourse.julialang.org/t/glmakie-masking-values-when-plotting-images/66508/7
`image` (with transparency)  and `plot` would be sufficient for my purposes, as I plot only small data. 

However, `heatmap` and `plot` are very un-optimal and shall still be fixed in my opinion.

jariji:
I'm not sure if this issue is only for heatmaps but I'm also having an issue with scatterplots. WGLMakie 0.4.7 is giving 100% CPU and 8% GPU usage on a few 1000-point scatters in Chrome.

SimonDanisch:
Can you post an MWE?

jariji:
```jl
using StaticArrays, WGLMakie, AlgebraOfGraphics

[...]

Issue #1412, opened 2021-10-28
Plotting 2000x2000 image is slow. A large amount of allocations is created.


MariusDrulea:
I have started to use GLMakie more intensively and noticed it is slow for the images I use. Bellow you can find a sample code.
heatmap, plot, image all takes about 0.15 seconds per call, and I have about 10 images to show, so this leads to 1.5 seconds, quite a lot. You can notice the amount of allocations per call: `1.00 M allocations: 50.724 MiB`. Looks like each value of the 2000x2000 image is copied and allocated into something else. This can be done in a more optimal way, a single allocation for all the values at once. I am currently debugging the issue, any hint is welcome.

```
julia> using GLMakie

julia> x = rand(2000, 2000)

julia> @time heatmap(x)
 57.865795 seconds (111.09 M allocations: 6.655 GiB, 2.91% gc time)

julia> @time heatmap(x)
  0.168210 seconds (1.01 M allocations: 50.836 MiB)

julia> @time heatmap(x)
  0.225403 seconds (1.01 M allocations: 50.836 MiB)

julia> @time plot(x)
  0.349197 seconds (1.29 M allocations: 67.006 MiB, 54.55% compilation time)

julia> @time plot(x)
  0.170603 seconds (1.00 M allocations: 50.724 MiB)

julia> @time plot(x)
  0.180033 seconds (1.00 M allocations: 50.724 MiB)

julia> @time plot(x)
  0.192559 seconds (1.00 M allocations: 50.724 MiB)

julia> @time image(x)
  1.026312 seconds (2.70 M allocations: 153.704 MiB, 3.97% gc time, 81.16% compilation time)

julia> @time image(x)
  0.152998 seconds (1.02 M allocations: 51.078 MiB)

julia> @time image(x)
  0.159101 seconds (1.02 M allocations: 51.078 MiB)
```

SimonDanisch:
>When I display a single 2000x2000 image in the REPL, the GPU stays close to 100% most of the time. To me this looks like an over-use of GPU. Too high update frequency?

You could actually try using your own renderloop:

Create a function like this one https://github.com/JuliaPlots/Makie.jl/blob/master/GLMakie/src/rendering.jl#L17=,
change the render criteria from time based to whenever you update the image and then replace the renderloop with:
```julia
GLMakie.set_window_config!(renderloop=your_renderloop)
```

SimonDanisch:
[...]

Issue #678, opened 2020-08-04
Excessive GPU usage in MacOS


dcabecinhas:
Opening any Makie window (even an empty scene) results in a "GPU utilization" of ~40%, an "Energy Impact" above 200 (numbers from Activity Monitor.app), and makes the laptop fan run at maximum speed.

Julia 1.5.0
Makie 0.11.1
MacOS 10.15.6
MacBook Pro (Retina, 13-inch, Late 2013) [Intel integrated graphics card]

SimonDanisch:
What GLMakie version?

SimonDanisch:
If you're not 0.1.9, try updating to that and run the timing again ;)

dcabecinhas:
It is happening with GLMakie 0.1.9. I'm pasting the relevant package versions.
Let me know if you need more info or have any suggestion regarding pinpointing the issue.

```
(@v1.5) pkg> st
Status `~/.julia/environments/v1.5/Project.toml`
  [537997a7] AbstractPlotting v0.12.6
  [f7f18e0c] GLFW v3.2.2
  [e9467ef8] GLMakie v0.1.9
  [ee78f7c6] Makie v0.11.1
```

dcabecinhas:
The issue seems to have been introduced in Makie 0.1.5.

With Makie 0.1.4 I can zoom and pan a plot and have only low single digit %s for both CPU and GPU.

With Makie 0.1.5 I get constant high % GPU and low % CPU when idling but ~100% CPU when zooming or panning a plot.

@SimonDanisch 

Manifest with no issues:
```
(tmp) pkg> st
Status `~/tmp/Project.toml`
  [537997a7] AbstractPlotting v0.10.11
  [e9467ef8] GLMakie v0.1.4 ⚲

(tmp) pkg> st -m
Status `~/tmp/Manifest.toml`
  [85c772de] AbstractNumbers v0.2.1
  [537997a7] AbstractPlotting v0.10.11
  [79e6a3ab] Adapt v2.0.2
  [6e34b625] Bzip2_jll v1.0.6+3
  [fa961155] CEnum v0.4.1
  [a2cac450] ColorBrewer v0.4.0
  [35d6a980] ColorSchemes v3.9.0
  [3da002f7] ColorTypes v0.10.8
  [c3611d14] ColorVectorSpace v0.8.5
  [5ae59095] Colors v0.12.3
  [e66e0078] CompilerSupportLibraries_jll v0.3.3+0
[...]

PR #2336, opened 2022-10-14
Render on demand

ffreyer:
# Description

This is an attempt at making a "render on demand" mode for GLMakie. The idea is have all renderobject input observables update a flag `requires_update` which is checked in the renderloop. Rendering only takes place if this flag is true for at least one (visible) renderobject. 

The basic idea already works, but it's currently not catching all updates. ~~For example changes to a mesh aren't tracked.~~ I guess the main task will be to find all of those.

## Type of change

## Checklist


SimonDanisch:
Cool, that might just work without too much effort :) Let me know if you have any questions, I could try to take a look as well...

ffreyer:
Texture, TextureBuffer, GLBuffer and VertexArray now track when they are updated and forward that to RenderObject. That's all the GPUArray subtypes other than GPUVector, which seems unused. I think with that the pr is basically done already.

I also went through every plot primitive to check if each uniform is triggering updates. I skipped the ones which I believe to be constant, irrelevant or inaccessible. 

<details>
  <summary>List of ignored attributes</summary>

## General
- doc_string (irrelevant)
- objectid (constant?)
- intensity (generally works if color is set to a float array first)
- shading (irrelevant, requires shader recompilation)

## scatter

- image (only works if passed as "image")
- gl_primitive

## Lines, Linesegments
- pattern (linestyle, doesn't work but I think that's indepedent of the pr)
- pattern_length
- gl_primitive

## mesh

- interpolate_in_fragment_shader (constant?)

## surface

- scale (constant?)
- color (constant?)
- vertices (constant)
- faces (constant)

## text

- intensity (not accessable?)
- shape (constant)
- scale_primitive (irrelevant?)
- color_norm (irrelevant w/o intesity)
- image (irrelevant)
- billboard 
- color_map (irrelevant w/o intensity)
- gl_primitive (constant)

## heatmap/image

- stroke_color (doesn't work anyway?)

## volume

- absorption (constant?)
- prerender (constant?)
- postrender (constant?)

## meshscatter

- image (broken in general)
- vertex_color (is this usable?)
- interpolate_in_fragment_shader (constant?)
</details>

ffreyer:
Should help with #678 (on any OS), maybe also #267 and #1683. Should also help with idle gpu utilization in #1427, but I don't think it'll help with #1412. 

Whether this fixes any of those issues is kind of subjective I guess. This pr should make raw rendering a little bit slower through the update tracking, but cuts down idle gpu usage to 0. So in an interactive or animated situation things shouldn't be any better, but otherwise they should be a lot better.

For example `arrows(rand(Point3f, 1_000_000), rand(Point3f, 1_000_000))` maxes out my GPU while interacting, but has 0% GPU utilization once I stop.

ffreyer:
I did a bit of benchmarking with
```julia
scene = Scene()
for _ in 1:100
    scatter!(scene, rand(10), marker = Rect)
end
display(scene, render_on_demand = false)

function update!(scene)
    for p in scene.plots
        p.ssao[] = !p.ssao[]
        p.overdraw[] = !p.overdraw[]
        p.color[] = :red
        p.markersize[] = 10f0
        p.marker[] = Rect
    end
end

@benchmark update!($scene)
```
That looks to be 3-8% slower with this pr. (Min , mean, median time going from (1.38, 1.96, 1.94) ms -> (1.43, 2.07, 2.03) ms)

ffreyer:
I added some code to disable update tracking when the other render loops are used. (Those can't be changed dynamically, right?)

I also noticed that I wasn't disabling `render_on_demand` correctly in my benchmark, but it doesn't change much. My timing were a bit worse across the board but the pr was still 5-10% slower. With the latest commit it's now on par with `fps_renderloop`.

GLMakie/src/GLAbstraction/GLTypes.jl
```diff
@@ -345,6 +365,11 @@ function RenderObject(
 
     switch_context!(context)
 
+    # This is a lazy workaround for disabling updates of `requires_update` when
+    # not rendering on demand. A cleaner implementation should probably go 
+    # through @gen_defaults! and adjust constructors instead.
+    track_updates = to_value(pop!(data, :track_updates, true))
```
SimonDanisch, line 371: `@gen_defaults` is pretty much deprecated (without a new implementation yet, but I think it's pretty clear that it's pretty redundant), so I guess that's fair ;)

ffreyer: The way it works atm is that GPUArrays always attach the update trackers when they get created. When the renderobject is created it either removes the tracking (i.e. observer functions) or connects the remaining uniforms based on what would usually be interpreted as a uniform. Seems quite hacky to me :shrug: 

A clean implementation wouldn't connect (and maybe also not define) the tracking observables in GPUArrays. I think with how things are organized atm `@gen_defaults` would need to be adjusted to forward that information. But I don't really want to mess with that macro. I'd probably just break it

SimonDanisch: I meant, `@gen_defaults` needs to be refactored / removed going forward, so whenever we do that, we can clean up the implementation of `track_updates` ;)

SimonDanisch: In other words: please don't waste a second on `@gen_defaults` ;)

ffreyer: Ok I think the pr is ready then

GLMakie/src/GLAbstraction/AbstractGPUArray.jl
```diff
@@ -193,8 +193,10 @@ max_dim(t)       = error("max_dim not implemented for: $(typeof(t)). This happen
 
 function (::Type{T})(x::Observable; kw...) where T <: GPUArray
     gpu_mem = T(x[]; kw...)
-    on(x-> update!(gpu_mem, x), x)
-    gpu_mem
+    # TODO merge these and handle update tracking during contruction
```
SimonDanisch, line 196: Why do we have this todo, and not just:
```julia
  on(data) do new_data
      update!(gpu_mem, new_data)
      gpu_mem.requires_update[] = true
  end
```
(I already have this locally, will push some changes like this later)

GLMakie/src/GLAbstraction/GLTypes.jl
```diff
@@ -289,11 +297,14 @@ mutable struct RenderObject{Pre}
     prerenderfunction::Pre
     postrenderfunction
     id::UInt32
+    requires_update::Bool
```
SimonDanisch, line 300: I think we should consistently use the same type for `requires_update`, and I don't think we need an observable, so maybe use `Base.RefValue` consistently instead?

SimonDanisch: Ah, I  see, some do need to be an observable..So maybe use Observable consitently?

GLMakie/src/GLAbstraction/GLTypes.jl
```diff
@@ -303,12 +314,46 @@ mutable struct RenderObject{Pre}
...
+                    end
+                elseif uniform isa GPUArray
+                    on(uniform.requires_update) do _
+                        robj.requires_update = true
+                    end
+                end
+            end
+            on(_ -> robj.requires_update = true, vertexarray.requires_update)
+        else
+            # remove tracking from GPUArrays
```
SimonDanisch, line 339: not sure why that's needed, could you elaborate the comment?

GLMakie/src/screen.jl
```diff
@@ -663,10 +667,45 @@ function fps_renderloop(screen::Screen)
     end
 end
 
+
+function requires_update(screen::Screen)
+    for (_, _, robj) in screen.renderlist
+        visible = Bool(to_value(get(robj.uniforms, :visible, true)))
```
SimonDanisch, line 673: I think we may want to use this PR to change the visible field into a well typed, mandatory field for RenderObject.

GLMakie/src/GLAbstraction/AbstractGPUArray.jl
ffreyer: I separated them so the update tracking could be removed again when using fps_renderloop

GLMakie/src/GLAbstraction/GLTypes.jl
SimonDanisch: Hm, it feels cleaner to me, to just let the GPUArrays track their updates, which should be very cheap, but then not connect them further if we don't track updates...

GLMakie/src/GLAbstraction/GLTypes.jl
ffreyer: For `GPUArray`we either need an observable to forward to the information to the renderobject or we need to search through a bunch of stuff every potential frame. Or we need to make their updates aware of the renderobject but that might be a lot of work?

Otherwise I went with Bool since the structs are already mutable. I think that's a bit faster than a Refvalue in a mutable struct too...

GLMakie/src/screen.jl
ffreyer: Sounds good.

GLMakie/src/GLAbstraction/GLTypes.jl
ffreyer: I think I actually removed all the tracking in one go so I'm not sure how much the GPUArrays contribute. The test case I used also didn't trigger updates there, but in simple observables since that should have a higher ratio of tracking cost / total update cost and I wanted to know how bad it can get. 

I was planning to test having GPUArray updates wait for frames in the future, which would probably require something like `requires_update` too. So maybe this will end up as always on/tracking later anyway...

ffreyer:
```julia
using BenchmarkTools, GLMakie

scene = Scene()
for _ in 1:100
    heatmap!(scene, rand(16, 16))
end
display(GLMakie.Screen(render_on_demand = false), scene)

function update!(scene)
    for p in scene.plots
        notify(p[3])
    end
end

@benchmark update!($scene)
```

### With cleanup

```
BenchmarkTools.Trial: 9326 samples with 1 evaluation.
 Range (min … max):  381.502 μs …  32.987 ms  ┊ GC (min … max): 0.00% … 98.11%
 Time  (median):     502.491 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   532.654 μs ± 812.194 μs  ┊ GC (mean ± σ):  3.84% ±  2.49%

      ▁           ▃▄▇█▄▃▁▁▁▂▁▁                                   
  ▁▁▅▄█▅▄▄▄▃▄▄▄▄▄▆█████████████▇▇▇▆▇▆▆▅▅▅▅▅▄▄▄▃▃▃▃▃▃▂▂▂▂▂▂▁▂▁▁▁ ▄
  382 μs           Histogram: frequency by time          694 μs <

 Memory estimate: 121.88 KiB, allocs estimate: 900.
```

### Without cleanup (the else branch commented out)

```
BenchmarkTools.Trial: 4238 samples with 1 evaluation.
 Range (min … max):  803.314 μs …  34.882 ms  ┊ GC (min … max): 0.00% … 96.12%
 Time  (median):       1.153 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.174 ms ± 745.947 μs  ┊ GC (mean ± σ):  1.33% ±  2.08%

          ▂▃▅▅▅█▃▃▃▄▃▃▃▃▄▄▄▃▃▄▃▃▇▃▅▇▄▃▄▄▃▃▁▂                     
  ▁▄▄▆▄▅▅▃███████████████████████████████████▆▇▆▅▅▄▄▃▄▄▃▂▂▃▃▂▂▂ ▅
  803 μs           Histogram: frequency by time          1.6 ms <

 Memory estimate: 128.12 KiB, allocs estimate: 1300.
```

### Without cleanup & single observer function

```
BenchmarkTools.Trial: 4747 samples with 1 evaluation.
 Range (min … max):  757.975 μs …  31.849 ms  ┊ GC (min … max): 0.00% … 96.00%
 Time  (median):       1.006 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.048 ms ± 783.416 μs  ┊ GC (mean ± σ):  1.84% ±  2.42%

          ▃▄▇▅▅█▅▅▅▂▃ ▃▁▁▁▁ ▁▁▂▁▂▁▁  ▁                           
  ▅▃▆▅▄▆▅▆████████████████████████████████▆▆▆▄▄▅▆▄▄▄▃▃▃▃▂▂▂▂▁▂▂ ▅
  758 μs           Histogram: frequency by time         1.43 ms <

 Memory estimate: 128.12 KiB, allocs estimate: 1300.
```

That's much more than I thought it would be...

GLMakie/src/screen.jl
ffreyer: Not sure if I did what you had in mind but I moved it to a field.

I also had the logic for re-rendering with visible wrong. I tried to set it up in a way that invisible renderobjects can't trigger updates, but the way I did it caused `p.visible[] = false` to not trigger a re-render. I fixed that by moving the check to the renderobject updater functions.

ffreyer:
I reran the benchmarks I posted before and now things are more equal. Maybe the things I have running in the background affect this, not sure. To summarize:

```
Min times on master:               754, 774, 759µs
Min times on pr w/ fps_renderloop: 743, 740, 745µs
Min times on pr w/ on_demand:      848, 844, 852µs
```

So in a workflow with lots of GPUArray updates, the pr is now does slightly better than master with `fps_renderloop` and ~15% worse with the on-demand renderloop. I restarted Julia once after switching branches just to be save.

SimonDanisch:
Merged in https://github.com/MakieOrg/Makie.jl/pull/2397

MariusDrulea:
> Should help with #678 (on any OS), maybe also #267 and #1683. Should also help with idle gpu utilization in #1427, but I don't think it'll help with #1412.

I did some tests and plotting a 3D volume `plot(rand(1000, 1000, 1000))` works nice now, 0% GPU utilization when the volume is not manipulated. Plotting a 2D matrix `plot(rand(2000, 2000))` however still keeps the GPU busy even there is no interaction on the image, just like mentioned above.

PR #2346, opened 2022-10-20
display png via html preserving size regardless of pixel density

jkrumbiegel:
# Description

Fixes issue https://github.com/MakieOrg/Makie.jl/issues/2264

## Type of change

Adds the ability to show scenes with `MIME"text/html"` and related types, using the png output and setting the correct width/height on the result.

## Checklist

## Example images

This is in VSCode, displaying with different px_per_unit values.
Note that the relative size of the image doesn't change.
You can see that clarity is ideal at 2 px per unit on my retina screen, which makes sense. However, for zooming in, it can be beneficial to go higher.

image
image
image
image

Here's the same image shown with mime type svg, as you can see the apparent size is the same:

image

When we display with mime type png like before, we can see that the image blows up in size, filling the available plot pane space. This is not desirable, as the pixel density becomes lower again (blurriness) and the text becomes too large:

image

### DIFF
--- a/CairoMakie/src/display.jl
+++ b/CairoMakie/src/display.jl
@@ -99,10 +99,21 @@ function Makie.backend_show(screen::Screen{IMAGE}, io::IO, ::MIME"image/png", sc
     return screen
 end
 
+function Makie.backend_show(screen::Screen{IMAGE}, io::IO, ::Union{MIME"text/html",MIME"application/vnd.webio.application+html",MIME"application/prs.juno.plotpane+html",MIME"juliavscode/html"}, scene::Scene)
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Here's another example from a Pluto notebook. As you can see, the image can be made crisper without increasing its size in the new html png mode:

grafik

jkrumbiegel:
I think this is good to go in terms of functionality. I tested in VSCode and Pluto and it seemed to work well.

I do still want to write a couple paragraphs of docs about this but a review can already happen now.

jkrumbiegel:
merged into https://github.com/MakieOrg/Makie.jl/pull/2831

Issue #2289, opened 2022-09-20
`density` with gradient color cannot be added to `Legend`


fatteneder:
### MWE

```julia
using GLMakie
GLMakie.activate!()

f = Figure()
ax = Axis(f[1,1])

density!(ax, randn(200) .- 2sin(2/3*pi),
         color = :x, colormap = :thermal, colorrange = (-5, 5),
         strokewidth = 1, strokecolor = :black, label="density")

Legend(f[1,2], ax)

f
```
throws
```julia
julia> include("new_mwe.jl")
ERROR: LoadError: Unknown color: x
```

The above works if any other `<:Colorant` is given and it then plots a little square with that color as the marker in the `Legend`.

Should we plot a squeezed colorbar as the legend marker in this case?

Btw: I think the same question applies to a `heatmap`. However, there isn't an overload for `legendelements` yet.

fatteneder:
A squeezed colorbar is perhaps not ideal because the markers in the legend are usually relatively small compared to a real colorbar.
Also: If you have two `density` plots and you put each one into a `Legend`, then having the same squeezed colorbar as a marker would not help you in telling them apart in the plot.

A simpler solution, which is also employed by `gnuplot`, is to put only the label and no marker in the legend.

jkrumbiegel:
That's why I never implemented automatic legend items for things with continuous color maps. It's just not obvious to me what the behavior should be like.

fatteneder:
Imitating `gnuplot`s behavior is not an option?
The only reason that would be useful is to have a target you can click for the hide and show functionality of #2276, assuming it gets merged.

jkrumbiegel:
Do you have some visual examples? Never used gnuplot

fatteneder:
Here is a `gnuplot` demo which you can reproduce with
```sh
$ gnuplot
gnuplot> splot cos(x) * cos(y) w pm3d, sin(x) * sin(y) w p
```

![demo_gnuplot](https://user-images.githubusercontent.com/26469701/191803123-167b5fa3-2448-4e97-a023-debbdb4cf264.gif)

Explanation:
Plotting with palette-mapped 3d (that's what `w pm3d` says) adds a colorbar as well as a legend entry with `cos(x) * cos(y)`, but with no marker (top right).
Plotting with points (that's what `w p` says) does not add a colorbar, but adds a legend entry with a little purple cross marker.

jkrumbiegel:
[...]

Issue #1156, opened 2021-07-20
Can't construct legend from stacked barplot axis


grahamas:
Since a stacked barplot can't be constructed as a sequence of labeled plots, I'm not sure how to construct a legend other than from the axis. However, `Legend(fig, stack_ax)` cannot find plots with labels, so fails.

```julia
using CairoMakie
tbl = (x = [1, 1, 1, 2, 2, 2, 3, 3, 3],
       height = 0.1:0.1:0.9,
       grp = [1, 2, 3, 1, 2, 3, 1, 2, 3],
       grp1 = [1, 2, 2, 1, 1, 2, 1, 1, 2],
       grp2 = [1, 1, 2, 1, 2, 1, 1, 2, 1]
)

fig = Figure()
fig[1,1] = ax = Axis(fig)
barplot!(ax, tbl.x, tbl.height,
        stack = tbl.grp,
        color = tbl.grp,
        axis = (xticks = (1:3, ["left", "middle", "right"]),
                title = "Stacked bars"),
        )
fig[1,2] = Legend(fig, ax)
```
gives
```
ERROR: There are no plots with labels in the given axis that can be put in the legend. Supply labels to plotting functions like `plot(args...; label = "My label")`
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] layoutable(::Type{Legend}, fig_or_scene::Figure, axis::Axis, title::String; merge::Bool, unique::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Makie.MakieLayout ~/.julia/packages/Makie/xbI6d/src/makielayout/layoutables/legend.jl:539
 [3] layoutable(::Type{Legend}, fig_or_scene::Figure, axis::Axis, title::String)
   @ Makie.MakieLayout ~/.julia/packages/Makie/xbI6d/src/makielayout/layoutables/legend.jl:538
 [4] _layoutable(::Type{Legend}, ::Figure, ::Axis, ::Vararg{Any, N} where N; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Makie.MakieLayout ~/.julia/packages/Makie/xbI6d/src/makielayout/layoutables.jl:69
 [5] _layoutable(::Type{Legend}, ::Figure, ::Axis, ::Vararg{Any, N} where N)
[...]

PR #2348, opened 2022-10-20
Add fallback for `legendelements`

fatteneder:
# Description

`legendelements` is not implemented for all `<:AbstractPlot`s, because it is not clear (or decided on) what kind of marker to put for every type next to a label.
This PR adds a fallback where the marker is skipped and only the label is drawn. This is inspired by `gnuplot`s behavior which does not plot markers for gradient plots (cf. https://github.com/MakieOrg/Makie.jl/issues/2289#issuecomment-1255284570).

Here is a MWE:
```julia
using GLMakie
f = Figure()
ax = Axis3(f[1,1])
meshscatter!(ax, rand(10),rand(10),rand(10), label="test")
Legend(f[1,2],ax)
display(f)
```
image

I have to admit, it looks irritating if there is only one element with extra space to the left of the label. Maybe we want to shrink the box if there are only elements with no marker.

I am happy for any input on this. I am mainly proposing this, because it would be helpful to get #2276 working for all `<:AbstractPlot` types by default.

Related: #2289, #1703, #1156

## Type of change

## Checklist


SimonDanisch:
Should we at least use the color of the object?

fatteneder:
The open question is: What color do you put for something like a `heatmap` with a gradient?

asinghvi17:
> The open question is: What color do you put for something like a heatmap with a gradient?

Another heatmap, of course!

PGFPlots does this for surfaces; no reason we can't do the same.

Screen Shot 2023-01-15 at 9 23 18 PM

fatteneder:
> Another heatmap, of course!

Yeah, we probably should.

I am quite busy at the moment, but hopefully I can tackle this next weekend or so.

ffreyer:
I don't know when or where, but this has been implemented

Issue #2331, opened 2022-10-13
VSCode Plot Pane not working with CairoMakie v0.9.0


kfrb:
After updating to CairoMakie v0.9.0, plots in VSCode no longer open in the Plot Pane, but as a separate png file.

I tried this example
```julia
using CairoMakie
lines(1:10, rand(10))
```
in a project with only CairoMakie installed
```julia-repl
pkg> status
  [13f3f980] CairoMakie v0.9.0
```
and with the following versions:
```julia-repl
julia> versioninfo()
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: 16 × Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake)
  Threads: 1 on 16 virtual cores
```
VSCode:
```
Version: 1.72.1 (Universal)
Commit: 129500ee4c8ab7263461ffe327268ba56b9f210d
Date: 2022-10-10T17:25:13.482Z (2 days ago)
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Darwin x64 21.6.0
Sandboxed: No
```
with VSCode Julia Extension `v1.37.2`

SimonDanisch:
Are you sure you have the Plotpane enabled?

SimonDanisch:
Or is this in the repl by any chance? I think this depends on your vscode repl settings...

kfrb:
Hello Simon, thanks for your fast reply!
I checked the VSCode settings and everything should be fine.
The plots show in the Plot Pane again by downgrading with
```julia-repl
pkg> add CairoMakie@0.8.13
```
in the example project.
After that, updating with
```julia-repl
pkg> update
```
results in the plots opening as png files.

SimonDanisch:
Really? That'd surprise me quite a bit...How did you eval the file? Have you set file eval to eval in REPL? VScode has quite a few options like that..

kfrb:
I have tried all `Julia: Execute [...]` commands of the command palette. What else can I try? In what other way can code be evaluated? I'm not sure which REPL settings play a role there as well, but I use almost all default settings otherwise.

[...]

PR #2353, opened 2022-10-21
bring back inline! :(

SimonDanisch:
Guess we couldn't remove it yet, because I forgot the edge case of VSCode repl exposing a plotpane, but directly calling `display(eval_result)`, which made `Makie@0.18` always open a window, without a way to switch that behaviour.

We really want a better solution, especially since `inline!(false)` doesn't currently always open a window, since we can't figure out how to always open a window, when VSCode thinks everything should go into the plotpane ;) 
Fixes https://github.com/MakieOrg/Makie.jl/issues/2331

### DIFF
--- a/CairoMakie/src/screen.jl
+++ b/CairoMakie/src/screen.jl
@@ -98,6 +98,8 @@ function device_scaling_factor(surface::Cairo.CairoSurface, sc::ScreenConfig)
     return is_vector_backend(surface) ? sc.pt_per_unit : sc.px_per_unit
 end
 
+const LAST_INLINE = Ref(true)
+
 """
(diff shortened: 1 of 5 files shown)


jkrumbiegel:
I guess we could also call it `force_inline!` so that the negation doesn't promise a window. On the other hand, why introduce new syntax just to remove it again soon-ish.

SimonDanisch:
I don't want to rename it, since I really don't want to keep this solution in the future. Renaming the stopgap solution would be too much code churn for my taste ;)

jkrumbiegel:
Yeah I figured, I agree

PR #2356, opened 2022-10-22
Brackets and bezier lines

jkrumbiegel:
- allow cairomakie to draw bezierpath lines, add convert to vector of points for bp
- add a bracket recipe

image

# Description

Fixes # (issue)

## Type of change

## Checklist


asinghvi17:
Looks pretty good to me - can we merge this?

jkrumbiegel:
I think so, the only remaining question was what to do with the conversion method for BezierPath. I'm converting to a line essentially, so that glmakie can fall back to that. But that means you can't use `poly` with a BezierPath. Maybe we should just tackle that later though and just keep the lines for now.

asinghvi17:
We could override the `convert_arguments` for Poly specifically, if the only adjustments needing to be made are that the path should be closed.  IMO, that could be tackled later fairly easily.

PS: this will help out in a lot of other places, since you could do e.g. bars, arrows, etc with different styles.

src/basic_recipes/bracket.jl
```diff
@@ -0,0 +1,146 @@
+"""
+    bracket(x1, y1, x2, y2; kwargs...)
+    bracket(x1s, y1s, x2s, y2s; kwargs...)
+    bracket(point1, point2; kwargs...)
+    bracket(vec_of_point_tuples; kwargs...)
+
+Draws a bracket between each pair of points (x1, y1) and (x2, y2) with a text label at the midpoint.
+
+By default each label is rotated parallel to the line between the bracket points.
+
```
asinghvi17, line 10:
```suggestion

In order to implement a new style for brackets, overload `Makie.bracket_bezierpath(::Val{:mybracketstyle}, p1, p2, d, width)`, 
which should return a [`BezierPath`](@ref).
```

jkrumbiegel: I don't want to make that API, yet, in case I find there's more to it than I currently have.

PR #2372, opened 2022-10-26
A little experiment towards better clean up + more typed observables

SimonDanisch:
I rediscovered while looking at the GLMakie backend code, that we actually are already pretty strict about atomic plot attributes, in the sense that the output type of `convert_attributes` is not allowed to change. So, e.g. if we set color to a single color, we may never change it to an array of colors, alowing to restrict color to `Observable{RGBAf}`.

So, at least for atomic plots, we should be able to always create strictly typed observables of the type `typeof(conver_arguments(...))`.
In this PR, I'm rewriting `Attributes` to lazily convert via `convert_attributes` to always return a typed observable...
Still needs to be optional though since `Attributes` are used for so many things, but with a bit larger refactor, we may be able to flip the switch on that.
Still an experiment, but may greatly help with not doing conversions multiple times, have better typed Observables and have better ways to completely clean up the observable connections of a plot.
If this works out well, we may want to make our own Attribute type for atomic plots (which then can also finally lock in the set of allowed attributes), and use a simpler type for the rest of the `Attribute` uses.

### DIFF
--- a/GLMakie/src/drawing_primitives.jl
+++ b/GLMakie/src/drawing_primitives.jl
@@ -77,24 +77,26 @@ function connect_camera!(gl_attributes, cam, space = gl_attributes[:space])
     return nothing
 end
 
-function cached_robj!(robj_func, screen, scene, x::AbstractPlot)
+function cached_robj!(robj_func, screen, scene, @nospecialize(x::AbstractPlot))
     # poll inside functions to make wait on compile less prominent
     pollevents(screen)
+    ignore = Set([
+        :transformation, :tickranges, :ticklabels, :raw, :SSAO,
+        :lightposition, :material,
+        :inspector_label, :inspector_hover, :inspector_clear
+    ])
     robj = get!(screen.cache, objectid(x)) do
-        filtered = filter(x.attributes) do (k, v)
-            !in(k, (
-                :transformation, :tickranges, :ticklabels, :raw, :SSAO, 
-                :lightposition, :material, 
-                :inspector_label, :inspector_hover, :inspector_clear
-            ))
-        end
+        gl_attributes = Dict{Symbol, Any}()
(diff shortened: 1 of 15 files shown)


jkrumbiegel:
But is it good that you can't change `color = rand(10)` to `color = :red`? Or am I misunderstanding what `So, e.g. if we set color to a single color, we may never change it to an array of colors, alowing to restrict color to Observable{RGBAf}` means

SimonDanisch:
>But is it good that you can't change color = rand(10) to color = :red interactively? 

You can't do that right now 🤷

ffreyer:
To give some context:

https://github.com/MakieOrg/Makie.jl/blob/e0cb40096896c3f43a390cd0523138ad644e41fc/GLMakie/assets/shader/sprites.vert#L26-L28

lines like these get replaced before shader compilation with something like `type_of_attribute name_of_attribute;`. So if the renderobject gets generated with an RGBAf color, you'd get something like `vec4 color;` there. After shader compilation you can change the value but not the type.
Making this dynamic would either require converting to arrays beforehand or getting shaders to recompile on type changes.

SimonDanisch:
Yeah, and WGLMakie & GLMakie at least both have a version of:
https://github.com/MakieOrg/Makie.jl/blob/master/GLMakie/src/drawing_primitives.jl#L163
```julia
function lift_convert_inner(value, key, plot_key, plot)
    return lift(value) do value
        return convert_attribute(value, key, plot_key)
    end
end
```
which is pretty much identical to:
```julia
obs = Observable(convert_attribute(value, key, plot_key)) # creates strongly typed obs
map!(x-> convert_attribute(x, ...), obs)
```

SimonDanisch:
Closing this experiment for now

PR #2376, opened 2022-10-26
Fix `TypeError: attribute.onUploadCallback is not a function`

SimonDanisch:
So, me and others have reported spurious errors a la:
```julia
Stack trace:
    TypeError: attribute.onUploadCallback is not a function
        at createBuffer (https://unpkg.com/three@0.136.0/build/three.js:9462:14)
        at Object.update (https://unpkg.com/three@0.136.0/build/three.js:9556:28)
        at Object.setup (https://unpkg.com/three@0.136.0/build/three.js:10732:16)
        at WebGLRenderer.renderBufferDirect (https://unpkg.com/three@0.136.0/build/three.js:19427:18)
        at renderObject (https://unpkg.com/three@0.136.0/build/three.js:19784:11)
        at renderObjects (https://unpkg.com/three@0.136.0/build/three.js:19769:6)
```

The weird thing is, that it sometimes happens but not all times.
I would suspect different ThreeJS versions, but as far as I can tell, we pin the ThreeJS version...
I debugged this issue on [sd/new-jsserve](https://github.com/MakieOrg/Makie.jl/compare/sd/new-jsserve), where the issue was happening more consistently and there it turned out, that `buffer_geometry.setIndex` requires a `BufferAttribute`, which is actually documented like that and looking at the threejs code, it really looks like a requirement.
Really weird, that it works without it too...sometimes...
But this is JS after all 🤷 
@SebastianM-C and especially @mashu, could you try this out if it fixes the issue?
`]add MakieCore#sd/fix-onUploadCallback Makie#sd/fix-onUploadCallback WGLMakie#sd/fix-onUploadCallback`


mashu:
So I made small MWE

First the project file is
```
name = "MWE"
uuid = "4651299b-9767-4d32-a76b-6bae303ba761"
authors = ["Mateusz Kaduk <mateusz.kaduk@gmail.com>"]
version = "0.1.0"

[deps]
DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
GraphMakie = "1ecd5474-83a3-4783-bb4f-06765db800d2"
Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
MakieCore = "20f20a25-4f0e-4fdf-b5d1-57303727442b"
NetworkLayout = "46757867-2c16-5918-afeb-47bfcb05e46a"
Pluto = "c3e4b0f8-55cb-11ea-2926-15256bba5781"
Probably = "2172800d-0309-5a57-a84f-d50c94757422"
WGLMakie = "276b4fcb-3e11-5398-bf8b-a0c2d153d008"
```
Then code in src/MWE.jl is
```
module MWE
using GraphMakie
using WGLMakie
using NetworkLayout
using Graphs
using Probably
using DataStructures

function mwe()
    g = complete_graph(5)
    graphplot(g)
end

end # module MWE
```

You **need to use sysimage to reproduce this issue**, otherwise everything works as expected.
When compiling WGLMakie as sysimage in VSCode Ctrl+Shift+B and enter to build sysimage with current project.
This after completion will restart REPL and you can Ctrl+Enter selected block to execute module code than call `MWE.mwe()` from REPL you get JS error.

Updating with above fix and than building sysimage, results in empty white blank figure and no error on plotting.
There is error when zooming on empty area, but that's probably nothing is there and consequence of above that plot hasn't been populated at all.

You can't test whether fix works without building sysimage, as WGLMakie always works without sysimage.
Hope that helps.

SimonDanisch:
Hm, when I think about it, it could also be the HTTP version, considering that @SebastianM-C had these issues without a system image (I believe?).

SebastianM-C:
Oh, yeah. I think that it was HTTP@v1

SimonDanisch:
Did you force it somehow ? I think the currently released JSServe shouldn't be compatible with `HTTP@v1`

SimonDanisch:
Integrated into https://github.com/MakieOrg/Makie.jl/pull/2428

PR #2384, opened 2022-10-29
move Arrows and Wireframe to MakieCore

jw3126:
This is kind of a follow-up to:

https://github.com/MakieOrg/Makie.jl/pull/2334

### DIFF
--- a/src/basic_recipes/arrows.jl
+++ b/src/basic_recipes/arrows.jl
@@ -21,31 +21,12 @@ grid.
 ## Attributes
 $(ATTRIBUTES)
 """
-@recipe(Arrows, points, directions) do scene
-    attr = merge!(
-        default_theme(scene),
-        Attributes(
-            arrowhead = automatic,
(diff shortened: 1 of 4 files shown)


src/basic_recipes/arrows.jl
```diff
@@ -21,31 +21,12 @@ grid.
 ## Attributes
 $(ATTRIBUTES)
 """
```
jw3126, line 23: In this PR I kept the docs in Makie.jl. 
The advantage is that the docstring extensions machinery is avaible in Makie but not in MakieCore.

The only observable difference is that before `arrows` and `arrows!` had exactly the same doc string. After this PR the `arrows` docstring stays the same, but `arrows!` docstring just says "see arrows".

ffreyer:
Should we just move all of them?

jw3126:
> Should we just move all of them?

My opinion is that every recipe that does not add dependencies to MakieCore should be declared there.

PR #2387, opened 2022-10-31
Replace `textsize` with `fontsize` everywhere

jkrumbiegel:
# Description

We had both `textsize` and `fontsize` in use for quite a while, which was always confusing.
With this PR, `textsize` is removed everywhere.

I've added error messages in two central places, plot recipes and Block creation functions, which should catch all uses of the keyword and prompt users to search and replace. With that, migration should not take more than a few seconds so I believe this breaking change is worth it in terms of API clarity.

We can discuss if the error policy is too drastic because it will also apply to user-defined recipes and Blocks. However, I'd argue that it's an improvement for the ecosystem if all uses of `textsize` are purged, unless we want it to stick around as a ghost for years.

The PR also makes `text` inherit `fontsize` from the main theme setting by default, which it arguably should have done for a while.

## Type of change

## Checklist


Moelf:
missing breaking tag

himcraft:
Will this be merged into 0.18.4 with the Rich Text PR?

jkrumbiegel:
This should go into a new breaking version with rich text, so probably 0.19

Issue #2396, opened 2022-11-03
CairoMakie saving raster to .svg depending on content


jaakkor2:
In this example, file `debug2.svg` is 75 kB and contains the png raster of the output.  `debug1.svg` is 20 kB and contains the vector graphics as intended.
```
using CairoMakie
CairoMakie.activate!(type="svg")
using Makie.GeometryBasics

item1 = Point2.([[0,0],[1,0],[0,1],[0,0]])
item2 = Point2.([[0,0],[1,0],[0,-1],[0,0]])
fig = Figure()
ax = Makie.Axis(fig[1,1])
Makie.poly!(ax, MultiPolygon([Polygon(item1)]), color = :red)
save("debug1.svg", fig) # vector (ok)
Makie.poly!(ax, Polygon(item2), color = :blue)
save("debug2.svg", fig) # png raster inside .svg (why?)
```

Project has only CairoMakie, manifest lists CairoMakie 0.9.1, Cairo 1.0.5, Cairo_jll 1.16.1+1, Makie 0.18.1, MakieCore 0.5.1, GeometryBasics 0.4.2, GeoInterface 1.0.1, Julia 1.8.2.

jkrumbiegel:
`poly` has special override functions in CairoMakie so that it doesn't need to fall back to `mesh` which it is based on in general. If `mesh` is hit, then the whole cairo context gets rasterized as mesh drawing is not supported in svg. So one of your cases must hit mesh even though it's not supposed to.

PR #2398, opened 2022-11-03
Don't mesh a polygon in CairoMakie

jaakkor2:
# CairoMakie plot of a polygon is a mesh resulting in a raster .svg file

Fixes https://github.com/MakieOrg/Makie.jl/issues/2396

Treat a polygon same way as an array of polygons

## Type of change

## Checklist


jkrumbiegel:
Would be great if you could add a plot with a couple different `poly` versions (including your new one) to this file to ensure they don't get rasterized again in the future.

https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/test/rasterization_tests.jl

jaakkor2:
Thank you @jkrumbiegel for pointing out the location of the rasterization tests.  I now simply added a simplest polygon to the existing testset.

fatteneder:
This PR broke the 'polygon with hole' example #2456

jkrumbiegel:
Ah damn it seemed not to have a visual regression test. It shouldn't be too hard to fix, sometimes holes don't work right if the interiors are not the opposite winding order, which they should have but don't always have. Not sure what the problem was exactly in the docs case.

PR #2399, opened 2022-11-04
clean up interactions + remove close(obs)

SimonDanisch:
While cleaning up our Observable handling, I noticed that close isn't used anymore.
When investigating the usage of `close(obs)`, I also tripped over these abstractly typed interactions, which probably isn't super bad, but can also not be good since they're somewhat performance sensitive.


ffreyer:
Why not just make the types mutable?

SimonDanisch:
Because the refs are passed around ;)

PR #2400, opened 2022-11-04
Add Fontconfig font search to the rich-text scheme

asinghvi17:
# Description

This PR builds off jk/rich-text.  It doesn't modify that code at all, but introduces a method to use Fontconfig to find and create font sets.

Basically, what this does is introduce 3 main methods.

1. The ability to obtain all styles in a font family (`populate_font_family`) - returns a NamedTuple of Symbol styles => FTFont (Makie.NativeFont).
2. The ability to search for a particular style within a font family, along with some basic fallbacks (regular can be roman or book, for example, if there is no style called `regular`) - `search_font`
3. The ability to cache font families and regularize some names, so that `(:regular, :bold, :italic, :bold_italic)` are always available, even if the font has obliques instead of italics.  Individual styles can still be referred to by their original names.  This is the `FontFamily` interface, which is, additionally, editable.  It also automatically searches for a style if it can't exactly match, making it a bit more dynamic.

This results in the following:

```julia
FontFamily("TeX Gyre Heros Makie")
FontFamily("TeX Gyre Heros Makie", Attributes with 5 entries:
  bold => FTFont (family = TeX Gyre Heros Makie, style = Bold)
  bold italic => FTFont (family = TeX Gyre Heros Makie, style = Bold Italic)
  bold_italic => FTFont (family = TeX Gyre Heros Makie, style = Bold Italic)
  italic => FTFont (family = TeX Gyre Heros Makie, style = Italic)
  regular => FTFont (family = TeX Gyre Heros Makie, style = Regular))
```
and `fonts = "TeX Gyre Heros Makie"` is equivalent to `fonts = FontFamily("TeX Gyre Heros Makie")`, which is what has been provided in the default theme.  This way, people can access the exact font style (`Extracondensed Semibold` or what have you) as well as some basic shortcuts.

## Type of change

## Checklist


SimonDanisch:
Cool! Can we move most of this to FontConfig.jl?

SimonDanisch:
It would be important to make Fontconfig well tested across platforms in the Fontconfig.jl CI, to actually take it on as a dependency ;)

asinghvi17:
Definitely!  The only code which would have to live in Makie would be the extension of `to_font`, really.

I seem to recall you mentioning that you had some issues with Fontconfig.jl on Ubuntu @SimonDanisch - have those been resolved since?

Looking at CI: I think the tests as they stand now are pretty alright?  Of course, some tests for the new code would be added.  

It looks like Travis CI tests Linux, Windows and OSX for Fontconfig...which should be sufficient?  If necessary I could add Github Actions as well, but it doesn't seem needed yet.  Will definitely have it test on Julia 1.3, release and nightly though.

SimonDanisch:
> have those been resolved since?

I have no idea^^ That's why it'd be nice to have some relevant tests run all platforms :)

>I think the tests as they stand now are pretty alright 

That'd be great :) Haven't looked at them yet ;) 

>If necessary I could add Github Actions

I think we should get rid of Travis and test ubuntu, linux and osx with github actions!
And I'd just test on 1.6 and higher ;)

SimonDanisch:
>I think the tests as they stand now are pretty alright

Hm...they only seem to test that nothing is crashing:
https://github.com/JuliaGraphics/Fontconfig.jl/blob/master/test/runtests.jl
Doesn't even make sure, that fonts can be found etc, so I think they're not really great :D 
From what I remember, the main cross platform issue was, that it wasn't finding any fonts on some platforms.

I think we should at least test something like this:
https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/blob/master/test/runtests.jl#L240-L271

Issue #2406, opened 2022-11-07
CairoMakie svg id assimilation and salting is broken


jkrumbiegel:
It's enough to do:

```julia
using CairoMakie
f, ax, sc = scatter(1:10)
save("test1.svg", f)
save("test2.svg", f)
```

The two files will not be the same because the corrections in the show function relating to ids are not applied anymore.

PR #2407, opened 2022-11-07
fix salting

SimonDanisch:
Fix https://github.com/MakieOrg/Makie.jl/issues/2406


SimonDanisch:
The GLMakie error will be fixed in #2397

PR #2408, opened 2022-11-08
add stephist

Moelf:
# Description

helps #368

Looks stupid right now because I just copied the `Hist`, probably can be simplified, welcome comments

## Type of change

## Checklist


Moelf:
```julia
julia> f = stephist(rand(10000))
FigureAxisPlot()
```

image

src/stats/hist.jl
```diff
@@ -8,6 +8,60 @@ function convert_arguments(P::Type{<:AbstractPlot}, h::StatsBase.Histogram{<:Any
...
+        offset = 0.0,
+        scale_to = nothing,
+
+        flip_labels_at = Inf,
+        label_color = theme(scene, :textcolor),
+        over_background_color = automatic,
+        over_bar_color = automatic,
+        label_offset = 5,
+        label_font = theme(scene, :font),
+        label_size = 20,
```
lazarusA, line 27: maybe consider adding `linestyle`.

Moelf:
if people are feeling okay with the implementation I can start adding News, Docs and test samples

Moelf:
What's the core maintainers' opinion on picking up a histogram package as a dependency? I've grown tired of re-writing crippled binning logic in multiple places :(

Moelf:
this is ready I think. Looks like we don't do reference image for this kind of non basic plots? I checked waterfall and tricontourf PRs.

Let me know if we want anything else

jkrumbiegel:
> Looks like we don't do reference image for this kind of non basic plots? 

Would still be great if you added a reference image, could be one image with a couple of subplots that expose different settings (ideally thick lines so that the visual differ would pick up on it if anything broke).

Also, the docs build is currently broken because of an `automatic` value being passed too far into the GLMakie backend it seems?

Moelf:
Do you have an example of PR adding references imagine?

jkrumbiegel:
The PRs don't add the images themselves, just the code. You can check them in the ReferenceTests folder, all the @reference macros

Moelf:
Am I doing it correctly with reference?

Moelf:
can't seem to find which automatic is too far

lazarusA:
@Moelf this looks like almost there. Maybe a final sprint to make it possible.

Moelf:
yeah I just don't understand what `Automatic()` is getting propagated, since I pretty much took a subset from regular histogram

Moelf:
https://github.com/MakieOrg/Makie.jl/actions/runs/4018541242/jobs/6904286091#step:7:613

I think we have made progress, now only the Reference saving is not working, due to what seems to be because it's trying to save Observable?

SimonDanisch:
I cleaned up the attributes! I think you copied a few attributes from hist, without using them?
If I removed anything that wasn't supposed to be removed I can revert the commit ;) 
But the error was due to passing `label_font` to `stairs`, which doesn't support that attribute

asinghvi17:
@cossio that'll be done in #2636

PR #2436, opened 2022-11-22
Only apply axis/scene transformations to :data space plots

### LINKED ISSUE #2065: Transform is applied even if `space != :data`
@ffreyer these two should be the same I think?

```julia
scatter(Point(100, 100), axis = (xscale = log10, yscale = log10), space = :pixel)
```

grafik

```julia
scatter(Point(100, 100), space = :pixel)
```

grafik

### LINKED ISSUE #2424: text bounding box in an axis
There appear to be some issues with text bounding boxes when used within an axis. Especially when the axis is scaled by a transform. 

Expected behavior
```
using CairoMakie, StableRNGs
x = cumsum(randn(StableRNG(1), 100).^2)
f = lines(x,  axis=(yscale=log10,))
lbl = text!(f.axis, text="Direct Labels are Great!", 50, 50, align=(:right,:bottom))
bb = boundingbox(lbl)
wireframe!(f.axis, bb, color=:red, space=:data)
f
```
shows a bounding box around the text string.

Instead, this shows: 
download

Changing `wireframe!(f.axis, bb, color=:red, space=:data)` to `wireframe!(f.axis.scene, bb, color=:red, space=:pixel)` shows no bounding box. Adding `campixel(f.axis.scene)` does not work either. Plotting `text!` into `f.axis.scene` also made no difference. 

Where we do see somewhat expected behavior 
```
using CairoMakie, StableRNGs                                                                                                                                                                                                 
x = cumsum(randn(StableRNG(1), 100).^2)
f = lines(x) # CHANGE, no log10 scale...
scene = campixel(f.axis.scene) # CHANGE
lbl = text!(scene, 200, 200, text="Direct Labels are Great!", align=(:right,:bottom),space=:pixel)                 
bb = boundingbox(lbl)                                                                               
wireframe!(scene, bb, color=:red, space=:pixel)         
f
```
download

This does work, but requires _pixel coordinates_ for the text rather than data coordinates. 

This also fails with specifying a logscale y axis...

```
using CairoMakie, StableRNGs                                                                                                                                                                                                 
x = cumsum(randn(StableRNG(1), 100).^2)
f = lines(x, axis=(yscale=log10,)) # CHANGE, log10 scale...
scene = campixel(f.axis.scene) # CHANGE
lbl = text!(scene, 200, 200, text="Direct Labels are Great!", align=(:right,:bottom),space=:pixel)                 
bb = boundingbox(lbl)                                                                               
wireframe!(scene, bb, color=:red, space=:pixel)         
f
```

download

This was verified on the most recent versions I could update to of CairoMakie (CairoMakie v0.9.2) and Makie (Makie v0.18.2)

See origin of this discussion here:
https://discourse.julialang.org/t/automatic-direct-labels-for-makie-with-voronoi-tessellation/90247/4

dgleich:
# Description

Fixes #2424 Fixes #2065 

This PR adjusts the behavior of the axis/scene transformations to only apply to plots in the :data space. 

## Notes

Examined and editing almost all places where "apply_transform" was used to switch to the new three parameter version suggested in #2065. 

There were some places where the existing behavior was left. This is due to a quick survey of the surrounding logic suggesting that the existing code assumes things are transformed. Places where this happens:

NO CHANGE src/basic_recipes/hvspan.jl:_apply_x_transform(t::Tuple, v) = apply_transform(t[1], v)
NO CHANGE src/basic_recipes/hvspan.jl:_apply_y_transform(t::Tuple, v) = apply_transform(t[2], v)
NO CHANGE src/layouting/data_limits.jl:        point_t = apply_transform(trans_func, point)
NO CHANGE src/layouting/data_limits.jl:    (to_ndim(Point3f, project(model, apply_transform(trans_func, point)), 0f0) for point in points)
NO CHANGE src/makielayout/blocks/axis.jl:    tlims = Makie.apply_transform(t, lims)

## Type of change

## Checklist

### DIFF
--- a/src/layouting/transformation.jl
+++ b/src/layouting/transformation.jl
@@ -196,6 +196,16 @@ transform_func(x) = transform_func_obs(x)[]
 transform_func_obs(x) = transformation(x).transform_func
 
 """
+    apply_transform(f, data, space)
+Apply the data transform func to the data if the space matches one
+of the the transformation spaces (currently only :data is transformed)
+"""
+apply_transform(f, data, space) = space == :data ? apply_transform(f, data) : data  
+function apply_transform(f::Observable, data::Observable, space::Observable) 
+    return lift((f, d, s)-> apply_transform(f, d, s), f, data, space)
+end 
+
+"""
     apply_transform(f, data)
 Apply the data transform func to the data
 """
@@ -373,3 +383,5 @@ end
 # by heatmaps or images
 zvalue2d(x)::Float32 = Makie.translation(x)[][3] + zvalue2d(x.parent)
 zvalue2d(::Nothing)::Float32 = 0f0
+
+
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -437,7 +437,7 @@ function draw_glyph_collection(scene, ctx, position, glyph_collection, rotation,
 
     glyph_pos = let
         transform_func = scene.transformation.transform_func[]
-        p = Makie.apply_transform(transform_func, position)
+        p = Makie.apply_transform(transform_func, position, space)
 
         Makie.clip_to_space(scene.camera, markerspace) *
         Makie.space_to_clip(scene.camera, space) *
@@ -797,7 +797,7 @@ function draw_mesh3D(
     # and have `func` be fully typed inside closure
     vs = broadcast(meshpoints, (func,)) do v, f
         # Should v get a nan2zero?
-        v = Makie.apply_transform(f, v)
+        v = Makie.apply_transform(f, v, space)
         p4d = to_ndim(Vec4f, scale .* to_ndim(Vec3f, v, 0f0), 1f0)
         view * (model * p4d .+ to_ndim(Vec4f, pos, 0f0))
     end
--- a/CairoMakie/src/utils.jl
+++ b/CairoMakie/src/utils.jl
@@ -4,7 +4,7 @@
 
 function project_position(scene, transform_func::T, space, point, model, yflip::Bool = true) where T
     # use transform func
-    point = Makie.apply_transform(transform_func, point)
+    point = Makie.apply_transform(transform_func, point, space)
     _project_position(scene, space, point, model, yflip)
 end
 
--- a/GLMakie/src/drawing_primitives.jl
+++ b/GLMakie/src/drawing_primitives.jl
@@ -200,7 +200,7 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::Union{Scatte
         gl_attributes[:shading] = to_value(get(gl_attributes, :shading, true))
         marker = lift_convert(:marker, pop!(gl_attributes, :marker), x)
 
-        space = get(gl_attributes, :space, :data)
+        space = get(gl_attributes, :space, :data) # needs to happen before connect_camera! call
         positions = handle_view(x[1], gl_attributes)
         positions = apply_transform(transform_func_obs(x), positions, space)
 
@@ -260,7 +260,7 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::Lines))
             linewidth = gl_attributes[:thickness]
             data[:pattern] = ls .* (to_value(linewidth) * 0.25)
         end
-        space = get(gl_attributes, :space, :data)
+        space = get(gl_attributes, :space, :data) # needs to happen before connect_camera! call
         positions = handle_view(x[1], data)
         positions = apply_transform(transform_func_obs(x), positions, space)
         handle_intensities!(data)
@@ -280,7 +280,7 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::LineSegments
             linewidth = gl_attributes[:thickness]
             data[:pattern] = ls .* (to_value(linewidth) * 0.25)
         end
-        space = get(gl_attributes, :space, :data)
+        space = get(gl_attributes, :space, :data) # needs to happen before connect_camera! call
         positions = handle_view(x.converted[1], data)
         positions = apply_transform(transform_func_obs(x), positions, space)
         if haskey(data, :color) && data[:color][] isa AbstractVector{<: Number}
@@ -304,7 +304,7 @@ function draw_atomic(screen::Screen, scene::Scene,
 
         transfunc =  Makie.transform_func_obs(scene)
         pos = gl_attributes[:position]
-        space = get(gl_attributes, :space, Observable(:data))
+        space = get(gl_attributes, :space, Observable(:data)) # needs to happen before connect_camera! call
         markerspace = gl_attributes[:markerspace]
         offset = pop!(gl_attributes, :offset, Vec2f(0))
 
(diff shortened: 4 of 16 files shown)


src/layouting/transformation.jl
```diff
@@ -195,6 +195,16 @@ transformationmatrix(x) = transformation(x).model
 transform_func(x) = transform_func_obs(x)[]
 transform_func_obs(x) = transformation(x).transform_func
 
+"""
+    apply_transform(f, data, space)
+Apply the data transform func to the data if the space matches one
+of the the transformation spaces (currently only :data is transformed)
+"""
+apply_transform(f, data, space) = space == :data ? apply_transform(f, data) : data
```
ffreyer, line 203: Just to point it out - this might need to be `copy(data)` if there are any mutating operations done on the output. I don't think there are, since we generally go through observables which create copies, but I thought I'd mention it anyway.

src/makielayout/interactions.jl
```diff
@@ -157,8 +157,10 @@ function process_interaction(r::RectangleZoom, event::MouseEvent, ax::Axis)
     end
 
     if event.type === MouseEventTypes.leftdragstart
-        data = Makie.apply_transform(inv_transf, event.data)
-        prev_data = Makie.apply_transform(inv_transf, event.prev_data)
+        # because these are designed to alter the axis, they 
+        # are kept in data space. 
+        data = Makie.apply_transform(inv_transf, event.data, :data)
+        prev_data = Makie.apply_transform(inv_transf, event.prev_data, :data)
```
ffreyer, line 163: I think all of the ones in this file can just continue being `Makie.apply_transform(inv_transf, event.data)` etc.

src/layouting/transformation.jl
dgleich: The identity transforms make a point of returning `data` (or `x`) directly (and verifies that by checking `===` in one of the test cases), so that behavior was mirrored here. I would think any issues with mutating the data would quickly show up in that regard. Let me know if you think this needs any changes.

src/makielayout/interactions.jl
dgleich: I reverted all the changes to this file.

dgleich:
I spotted a few other places in GLMakie.jl where I wasn't wrapping the space observable in a map. (When I was testing locally, this initially caused the log10 heatmap reference image to fail before I fixed it.) The other cases are in `function mesh_inner(screen::Screen, mesh, transfunc, gl_attributes)` and `function draw_atomic(screen::Screen, scene::Scene, x::Surface)` which I'm guessing don't have tests with `log10` scaled axis. I've edited them locally and am just running tests again before updating the PR. Will fix that and update shortly

src/layouting/transformation.jl
jkrumbiegel: Yeah, I did that at the time because I thought the overhead was worse. And because I didn't see the need for mutation afterwards, but still one should be careful

dgleich:
Okay, I fixed the other cases... GLMakie.jl passes tests locally.

src/layouting/transformation.jl
dgleich: Is there a change (documentation note maybe?) or test you would like to insert here to make sure this stays consistent in the future?

SimonDanisch:
Should we make this non breaking by having `apply_transform(f, data, space = :data)` ?

jkrumbiegel:
Is `apply_transform` exported?

SimonDanisch:
It's used in other packages, e.g. GeoMakie 🤷

jkrumbiegel:
Hm ok maybe it's the path of least resistance then?

SimonDanisch:
Yeah would be nice to not break GeoMakie if it's so easy to avoid...

ffreyer:
This shouldn't break anything. All the other spaces should exclude transformation function by definition. `space = :pixel` means a 1:1 corresponds between (x, y) inputs and pixel positions. `space = :clip` is strictly a [-1, 1] x [-1, 1] x [0, 1] coordinate system. `space = :relative` is a [0, 1] x [0, 1] x [0, 1] coordinate system. Having a transformation function other than identity will break these conventions.

SimonDanisch:
The breaking change is that `apply_transform` needs 3 arguments now, as far as I can tell...

ffreyer:
> ```julia
> apply_transform(f, data, space) = space == :data ? apply_transform(f, data) : data
> ```

Both exist. For implementing new functions you'd still extend the 2 argument version, but for usage it should generally be the 3 argument version.

SimonDanisch:
Jeez, you're right... I really need to up my github diff reading game ;) 
GeoMakie passes tests on this branch :) 

@dgleich, could you add a minimal test plot to: https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/short_tests.jl#L247, which now works with your changes?

Just add a block like: 
```julia
@reference_test "array of rects poly" begin
    plot_that_looks_correct_now(...)
end
```

dgleich:
@SimonDanisch Added! Was going to inquire if you wanted something like that too :)

SimonDanisch:
This looks wrong, no?
image

SimonDanisch:
This is the picture the CI created from the current run... And if I check out the PR locally, I get the same... How did you produce the picture?

dgleich:
Hmm... the CairoMakie images I downloaded here look okay and ran okay locally... 

https://github.com/MakieOrg/Makie.jl/actions/runs/3565833787

It looks like something changed in GLMakie.jl -- let me take a quick look.

dgleich:
Okay, tests are useful :)

The poly command for GLMakie.jl is the one that gives the problem. (This works with lines and scatter...) I'm still debugging.

jkrumbiegel:
In CairoMakie, `poly` is drawn directly while it is drawn as nested `mesh` + `wireframe` in GLMakie

dgleich:
Found the issue -- `connect_camera!` deletes the `:space` attribute in `gl_attributes` 

I can work around this. (e.g. the routines that _did_ work pulled out the space attribute before connect_camera.)

Is `GLMakie.mesh_inner` used in other packages? The easiest way to patch this is to give that the space parameter before connect_camera deletes it. That's what I did in the most recent set of edits I'll push after I double check the test works :)

dgleich:
Okay, new code passes the simple test. I added another test that tries to plot an image into a transformed axis with pixel... it fails on master and works with the new code (yay!)

SimonDanisch:
Ok, this looks good now: 
image
I think we should be able to merge this if CI runs through with the new reference image :)

dgleich:
Looking at fixing it now. I think I had a tiny typo in the new test and I’m
being more careful with the final edits.

dgleich:
Okay, I fixed the typo which passes GLMakie.jl tests locally.

SimonDanisch:
Great :) Thanks for hanging in there and pushing this through! ;) 
Sounds like the CairoMakie failure should be turned into an issue at least?

dgleich:
Okay -- I was again too hasty! Turns out the CairoMakie failure _is_ related to this request. Let me see if I can fix that one up too...

The existing code here:
```
if interpolate
        if !regular_grid
            error("$(typeof(primitive).parameters[1]) with interpolate = true with a non-regular grid is not supported right now.")
        end
        if !identity_transform
            error("$(typeof(primitive).parameters[1]) with interpolate = true with a non-identity transform is not supported right now.")
        end
    end
```

should also check if we are doing pixel-space plots...

SimonDanisch:
Sounds good to me!

PR #2437, opened 2022-11-23
Fix repeated rendering with invisible objects

ffreyer:
# Description

Fixes comment in #2336

An invisible object currently never resets it update requests. The pr fixes that. (Going from visible -> invisible needs to trigger a render, so the request needs to be true for an object that just became visible. We then need to reset every objects request, whether they rendered (are visible) or not.)

## Type of change

### DIFF
--- a/GLMakie/src/GLAbstraction/GLRender.jl
+++ b/GLMakie/src/GLAbstraction/GLRender.jl
@@ -55,9 +55,9 @@ So rewriting this function could get us a lot of performance for scenes with
 a lot of objects.
 """
 function render(renderobject::RenderObject, vertexarray=renderobject.vertexarray)
+    renderobject.requires_update = false
+    
     if renderobject.visible
-        renderobject.requires_update = false
-
         renderobject.prerenderfunction()
         program = vertexarray.program
         glUseProgram(program.id)


ffreyer:
> WARNING: Makie.MakieLayout is deprecatedThe module `MakieLayout` has been removed and integrated into Makie, so simply replace all usage of `MakieLayout` with `Makie`.
  likely near none:1
WARNING: importing deprecated binding Makie.MakieLayout into GLMakie.

?

> libGL error: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: swrast
┌ Warning:     GLFW couldn't create an OpenGL window.
│     This likely means, you don't have an OpenGL capable Graphic Card,
│     or you don't have an OpenGL 3.3 capable video driver installed.
│     Have a look at the troubleshooting section in the GLMakie readme:
│     https://github.com/MakieOrg/Makie.jl/tree/master/GLMakie#troubleshooting-opengl.
└ @ GLMakie ~/work/Makie.jl/Makie.jl/GLMakie/src/screen.jl:224
ERROR: LoadError: LoadError: GLFWError (VERSION_UNAVAILABLE): GLX: Failed to create context: GLXBadFBConfig

And the GLFW error?

SimonDanisch:
GLFW error is this: https://github.com/MakieOrg/Makie.jl/pull/2439

>WARNING: importing deprecated binding Makie.MakieLayout into GLMakie.

I don't think we can avoid this until we remove the `Base.@deprecate_binding MakieLayout Makie true` ... Which I guess we decided not to for a while 🤷

PR #2439, opened 2022-11-23
revert CI to ubuntu-20.04

jkrumbiegel:


Moelf:
>The recent change to 22.04 for ubuntu-latest broke GLMakie's CI.

why?

SimonDanisch:
Not sure ... It has been hard to debug the exact reasons... I think ubuntu just doesn't have a great test suite for headless, software mode OpenGL, so it seems to break with every new ubuntu version (this has been happening quite reliably) 🤷

PR #2458, opened 2022-11-29
RFC: Fix polygon drawing in `CairoMakie`

### LINKED ISSUE #2456: `CairoMakie`: polygon with hole example not working
See the 2nd example at the docs page: https://docs.makie.org/v0.18.3/examples/plotting_functions/poly/index.html

Expected output: 
image

fatteneder:
# Description

Attempt to fix #2456 

The way I fix it is to add an `Cairo.clip` for a polygon's `interiors` before we draw its `exteriors`.
For the clip I am enclosing all the `interiors` paths in a big rectangle that is the size of the surface to 'invert' the clipping area (if that makes sense). I followed this guide here: https://www.cairographics.org/cookbook/outside_clipping/

Problem is that this now causes different outputs compared to `GLMakie` when one draws a self-intersecting polygon, e.g.
```julia
using CairoMakie
CairoMakie.activate!()
# using GLMakie
# GLMakie.activate!()

using Makie.GeometryBasics

f = Figure()
ax = Axis(f[1, 1])
p1 = Polygon(
    Point2f[(3, 0), (5, 0), (6, 1), (4, 1)],
    [Point2f[(3.75, 0.25), (10, 0.25), (10, 0.75), (4.25, 0.75)]]
)
poly!(ax, p1)
save("mwe_poly.png", f)
```

- `CairoMakie`
image

- `GLMakie`
image

Possible ways forward:
1. Merge this, but change the way polygons are drawn in `GLMakie`
2. Close this and undo #2398
3. Any other options?

## Type of change

## Checklist


asinghvi17:
It seems to me that CairoMakie is performing correctly, but this is probably not an input you'd see day to day.  It will definitely be better than what it is now, though!

SimonDanisch:
The added test seems to error?

fatteneder:
@asinghvi17 thanks for fixing the tests.
The failure on `v1.0` seems to be unrelated.

asinghvi17:
This is still needed I think - on current master, the output of the MWE looks like this in CairoMakie.
display


The solution for GLMakie might be to fix the triangulation somehow...

Issue #2454, opened 2022-11-28
`CairoMakie`'s png output empty for large enough surface plot


fatteneder:
# MWE

```julia
# using GLMakie
# GLMakie.activate!()
using CairoMakie
CairoMakie.activate!()

Nx, Ny = 512, 8001
x, y = collect(LinRange(-20,40,Nx)), collect(LinRange(-20,20,Ny))
xx, yy = repeat(x, 1, Ny), repeat(y', Nx, 1)
zz = @. sin(xx) * cos(yy)

f = Figure()
ax = Axis3(f[1,1])
surface!(ax, xx, yy, zz)

save("mwe.png", f)
```

- `GLMakie`:
![image](https://user-images.githubusercontent.com/26469701/204265007-8af9efd8-3d9e-438d-8f77-a9d230d7d44c.png)

- `CairoMakie.jl`:
![image](https://user-images.githubusercontent.com/26469701/204264921-6d639cbf-5976-43d7-a3a3-d233065af527.png)

`CairoMakie` seems to do something, because it takes a while for the script to complete.

jkrumbiegel:
have you tried `rasterize = 1` on `surface`?

fatteneder:
Another test says its the overall size of the surface, e.g. if `Nx*Ny >= 2560000 (= 512 * 5000)` then it refuses to plot.

SimonDanisch:
There's another weird limit like that: https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/primitives.jl#L604
Above that threshold, cairo also just stops rendering without any error....

jkrumbiegel:
Ok but the limit for a single texture makes more sense to me than a count limit for drawing meshes on a bitmap. Unless before the drawing is over, there is some state being accumulated where the count matters. Maybe one can flush that state before the drawing is over?

fatteneder:
With this patch here to read `cairo_pattern_status` and the above MWE I get
```julia
julia> include("mwe.jl")
cairo_pattern_status = no error has occurred
cairo_pattern_status = no error has occurred
cairo_pattern_status = no error has occurred
cairo_pattern_status = out of memory
CairoMakie.Screen{IMAGE}
```

fatteneder:
Not sure why there are 4 calls to `mesh2d`, but I guess it doesn't matter.

I also found two threads discussion some 32k pixel limits:
[...]

PR #2459, opened 2022-11-30
Fix CairoMakie mesh overflow

jkrumbiegel:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/2454

It seems that just adding patches to a single pattern causes an overflow at some point, corrupting Cairo's internals.
For bitmaps, it should make no difference if each triangle is drawn with a separate patch, because these aren't anti-aliased anyway so they can be drawn sequentially without problem. I'm not sure if the file size of vector graphics containing these will be higher, but in that case one could probably still flatten to a single bitmap by setting `rasterize = true` to the `surface` call.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -713,9 +713,9 @@ function draw_mesh2D(scene, screen, per_face_cols, space::Symbol,
     ctx = screen.context
     # Priorize colors of the mesh if present
     # This is a hack, which needs cleaning up in the Mesh plot type!
-    pattern = Cairo.CairoPatternMesh()
 
     for (f, (c1, c2, c3)) in zip(fs, per_face_cols)
+        pattern = Cairo.CairoPatternMesh()
         t1, t2, t3 =  project_position.(scene, space, vs[f], (model,)) #triangle points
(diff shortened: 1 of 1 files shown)


fatteneder:
This does not eliminate the `out of memory` error entirely, or?
If I run the MWE from #2454 with an even much higher resolution then this could happen again, or?

Would it make sense to figure out this 'limit' and put a check in like we have it images/heatmaps?

jkrumbiegel:
The higher the mesh resolution the less likely the bug I think. It's either due to number of patches within pattern (fixed completely if true) or due to pattern surface area (could still be hit with a super large triangle I guess?)

fatteneder:
I tested the MWE with `Nx, Ny = 5000, 5000` and it works now.

But I also saw silent crashes of julia with `Nx, Ny = 8000`. However, that happened also with `GLMakie` (although with higher `N`s), likely because I ran out of RAM and swap (only using 8GB + 8GB atm).

jkrumbiegel:
Yeah that sounds like a different issue again I think, I'm going to merge this one and we can open another issue for the other crashes.

PR #2474, opened 2022-12-07
Add Search Page and tweak Result Ordering

### LINKED ISSUE #2448: Improve search result ordering, make results more concise.
The Makie documentation search is lightning fast, which is lovely. But there are two problems that make it a whole lot less useful that it can be.

## Issue number 1: Search result ordering is far from optimal
Currently, more or less whatever I search generally returns results from the API section first:

Returning the API is especially useless because it is the same help that you get from the docstrings (right?), so that it is (at least for me) _**never**_ what I go to the full documentation to find.

Other times, the first result is hardly the most relevant:

This issue is made more painful from the second issue (verbose results), so that one often needs to scroll significantly to get to the relevant results.

Can the ordering of the results have some bias? Perhaps each file could get a particular "general relevance rating" for the search results? Or biased by the most frequently visited pages of the documentation (with some "forgetting factor" to remove initial bias from appearing first)?

## Issue number 2: Results are too verbose
This is all _a single_ result:

It is generally not that useful to see all sentences containing the word, as it is pretty much all code, and it can be hard to quickly parse the differences (particularly for newer users). I would **much** prefer seing something like "`**Filename**` (next line) `17 hits found`". This would make it possible to get an overview of the results, rather than the currently verbose and frankly unhelpful display of the hits. 

Alternativly, simply not showing _code_ in the results could make the display of hits go from annoying to useful.

Fixing this would greatly alleviate the pain from issue number 1, but fixing both would clearly be worth some effort IMO.

bluzukk:
# Description

Fixes #2448

**Result Ordering**
This request tweaks the stork configuration such that the result order is optimized. This is done by using the title_boost option, which determines how much a result will be boosted if the search query matches the title.

**Search Page**
This pull request adds a search-page with more space for displaying the results, which can be accessed by pressing enter in the search-box.

**Additional Stork Configuration File**
This request adds an additional config file for stork such that the results displayed on the search-page and the results displayed in the search-box can be configured independently. This makes it possible to have a more verbose search result being displayed on the search-page. 
The search-box now shows 3 (instead of 5) occurrences of the search query, while the search-page shows 7.
Please share your thoughts about how many occurrences should be displayed in the respective areas.

**Navbar Update**
I adjusted the Navbar CSS slightly to improve visibility of search-box results (see screenshots). 
This prevents the "cut off" the search results and also prevents a "double scrollbar" within the search results.

## Type of change

## Checklist

### Result Ordering 
The old version (left) and the new version (right).
Note that the search-box now adjusts its size depending on the view height.
<p>
    
   
</p>

### Search Page
search-page

### DIFF
--- a/docs/makedocs.jl
+++ b/docs/makedocs.jl
@@ -172,10 +172,18 @@ function populate_stork_config(deploydecision)
     finally
         cd(wd)
     end
-    cp("__site/libs/stork/config.toml", "__site/libs/stork/config_filled.toml", force = true)
+    cp("__site/libs/stork/config_box.toml", "__site/libs/stork/config_box_filled.toml", force = true)
+    cp("__site/libs/stork/config_page.toml", "__site/libs/stork/config_page_filled.toml", force = true)
 
-    toml = TOML.parsefile("__site/libs/stork/config.toml")
-    open("__site/libs/stork/config_filled.toml", "w") do io
+    toml = TOML.parsefile("__site/libs/stork/config_box.toml")
+    open("__site/libs/stork/config_box_filled.toml", "w") do io
+        toml["input"]["files"] = map(Dict ∘ pairs, sites)
+        subf = deploydecision.subfolder
+        toml["input"]["url_prefix"] = isempty(subf) ? "" : "/" * subf * "/" # then url without / prefix
+        TOML.print(io, toml, sorted = true)
+    end
+    toml = TOML.parsefile("__site/libs/stork/config_page.toml")
+    open("__site/libs/stork/config_page_filled.toml", "w") do io
         toml["input"]["files"] = map(Dict ∘ pairs, sites)
         subf = deploydecision.subfolder
         toml["input"]["url_prefix"] = isempty(subf) ? "" : "/" * subf * "/" # then url without / prefix
@@ -189,7 +197,8 @@ function run_stork()
     wd = pwd()
     try
         cd("__site/libs/stork")
-        run(`$stork build --input config_filled.toml --output index.st`)
+        run(`$stork build --input config_box_filled.toml --output index_box.st`)
+        run(`$stork build --input config_page_filled.toml --output index_page.st`)
     finally
         cd(wd)
     end
--- a/docs/makedocs.jl
+++ b/docs/makedocs.jl
@@ -174,7 +174,6 @@ function populate_stork_config(deploydecision)
     end
 
     for file in ["config_box", "config_page"]
-        println("__site/libs/stork/$(file).toml")
         cp("__site/libs/stork/$(file).toml", "__site/libs/stork/$(file)_filled.toml", force = true)
 
         toml = TOML.parsefile("__site/libs/stork/$(file).toml")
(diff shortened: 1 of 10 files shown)


docs/makedocs.jl
```diff
@@ -172,10 +172,18 @@ function populate_stork_config(deploydecision)
     finally
         cd(wd)
     end
-    cp("__site/libs/stork/config.toml", "__site/libs/stork/config_filled.toml", force = true)
+    cp("__site/libs/stork/config_box.toml", "__site/libs/stork/config_box_filled.toml", force = true)
+    cp("__site/libs/stork/config_page.toml", "__site/libs/stork/config_page_filled.toml", force = true)
 
-    toml = TOML.parsefile("__site/libs/stork/config.toml")
-    open("__site/libs/stork/config_filled.toml", "w") do io
+    toml = TOML.parsefile("__site/libs/stork/config_box.toml")
```
jkrumbiegel, line 178: This should be a loop if the new section does the same as the old, otherwise it's hard to see if there might be differences (I didn't see any)

jkrumbiegel (changes requested):
Very cool, thanks for the PR! While trying it out locally it mostly seemed to work well, the title boosting is nice and the search page as well. The amount of text in search box and search page seems reasonable to me. The jumping to the closest id in the document doesn't seem to be that exact, but that's probably a browser / css / html issue that doesn't need to be solved here.

One issue I noticed, though, is that searching on a page in a subdirectory appends the path of that page to the search result, leading to a 404. The live page doesn't do that so something must have changed in that regard.

docs/makedocs.jl
bluzukk: Yeah, it does the same as the old but for two files. I will make it a loop

bluzukk:
Thanks for your reply and the review!


> One issue I noticed, though, is that searching on a page in a subdirectory appends the path of that page to the search result, leading to a 404. The live page doesn't do that so something must have changed in that regard.

Yes, this issue also occurs if I try it out locally.
However, if I build the docs using the Makie master branch this happens too!
I assume that the problem is resolved if the docs are deployed to the live page.

I did dig a little deeper into this:
The problem is the "url_prefix" which gets added to `config_filled.toml` :

```python
    toml["input"]["url_prefix"] = isempty(subf) ? "" : "/" * subf * "/" # then url without / prefix
```

since `subf` is always empty when deploying locally the "url_prefix" is empty but it should be "/" to make all linking from subdirectories work. The following solves this problem :) (Tested it on Makie:Master and this Version)

```python
    toml["input"]["url_prefix"] = isempty(subf) ? "/" : "/" * subf * "/" # then url without / prefix
```

I fixed it in [5e696b3](https://github.com/MakieOrg/Makie.jl/pull/2474/commits/5e696b3d9b27787145545fbc43acd7a9713b466d)

jkrumbiegel (approved):

SimonDanisch (approved):

KronosTheLate:
Awesome, thanks for this ^_^

PR #2490, opened 2022-12-12
Add version warner to all old docs

jkrumbiegel:
I spliced the necessary js and css into all old doc builds. I tested locally and it seemed to work ok, although we won't have a live test before merging this.

The same stuff needs to go into a separate PR so that it also appears in new builds, but for now I thought a warning in old versions was the priority.

Looks like this:

image

I changed the Documenter css a bit so the warning can't be removed, but it also doesn't cover anything, it just sits atop the masthead. I always found it a bit annoying clicking away the warning just for it to reappear on every page.


jkrumbiegel:
Saving the code to rewrite here just in case I need to reference it again:

```julia
using Gumbo: Gumbo, parsehtml, HTMLElement, setattr!


for folder in readdir()
    (isdir(folder) && !success(`test -L $folder`)) || continue
    version = try
        VersionNumber(folder)
    catch
        continue
    end
    version > VersionNumber("v0.15") || continue
    @show version

    for (root, dirs, files) in walkdir(folder)
        for file in files
            endswith(file, ".html") || continue
            rpath = relpath(joinpath(root, file), folder)
            @info rpath
            html = parsehtml(read(joinpath(root, file), String))
            head::HTMLElement{:head} = html.root.children[1]

            for jsfile in ["siteinfo.js", "libs/warner.js", "/versions.js"]
                reljs = if startswith(jsfile, "/")
                    jsfile
                else
                    relpath(jsfile, normpath(splitdir(rpath)[1]))
                end
                @show reljs
                jsel = HTMLElement(:script)
                setattr!(jsel, "src", reljs)
                setattr!(jsel, "type", "text/javascript")
                @show jsel
                push!(head, jsel)
            end

            open(joinpath(root, file), "w") do f
                return print(f, html)
            end
        end
    end
end
```

PR #2491, opened 2022-12-12
Fix searchpage redirect not respecting prefix

### LINKED ISSUE #2486: Search page redirect doesn't respect prefix
When you search something on `docs.makie.org/dev` and press enter, you get to `https://docs.makie.org/search?q=` so the `/dev` prefix is missing. This leads to a 404. @bluzukk

bluzukk:
# Description

Fixes #2486 

This fixes the redirect to the searchpage by using the `deploydecision.subfolder `variable. 
It works locally and now hopefully on the live page too.

## Type of change

### DIFF
--- a/docs/_layout/foot.html
+++ b/docs/_layout/foot.html
@@ -36,6 +36,9 @@
       stork.register("makiesearch", `${storkfolder}/index_page.st`);
     </script>
     {{else}}
+    <!-- Franklin.jl ignores page variables inside script tags therefore we
+    need this placeholder to get the prefix variable -->
+    <label id="prefix-placeholder" style="display:none">{{prefix}}</label>
     <script>
       storkfolder = document.querySelector("#stork-library-path-placeholder").src;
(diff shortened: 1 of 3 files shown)


bluzukk:
Is there a way I can test such changes in some live environment or create a preview?
Like this https://github.com/MakieOrg/Makie.jl/tree/gh-pages/previews/PR2428 for example?

jkrumbiegel:
Can you download the workflow artifact? When you unzip that you can serve it locally, for example with LiveServer.jl

bluzukk:
Unfortunately I can not download the workflow artifact

jkrumbiegel:
That's impractical.. The preview builds are only made for members of the org because they need key access.

bluzukk:
Oh I indeed can download the artifact! I just could not find it... I will try it with LiveServer.jl

bluzukk:
Can confirm its working.

jkrumbiegel (approved):

PR #2495, opened 2022-12-13
Revert "Fix searchpage redirect not respecting prefix"

jkrumbiegel:
Reverts MakieOrg/Makie.jl#2491

This broke https://docs.makie.org/dev/ because the paths are now all double-prefixed:

image

### DIFF
--- a/docs/_layout/foot.html
+++ b/docs/_layout/foot.html
@@ -36,9 +36,6 @@
       stork.register("makiesearch", `${storkfolder}/index_page.st`);
     </script>
     {{else}}
-    <!-- Franklin.jl ignores page variables inside script tags therefore we
-    need this placeholder to get the prefix variable -->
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
I remember vaguely that I left the Franklin prefix variable blank, because I wanted to have a movable build (which would stay intact when moved in the gh-pages folder hierarchy). Before, we had problems linking to the current version as `stable` but all links had hardcoded versions in them from the prefix, so I rewrote all resources to relative links.

bluzukk:
This is my bad. Im sorry! 

So the _prepath_ variabe needs to be blank (also the franklin Website states this or css will break). 

What I didnt know when doing my changes is that the _prefix_ variable is appenrently an alias for the _prepath_ variable. 

So the solution to this is to rename the variable _prefix_ to something else.

Issue #832, opened 2021-02-01
Feature request: contour labels


briochemc:
I don't think one can annotate contours with values in Makie (yet), for example (matplotlib):

<img width=50% src="https://matplotlib.org/3.3.3/_images/sphx_glr_contour_demo_001.png">

I could not find an existing issue for this, so I thought I'd create one! 😃 

---

FWIW, I only found [this discourse post](https://discourse.julialang.org/t/makie-jl-plotting-contour-with-fillrange-true-above-lines/39830/6) with a potential workaround, with this plot:

<img width=50% src="https://aws1.discourse-cdn.com/business5/uploads/julialang/original/2X/d/d55de321c4d7d36c52319a76555617afb4cf5059.png">

jkrumbiegel:
For this we would need this PR https://github.com/JuliaPlots/AbstractPlotting.jl/pull/430 to be merged, which is still stuck at the GLMakie implementation

crstnbr:
Came up on discourse again: https://discourse.julialang.org/t/how-to-add-value-labels-on-top-of-contours-in-makie/59242

Now that #430 is merged this should be possible somehow?

SimonDanisch:
Yeah it does work nicely now:
```
using GLMakie

np = 500
nc = 200
f(x) = 2*x+1
x = LinRange(-1,1,f(np))
y = rand(f(np))*2 .- 1
x_c = LinRange(-nc/np,nc/np,f(nc))
y_c = LinRange(-nc/np,nc/np,f(nc))
c_mat = @. x_c^2 + y_c'^2
fig, ax, hp = heatmap(x_c, y_c, c_mat)
cplot = contour!(ax,x_c,y_c,c_mat, linewidth=2, color=:white)

beginnings = Point2f0[]; colors = RGBAf0[]
# First plot in contour is the line plot, first arguments are the points of the contour
segments = cplot.plots[1][1][]
for (i, p) in enumerate(segments)
    # the segments are separated by NaN, which signals that a new contour starts
    if isnan(p)
        push!(beginnings, segments[i-1])
    end
end
sc = scatter!(ax, beginnings, markersize=50, color=(:white, 0.1), strokecolor=:white)
translate!(sc, 0, 0, 1)
# Reshuffle the plot order, so that the scatter plot gets drawn before the line plot
delete!(ax, sc)
delete!(ax, cplot)
push!(ax.scene, sc)
push!(ax.scene, cplot)
[...]

Issue #1653, opened 2022-02-08
Bounding Boxes


ffreyer:
I wanted to collect some issues/thoughts on bounding boxes. I'll work on them after #1596 gets merged.

#### Bounding Boxes ignore marker sizes

They probably shouldn't...

```julia
fig, ax, p = meshscatter(Point2f[(0,0), (1,1)], marker = Rect2f(-0.5, -0.5, 1, 1))
bb = boundingbox(p)
# GeometryBasics.HyperRectangle{3, Float32}(Float32[0.0, 0.0, 0.0], Float32[1.0, 1.0, 0.0])
lines!(ax, Rect2f(bb), color = :red)
```

![Screenshot from 2022-02-08 15-53-52](https://user-images.githubusercontent.com/10947937/153013306-304bb2a5-9883-4161-a7d3-752c749b54de.png)

#### User facing text plots have no bounding box

```julia
fig, ax, p = text("Test")
bb = boundingbox(p)
# GeometryBasics.HyperRectangle{3, Float32}(Float32[0.0, 0.0, 0.0], Float32[0.0, 0.0, 0.0])
```

#### Internal text bounding boxes always have pixel units widths and data space origins

They should be in whatever space the primary plot is in 

```julia
fig, ax, p = text("Test", position = Point2f(7.3, -0.3))
bb = boundingbox(p.plots[1])
# GeometryBasics.HyperRectangle{3, Float32}(Float32[7.3, -0.3, 0.0], Float32[42.36328, 23.28125, 0.0])
```

#### Bounding boxes in different spaces

Sometimes it may be useful to get a bounding box in a different space than the plot is in. For example, if you want to annotate a point in 3D with a white background pane it would be useful to plot the text at a 3D data space position and the pane using its screen/pixel space bounding box.

#### Per-marker bounding boxes

Getting bounding boxes per marker or whatever makes sense for a given plot might also be useful. I think it would be necessary for a ray based point picking system and probably for a collision system. It can also be useful to highlight hovered objects with `pick`.

koehlerson:
Let me know if I should open a separate issue for it, but it seems that the `space=:pixel` option doesn't work in recipes 
```julia
t = Makie.text!(Ele,"$id", position=position, textsize=Ele[:textsize])
bb = Makie.boundingbox(t)
Makie.wireframe!(Ele,bb,space=:pixel)
```
produces 
[...]

Issue #1405, opened 2021-10-24
Heatmap with log scale colorbar (cscale) ?


ederag:
Reposted from [discourse](https://discourse.julialang.org/t/heatmap-with-log-scale-colorbar-cscale/63018),
with updated CairoMakie, figures and link to documentation.

Makie is awesome, looks like everything is doable :slightly_smiling_face:,
for instance, this heatmap with log scale color axis (exactly what I wanted):

![image](https://user-images.githubusercontent.com/10962809/138586021-2c718073-6545-4ab6-8742-c6270366419f.png)

<details>
 <summary>Code for the above figure</summary>

```julia
let
	# cf. https://github.com/JuliaPlots/Makie.jl/issues/822#issuecomment-769684652
	# with scale argument that is required now
	struct LogMinorTicks end
	
    function MakieLayout.get_minor_tickvalues(
			::LogMinorTicks, scale, tickvalues, vmin, vmax
	)
    	vals = Float64[]
		extended_tickvalues = [
			tickvalues[1] - (tickvalues[2] - tickvalues[1]);
			tickvalues;
			tickvalues[end] + (tickvalues[end] - tickvalues[end-1]);
		]
    	for (lo, hi) in zip(
				@view(extended_tickvalues[1:end-1]),
				@view(extended_tickvalues[2:end])
			)
        	interval = hi-lo
        	steps = log10.(LinRange(10^lo, 10^hi, 11))
        	append!(vals, steps[2:end-1])
    	end
		return filter(x -> vmin < x < vmax, vals)
	end
	
	custom_formatter(values) = map(
		v -> "10" * Makie.UnicodeFun.to_superscript(round(Int64, v)),
		values
	)
	
	x = 10.0.^(1:0.1:4)
	y = 1.0:0.1:5.0
	data = x .* ones(Float64, 1, length(y))
	clims = (1e1, 1e4)  # natural display
	fig = Figure()
	ax, hm = heatmap(fig[1, 1], x, y, log10.(data), colorrange=log10.(clims),
		axis=(;xscale=log10,
			   xminorticksvisible=true,
			   xminorticks=IntervalsBetween(9))
	         )
	cb = Colorbar(fig[1, 2], hm;
		tickformat=custom_formatter,
		minorticksvisible=true,
		minorticks=LogMinorTicks()
	)
	fig
end
```
</details>

But it was a bit involved for a newcomer. Here is how I would have expected it to work:

```julia
[...]

PR #2496, opened 2022-12-13
implement contour labels

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/832.

~~Very early stage implementation, much to be discussed.~~

# 2D

```julia
using CairoMakie

main() = begin
  paraboloid = (x, y) -> 10(x^2 + y^2)
  x = range(-3, 3; length = 100)
  y = range(-3, 3; length = 200)
  z = paraboloid.(x, y')

  fig, ax, hm = heatmap(x, y, z; interpolate = true)
  Colorbar(fig[1, 2], hm)

  contour!(
    ax, x, y, z;
    color = :red, levels = 0:10:100, labels = true,
    labelsize = 15, labelfont = :bold, labelcolor = :orange
  )

  save("contour_labels_2d.png", fig)
  fig
end

main()
```
contour_labels_2d

# 3D

~~currently failing, transformation or projection issue, comments welcome :)~~

```julia
using CairoMakie

main() = begin
  fig = Figure()
  ax = Axis3(fig[1, 1])

  xs = ys = range(-.5, .5; length = 100)
  zs = @. √(xs^2 + ys'^2)

  levels = .025:.05:.475
  contour3d!(-zs; levels = -levels, labels = true, color = :blue)  # labels inherit linecolor by default
  contour3d!(+zs; levels = +levels, labels = true, color = :red, labelcolor = :black)

  save("contour_labels_3d.png", fig)
  fig
end

main()
```
contour_labels_3d

- [x] rotate labels automagically to align on the local tangent ?
- [x] dynamic contour label attributes (fontsize, color, ...) ?
- [x] ~~check~~ fix contour3d

## Type of change

## Checklist


t-bltg:
@jkrumbiegel, I'm a bit stuck in 3D.
The text bounding boxes are not correct, and I cannot find a way to fix it.
Using `wireframe!(plot, bb, space = :pixel)` as commented in the PR code draws the correct bboxes in the 2D examples, but renders non sense in 3D.
I tried to `lift(scene.camera.projectionview, scene.px_area)` to recompute the bounding boxes but it fails.
I think it might be related to https://github.com/MakieOrg/Makie.jl/issues/1653, but I have no ideas left ...
Could you advise ?

jkrumbiegel:
What do the wrong boundingboxes look like? Could also always still be a bug with space=:pixel I guess, unless you think the boundingboxes themselves are wrong somehow

t-bltg:
@jkrumbiegel:

contour_labels_2d

contour_labels_3d

t-bltg:
@jkrumbiegel, there is a weird effect of `wireframe!` on the limits in 3D, if I force axes limits:
```julia
xlims!(ax, 0, 100)
ylims!(ax, 0, 100)
zlims!(ax, -0.5, 0.5)
```
this is what I get:
contour_labels_3d

t-bltg:
If I remove the `scene.camera.projectionview, scene.px_area` from the listeners on bboxes computation, then no bounding box is drawn in 3D.

This explain at least why the label masking in 3D doesn't appear: I think a masked line is drawn on top of an un-masked line (before the projection matrix changes), thus hidding the masking effect.

jkrumbiegel:
Ah Axis3 probably just doesn't ignore limits of plots with space=:pixel yet, there's still stuff to clean up.

On the second image the boxes don't look too bad right? Although not all of them are filled?

t-bltg:
Is there a way to delay stuff until the final projection-view is computed ?

jkrumbiegel:
No, you should update whenever that matrix changes to be sure you've got the correct boundingboxes

t-bltg:
I tried a few things, but without knowing the current position in the plot pipeline I don't know how to tackle this.

src/basic_recipes/contours.jl
```diff
@@ -172,18 +189,63 @@ function plot!(plot::T) where T <: Union{Contour, Contour3d}
...
         xv, yv = to_vector(x, size(z,1), t), to_vector(y, size(z,2), t)
         contours = Contours.contours(xv, yv, z,  convert(Vector{eltype(z)}, levels))
-        contourlines(T, contours, level_colors)
+        contourlines(T, contours, level_colors, labels)
+    end
+    scene = parent_scene(plot)
+
+    texts = lift(labels, label_attributes, color, result) do labels, label_attributes, color, (_, _, str_pos)
+        labels || return
+        map(str_pos) do (str, (p1, p2, p3))
```
jkrumbiegel, line 205: Here's the problem in the logic. You're reacting to `labels, label_attributes, color, result` to set the rotation, and you actually do `text!(scene...` within the lift block. You should never do that because you really only want one `text` object, but this is supposed to be updated.

So make an empty `text` plot before this `lift` block which takes an observable of a vector of strings as input. These can be empty at first.

Then you do `onany(labels, label_attributes, whatever_you_need..., scene.camera.projectionview, scene.px_area)` and in that block you first compute the labels and their positions in the scene. Then you update the `text` plot with that. Now, in the same block, you will be able to query the new boundingboxes from the `Vector{GlyphCollection}` underlying the `text` plot because it has just been updated. With these boundingboxes, you null out the affected points in the contour lines copy.

t-bltg: Ok, thanks for the hints, I will try moving the `text!` out of the `lift` block and change the approach.

One problem is that you use a single text elements, and I couldn't find a way to access individual text elements i.e. a single glyph collection, to be able to compute individual bboxes.

jkrumbiegel: Ok yeah this is not yet so well developed, but I made it work like this:

```julia
s = Scene(camera = campixel!)
points = Point2f[(100, 100), (200, 200)]
t = text!(s, points, text = ["One string", "Another longer string"], rotation = [0, 0.5pi])
glyphcollections = t.plots[1][1][]
Makie.broadcast_foreach(glyphcollections, points, Makie.to_rotation(t.rotation[])) do gc, point, rot
    wireframe!(s, Makie.boundingbox(gc, Makie.to_ndim(Point3f, point, 0), rot))
end
s
```

image

jkrumbiegel: You might even want to do it differently and compute the boundingbox before rotation, then rotate it, because otherwise the bbox is pretty big for long oblique strings. But that can be an optimization for later.

```julia
s = Scene(camera = campixel!)
points = Point2f[(100, 100), (200, 200)]
t = text!(s, points, text = ["One string", "Another longer string"], rotation = [0, 0.25pi])
glyphcollections = t.plots[1][1][]
Makie.broadcast_foreach(glyphcollections, points, Makie.to_rotation(t.rotation[])) do gc, point, rot
    wireframe!(s, Makie.boundingbox(gc, Makie.to_ndim(Point3f, point, 0), rot))
end
s
```
image

t-bltg: > You might even want to do it differently and compute the boundingbox before rotation, then rotate it, because otherwise the bbox is pretty big for long oblique strings.

Ha yeah, the `30` label in 2d.

jkrumbiegel: 3D seems super annoying as everything overlaps :)

t-bltg: So I've reworked using the approach described in https://github.com/MakieOrg/Makie.jl/pull/2496#discussion_r1049424057.

It still fails 3D:

contour_labels_2d
contour_labels_3d

jkrumbiegel: Looks weird, the boundingboxes are half correct, half in the wrong place but in the same configuration, both sets of labels are in the same place but ever so slightly shifted. Didn't you have label placement calculated correctly before?

t-bltg: Right, wrong position type.

contour_labels_3d

Now, were back to the pipeline update problem described in https://github.com/MakieOrg/Makie.jl/pull/2496#issuecomment-1352126622, and the project to `:pixel` space in 3D.

jkrumbiegel: You still have duplicated/shifted boundingboxes. How are you drawing those again?

t-bltg: `wireframe!(plot, bb, space = :pixel)`, and the bounding box `bb` updated on `lift(scene.camera.projectionview, scene.px_area, labels)`.

t-bltg: Also, the `masked_lines = lift(...)` block gets triggered `22` times which is a lot, and might hurt performance on a big dataset.

t-bltg: Ooooh ! After reading [this](https://github.com/MakieOrg/Makie.jl/blob/6da4add742f7a3b465b8c5c3272284e8e81e599c/src/makielayout/blocks/label.jl#L20) it seems that dropping the `z` component of the bounding box in `3D` just works !

Makie would benefit from having commented code. Are we supposed to discover things like this on our own ?

contour_labels_3d

jkrumbiegel:
Very cool, it seems to work quite well overall! One thing I noticed just now is that rotating around in 3D doesn't adjust the label rotations:

grafik

t-bltg:
> One thing I noticed just now is that rotating around in 3D doesn't adjust the label rotations:

Fixed.

t-bltg:
I'm not `100%` happy with the `label_attributes` name.
Are there alternative ideas ?

jkrumbiegel:
You will not want to allow setting all attributes anyway, so I usually prefer spelling them out exactly. You won't do align, justification, rotation, position, so what remains is `labelcolor`, `labelfont`, `labelsize`?

t-bltg:
Great.

For consistency with the neighbor source file `src/basic_recipes/barplot.jl`, I've used underscores e.g. `label_color`.

I also find a bit inconsistent the use of `label_size` instead of `label_fontsize`, but shorter is also better.

jkrumbiegel:
yeah it's a tradeoff, I'm also ok with `label_fontsize`. In `Axis` it's all without underscore like `xticklabelsize` 🤷

t-bltg:
So in `Makie` are there more kw with underscores or not ?
Let's minimize those inconsistencies.

jkrumbiegel:
That error means you passed the attribute `labelfont` to a plot primitive and it wasn't handled before reaching GLMakie. For `text` it needs to be called `font` so that's probably where the mismatch lies.

t-bltg:
Ok, I think this is fixed, I've updated https://github.com/MakieOrg/Makie.jl/pull/2496#issue-1494198895.

t-bltg:
I don't understand why text cannot accept and observable in `GLMakie`, MRE:
```julia
julia> using GLMakie
julia> r = LinRange(-1, 1, 100)
julia> cube = [(x.^2 + y.^2 + z.^2) for x = r, y = r, z = r]
julia> contour(cube, alpha=0.5)
┌ Warning: Can't convert Observable{Symbol} to opengl, for attribute labelfont
└ @ GLMakie.GLAbstraction [...]/.julia/packages/GLMakie/IZlSz/src/GLAbstraction/GLTypes.jl:427
ERROR: MethodError: no method matching gl_convert(::Symbol)
Closest candidates are:
  gl_convert(::T) where T<:ColorTypes.Colorant at [...]/.julia/packages/GLMakie/IZlSz/src/GLAbstraction/GLUniforms.jl:194
  gl_convert(::T) where T<:Number at [...]/.julia/packages/GLMakie/IZlSz/src/GLAbstraction/GLUniforms.jl:193
  gl_convert(::Quaternion) at [...]/.julia/packages/GLMakie/IZlSz/src/glshaders/particles.j
```

t-bltg:
The solution is to ignore those attributes in the context of a volume plot.

jkrumbiegel:
This one should also get examples in the documentation to make it discoverable. What are your thoughts about the current implementation, is that already mergeable from your point of view?

t-bltg:
> This one should also get examples in the documentation to make it discoverable.

Right, should it go in [contour.md](https://github.com/MakieOrg/Makie.jl/blob/master/docs/examples/plotting_functions/contour.md) ?

> is that already mergeable from your point of view

Yes, I guess that the 2D/3D implementation is robust enough, if we agree that the volume doesn't need labels. The only concern is the number of passes in `3D` but I guess this in a general `Makie` "problem" unrelated to this PR.

jkrumbiegel:
Does the number of passes also increase a lot when labels are not used? I think in that case it's ok and can be fixed later. If it makes every contour plot a lot slower, that wouldn't be great.

And yes, contour.md seems like a good place.

t-bltg:
> Does the number of passes also increase a lot when labels are not used?

No, identical. Also, it might be minor, since we reuse text vector and use `empty!` on them which is efficient and minimizes allocs.

I'm going to change the label position to the middle of the line segment. Do we have a default text formatter in Makie that I can use ? I'd like to limit the numbers of significant digits so that we don't pollute the plot with tons of labels. Maybe 2 digits at most is enough ? We could also add a step like label it once every `...` contour ?

t-bltg:
Here is the added doc example:

contour_labels_docs

jkrumbiegel:
How are the label placements on each contour actually determined?

t-bltg:
The label are currently positioned around half of the segment length.

jkrumbiegel:
Ok, I assumed something like that. For lines that end at the image boundary this will probably be a good choice, as the label would on average be further away from the edge. For loops, it can be in any arbitrary position. I wonder what other plotting libraries do to calculate good positions for the full set of labels, I doubt they rely on the essentially arbitrary point order in the contour vectors. But this doesn't have to be solved in this first PR.

t-bltg:
By the way I'd like the colors to be `log10` scaled in the docs example, how can I achieve that ?

I tried using `Makie.sampler` but without success ...

jkrumbiegel:
There currently is no transformation step between numeric values and colormap lookup. You can transform the colormap itself but that is not great because you'll get resolution problems for log scales.

t-bltg:
Well, I think I'll retain the following, as it's quite readable:

```julia
using CairoMakie
himmelblau(x, y) = (x^2 + y - 11)^2 + (x + y^2 - 7)^2
x = y = range(-6, 6; length=100)
z = himmelblau.(x, y')

levels = 10.0.^range(0.3, 3.5; length=10)
colormap = Makie.sampler(:hsv, 100; scaling=Makie.Scaling(x -> x^(1 / 10), nothing))
f, ax, ct = contour(x, y, z; labels=true, levels, colormap)
f
```

contour_labels_docs

jkrumbiegel:
Huh I didn't even know about that scaling thing!

t-bltg:
Yeah, maybe indirectly related to https://github.com/MakieOrg/Makie.jl/issues/1405, but left for another PR.

src/basic_recipes/contours.jl
```diff
@@ -45,32 +51,50 @@ $(ATTRIBUTES)
...
-    result = Point2f[]
+angle(p1::Union{Vec2f,Point2f}, p2::Union{Vec2f,Point2f})::Float32 =
+    atan(p2[2] - p1[2], p2[1] - p1[1])  # result in [-π, π]
+
+function label_info(lev, vertices, col)
+    mid = ceil(Int, 0.5f0 * length(vertices))
+    pts = (vertices[max(firstindex(vertices), mid - 1)], vertices[mid], vertices[min(mid + 1, lastindex(vertices))])
+    lev_short = round(lev; digits = 2)
+    (
+        string(isinteger(lev_short) ? round(Int, lev_short) : lev_short),
```
asinghvi17, line 62: It would be nice to allow the user to pass their own formatter through to here, in the same style as `ax.ticklabelformat`.

t-bltg: Good idea I'll add that.

t-bltg: Implemented.

lazarusA:
oh, this is a good one!. A pitty that is still not ready to merge.

t-bltg:
What makes you say it isn't ?

lazarusA:
apologies, maybe my mistake, I  just the CI failing. And your 3d wireframe example, I don't see the current example working.

asinghvi17:
@t-bltg mind if I merge master in?

t-bltg:
No worries.

The up-to-date examples (2D/3D) are in the first post (https://github.com/MakieOrg/Makie.jl/pull/2496#issue-1494198895).

The failing CI is due to missing references images, which need to be merged by a maintainer.

t-bltg:
Pending CI, can we get this PR in ?

SimonDanisch:
Thank you so much for this great PR :) Sorry for taking so long...

t-bltg:
Amazing, and many thanks to @jkrumbiegel for the hints and guidance towards (hopefully) quality code !

PR #2497, opened 2022-12-13
Add show_data for band

ffreyer:
# Description

Adds a custom `show_data` for `band` which displays the x position of the mouse and the relevant (interpolated) min and max value of the band at that x value. This also affects `density` and whatever else relies on `band`. Currently this shows a bounding box for the underlying mesh.

I'm making the assumptions that
- `band` is used in 2D
- x values are sorted in `plot.converted`
- `plots.converted` is always 2 Point2f arrays, with the first being low values and the second being high values
- x values between low and high points match

Example with `density(rand(100))`: (3rd commit)

Screenshot from 2022-12-14 11-41-57

## Type of change

Well technically it's changing the current behavior but a bounding box is not a useful thing to get here...

## Checklist

- ~~Added reference image tests for new plotting functions, recipes, visual options, etc.~~


jkrumbiegel:
Looks good, the only thing is that I think `(0.345, 0.000..0.932)` like in the example is a bit hard to read. Maybe spaces around the `..` like `(0.345, 0.000 .. 0.932)`? Or any other suggestions?

ffreyer:
Do we have (x, x, y) bands? I don't see a keyword or plot method for it.

ffreyer:
Oh it's a raw input of Points. I'll have to tweak this then.

> Looks good, the only thing is that I think `(0.345, 0.000..0.932)` like in the example is a bit hard to read. Maybe spaces around the `..` like `(0.345, 0.000 .. 0.932)`? Or any other suggestions?

Some thoughts:

-  `($x, $ymin, $ymax)` (or $xmin, $xmax, $y) was my first thought, but that's even worse for readability imo
- Two lines like `"($x, $ymin)\n($x, $ymax)"` would be another option
- The `$ymin..$ymax` matches the Interval notation which I thought was rather fitting for band. I guess we could also have `($ymin, $ymax)` or `"$ymin <= y <= $ymax"`
- Displaying the current position rather than band min/max would be an option, but I think less useful? I think that's more something for a separate function/tool which just shows `mouseposition()` regardless.
- Having two or even three (top, bottom, current pos) tooltips would require reworking `tooltip!` and is too much noise I think.

ffreyer:
I got stuck for a while trying to figure out a direct way to find the line I need to draw and ultimately just used an iterative method. Now it can handle densities in `:x` and `:y` direction and also bands like this:

```julia
inner = [(1 + rand()) * Point2f(cos(phi), sin(phi)) for phi in range(0, 2pi, length=51)]
outer = [(4 + rand()) * Point2f(cos(phi), sin(phi)) for phi in range(0, 2pi, length=51)]
fig, ax, p = band(inner, outer)
di = DataInspector(fig)
fig
```

Screenshot from 2022-12-25 03-30-40

You can also swap outer and inner without issues. (It will break if the quads look like `|><|`, e.g. if you swap lows and highs throughout the band. At this point band is malformed anyway though)

PR #2500, opened 2022-12-13
Fix search-page redirect not respecting prefix

### LINKED ISSUE #2486: Search page redirect doesn't respect prefix

bluzukk:
# Description

Fixes #2486 (issue)

PR #2491 did not fix the issue since the Franklin variable `prefix` is reserved  and an alias for `prepath`.
Instead of renaming the variable I found a better solution using relative paths by making use of the 
`make_links_relative() ` function in `makedocs.jl`

## Type of change

### DIFF
--- a/docs/_layout/foot.html
+++ b/docs/_layout/foot.html
@@ -36,6 +36,7 @@
       stork.register("makiesearch", `${storkfolder}/index_page.st`);
     </script>
     {{else}}
+    <img id="search-path-placeholder" src="/search" style="display:none"/>
     <script>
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Thanks for the fix, I should improve the situation with this whole prepath business at some point now that we already have two of these rather ugly workarounds, but for now it will have to do.

jkrumbiegel (approved):

PR #2501, opened 2022-12-13
Improve `bar_labels`: `LaTeXString`s and `fontsize`

greimel:
Unfortunately this doesn't work yet because 
```julia
text([1], [1]; text = [L"a"], align = Vec2f[[0, 0]])
```
doesn't work.

Here's an example
```julia
barplot([1, 2], [0.5, 0.2], bar_labels = [L"a", "b"])
```

fig


src/basic_recipes/barplot.jl
```diff
@@ -44,7 +44,7 @@ $(ATTRIBUTES)
         color_over_bar = automatic,
         label_offset = 5,
         label_font = theme(scene, :font),
-        label_size = 20,
```
greimel, line 47: @SimonDanisch you added `bar_labels` in https://github.com/MakieOrg/Makie.jl/pull/1069. Do you remember, if there was a reason you hard-coded the font size? I think it looks nicer, if the labels have the same font size as the axis labels, tick labels etc.

src/basic_recipes/barplot.jl
```diff
@@ -260,6 +260,6 @@ function Makie.plot!(p::BarPlot)
         highclip = p.highclip, lowclip = p.lowclip, nan_color = p.nan_color,
     )
     if !isnothing(p.bar_labels[])
-        text!(p, labels; align=label_aligns, offset=label_offsets, color=label_colors, font=p.label_font, fontsize=p.label_size, rotation=p.label_rotation)
```
greimel, line 263: to be reverted before merging (`align` needs to be implemented for plotting `LaTeXString`s)

src/basic_recipes/barplot.jl
jkrumbiegel: This was the old global default, but it should inherit the new value like all other functions

greimel:
That's how it looks now

fig2

```julia
let
	fig = Figure()
	lab1 = L"a - b"
	lab2 = L"\frac{a}{b} - \sqrt{b}"

	barplot(fig[1,1], [1, 2], [0.5, 0.2], bar_labels = [lab1, lab1], flip_labels_at = 0.3, direction=:x)
	barplot(fig[1,2], [1, 2], [0.5, 0.2], bar_labels = [lab1, lab1], flip_labels_at = 0.3)
	barplot(fig[2,1], [1, 2], [0.5, 0.2], bar_labels = [lab2, lab2], flip_labels_at = 0.3, direction=:x)
	barplot(fig[2,2], [1, 2], [0.5, 0.2], bar_labels = [lab2, lab2], flip_labels_at = 0.3)

	text(fig[1,3], [1], [1]; text = [lab2], align = [(-1, 1)])
	fig
	save("/tmp/fig.png", fig); run(`open /tmp/fig.png`)
end
```

* Do you have an idea, how the hlines in fraction and square root can be fixed? (probably also cc  @Kolaru)
* What do you think about factoring out `get_xshift` and `get_yshift`? I've just seen that this is used in `layouting.jl` as well. So one could share this functionality even more widely.

Kolaru:
> * Do you have an idea, how the hlines in fraction and square root can be fixed? (probably also cc  @Kolaru)

Somehow Makie is rendering the line at the incorrect position. I spent some time on it but couldn't quite understand this drawing pipeline and were the problem could come from. Due to how it looks, I suspect it has something to do with scaling of the lines at some point.

greimel:
I narrowed down the offender: it's the `align` keyword

https://user-images.githubusercontent.com/6280307/207872450-bb250387-4513-44ca-9e71-13c7179a9fbf.mov

jkrumbiegel:
> I narrowed down the offender: it's the `align` keyword
> Screen.Recording.2022-12-15.at.14.32.42.mov

Aren't you demonstrating in your video that it's offset and not align? :)

greimel:
No, it doesn't. I am not using observables there. The plot is redrawn completely at each slider change.

jkrumbiegel:
Ok then offset is just completely ignored for linesegments calculation :)

greimel:
Thanks for the hint! 

Let's see if I manage to digest all these abbreviations.

jkrumbiegel:
Sure!

greimel:
Does this need a reference test?

greimel:
friendly bump - is there something I can do do help getting this merged?

SimonDanisch:
Ah, yeah a reference test would be nice if it's not too much work ;)
Just add it too the bottom here: https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/primitives.jl

SimonDanisch:
Thank you!

PR #2504, opened 2022-12-17
Update recipes.md

jw3126:


docs/documentation/recipes.md
```diff
@@ -6,7 +6,7 @@ Recipes allow you to extend `Makie` with your own custom types and plotting comm
     If you're a package developer, it's possible to add recipes without adding all of
     `Makie.jl` as a dependency. Instead, you can use the `MakieCore` package, which
     is a lightweight package which provides all the necessary elements to create a recipe,
-    such as the `@recipe` macro, `convert_arguments` and `convert_attributes` functions,
+    such as the `@recipe` macro, `convert_arguments` function
```
jw3126, line 9: AFAICT there is no such thing as a `convert_attributes` function in Makie?

jw3126: Ok I was confused

SimonDanisch: Ah, but I guess you're right, it's called `convert_attribute`, so should be changed ;)

jw3126: The `convert_attribute` docstring is also super confusing. I will clean it up and fix the typo.

PR #2507, opened 2022-12-18
rework convert_attribute docstring

jw3126:
supersedes https://github.com/MakieOrg/Makie.jl/pull/2504


MakieCore/src/conversion.jl
```diff
@@ -1,5 +1,19 @@
...
+
+# Example
+```jldoctest
+julia> using Makie
+
+julia> Makie.convert_attribute(:black, key"color"())
+RGBA{Float32}(0.0f0,0.0f0,0.0f0,1.0f0)
+```
+"""
 function convert_attribute end
```
jw3126, line 17: I have a couple of questions about this function. Is this an internal function? Is the user supposed to overload it? If so are there typical cases, when a user might want to overload this?

jw3126: For instance I might have a custom recipe `particles` that takes attributes
`photoncolor` and `electroncolor.
Should I overload 
```julia
MakieCore.convert_attributes(value, ::key"photoncolor", ::key"phsp") = convert_attributes(value, ::key"color")
MakieCore.convert_attributes(value, ::key"electroncolor", ::key"phsp") = convert_attributes(value, ::key"color")
```
?

SimonDanisch: No, this is only for atomic plot types (like scatter, mesh, etc) and gets called in the backend.
You can use `to_color` in your recipe for such cases...
Overloading is only useful, if you have your own custom color type and want to use it directly as a color attribute!

SimonDanisch:
Thanks!

PR #2511, opened 2022-12-21
prevent potential type instabilities

t-bltg:
# Description

I haven't checked if actual type instabilities actually occurred, but it jumped out at me.

## Type of change

Non-functional change.


t-bltg:
GLMakie's ci seems KO also on master (same error: https://github.com/MakieOrg/Makie.jl/actions/runs/3748259620/jobs/6365399666#step:7:698).

SimonDanisch:
Somehow uploading the reference images didn't get updated after merging https://github.com/MakieOrg/Makie.jl/pull/2498 or got reverted, not sure, I specifically re-run the CI to check if everything worked and it did!
Lets see what it does now, after I re-uploaded the reference images...

Issue #2519, opened 2022-12-24
Inconsistent formatting and code styles


t-bltg:
Would you consider adding a `JuliaFormatter` action workflow and an initial format of the `Makie` repo ?

The code style across source file is inconsistent (not only from the margins pov, but also operator spacing, function one-liners, function declarations, indentation, ...).

I must admit this has bothered me for a while, and it is imo a bit of a pain to work with: it would help a lot if we had an initial format run, and added a format check action for PRs (not auto format PRs, because they pollute the git history), together with a custom `.JuliaFormatter.toml` file.

It's not a matter of whatever style is chosen (we don't care which one would be chosen): being inconsistent is not only disruptive but can also lead to endless discussion and arguing about those.

The `JuliaFormatter` tool might not be perfect yet, but we get decent results these days.

SimonDanisch:
Yeah, I planned to do that once upon a time, but at that point, `JuliaFormatter` was still erroring on the `Makie` repo, so I never went through with it... If `JuliaFormatter` is more mature now and runs without error, we could think about adding it :)

t-bltg:
Thanks, I'll try to come up with something.
Do you have a preferred style in mind ?

Regarding the bugs that might occur, I prefer to fix `JuliaFormatter` bugs and go on with this.

SimonDanisch:
I already pushed a Juliafromatter.toml when I tried to add it to the repo:
https://github.com/MakieOrg/Makie.jl/blob/master/.JuliaFormatter.toml

t-bltg:
Ho, I missed that :sweat_smile:, thx.

ffreyer:
We got Runic now

PR #2493, opened 2022-12-13
add default `colorscale` attribute

t-bltg:
# Description

~~Partially~~ fix #1405.

# First example

<details>
<summary>old approach</summary>

```julia
using CairoMakie

x = 10.0.^(1:0.1:4)
y = 1.0:0.1:5.0
z = broadcast((x, y) -> log10(x), x, y')  # NOTE: the scaling is also applied here

axis = (; scale = log10, minorticksvisible = true, minorticks = IntervalsBetween(9))

fig, ax, hm = heatmap(x, y, z; axis = NamedTuple(Symbol(:x, k) => v for (k, v) in pairs(axis)))
Colorbar(fig[1, 2], hm; minorticksvisible = true, minorticks = IntervalsBetween(9)

Makie.save("log_cbar.png", fig)
```

log_cbar

`test CairoMakie` passes locally.

~~I also believe that the user shouldn't be responsible for scaling height values (i.e. using `(x, y) -> x` should be scaled by `xscale` and `yscale`, maybe as suggested by adding a `cscale` or `zscale` for 2D `Axis`, but this is left for another issue / discussion).~~

</details>

~~Unscale the `Colorbar` axis limits, such that the following yields a correctly scaled colorbar:~~

Add a `colorscale` attribute, and make colorbar inherit this attribute by default. The rationale is that the colorbar colors are linked to the plot colors (or `z` values, heights). Consequently, only setting the `scale` attribute of the colorbar when creating it is not sufficient to correctly render scaled colors. This PR makes Makie consistent with other libraries such as `gnuplot` or `pyplot`.

**new approach**
```julia
using CairoMakie

x = 10.0.^(1:0.1:4);
y = 1.0:0.1:5.0;
z = broadcast((x, y) -> x, x, y');  # NOTE: don't apply the transformation here

axis = (; xscale = log10, xminorticksvisible = true, xminorticks = IntervalsBetween(9));

fig, ax, hm = heatmap(x, y, z; colorscale = log10, axis);  # NOTE: added `colorscale` attr
Colorbar(fig[1, 2], hm; minorticksvisible = true, minorticks = IntervalsBetween(9));  # inherits `hm.colorscale` by default

display(fig)
Makie.save("log_cbar.png", fig);
```
log_cbar

# Python version

```julia
import PythonPlot

main() = begin
  x = 10.0.^(1:0.1:4)
  y = 1.0:0.1:5.0
  z = broadcast((x, y) -> x, x, y')
[...]

PR #2520, opened 2022-12-24
Fmt

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/2519

- Add a format check action, running on PRs.
- Apply an initial format.

I've ran on a single format error on https://github.com/MakieOrg/Makie.jl/blob/master/src/camera/camera.jl#L32-L34, which was trivial to fix, and I'm going to open an issue + PR in `JuliaFormatter` to fix this, but this is non-blocking.

## Type of change

- [x] Non functional change - format, code style

## Checklist


CairoMakie/src/CairoMakie.jl
```diff
@@ -2,9 +2,9 @@ module CairoMakie
 
 using Makie, LinearAlgebra
 using Colors, GeometryBasics, FileIO
-import SHA
```
SimonDanisch, line 5: I'm not a big fan of this :D Can we specify that in the formatter.toml?

t-bltg: That is the consequence of https://github.com/MakieOrg/Makie.jl/blob/f124a016d2662ec12d31dac4098b7a92622bd3b1/.JuliaFormatter.toml#L3.

Remove ?

SimonDanisch: Haha, surely! Totally forgot about the original content of that file!

t-bltg: What I would propose is try to use the minimal number of options with the "yas" style, and add them after if needed :)

CairoMakie/src/display.jl
```diff
@@ -133,7 +127,7 @@ end
 
 Converts anything like `"png", :png, "image/png", MIME"image/png"()` to `"image/png"`.
 """
-function to_mime_string(mime::Union{String, Symbol, MIME})
+function to_mime_string(mime::Union{String,Symbol,MIME})
```
ffreyer, line 130: I'd prefer this and also default args like `get_all_plots(scene, plots=AbstractPlot[])` to keep the spaces. Helps with readability and produces nicer line breaks in github diffs imo.

t-bltg: Thanks, I've set `whitespace_in_kwargs = true` and `whitespace_typedefs = true`.

src/interaction/inspector.jl
```diff
@@ -30,28 +35,29 @@ function bbox2string(bbox::Rect2)
...
 color2text(c::Symbol) = string(c)
 color2text(c) = color2text(to_color(c))
-function color2text(c::RGBAf)
+color2text(c::RGBAf) =
     if c.alpha == 1.0
         @sprintf("RGB(%0.2f, %0.2f, %0.2f)", c.r, c.g, c.b)
     else
         @sprintf("RGBA(%0.2f, %0.2f, %0.2f, %0.2f)", c.r, c.g, c.b, c.alpha)
     end
-end
```
ffreyer, line 46: I don't like the free floating if-block. Imo the short form is nice when it's a one-liner but as soon as it's not I think the `function` Syntax becomes more readable. I know there are a lot of these in Makie so maybe I'm on my own with that opinion...

src/interaction/inspector.jl
```diff
@@ -30,28 +35,29 @@ function bbox2string(bbox::Rect2)
...
 
 color2text(name, i::Integer, j::Integer, c) = "$name[$i, $j] = $(color2text(c))"
 function color2text(name, i, j, c)
     idxs = @sprintf("%0.2f, %0.2f", i, j)
     "$name[$idxs] = $(color2text(c))"
```
ffreyer, line 58: How come this didn't get a `return`?

src/interaction/inspector.jl
```diff
@@ -426,14 +416,10 @@ function update_tooltip_alignment!(inspector, proj_pos)
...
 end
 
-
-
 ################################################################################
 ### show_data for primitive plots
 ################################################################################
 
-
-
```
ffreyer, line 436: Do we want larger spacings like this to get removed? I find them quite helpful to make sections easy to notice as you scroll by.

src/interaction/inspector.jl
```diff
@@ -742,20 +738,20 @@ function show_imagelike(inspector, plot, name, edge_based)
...
-    t = clamp(b+1, 2, size(img, 2))
-    z = ((r-i) * img[l, b] + (i-l) * img[r, b]) * (t-j) +
-        ((r-i) * img[l, t] + (i-l) * img[r, t]) * (j-b)
+    l = clamp(floor(Int, i), 1, size(img, 1) - 1)
+    r = clamp(l + 1, 2, size(img, 1))
+    b = clamp(floor(Int, j), 1, size(img, 2) - 1)
+    t = clamp(b + 1, 2, size(img, 2))
+    z =
+        ((r - i) * img[l, b] + (i - l) * img[r, b]) * (t - j) +
+        ((r - i) * img[l, t] + (i - l) * img[r, t]) * (j - b)
```
ffreyer, line 754: I guess this is very subjective and nit-picky but I think in some situations (like here) it's nicer to have e.g. `(t-j)` instead of `(t - j)`. The former sort of establishes groups and gives you less separate things to look at which, in my opinion, helps readability. Maybe this is something to keep unformated?

src/interaction/iodevices.jl
```diff
@@ -4,206 +4,205 @@ represent keyboard buttons.
 """
 module Keyboard
```
ffreyer, line 5: I don't think we should lose the indention for modules with how we use them here.

src/makielayout/blocks/axis.jl
```diff
@@ -281,118 +340,283 @@ function initialize_block!(ax::Axis; palette = nothing)
...
-            p1 = Point2f(left(r) - 0.5sw, y)
-            p2 = Point2f(right(r) + 0.5sw, y)
-            return [p1, p2]
-        else
-            y = top(r)
-            p1 = Point2f(left(r) - 0.5sw, y)
-            p2 = Point2f(right(r) + 0.5sw, y)
-            return [p1, p2]
+    xoppositelinepoints =
+        lift(scene.px_area, ax.spinewidth, ax.xaxisposition; ignore_equal_values = true) do r, sw, xaxpos
```
ffreyer, line 501: These line breaks are a bit weird to me. Shouldn't this become 
```julia
... = lift(
    ...
) do ...
    ...
end
```
if it overflows?

src/interaction/inspector.jl
ffreyer: The `"$name[$idxs] = $(color2text(c))"`

src/basic_recipes/arrows.jl
```diff
@@ -51,75 +51,97 @@ function arrow_tail(N, marker::Automatic, quality)
...
-        fxaa, ssao, transparency, visible, inspectable
+        diffuse,
+        specular,
+        shininess,
+        fxaa,
+        ssao,
+        transparency,
+        visible,
+        inspectable,
     )
```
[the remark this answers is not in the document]
t-bltg: As a workaround, would could split these manually into:
```julia
@extract arrowplot (points, directions, colormap, normalize, align)
@extract arrowplot (arrowtail, color, linecolor, linestyle, linewidth)
...
```
with each line respecting the margin (I don't think there is a performance penalty).

src/interaction/inspector.jl
t-bltg: That's a `JuliaFormatter` bug I think, I'll fix it.

src/interaction/inspector.jl
t-bltg: I've disabled `always_use_return = true` on purpose.
IMO this is pretty minor, and I think we should make the formatter the least intrusive as possible.

src/makielayout/blocks/axis.jl
t-bltg: That's a bit weird yes.

src/interaction/inspector.jl
t-bltg: That's the effect of `remove_extra_newlines = true` and it can be disabled.

t-bltg:
I taken another approach, since there were too many controversial changes as @ffreyer pointed out.

I propose instead to use the `minimal` style, which e.g. doesn't take into account margins.

I'm also annoyed by the forced trailing zeros (`20f0 -> 20.0f0`) (see https://github.com/domluna/JuliaFormatter.jl/pull/672), hopefully this will be resolved quickly.

src/basic_recipes/arrows.jl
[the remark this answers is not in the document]
t-bltg: Fixed by using style = "minimal" (very large margin).

GLMakie/experiments/shaderabstr.jl
```diff
@@ -43,88 +43,93 @@ p = ShaderAbstractions.Program(
...
 
-    vec4 get_$(name)(){
-        return texture($(name)_texture, fragment_$(name)_index);
-    }
-    """)
+vec4 get_$(name)(){
+    return texture($(name)_texture, fragment_$(name)_index);
+}
+""",
+    )
```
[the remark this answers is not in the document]
t-bltg: This is a bit annoying yes (https://github.com/domluna/JuliaFormatter.jl/issues/501).

CairoMakie/src/overrides.jl
```diff
@@ -36,8 +41,16 @@ function draw_poly(scene::Scene, screen::Screen, poly, points::Vector{<:Point2})
...
+function draw_poly(
+    scene::Scene,
+    screen::Screen,
+    poly,
+    points::Vector{<:Point2},
+    color::Union{Symbol, Colorant, Makie.AbstractPattern},
+    model,
+    strokecolor,
+    strokewidth,
+)
```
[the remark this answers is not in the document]
t-bltg: I think this is a matter of taste.

src/interaction/inspector.jl
t-bltg: Not sure, I personally find this quite readable, and consistent with the rest.

CairoMakie/src/overrides.jl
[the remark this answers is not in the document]
ffreyer: Sure, that's true for pretty much all of these. I'm mostly pointing this stuff out so it can be discussed and we can figure out what we actually want

src/interaction/inspector.jl
ffreyer: Huh. I saw a bunch of other places get a return, so I got confused why there wasn't one here. Maybe I was looking at different commits.

t-bltg: Yeah, sorry about the changes, I'm trying to propose something not too disruptive for a start, which could maybe be polished later ...

ffreyer:
I'll say it again here so it's more visible: Most of my complaints are just my opinions on code style. My main reason for posting them is so we can discuss those points and figure out what styling we want. It doesn't need to be what I suggest. But I think this pr is the place to discuss that.

t-bltg:
Code style is always opinionated.
I'm glad we can have those discussions so we can come up with an acceptable solution :)

t-bltg:
Rebased against latest master.

Should this go in (based on https://github.com/MakieOrg/Makie.jl/pull/2493#issuecomment-1440511329) ?
This is exactly the purpose of this MR, not to waste time and endlessly discuss code style in other PRs.

t-bltg:
Superseded by https://github.com/MakieOrg/Makie.jl/pull/5078.

PR #2428, opened 2022-11-17
Use new JSServe

SimonDanisch:

PR #2524, opened 2022-12-26
Remove ElectronDisplay.jl and use Electron directly

SimonDanisch:


SimonDanisch:
Taken from https://github.com/MakieOrg/Makie.jl/pull/2428, so that we have this in the benchmark code for testing against master

PR #2525, opened 2022-12-26
Small fixes

ffreyer:
# Description

A bunch of simpler fixes, see commits

## Type of change

### DIFF
--- a/src/layouting/layouting.jl
+++ b/src/layouting/layouting.jl
@@ -8,7 +8,9 @@ end
 
 function attribute_per_char(string, attribute)
     n_words = 0
-    if attribute isa Vector
+    if attribute isa GeometryBasics.StaticArray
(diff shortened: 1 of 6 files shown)


src/layouting/layouting.jl
```diff
@@ -8,7 +8,7 @@ end
 
 function attribute_per_char(string, attribute)
     n_words = 0
-    if attribute isa AbstractVector
+    if attribute isa Vector
```
ffreyer, line 11: I'm not sure if this is fine... It now treats `Point` and `Vec` as constant per character, but also ranges which might be bad?

ffreyer: I'll just switch to adding an extra branch for StaticArrays instead

SimonDanisch:
Thanks!

Issue #2522, opened 2022-12-25
Fixing retina screen display and allowing to save upscaled GLMakie images


jkrumbiegel:
I think both of these issues are connected, and I'd like to get some help in figuring out the solution.

Basically, currently there is no implementation for retina screens. What happens is that when it's detected that a screen has a retina / hi-dpi factor, the window resolution is reduced by that factor. On Mac at least that means that a 800x600 window is requested as a 400x300 window. Because of retina framebuffers, the backing buffer is actually 800x600 again, but because the effective window size is 400x300, the window is small and text is tiny.

In order to fix this, I think we'd need a way to decouple framebuffer and window size. This means, one should be able to draw any window at an arbitrary framebuffer resolution. For retina windows, one would choose 2 as the correct factor, but for saving out high-resolution images instead of being limited to monitor resolution, one could choose 3, 4 or 5 (not displaying that, just rendering it).

However, I don't understand the pipeline well enough to implement it, in my preliminary tests I only managed to correct the window size on Mac, but things were still too small visually.

PR #2529, opened 2022-12-27
Add px_per_unit to GLMakie

ffreyer:
# Description

This is a prototype for #2522 though I don't have a mac to test anything with. Feel free to mess around with this @jkrumbiegel 

The pr adds `px_per_unit` to the GLMakie screen config, which acts as a multiplier between the window size and the framebuffer size. So with this you can have 2x2 pixels per window resolution pixel by doing

```julia
...
display(GLMakie.Screen(px_per_unit = 2), fig)
```

TODO:
- fix the black outline of LScenes with ssao
- check if any of the screen/framebuffer size calls are expensive and if so avoid them (as far as I can tell GLFW.Get... calls are fairly expensive, but getting the size of a texture is now)
- general cleanup

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/postprocessing.jl
+++ b/GLMakie/src/postprocessing.jl
@@ -157,6 +157,7 @@ function ssao_postprocessor(framebuffer, shader_cache)
     full_render = screen -> begin
         fb = screen.framebuffer
         w, h = size(fb)
+        scale = screen.config.px_per_unit
 
         # Setup rendering
         # SSAO - calculate occlusion
@@ -170,8 +171,7 @@ function ssao_postprocessor(framebuffer, shader_cache)
             # scenes. It should be a leaf scene to avoid repeatedly shading
             # the same region (though this is not guaranteed...)
             isempty(scene.children) || continue
-            a = pixelarea(scene)[]
-            glScissor(minimum(a)..., widths(a)...)
+            glScissor(trunc_scale(scale, scene)...)
             # update uniforms
             data1[:projection] = scene.camera.projection[]
             data1[:bias] = scene.ssao.bias[]
@@ -185,7 +185,8 @@ function ssao_postprocessor(framebuffer, shader_cache)
             # Select the area of one leaf scene
             isempty(scene.children) || continue
             a = pixelarea(scene)[]
-            glScissor(minimum(a)..., widths(a)...)
+            # glScissor(minimum(a)..., widths(a)...)
+            glScissor(trunc_scale(scale, scene)...)
             # update uniforms
             data2[:blur_range] = scene.ssao.blur
             GLAbstraction.render(pass2)
@@ -285,8 +286,7 @@ function to_screen_postprocessor(framebuffer, shader_cache, screen_fb_id = nothi
     pass.postrenderfunction = () -> draw_fullscreen(pass.vertexarray.id)
 
     full_render = screen -> begin
-        fb = screen.framebuffer
-        w, h = size(fb)
+        w, h = size(screen)
 
         # transfer everything to the screen
         default_id = isnothing(screen_fb_id) ? 0 : screen_fb_id[]
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -359,9 +359,15 @@ function Screen(;
     # Screen config is managed by the current active theme, so managed by Makie
     config = Makie.merge_screen_config(ScreenConfig, screen_config)
     screen = screen_from_pool(config.debugging)
+    
+    # to avoid error in resize!
+    if !isdefined(screen, :config)
+        screen.config = config
+    end
     if !isnothing(resolution)
         resize!(screen, resolution...)
     end
+    
     apply_config!(screen, config; visible=visible, start_renderloop=start_renderloop)
     return screen
 end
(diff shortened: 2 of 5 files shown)


jkrumbiegel:
Nice! Thank you :) it currently fails during loading though, same error as in CI

jkrumbiegel:
Ok so this doesn't quite solve the problem, yet, as the size of the elements is still too small, although I can now render it at higher quality. Here's an example with `px_per_unit = 0.5`, the apparent size of everything is still the same, it's just blurrier. 

grafik

In the end, I need the same apparent scaling on a normal dpi monitor with 800x600 pixels, where text is 16px high, just with twice the pixels.

ffreyer:
Should the window just be double the size?l

jkrumbiegel:
Right now it is such that when a retina factor is detected, the window size is reduced by the same factor. For example, 800x600 figure gets a 400x300 size window, backed by a 800x600 pixel buffer. What I really need is a 800x600 window with a 1600x1200 buffer, but with the same "real estate" for the figure as in a non-retina window, just sharper.

GLMakie/src/screen.jl
```diff
@@ -590,24 +597,21 @@ function closeall()
...
-        w, h = resolution ./ retina_scale
-        if oldsize == (w, h)
-            return
-        end
-        GLFW.SetWindowSize(window, round(Int, w), round(Int, h))
+        windowsize(window) == (w, h) && return
+        GLFW.SetWindowSize(window, w, h)
     end
+    return
 end
```
ffreyer, line 607: I thought the reduction in window size was happening here...

Looking at GLFW docs, there seems to be a retina window hint which we don't have in GLFW.jl

> `GLFW_COCOA_RETINA_FRAMEBUFFER` specifies whether to use full resolution framebuffers on Retina displays. Possible values are `GLFW_TRUE` and `GLFW_FALSE`. This is ignored on other platforms.

https://www.glfw.org/docs/latest/window_guide.html

and there is also a compile flag

> `GLFW_USE_RETINA` determines whether windows will use the full resolution of Retina displays.

https://www.glfw.org/docs/3.1/compile.html

Maybe we also need some changes in GLFW?

GLMakie/src/postprocessing.jl
```diff
@@ -285,8 +286,7 @@ function to_screen_postprocessor(framebuffer, shader_cache, screen_fb_id = nothi
     pass.postrenderfunction = () -> draw_fullscreen(pass.vertexarray.id)
 
     full_render = screen -> begin
-        fb = screen.framebuffer
-        w, h = size(fb)
```
ffreyer, line 289: There is also a window hint

> `GLFW_SCALE_TO_MONITOR` specified whether the window content area should be resized based on the [monitor content scale](https://www.glfw.org/docs/3.3/monitor_guide.html#monitor_scale) of any monitor it is placed on. This includes the initial placement when the window is created. Possible values are GLFW_TRUE and GLFW_FALSE.

which we do have available. You could try adding it here
https://github.com/MakieOrg/Makie.jl/blob/2fcee32cad5938b50046289f1b3edba3d3b80325/GLMakie/src/screen.jl#L200-L213

If this changes the size of the final framebuffer you will probably need to adjust this to something like `2 .* size(screen)`. (I'm guessing `GLFW.GetFramebufferSize(screen.glscreen)` will return the correct size, but it would be better to avoid calling that all the time. If we don't need `px_per_unit` we can probably just revert to the old code.)

GLMakie/src/screen.jl
jkrumbiegel: I think retina is already used correctly, you specify a window size of 400x300 and you get a frame buffer size of 800x600. It's just that the size of everything should be doubled in pixels so that it looks normal.

You can use the retina flag to disable that behavior, then you'd get a pixelated window though, I think.

GLMakie/src/postprocessing.jl
jkrumbiegel: I don't have a second monitor to test with, but it sounds reasonable that pixel density should be automatic and depend on the current monitor. Unless you want to override it, or render out hi res pngs.

jkrumbiegel:
I found some code that looks like it solves the same problem: https://web.eecs.umich.edu/~sugih/courses/eecs487/glut-howto/glfw/sample.cpp

This does the "scaling" that I was talking about if I understand correctly, with the `glOrtho(0.0, (float)fbwidth, 0.0, (float)fbheight, -1.f, 1.f);` call. 

```c
/* Called when framebuffer is resized, e.g., when window is resized
 * OR when the same size buffer is moved across Retina and non-Retina displays
 * when running Mac OS X.
 * NOT called automatically when window is first created.
 * Called by GLFW BEFORE reshape().
*/
void
fbreshape(GLFWwindow *wd, int w, int h)
{
  /* save new framebuffer dimensions */
  fbwidth = w;
  fbheight = h;

  /* do an orthographic parallel projection with the view volume
     set to first quadrant, fixed to the initial window dimension */
  glMatrixMode(GL_PROJECTION);
  glLoadIdentity();
  glOrtho(0.0, (float)fbwidth, 0.0, (float)fbheight, -1.f, 1.f);

  /* Tell OpenGL to use the whole window for drawing.
     Note that we don't resize the view volume, so
     the viewport will show the whole view volume 
     shrunk/stretched to fit the current view port. */
  glViewport(0, 0, (GLsizei) fbwidth, (GLsizei) fbheight);

  init_data();
  
  return;
}
```

ffreyer:
Closed in favor of #2544

PR #2533, opened 2022-12-29
use nightly for benchmarks

SimonDanisch:
Just for fun! Although, this won't compare to older julia versions, but lets see what will happen ;)


jkrumbiegel:
Here's this PR and the last with the normal Julia version side by side (just a mobile phone screen shot)

Julia nightly 

image

Julia stable

image

SimonDanisch:
without cache:
image
image
with cache:
image
image

Hm, a bit faster, but not as fast as expected ^^ I guess one might as well just use `julia-actions/cache@v1`

Issue #2205, opened 2022-08-08
`linewidth` for `linesegments` in CairoMakie


hexaeder:
In CairoMakie `0.8.13`, the line ends plottet by `linesegments` look differently depending on whether `linewidth` is given as a vector or not.
```julia
segs = Point2f.([(0, 0), (0,1), (1,0), (1,1)])
linesegments(segs; linewidth = 25)
linesegments(segs; linewidth = [10, 25])
```
results in
<img width="300" alt="grafik" src="https://user-images.githubusercontent.com/35867212/183392458-c77bc127-a55a-4534-9b84-1d1387d9533c.png">

jkrumbiegel:
Not sure why this changed exactly, but in CairoMakie there's been a hack for quite some time which changes the line caps to round if it needs to draw connected lines out of multiple segments (because butted caps give ugly gaps). This shouldn't apply to linesegments though I'd say.

jkrumbiegel:
Ideally one could of course pass linecap style manually, but glmakie doesn't have the capability to draw different ones as far as I know.

hexaeder:
> Not sure why this changed exactly

This behaviour might have been like this forever -- I didn't test it with previous versions. I just noticed while creating a new GraphMakie reference test and was confused... for my usecase it doesn't actually mater.

ffreyer:
The behavior is the same in 0.21 now, after #2536, and you can control which `linecap` is chosen.

Issue #871, opened 2021-03-11
Allow to set line join and cap in CairoMakie


devmotion:
It would be nice to be able to use a different line join style (`Cairo.CAIRO_LINE_JOIN_BEVEL` instead of the default `Cairo.CAIRO_LINE_JOIN_MITER`) with CairoMakie, and possibly also a different line cap.

This can be useful to avoid visually irritating and seemingly incorrect values: when I plotted a stochastic trajectory that is constrained to the domain [0, 1] I ended up with a plot that seemed to contain negative values. The same did not happen in a scatter plot, and after a while I realized that it was caused by the default line join. For my application I reimplemented https://github.com/JuliaPlots/CairoMakie.jl/blob/c54dd87ec687deed1e8bca2a0faf702cbb46be38/src/infrastructure.jl#L278-L286 and added `Cairo.set_line_join(screen.context, Cairo.CAIRO_LINE_JOIN_BEVEL)` before https://github.com/JuliaPlots/CairoMakie.jl/blob/c54dd87ec687deed1e8bca2a0faf702cbb46be38/src/infrastructure.jl#L283, but it would be nice if this could be set by a user more generally. Probably this should be done differently since I assume sometimes one might even want to use different settings for different line plots in the same figure.

jkrumbiegel:
yeah this would be good, GLMakie doesn't have other line joins that's why it's not in AbstractPlotting and also not in CairoMakie. But I think we could add it to AbstractPlotting with the caveat that it will not have an effect on anything but CairoMakie, I guess. cc @SimonDanisch

PR #2536, opened 2022-12-30
add linecaps (GLMakie, CairoMakie)

### LINKED ISSUE #1119: Linesegments - Missing corners in segment drawing
I noticed a weird behavior when plotting segments with corners with ```linesegments```. The issue is that the corners seem to be rounded only if we explicitly provide a color for the segments and are otherwise missing. To exemplify this, if we want to draw the triangle {(0,0), (0,1), (1,1)}, then
```julia
using Makie
segments = [0.0 0.0; 
            0.0 1.0; 
            0.0 1.0; 
            1.0 1.0; 
            1.0 1.0; 
            0.0 0.0]
color = [:red, :red, :red]
fig = Figure()
ax1 = Axis(fig[1,1]) 
ax2 = Axis(fig[1,2])
linesegments!(ax1, segments, color=:red, linewidth=5) # Missing corners
linesegments!(ax2, segments, color=color, linewidth=5) # Round corners
```
corners

The last two differ just from the "filling" in the corners, which is almost imperceptible at default linewidth. Is this an intended feature or just a bug with colors?

### LINKED ISSUE #1541: allow for different line widths in lines
At the moment is possible to do several lines at once by [NaN separation](https://makie.juliaplots.org/stable/examples/plotting_functions/lines/index.html). Different colours also works, however passing a set of values for different line widths is not supported, as in linesegments. It would be nice to have this feature for lines, which in all my tests is the fastest one to plot thousands of lines at once.

### LINKED ISSUE #2472: Account for arrow head in arrow length
This issue is follows the discussion in this discourse [thread](https://discourse.julialang.org/t/makie-3d-arrows-scale-in-a-strange-fashion/89371/5).

### Problem
Currently, 3D arrows (2D was not tested) are plotted without accounting for their head. Which means that the total length of an arrow corresponds to the length of the tail + the length of the head. In a more practical fashion, that means that 
```
arrows!(ax,Point3f[(0,0,0)],Point3f[(1,0,0)],arrowsize = 0.5)
```
will result in an overall shorter arrow than
```
arrows!(ax,Point3f[(0,0,0)],Point3f[(1,0,0)],arrowsize = 1.0)
```
It could be nice to have an option (so as to not be a breaking change) to have the total length of the arrow account for the head length. 

### Example
```julia
using GLMakie

fig = Figure()
ax = LScene(fig[1,1])
# Has to be added otherwise first arrows with single arrow fails
scatter!(ax,Point3f[(2,0,0),(0,2,0),(0,0,2)])
# Effect of arrowsize parameter
arrows!(ax,Point3f[(0,0,0)],Point3f[(1,0,0)],color = RGBAf(255,0,0,1),arrowsize = 0.5)
arrows!(ax,Point3f[(0,0,0)],Point3f[(1,0,0)],color = RGBAf(0,255,0,0.5),arrowsize = 1)
# Just for fun effect of align parameter (surprising ?)
poss = (:head, :lineend, :tailend, :headstart, :center,:origin)
for (z,pos) in enumerate(poss)
arrows!(ax,Point3f[(0,0,z)],Point3f[(1,0,0)],color = RGBAf(0,0,255,0.5),align=pos,arrowsize = 1)
end
display(GLMakie.Screen(),fig)
```
image

While I was at it, a bit of fiddling with position alignment showed a surprising behavior.

ffreyer:
# Description

This pr aims to add linecaps to `lines!` and `linesegments!` in GLMakie. 

Current prototype (second commit) should render caps for lines (triangular, rectangular or circular depending on the values in the shader.)

(Debug) Example for round line caps:

```julia
GLMakie.closeall(); display(GLMakie.Screen(), lines([1, 2, NaN, 4, 5], linewidth = 50))
```

Screenshot from 2022-12-30 10-51-45

## TODO

- [x] (Fix this) Line caps and lines use different antialiasing algorithms which seem to give slightly different results: (this is true for triangle and circle too)
image
- [x] Implementation for linesegments
- [x] add uniforms to adjust linecaps
- [x] add attributes for linecap
- [x] add per-vertex linewidth to `lines!` (closes #1541)
- [x] fix occasional bad triangle generation (with different linewidths the miter adjusted normals terminate triangles early sometimes, which results in a small kink in the line)
- [x] fix linestyle
- [x] add empty linecap (what Cairo calls butt, no extension past the start/end point of a line)
- [x] add linecap length and allow it to be negative to cut away from a line (This should help with keeping arrow markers and lines separate with arrows like #2472 suggests in 2D)
- ~~reuse line vertices for linecaps~~ this isn't easily possible since different line caps use different uvs...
- [x] cleanup
- [x] update attribute passthrough in recipes
- [x] add example to lines/linesegments docs
- [x] update recipe docstrings
- [x] check blocks for overlapping lines
- [x] prototype update for 2d arrows

closes #1119, #2205

Also related: #871

## Type of change

## Checklist


GLMakie/assets/shader/lines.frag
```diff
@@ -10,6 +10,7 @@ in vec4 f_color;
 in vec2 f_uv;
 in float f_thickness;
 flat in uvec2 f_id;
+flat in int f_type; // TODO bad for performance
```
ffreyer, line 13: Given the comment in the scatter shader

https://github.com/MakieOrg/Makie.jl/blob/4dab9955103282b6af93e5bc516a9797a5b8273f/GLMakie/assets/shader/distance_shape.frag#L26

I assume we should find a way around this?

GLMakie/src/drawing_primitives.jl
```diff
@@ -269,7 +270,7 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::Lines))
             data[:pattern] = ls
         else
             linewidth = gl_attributes[:thickness]
-            data[:pattern] = ls .* (to_value(linewidth) * 0.25)
+            data[:pattern] = ls * _mean(to_value(linewidth))
```
ffreyer, line 273: I took this as somewhat of a shortcut for now, because `linestyle` needs some cleanup anyway.

Example with variable linewidth:
```julia
GLMakie.closeall()
widths = [57.37044620677474, 27.847264587380923, 56.663525479541775, 47.302310563997985, 56.36164168328704, 43.014590186214285, 58.832167079498525, 27.208798997411307, 17.12575358453462, 45.95171757911098, 26.771344098755126, 35.85649823581235, 53.08232474276624, 27.217846025535966, 43.98443255191129, 37.389819316502695, 18.07005596531466, 59.569472267387056, 50.919270815270814, 36.74914882840677, 53.08834081120422, 45.44455789568474, 40.147375157549135, 22.166337026901253, 50.53203686576642, 15.238811525483367, 12.780634172815944, 12.980713770063876, 28.175981476856755, 37.542112019663485]
ys = [0.3672275443344566, 0.13769475910008522, 0.22445059716611304, 0.5883310148108587, 0.7893622240334566, 0.3129809823712394, 0.953606981656488, 0.2770419071274035, 0.3048558242752848, 0.6782285082882588, 0.5447836889282002, 0.19905515156197973, 0.19770277191975638, 0.537590947032379, 0.28027667604204765, 0.8287252657638743, 0.9336781225747905, 0.1371941303349402, 0.5973028266375866, 0.3561025405662749, 0.43500848119792546, 0.36410392908891054, 0.9117308440489152, 0.8008057243055526, 0.2349037873519133, 0.21294938231466431, 0.8334978136186971, 0.5097696792342837, 0.21488349759400982, 0.3434422610208452]
fig = Figure()
ax = Axis(fig[1, 1])
# p = linesegments!(ax, ys, linewidth = widths./10, linecap = 4, linecap_length = widths./10, color = :red)
# p = linesegments!(ax, ys, linewidth = widths./10, linecap = 4, linecap_length = widths./10, linestyle = :dot)
p = lines!(ax, ys, linewidth = widths./10, linecap = 0, linecap_length = widths, color = :red)
p = lines!(ax, ys, linewidth = widths./10, linecap = 0, linecap_length = widths, linestyle = :dash)
screen = display(GLMakie.Screen(), fig)
```

produces
Screenshot from 2022-12-31 17-07-54
Screenshot from 2022-12-31 17-08-04

and pixel perfect results with constant linewidth (tested explicitly for linewidth = 50 on a straight line with lines and linesegments looks the same).

There is however some artifacting with `:dash` that got more frequent but isn't new. From master:

Screenshot from 2022-12-31 17-19-36

ffreyer: Oh right, I also changed the scaling with linewidth in `lines!` (not here but in the shader) which makes `:dash` shorter notably at large linestyles. It's now 3lw on, 3lw off like the pattern requests. master has weird scaling, maybe quadratic? For linewidth 50, it seems to be about 40.1lw on, 40.1lw off. For linewidth 10 about 10.5lw on 10.5lw off
```julia
lines([0, 0], linewidth = 50, linestyle = :dash)
```
Screenshot from 2022-12-31 17-26-52

ffreyer:
Refimg test I added for linecaps:

Screenshot from 2022-12-31 18-55-16

This shows (nothing, :square, :round), default for lines, then (nothing, :square, :round) for linesegments. New default is `:square` like in PyPlot.jl

GLMakie/assets/shader/lines.frag
ffreyer: Did some benchmark with

```julia
let 
    GLMakie.closeall()
    ps = [2 * rand(Point2f) .- 1 for _ in 1:2000]

    function myrenderloop(screen)
        yield()
        GLMakie.GLFW.SwapInterval(0)
        GLMakie.pollevents(screen)
        GLMakie.render_frame(screen)
        GLMakie.GLFW.SwapBuffers(GLMakie.to_native(screen))
        yield()

        @time for _ in 1:10_000
            GLMakie.pollevents(screen)
            GLMakie.render_frame(screen)
            GLMakie.GLFW.SwapBuffers(GLMakie.to_native(screen))
        end

        while GLMakie.isopen(screen) && !screen.stop_renderloop
            GLMakie.pollevents(screen)
            sleep(0.1)
            yield()
        end

        return
    end

    scene = Scene()
    linesegments!(scene, ps, linewidth = 10)
    screen = GLMakie.Screen(renderloop = myrenderloop)
    display(screen, scene)
end
```

I get the same ~5% CI got for this branch vs master. 

I tried hacking together a two pass version which does the old lines in the first pass and caps in the second pass. This way we can swap out the `flat in int f_type` for `uniform int linecap` and `uniform int render_pass`. This ends up being about 10% slower than master though, doing both passes. Doing just the line pass is 3-4% slower.

SimonDanisch: I think that's ok for linecaps! Can we optionally disable it and get the 5% back for high perf use cases? We could also think about having a different recipe for that though 🤷 I think the default should make good lines, and performance should be the special case!

ffreyer: Reducing the number of options in the fragment shader ([5e206a9](https://github.com/MakieOrg/Makie.jl/pull/2536/commits/5e206a98e258097a1dbddb98374ed17a98175a3d)) helped a bit ... I think. I'm never sure with gpu benchmarking. When I tested it I got a bunch of benchmarks that were ~2% slower than master and then a bunch up to 5% slower :shrug:

ffreyer:
I still need to go through a bunch of recipes and adjust their attributes.

GLMakie/src/glshaders/lines.jl
```diff
@@ -133,7 +135,11 @@ function draw_linesegments(screen, positions::VectorTypes{T}, data::Dict) where
         end
         tex = GLAbstraction.Texture(ticks(pattern, 100), x_repeat = :repeat)
         data[:pattern] = tex
-        data[:pattern_length] = Float32(last(pattern))
+        @gen_defaults! data begin
+            pattern_length = Float32(last(pattern))
+            lastlen   = const_lift(sumlengths, positions) => GLBuffer
+            maxlength = const_lift(last, lastlen)
+        end
```
ffreyer, line 142: This was probably unnecessary...

ffreyer: It was

src/basic_recipes/poly.jl
```diff
@@ -59,14 +60,14 @@ function poly_convert(polygons::AbstractVector{<: AbstractVector{<: VecTypes}})
     end
 end
 
-to_line_segments(polygon) = convert_arguments(LineSegments, polygon)[1]
+to_line(polygon) = convert_arguments(Lines, polygon)[1]
```
ffreyer, line 63: I renamed and switched to `Lines` because this gets pushed into `lines` further down the pipeline. Without the change here we get duplicate points which results in some ugly corners on this branch. (Because joins still get drawn for zero length lines I guess)

Something I else I noticed here is that `Rect` doesn't make it to `plot!(plot::Poly{<: Tuple{<: Union{Polygon, AbstractVector{<: PolyElements}}}})`. It uses `plot!(plot::Poly{<: Tuple{Union{GeometryBasics.Mesh, GeometryPrimitive}}})` instead. That results in in every `Rect` having 4 bad corners instead of just one... (That should also be the case on master) I tried adding `Rect` explicitly to the nicer `plot!` method but it still used the other. I don't know why

GLMakie/assets/shader/lines.frag
ffreyer: I ran them again without background processes and I'm getting ~2% now ([812ff3f](https://github.com/MakieOrg/Makie.jl/pull/2536/commits/812ff3f1691da3c255c0c56cc1c173f17c1b333d))

GLMakie/src/drawing_primitives.jl
```diff
@@ -260,16 +260,29 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::Union{Scatte
...
+    # default to no cap
+    return 0
+end
+
 function draw_atomic(screen::Screen, scene::Scene, @nospecialize(x::Lines))
     return cached_robj!(screen, scene, x) do gl_attributes
         linestyle = pop!(gl_attributes, :linestyle)
+        linecap = pop!(gl_attributes, :linecap)
         data = Dict{Symbol, Any}(gl_attributes)
+        data[:linecap] = map(gl_linecap, linecap)
```
ffreyer, line 279: I'm doing the conversion here wrong. The linecap input Observable is strongly typed at this point (I assume from `lift_convert`) so when you try changing the value from `nothing` to `:round` it errors. How do I do this correctly?

SimonDanisch: You should turn that into a `convert_attributes` overload, so that lift_convert converts it to a stable type...So something like:
```julia
function convert_attribute(cap::Union{Nothing, Symbol}, ::key"linecap")
    cap == :square   && return 5
    cap == :round    && return 4
# TODO maybe error on wrong symbols?
    # default to no cap (nothing included)
    return 0
end
```

ffreyer: Would this also trigger for CairoMakie?

ffreyer: Yes. I did a `nothing -> :butt` conversion instead so CairoMakie can have more readable Symbols.

SimonDanisch: CairoMakie doesn't need to apply the convert ;)

ffreyer: Adding this results in getting numbers in CairoMakie though, which I'd rather not...

ffreyer:
This should be good to go

ffreyer:
https://discourse.julialang.org/t/y-axis-in-cairomakie-jl-not-vertical/93165 made me realize that a pixel offset attribute for lines is probably less niche than I first thought, so I'm reworking `linecap_length` into a more general `length_offset`.

@jkrumbiegel This would allow you to nudge the start and endpoints of a line around by some number of pixel. So for an axis you could make ticks start 1 frame thickness later, and shorten/lengthen some of the frame lines by one thickness to avoid overlap. Or for 2d arrows you could use this to stop the line before the marker.

ffreyer:
I added `length_offset` for lines and linesegments across the board with this refimg:
Screenshot from 2023-01-18 19-17-05
WGLMakie support is there but the refimg test doesn't match because WGLMakie can't handle linecaps and doesn't have line joins

ffreyer:
> GLMakie display: +10.76%, 19.38ms slower :x:  (4.54d, 0.00p, 4.17std)

This seems to be mostly display time, not render time. Running render-N-frames check returns (1.7% min, 8.5% max, 1.8% mean, 1.3% median) more time spend on rendering here. (I'm now running 30 timings at 1k frames and filter spikes > 20% mean.) I guess this spike in display time isn't really that surprising with me adding another buffer to upload.

<details><summary>Code</summary>
```julia
let
    for _ in 1:30
        GLMakie.closeall()
        ps = [2 * rand(Point2f) .- 1 for _ in 1:2000]

        function myrenderloop(screen)
            yield()
            GLMakie.GLFW.SwapInterval(0)
            GLMakie.pollevents(screen)
            GLMakie.render_frame(screen)
            GLMakie.GLFW.SwapBuffers(GLMakie.to_native(screen))
            yield()

            @time for _ in 1:1_000
                GLMakie.pollevents(screen)
                GLMakie.render_frame(screen)
                GLMakie.GLFW.SwapBuffers(GLMakie.to_native(screen))
            end

            while GLMakie.isopen(screen) && !screen.stop_renderloop
                GLMakie.pollevents(screen)
                sleep(0.1)
                yield()
            end

            return
        end

        scene = Scene()
        linesegments!(scene, ps, linewidth = 10, linecap = nothing)
        screen = GLMakie.Screen(renderloop = myrenderloop)
        display(screen, scene)
        GC.gc()
        sleep(1)
    end
end

using Statistics

function remove_spikes(ts, threshold = 0.2)
    m = mean(ts)
    return filter(t -> abs(t - m) < threshold * m, ts)
end
```
</details>

ffreyer:
In Cairo linecaps are applied to dashes as well. It's set up to take space away from the "off" part of dashes. I.e. (with pattern being the vector of float we generate from a linestyle)
```
  cap              cap      cap
 |---|------------|---|    |---|------
     ^            ^            ^
 pattern[0]   pattern[1]   pattern[2]
```
I've adjusted the linestyle lengths so that caps take space from the "on" part of a dash/dot instead of the "off" part. So it's not:
```
      cap      cap              cap
     |---|----|---|            |---|--
     ^            ^            ^
 pattern[0]   pattern[1]   pattern[2]
```

ffreyer:
This will interfere heavily with #2666 and I think it's better to redo this than #2666 so I'm making this a draft for now

SimonDanisch:
What do we do with this pr after https://github.com/MakieOrg/Makie.jl/pull/2666 got merged?

ffreyer:
Probably kill it and restart it. Most of this would need to be redone either way.

I've also been thinking it would be good to match Cairo linecaps in GLMakie. They put them on each dash/dot, this pr only did lin start/end.

SimonDanisch:
Ok, makes sense! I'll close it then ;)

PR #2537, opened 2022-12-30
Fix CairoMakie Gtk.jl example

tknopp:
# Description

The CairoMakie Gtk.jl example did not work.

However, this change alone was not sufficient. In https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/cairo-extension.jl#L74 I had to add
```
    typ == Cairo.CAIRO_SURFACE_TYPE_QUARTZ && return SVG # not sure which type to chose here.
```
I did not include that change in this PR, since I don't know how to properly handle other platforms than Mac.

## Type of change

### DIFF
--- a/CairoMakie/README.md
+++ b/CairoMakie/README.md
@@ -38,7 +38,8 @@ function drawonto(canvas, figure)
     @guarded draw(canvas) do _
         scene = figure.scene
         resize!(scene, Gtk.width(canvas), Gtk.height(canvas))
-        screen = CairoMakie.Screen(scene, Gtk.cairo_surface(canvas), getgc(canvas), nothing)
+        config = CairoMakie.ScreenConfig(1.0, 1.0, :good, true, true)
+        screen = CairoMakie.Screen(scene, config, Gtk.cairo_surface(canvas))
         CairoMakie.cairo_draw(screen, scene)
     end
 end


SimonDanisch:
I think `IMAGE` would be the right one ;)

tknopp:
Not sure because of 
https://github.com/JuliaGraphics/Cairo.jl/blob/2b7185c705f42205194c33b076e4e14206212ac0/src/constants.jl#L77
vs
https://github.com/JuliaGraphics/Cairo.jl/blob/2b7185c705f42205194c33b076e4e14206212ac0/src/constants.jl#L83
And `Cairo.CAIRO_SURFACE_TYPE_QUARTZ` seems to be the non-image version?

tknopp:
How should we fix it?

```
function get_render_type(surface::Cairo.CairoSurface)
    typ = ccall((:cairo_surface_get_type, Cairo.libcairo), Cint, (Ptr{Nothing},), surface.ptr)
    typ == Cairo.CAIRO_SURFACE_TYPE_PDF && return PDF
    typ == Cairo.CAIRO_SURFACE_TYPE_PS && return EPS
    typ == Cairo.CAIRO_SURFACE_TYPE_SVG && return SVG
    typ == Cairo.CAIRO_SURFACE_TYPE_IMAGE && return IMAGE
    return IMAGE
end
```

would be the most simple.

PR #2539, opened 2022-12-31
Remove unnecessary type declaration in ispressed

### LINKED ISSUE #2538: Missing method for ispressed and Exclusively
Hi,
I am working on some plot interactions with hotkeys and was following the documentation on this matter.

However, when combining `ispressed()` and `Exclusively`
```julia
rawHotkeys = Dict(
    "stepBack" => Exclusively(Keyboard.left),
    "stepForw" => Exclusively(Keyboard.right),
    "stepBackFull" => Exclusively((Keyboard.left_shift & Keyboard.left)),
    "stepForwFull" => Exclusively((Keyboard.left_shift & Keyboard.right)),
)

on(events(fig).keyboardbutton) do event
        ispressed(fig, rawHotkeys["stepBack"]) && step_back()
        ispressed(fig, rawHotkeys["stepForw"]) && step_forw()
        ispressed(fig, rawHotkeys["stepBackFull"]) && step_back()
        ispressed(fig, rawHotkeys["stepForwFull"]) && step_forw()
end
```

I got an MethodError:
```julia
MethodError: no method matching ispressed(::Figure, ::Exclusively)
```

This is for:
```
[ee78f7c6] Makie v0.19.0
[e9467ef8] GLMakie v0.8.0
```

Looking at source code, in the [crucial line](https://github.com/MakieOrg/Makie.jl/blob/9c9cc6acd56fa01e640665721e7bfec643db4162/src/interaction/events.jl#L266) its the only method variant that has type declaration on the first argument.
Extending `ispressed` with exactly the same code without type declaration solves the issue and interaction works as intended.

Is this an omission or is it left there deliberately?

mkoculak:
# Description

Fixes #2538 

Removes type declaration from the last `ispressed` method that has `Scene` in it. This solves the `MethodError` when `Figure` is passed as the first argument and `Exclusively` is used in the second (just as it is shown in the Documentation).

 I took a look at tests and they only pass `Event` object as the first argument.
Should this pull request also add a test on the `Figure` object, which is how these methods are used in the Documentation?

## Type of change

## Checklist

### DIFF
--- a/src/interaction/events.jl
+++ b/src/interaction/events.jl
@@ -263,7 +263,7 @@ ispressed(scene, key::Keyboard.Button) = ispressed(events(scene), key)
 ispressed(scene, op::And) = ispressed(scene, op.left) && ispressed(scene, op.right)
 ispressed(scene, op::Or)  = ispressed(scene, op.left) || ispressed(scene, op.right)
 ispressed(scene, op::Not) = !ispressed(scene, op.x)
-ispressed(scene::Scene, op::Exclusively) = ispressed(events(scene), op)
+ispressed(scene, op::Exclusively) = ispressed(events(scene), op)
 ispressed(e::Events, op::Exclusively) = op.x == union(e.keyboardstate, e.mousebuttonstate)
 
 # collections
--- a/src/interaction/events.jl
+++ b/src/interaction/events.jl
@@ -224,10 +224,10 @@ create_sets(s::Set) = [Set{Union{Keyboard.Button, Mouse.Button}}(s)]
 # ispressed and logic evaluation
 
 """
-    ispressed(scene, result::Bool)
-    ispressed(scene, button::Union{Mouse.Button, Keyboard.Button)
-    ispressed(scene, collection::Union{Set, Vector, Tuple})
-    ispressed(scene, op::BooleanOperator)
+    ispressed(parent, result::Bool)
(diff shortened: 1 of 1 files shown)


ffreyer (approved):

jkrumbiegel:
Maybe the arguments of all those should be renamed if it's not just a scene. Otherwise this is confusing

ffreyer:
`events` works with Scene, SceneLike (plots) and anything `get_scene` is implemented for, i.e. Figure, Axis, Axis3, Lscene, FigureAxisPlot, AxisPlot, maybe more. I don't know what a good name for all of those would be. If SceneLike didn't exist I would say that is a good name. DataInspector has `fig_or_block` and `figure_axis_or_scene` atm, but those aren't nice names either imo. Maybe something like `scene_container` would work?

jkrumbiegel:
Or just `x`, but `scene` implies `Scene`

mkoculak:
Yeah, I agree changing it to something else might reduce the confusion.
I glanced through other files in interaction folder and in some parts a variable name `parent` is used which might be fitting also here? `x` feels so generic that might also confuse users.

mkoculak:
Hey, I have made the said change from `scene` to `parent` and added explanation based on the @ffreyer earlier comment.
Please let me know if anything else should be changed.

PR #2546, opened 2023-01-03
fix picking

SimonDanisch:
overlooked this, which means that anything displayed via show wouldn't work with picking


SimonDanisch:
rate limits + download issues made the CI fail -.- Merging, since it's a very small change and the 1.0 CI went through!

PR #2547, opened 2023-01-03
Add example for arc plotting function

bluzukk:
# Description

I noticed that there are no examples for the arc() plotting function on the website so I created some.
I tested the changes to the website locally and it works as intended.

## Type of change

- [x] Improvement of the Documentation

## Checklist

### These are the examples I added:

First two very basic ones:
```julia
using CairoMakie
arc(Point2f(0), 1, -π, π)
```
example_3177995486812465662

```julia
using CairoMakie
f = Figure() 
Axis(f[1, 1])

for i in 1:10
    arc!(Point2f(0, i), i, -π, π)
end

f
```

example_12741175784456680574

And as a last example - creating a Farey sequence diagram to order 4:
```julia
using CairoMakie

f = Figure()
Axis(f[1, 1])

for i in 1:4
    radius = 1/(i*2)
    left = 1/(i*2)
    right = (i*2-1)/(i*2)
    arc!(Point2f(left, 0), radius, 0, π)
    arc!(Point2f(right, 0), radius, 0, π)
end
for i in 3:4
    radius = 1/(i*(i-1)*2)
    left = (1/i) + 1/(i*(i-1)*2)
    right = ((i-1)/i) - 1/(i*(i-1)*2)
    arc!(Point2f(left, 0), radius, 0, π)
    arc!(Point2f(right, 0), radius, 0, π)
end

f
```
example_446186802487841561

### DIFF
--- a/docs/examples/plotting_functions/arc.md
+++ b/docs/examples/plotting_functions/arc.md
@@ -1,0 +1,56 @@
+# arc
+
+{{doc arc}}
+
+## Examples
+
+\begin{examplefigure}{svg = true}
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+
+arc(Point2f(0), 1, -π, π)
+```
+\end{examplefigure}
+
+\begin{examplefigure}{svg = true}
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+f = Figure() 
+Axis(f[1, 1])
+
+for i in 1:10
+    arc!(Point2f(0, i), i, -π, π)
+end
+
+f
+```
+\end{examplefigure}
+
+\begin{examplefigure}{svg = true}
+```julia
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
I hadn't even really registered that this recipe exists, one question is if it's useful enough that we should keep it?

SimonDanisch:
I think it's kind of useful ;) Especially since we dont have anything like `GeometryBasics.Ellipse`

jkrumbiegel:
but this doesn't replace an ellipse, it can only do circles :) An ellipse recipe I would find useful, because people often want to plot those around data points

SimonDanisch:
@bluzukk thank you :) this is a great addition and I don't think we need to open the can of worms in this pr weather we need the function or not ;)

PR #2549, opened 2023-01-03
Assume IMAGE as the default render type in CairoMakie

tknopp:
As outlined in https://github.com/MakieOrg/Makie.jl/pull/2537 currently CairoMakie.jl cannot be used in combination with Gtk(4).jl. This is a simple fix that should hold for all the different Cairo backends that are possible in combination with Gtk.jl


SimonDanisch:
Ok, thank you, lets try it like this for now, I think it shouldn't introduce any problems for other use cases ;)

tknopp:
@SimonDanisch: Thank you very much Simon! On Mac/Linux I have tested this patch and it works. Would it be possible that you trigger a release? We are in the process of migrating from Winston.jl to CairoMakie.jl and it would be helpful for us to get a release so that we can finally switch.

tknopp:
No hurry, just wanted to trigger it. We want to switch in 2 or 3 weeks. I am just preparing that the switch works without unreleased branches.

jkrumbiegel:
Would be cool to see an example application, I imagine others would be interested in interplay with GTK as well

PR #2550, opened 2023-01-04
Prototype/Proof of concept fix for Float32 issues

### LINKED ISSUE #1373: Rounding error for large x values
```julia
y = cumsum(randn(1000))
x = 1:1000

f = Figure()
ax1 = Axis(f[1,1]) 
lines!(x, y)
ax1 = Axis(f[2,1]) 
lines!(x .+ Int(1e8), y)
f
```
bug

Makie v0.15.2
CairoMakie v0.6.5

ffreyer:
# Description

Fixes #1373 and related in `Axis`. Tangentially also fixes `range step cannot be zero` issues since I changed that to use Float64.

The idea of this is to introduce an `AutoScale` object that scales values to Float32 friendly range. The scaling is based on axis limits and applied to data before creating a plot object. 

Example:
```julia
ys = 1e9 .+ collect(1:10) # eps(1f9) = 64
xs = collect(1:10)
f, a, p = scatter(xs, ys)
autolimits!(a) # needed in REPL
```

Screenshot from 2023-01-04 09-18-00

Summary of what's happening with this pr:
1. empty Axis is created with default values (including a `1 * x + 0` scaling in AutoScale)
2. The plot is inserted with its input data reacting to `AutoScale`. Since there is no scaling yet it just duplicated data so far.
3. A limit update gets triggered which fetches the data limits of the plot. The scaling is inverted here to get the original input data
4. The update of final limit calls `update_limit!(autoscale, limits)` which notices that the limits are not in Float32 firendly range. The scaling gets updated
5. The scaling update triggers a recomputation of the scaled input data and camera matrices
6. Via the `autolimits!(a)` call the process starts over from (3) but the scaling is already fine, so no extra updates there.

Some Notes:
- I probably changed way more to Float64 than necessary. I just replaced everything in a few places.
- Maybe it makes sense to have this as a transform_func instead?
- This will probably break if plots are rotated. And not work in 3D for similar reasons. Though I think the Float32 Problems are mainly an issue in 2D.
- There are lots of signatures not caught in `map_autoscale`
- updates of `AutoScale` should retrigger limit updates if and only if limits are picked automatically.
- Adjusting `plot.input_args` is probably a bad idea since that `plot.converted` is sometimes used. E.g. in DataInspector...

## Type of change

## Checklist

### DIFF
--- a/src/layouting/data_limits.jl
+++ b/src/layouting/data_limits.jl
@@ -120,7 +120,7 @@ function point_iterator(x::Volume)
     extremata = map(extrema∘to_value, axes)
     minpoint = Point3f(first.(extremata)...)
     widths = last.(extremata) .- first.(extremata)
-    rect = Rect3f(minpoint, Vec3f(widths))
+    rect = Rect3(minpoint, Vec3(widths))
     return unique(decompose(Point, rect))
 end
 
@@ -174,7 +174,7 @@ end
 
 function update_boundingbox!(bb_ref, point)
     if all(isfinite, point)
-        vec = to_ndim(Vec3f, point, 0.0)
+        vec = to_ndim(Vec3, point, 0.0)
         bb_ref[] = update(bb_ref[], vec)
     end
 end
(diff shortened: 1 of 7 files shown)


ffreyer:
With #2573 this isn't necessary for CairoMakie and it would be better to move this rescaling to backends. A condition for rescaling would be something like `widths (cam_bbox) / center(cam_bbox) < 1000eps (center (cam_bbox))`. I think we could also get a measure from eyeposition and projectionview by checking if `abs (projectionview * (eyeposition +- 10^-5 * eyeposition)) > 1`.

Either way I think we can get around what this pr does.

PR #2559, opened 2023-01-05
allow skipping precompilation during development via ENV var

jkrumbiegel:


CairoMakie/src/CairoMakie.jl
```diff
@@ -1,3 +1,7 @@
+if get(ENV, "MAKIE_PRECOMPILE", "true") == "false"
+    __precompile__(false)
+end
```
t-bltg, line 3: Doesn't have this to be set inside the module definition ?

Xref: https://github.com/timholy/SnoopCompile.jl/issues/324.

jkrumbiegel: It's documented that it disables precompilation for the whole file. I think I was told on Slack to do it like this but not completely sure

t-bltg: I see.
Nitpicking, but these can be condensed to `__precompile__(get(ENV, "MAKIE_PRECOMPILE", "true") == "true")`.

jkrumbiegel: Thought about that but found it clearer to show that `__precompile__` is not called at all if the variable isn't set.

t-bltg: Negating the logic would also allow to use any string (e.g. an Int which is shorter to type) to disable precompilation.

SimonDanisch: maybe use `&&` then?

timholy:
Note there is also now https://github.com/timholy/SnoopCompile.jl/pull/334 (not yet released though, want to try a PkgEval run first).

SimonDanisch:
Decided here https://github.com/MakieOrg/Makie.jl/issues/2585#issuecomment-1381585230 to not continue this PR.

Issue #2298, opened 2022-09-23
range step cannot be zero


jariji:
Plotting small numbers can cause an error. I get this error fairly often. It complicates the construction of complex multi-axis plots when I have to deal with errors popping up in the middle so it would be nice if this could be avoided.

A reproducer was found [on Discourse](https://discourse.julialang.org/t/aog-range-step-cannot-be-zero/87624/12):

```jl
julia> using GLMakie

julia> b = 2.0^(-150)
7.006492321624085e-46

julia> Float32(b) # notice that `b` is the smallest `Float64` such that `Float32(b) != 0.0f0`
1.0f-45

julia> Float32(prevfloat(b)) # here you can see that efectively taking just the previous number bellow `b`, then `Float32(b) == 0.0`
0.0f0

julia> lines([b, b])
Error showing value of type Makie.FigureAxisPlot:
ERROR: ArgumentError: range step cannot be zero
Stacktrace:
  [1] (::Colon)(start::Float32, step::Float32, stop::Float32)
    @ Base ./twiceprecision.jl:412
  [2] get_minor_tickvalues(i::IntervalsBetween, scale::Function, tickvalues::Vector{Float32}, vmin::Float32, vmax::Float32)
    @ Makie ~/.julia/packages/Makie/Ppzqh/src/makielayout/lineaxis.jl:670
  [3] (::Makie.var"#1179#1192"{Attributes, Observable{Vector{Float32}}, Observable{Tuple{Float32, Float32}}})(tickvalues::Vector{Float32}, minorticks::IntervalsBetween)
    @ Makie ~/.julia/packages/Makie/Ppzqh/src/makielayout/lineaxis.jl:394
  [4] (::Observables.var"#callback#13"{Makie.var"#1179#1192"{Attributes, Observable{Vector{Float32}}, Observable{Tuple{Float32, Float32}}}, Tuple{Observable{Vector{Float32}}, Observable{Any}}})(x::Any)
    @ Observables ~/.julia/packages/Observables/ynr7h/src/Observables.jl:339
  [5] #invokelatest#2
    @ ./essentials.jl:729 [inlined]
  [6] invokelatest
    @ ./essentials.jl:726 [inlined]
  [7] notify
    @ ~/.julia/packages/Observables/ynr7h/src/Observables.jl:143 [inlined]
  [8] setindex!(observable::Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/ynr7h/src/Observables.jl:86
[...]

Issue #1579, opened 2022-01-11
poly function giving an ArgumentError: range step cannot be zero


mmkim1210:
Hello, I would like to plot a straightforward polygon as below, but I keep getting an error, `ERROR: ArgumentError: range step cannot be zero`. The entire stacktrace is pretty long, so I will not attach it here, but I am running the latest version of Julia and Makie. 
```julia
poly(Point2f[(2.32756952e8, 0.9), (2.32756952e8, 1), (2.32757008e8, 1), (2.32757008e8, 0.9)])
```

ffreyer:
This is another version of #931. You might be able to fix it by explicitly setting x limits before plotting, though the poly might have issues rendering as well. The other option is to shift this close to 0 and adjust the axis labels.

asinghvi17:
This should be fixed now with v0.21.
![iTerm2 tKyaAC](https://github.com/MakieOrg/Makie.jl/assets/32143268/b3023d48-796e-4920-950b-58d16478dc61)

Issue #1419, opened 2021-10-31
2d scatter! and poly! with lots of data, some circles are not drawn


nsajko:
The issue is that when I do `poly!`, only some circles are drawn. This is with GLMakie v0.4.7.

My code:

```
data_scatter = # ...

using Colors, GLMakie
set_window_config!(framerate = 8.0, pause_rendering = false, decorated = false)
fig = Figure()
ax = Axis(fig[1, 1], aspect = AxisAspect(1))
scatter!(ax, data_scatter, markersize = 2)
disks = # ...
poly!(ax, disks, color = colorant"rgba(255,0,0,0.125)", transparency = true)
```

(I tried with both `transparency = true` and `transparency = false`, it doesn't seem to make any difference.)

Info about my data from the repl:

```
julia> length.((data_scatter, disks))
(1000000, 375611)

julia> typeof.((data_scatter, disks))
(Vector{StaticArrays.SVector{2, Int32}}, Vector{Circle{Float32}})
```

nsajko:
Filtering the points and disks, so that there's only (54459, 20471) of them seems to be a workaround.

ffreyer:
From what I can tell it doesn't have much to do with the number of points/circles but with how large they are. Doing
```julia
fig = Figure()
ax = Axis(fig[1, 1], aspect = AxisAspect(1))
data_scatter = [(1 - (0.99e-6 * phi)^2) * Point2f(cos(0.01phi), sin(0.01phi)) for phi in 1:1_000_000]
scatter!(ax, data_scatter, markersize = 2)
disks = [Circle((1-(phi/376_000)^2) * Point2f(cos(0.05phi), sin(0.05phi)), 5f-4) for phi in 1:375611]
poly!(ax, disks, color = colorant"rgba(255,0,0,0.125)", transparency = true)
fig
```
looks pretty complete to me. But going to `1f-4` for circle radii makes them disappear. I'd guess that's a Float32 precision error...

nsajko:
> looks pretty complete to me.

@ffreyer Your example seems fine to me too (even though I don't really know how it's supposed to look). Here's an example that doesn't work so well (for me at least). The .txt file is Julia code defining the points and disks, and the screenshot zooms in on a particularly bad region.

[export_session.txt](https://github.com/JuliaPlots/Makie.jl/files/7452004/export_session.txt)


[...]

PR #2573, opened 2023-01-08
Fix float issues for CairoMakie by using Float64

ffreyer:
# Description

This is basically a light version of #2550, aiming to fix float precision errors for CairoMakie but not gl backends. Basically I'm just changing a bunch of variables to Float64 so we don't lose that precision. I will probably go through this again later to remove Float32 -> Float64 conversions.

Some related issues are collected in #2848.

Working
- [x] `scatter(1:10, 1e9 .+ (1:10))` in empty Scene with manual camera.projectionview
- [x] Axis
  - [x] autolimits of axis
  - [x] scatter
  - [x] lines
  - [x] linesegments
  - [x] mesh
  - [x] heatmap
  - [x] image
  - [x] meshscatter
  - [x] text
  - [x] poly
- [x] Axis3
  - [x] mesh
  - [x] surface
  - ~~volume~~
  - [x] meshscatter
- [x] revive GLMakie
- [x] revive WGLMakie
- [ ] revive RPRMakie?

TODO:
- [x] fix issues with varying types (either pin the converted type, allow automatic converting (like Rect2f -> Rect2{Float64}) or make converted type adjustable) - switched to Float64 for now
- [ ] switch to lazy converts (i.e. don't convert to Float64 by default. Keep the given (concrete?) type as long as possible)

test

## Type of change

- [x] Bug fix? I don't think there is a breaking change in this.

## Checklist

  - [x] Added unit test for "range step cannot be zero"


jkrumbiegel:
Very cool, this will enable easy use of DateTime as well, as that can be converted to Float64 losslessly if I'm not wrong

ffreyer:
Would it be better to have `plot.converted` to be more specific, i.e. just `Point{2, Float64}` or `Point{3, Float64}` for scatter, or allow eltypes from input arguments to carry through, i.e. also `Point{3, Float32}` to exist? The latter would reduce the number of conversions we do and reduce memory overhead but I guess it would hurt compilation/load times. I've also seen some type annotations here and there that would be invalid with option 2.

jkrumbiegel:
Hmm good question. In general, I think we've had issues because of converting things too eagerly. Everything is Float32 because GLMakie needs that eventually, but that has always robbed CairoMakie of using higher precision input data. Although the compilation latency is an argument. I think conversion should generally be the shortest path to a backend-consumable type.

ffreyer:
This should fix all the Float32 precision errors in plotting for CairoMakie only. Related issues include: #931, #1579 (error fixed, plotting only in CairoMakie), #396 (is this related?), #1419 (not sure), #780, #1373, #1196

In general this should fix `range step cannot be zero` errors in Axis and problems with transforms caused by lack of precision. Fixes: #1670, #2298, #1040

docs/documentation/recipes.md
```diff
@@ -238,7 +238,9 @@ function Makie.plot!(
             push!(bar_tos[], s.close)
         end
         append!(colors[], [x.close > x.open for x in stockvalues])
-        colors[] = colors[]
+        notify(linesegs)
+        notify(colors)
+        return
```
ffreyer, line 243: Before this worked without `notify(linesegs)` because Makie did not copy the Point2f array. Now/currently it converts to Float64, so it does copy and needs an Observable update

ffreyer:
>  → evaluating code [example_figure] in (examples/plotting_functions/ecdf.md)     
┌ Franklin Warning: in <examples/plotting_functions/ecdf.md>
│ There was an error of type 'ErrorException' when running a code block.
...

Took a while to understand why this is failing. This is also a problem on master, it just doesn't make itself known because I switched an `onany` to `map!` in here.

```julia
f, a, p = ecdfplot(rand(10), npoints = 4)
notify(p.input_args[1])
```

will trigger this. What happens is that during first construction Makie goes through a conversion path starting with `ECDF`, i.e. hitting https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/src/stats/ecdf.jl#L54-L62, then https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/src/stats/ecdf.jl#L9-L21, and then going into the regular conversion pipeline from `conversions.jl`. So here there is a conversion from input data to whatever ecdf produces to Points.

The notify however triggers a different pipeline. Starting from https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/src/interfaces.jl#L325-L335 it calls `convert_arguments` with `FinalType` which is `Combined{Makie.stairs}`. That means we skip the first two calls and directly call the conversion pipeline from `conversions.jl`. And if `npoints` is given that fails, because `stairs` doesn't handle that.

This also means that any old `f, a, p = ecdfplot(rand(10))` will become a `stairs` plot when updating arguments. For example:
```julia
fig = Figure()
ax = Axis(fig[1, 1])
ys = rand(10)
p = ecdfplot!(ax, ys, linewidth = 4, color = :black)
ecdfplot!(ax, ys, linewidth = 4, linestyle = :dot, color = :red)
notify(p.input_args[1])
display(fig)
```
Screenshot from 2023-01-14 16-12-42

ffreyer:
Before I completely forget what I did here - I think this is mostly done. I used Float64 as a converted eltype for now, mostly because I got annoyed with Rect conversions not working automatically. I have not tested RPRMakie because every example I tried was broken.

ffreyer:
The conversion from `plot.input_args` to `plot.converted` now targets Float64 types rather than Float32. A couple of other things now use Float64 too, like camera matrices and Axis limits. So in CairoMakie, you end up multiplying Float64 matrices and positions to get to pixel coordinates, which gives you Float64 precision. 

In GLMakie and WGLMakie there is now another conversion step going down to Float32 types, which loses that precision again. Things that need Float64 precision will plot in GLMakie and WGLMakie too, but it'll have rounding errors from reducing precision. That was kind of the case before too though. (Assuming you get around "range step cannot be zero")

I'm not sure what to do about GLMakie and WGLMakie yet. It would be nice too keep the solution to the backends and it would be good to not recompute Float32-safe positions every time the camera updates. The other draft pr I have is one attempt that avoids updating frequently, but would affect every backend. I was wondering if maybe we can setup a matrix decomposition to extract scales from `view`, but I haven't gotten far with that yet.

jkrumbiegel:
At least this PR alone would put us into a much better position for the big chunk of people who want to use CairoMakie + DateTimes (+ AlgebraOfGraphics). So I'm all for merging it if GLMakie and WGLMakie don't get worse than they already are in this respect.

ffreyer:
Eh, there are still some bugs to work out. rectangle zoom errors, cameras need more Float64...

GLMakie/src/GLAbstraction/GLUniforms.jl
```diff
@@ -15,11 +15,11 @@ opengl_postfix(T) = error("Object $T is not a supported uniform element type")
 
 
 opengl_prefix(x::Type{T}) where {T <: Union{FixedPoint, Float32, Float16}} = ""
-opengl_prefix(x::Type{T}) where {T <: Float64} = "d"
+opengl_prefix(x::Type{T}) where {T <: Float64} = error("Float64") # "d"
```
asinghvi17, line 18:
```suggestion
opengl_prefix(x::Type{T}) where {T <: Float64} = error("Float64 encountered in GLMakie internals!") # "d"

File: GLMakie/src/GLAbstraction/GLUniforms.jl [hunk 3]
```diff
@@ -15,11 +15,11 @@ opengl_postfix(T) = error("Object $T is not a supported uniform element type")
...
 opengl_prefix(x::Type{Cint}) = "i"
 opengl_prefix(x::Type{T}) where {T <: Union{Cuint, UInt8, UInt16}} = "u"
 
-opengl_postfix(x::Type{Float64}) = "dv"
+opengl_postfix(x::Type{Float64}) = error("Float64") # "dv"
```
asinghvi17, line 22:
```suggestion
opengl_prefix(x::Type{Float64}) = error("Float64 encountered in GLMakie internals!") # "d"

File: src/Makie.jl [hunk 4]
```diff
@@ -89,6 +89,13 @@ const RGBAf = RGBA{Float32}
 const RGBf = RGB{Float32}
 const NativeFont = FreeTypeAbstraction.FTFont
 
+const Vec2e = Vec2{Float64}
+const Vec3e = Vec3{Float64}
+const Vec4e = Vec4{Float64}
+const Point2e = Point2{Float64}
+const Point3e = Point3{Float64}
+const Point4e = Point4{Float64}
+
```
asinghvi17, line 98:
```suggestion
const Rect2e = Rect2{Float64}
const Rect3e = Rect3{Float64}
```
for completeness

src/camera/camera2d.jl
```diff
@@ -1,9 +1,9 @@
 struct Camera2D <: AbstractCamera
-    area::Observable{Rect2f}
-    zoomspeed::Observable{Float32}
+    area::Observable{Rect2{Float64}}
```
asinghvi17, line 2:
```suggestion
    area::Observable{Rect2e}
```
or do we want to move away from f's and e's entirely in the internals?

src/utilities/utilities.jl
```diff
@@ -330,7 +330,7 @@ Allows to supply `f`, which gets applied to every point.
 """
 function matrix_grid(f, x::AbstractArray, y::AbstractArray, z::AbstractMatrix)
     g = map(CartesianIndices(z)) do i
-        return f(Point3f(get_dim(x, i, 1, size(z)), get_dim(y, i, 2, size(z)), z[i]))
+        return f(Point3(get_dim(x, i, 1, size(z)), get_dim(y, i, 2, size(z)), z[i]))
```
asinghvi17, line 333: This might return a `Point3{Int}` or similar in the right circumstance; do we want to allow that?

ffreyer: I think more generally the question of "What do we want to convert to" is still open. For `plot.converted` we have 3 options:
1. Always convert to Float64 derived types
2. Keep whatever type comes from the first conversion 
3. Use generic Any types

1 is easy and what I set up for now. For GLMakie and WGLMakie it leads to extra conversions later down the line though, which I don't really like.

2 will require all the possible `plot.converted` types to be able to convert automatically. Currently at least `Rect` does not do that. It might also be bad for setting up empty plots, though I guess it's not that much of a problem.

3 would probably suck for performance?

SimonDanisch: I've been pushing for lazy converts in my plot object refsctors (2)

src/conversions.jl
```diff
@@ -548,13 +548,7 @@ function convert_arguments(::Type{<:Mesh}, mesh::GeometryBasics.Mesh{N}) where {
...
     end
     # TODO Float64?
-    # If already correct eltypes for GL, we can pass the mesh through as is
-    if eltype(metafree(coordinates(mesh))) == Point{N, Float32} && eltype(faces(mesh)) == GLTriangleFace
-        return (mesh,)
-    else
-        # Else, we need to convert it!
-        return (GeometryBasics.mesh(mesh, pointtype=Point{N, Float32}, facetype=GLTriangleFace),)
-    end
+    return (GeometryBasics.mesh(mesh, pointtype=Point{N, Float64}, facetype=GLTriangleFace),)
```
SimonDanisch, line 551: Why did you remove this? I just recently added this to avoid unnecessary conversions, which has been slow, and also materialized any `Buffer` inside the mesh, breaking this feature: https://docs.makie.org/stable/examples/plotting_functions/mesh/index.html#using_geometrybasicsmesh_and_buffersampler_type

src/utilities/utilities.jl
```diff
@@ -364,3 +364,64 @@ end
...
+    elseif e.head == :parameters # keyword args
+        return _to_type(e.args[1])
+    elseif e.head == :kw # default value
+        return _to_type(e.args[1])
+    else
+        dump(e)
+        error("Failed to parse expression.")
+        return ""
+    end
+end
```
ffreyer, line 427: I put this together to have an easier time figuring out function call chains. You can put `@trace` before a function (full or inline) and it'll inject some printing which will print the function header with types, the file and the line number. Should we keep this as a debug tool?

ffreyer:
I changed things to allow Float32 and Float64 to pass through.  Also did some cleanup (write out `return`, changed some Number -> Real types, removed an effectively duplicate method). 

https://github.com/JuliaGeometry/GeometryBasics.jl/pull/195 was necessary for the pr at some point, not sure if it still is. It would help avoid conversion errors either way though.

j-fu:
Will this also allow to plot data with ranges exceeding the range of Float32 ?

gvdeynde:
@ffreyer do you have any idea whether this is still "under active development"? This is really a breaking thing for me (I work a lot in loglog scales with data ranges that fall outside Float32 range...). Not pushing you or the Makie community but I would like to make a decision whether to wait for the coolest visualisation tool to get this feature or to go on with Plots.jl... I wish I could help, but I don't think my Julia nor Makie skills are up to par for this...

ffreyer:
This pr isn't active anymore but we are planning to work on the problem. Most likely in new prs. #3550 and #3565 are aimed towards this, #3226 might become the new main pr

PR #2577, opened 2023-01-09
Add a few more precompiles

timholy:
On my machine, this drops TTFP from 1.75s to just under 1s.

Note the second commit changes an internal type to `IdDict`. Happy to omit that if it would somehow be problematic.

### DIFF
--- a/GLMakie/src/precompiles.jl
+++ b/GLMakie/src/precompiles.jl
@@ -10,15 +10,22 @@ macro compile(block)
 end
 
 let
-    @precompile_all_calls begin
-        GLMakie.activate!()
-        screen = GLMakie.singleton_screen(false)
-        close(screen)
-        destroy!(screen)
-        base_path = normpath(joinpath(dirname(pathof(Makie)), "..", "precompile"))
-        shared_precompile = joinpath(base_path, "shared-precompile.jl")
-        include(shared_precompile)
+    @precompile_setup begin
+        x = rand(5)
+        @precompile_all_calls begin
+            GLMakie.activate!()
+            screen = GLMakie.singleton_screen(false)
+            close(screen)
+            destroy!(screen)
+            base_path = normpath(joinpath(dirname(pathof(Makie)), "..", "precompile"))
+            shared_precompile = joinpath(base_path, "shared-precompile.jl")
+            include(shared_precompile)
+            try
+                display(plot(x))
+            catch
+            end
+            closeall()
+        end
     end
-    closeall()
     nothing
 end
--- a/src/makielayout/types.jl
+++ b/src/makielayout/types.jl
@@ -10,10 +10,10 @@ struct DataAspect end
 
 
 struct Cycler
-    counters::Dict{Type, Int}
+    counters::IdDict{Type, Int}
 end
 
-Cycler() = Cycler(Dict{Type, Int}())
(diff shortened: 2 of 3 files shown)


jkrumbiegel:
Just out of interest, what is the benefit of `IdDict` with respect to latency?

timholy:
Much of the benefit is the `@nospecialize`s: https://github.com/JuliaLang/julia/blob/master/base/iddict.jl#L86-L110. For packages like MethodAnalysis which sometimes handle hundreds or thousands of types, I've literally seen a change from 10s of seconds to milliseconds by switching from `Dict` to `IdDict` to avoid excess specialization.

But also, `IdDict` uses pointers for comparisons, which is great when the keys are `DataType`s (`DataType`s are unique so a pointer is sufficient). In contrast, for type-keys `Dict` will use the whole subtyping machinery (`isequal`, for types, rapidly dispatches to `ccall(:jl_types_equal, ...)`) which is a lot slower.

timholy:
One thing to note about this PR: a window will pop up and then close during precompilation. If you don't like that, a remedy is to set `SnoopPrecompile.verbose[] = true` before running the block, seeing what's being precompiled, and then insert manual `precompile` directives. That has disadvantages, though, because Julia keeps changing: for example, before Julia 1.9 we had `kwfunc`s for individual methods, now we have a generic `Core.kwcall`. So the appropriate set of `precompile` directives is version-dependent and thus annoying to maintain.

A more robust but perhaps slightly less effective approach might be to look in more detail what `display` actually does, and then execute a reduced workload that forces compilation of as much of the pipeline as possible without actually opening a window.

jkrumbiegel:
> now we have a generic Core.kwcall

Oh that sounds interesting, can this relieve some of the latency problems connected to keyword arguments? It seemed to me that one couldn't `@nospecialize` a method's keyword arguments, and we use a lot of keyword arguments everywhere.

timholy:
That may have been fixed in recent Julia versions (I filed issues about it and they are closed). I'm unsure how it works with kwcall, though.

jkrumbiegel:
Oh I don't really care about `kwcall` itself, I just assumed that it might have changed / been introduced because of the latency problems with keyword args. So you're saying `@nospecialize` on kwargs works now?

timholy:
I think so, but worth testing.

timholy:
> it might have changed / been introduced because of the latency problems with keyword args

At least at one point it seemed like it was going to make the new (de)serializer simpler. I'm not certain the reasons for that are still true, but it's nice not to have a gensymmed name that isn't reproducible. So I think it's a nice change.

jkrumbiegel:
Hm this looks similar on 1.8 and 1.9, so I assume nothing changed in that regard? Or I'm using `@nospecialize` incorrectly for that purpose. But I assumed that there would be no further compilation for different keyword arg types after the first run.

grafik

SimonDanisch:
@timholy you can use `display(plot; visible=false)` which turns out has been broken actually :D 
Fixing it in https://github.com/MakieOrg/Makie.jl/pull/2587

SimonDanisch:
merged in https://github.com/MakieOrg/Makie.jl/pull/2587

timholy:
> But I assumed that there would be no further compilation for different keyword arg types after the first run.

Maybe try `Base.@constprop :none` too?

PR #2590, opened 2023-01-12
[CairoMakie] Special-case multipolygons as well as polygons

asinghvi17:
# Description

Fixes a lot of issues where plots with multipolygons (a) take a long time to save and (b) have a huge file size in vector formats.  This is especially common in geographic plotting.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/overrides.jl
+++ b/CairoMakie/src/overrides.jl
@@ -72,6 +72,7 @@ end
 
 draw_poly(scene::Scene, screen::Screen, poly, rect::Rect2) = draw_poly(scene, screen, poly, [rect])
 
+
 function draw_poly(scene::Scene, screen::Screen, poly, rects::Vector{<:Rect2})
     model = poly.model[]
     space = to_value(get(poly, :space, :data))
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
Instead of copying this stuff here:

```julia
color = poly.color[]
    if color isa AbstractArray{<:Number}
        color = numbers_to_colors(color, poly)
    elseif color isa String
        # string is erroneously broadcasted as chars otherwise
        color = to_color(color)
    end
    strokecolor = poly.strokecolor[]
    if strokecolor isa AbstractArray{<:Number}
        strokecolor = numbers_to_colors(strokecolor, poly)
    elseif strokecolor isa String
        # string is erroneously broadcasted as chars otherwise
        strokecolor = to_color(strokecolor)
    end
``` 

it should be refactored, it is a remnant of what the code was like before we centralized more of these conversions in Makie.jl

asinghvi17:
Looks like the CI failure was an HTTP timeout, unrelated to the changes made here.  CairoMakie CI also works on 1.6 so I think this is good to merge now.

jkrumbiegel:
I think there should be a `to_color` conversion first and then we continue with the value from that. No `isa`, the drawing variants should be done with methods then.

asinghvi17:
Isn't that what this is doing?  A vector of numbers passed to `color` will reach the backend plotting function as that same vector, so it has to be converted somewhere.  Unless you mean that the functionality should be lifted to some other method, like `to_color(plot, color)`, as a general rule?

jkrumbiegel:
That method already kind of exists and is used that way in the other plot primitives. Using colormap etc automatically

asinghvi17:
Scatter and lines in Cairo basically do the same thing (if color isa Array{Number} then numbers_to_colors(...)), see below:

lines
https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/CairoMakie/src/primitives.jl#L33-L35

scatter
https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/CairoMakie/src/primitives.jl#L188-L192

so if I understand, basically what you're saying is to remove the special-case from Strings, which I agree with and is what I did.

jkrumbiegel:
I think Simon wanted it more like `color = to_color(primitive)` or maybe `color = to_color(primitive, primitive.color[])` anyway, this was supposed to be handled in Makie and not the backends separately. It's not too big of an issue to continue with the code here, but that would just create even more of what we don't really want. Maybe @SimonDanisch should clarify which mode he prefers.

SimonDanisch:
Had to look at the code to decide, and while figuring out how to do it just fixed it 🤷 https://github.com/MakieOrg/Makie.jl/pull/2599

PR #2594, opened 2023-01-13
add relative text docs

jkrumbiegel:


ffreyer:
Would it be worth mentioning that `:relative` is specifically relative to the scene a plot is in?

PR #2596, opened 2023-01-13
docs improvements

jkrumbiegel:


SimonDanisch:
Do we still want to merge this?

jkrumbiegel:
it's kind of a stupid workaround for slowdowns due to repeated executions of certain pages in the build due to Franklin behavior. For example determining the env status is surprisingly slow, and I had seen that show up as a large block in a profile once while working interactively with the docs. But I don't quite remember if it made any difference for the actual build.

PR #2598, opened 2023-01-15
Fix mesh/surface with NaN points

asinghvi17:
# Description

- Skip mesh cells which have at least 1 point which has some NaN component.
- Allow NaNs to be passed on from surface to mesh in CairoMakie.

Fixes MakieOrg/GeoMakie.jl#133

## Type of change

## Checklist


ffreyer:
This generates some black triangles around the NaN values, probably because normals at the edge become NaN:

Screenshot from 2023-01-15 18-55-17

It also happens in GLMakie to a lesser degree. I think there NaN points are ignored when generating normals. https://github.com/MakieOrg/Makie.jl/blob/db55264501ffbf68154a6fa4268433c3dde1e0c2/GLMakie/assets/shader/util.vert#L264-L288. 

Screenshot from 2023-01-15 19-13-12

asinghvi17:
Good point - will probably reimplement the normal finding in CairoMakie then, for ease of use.

SimonDanisch:
Would be nice to put that into makie, since Wglmakie also calculates normals on the cpu

CairoMakie/src/primitives.jl
```diff
@@ -928,12 +946,12 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Maki
...
 
 function surface2mesh(xs, ys, zs::AbstractMatrix)
-    ps = Makie.matrix_grid(p-> nan2zero.(p), xs, ys, zs)
+    ps = Makie.matrix_grid(identity, xs, ys, zs)
     rect = Tesselation(Rect2f(0, 0, 1, 1), size(zs))
     faces = decompose(QuadFace{Int}, rect)
     uv = map(x-> Vec2f(1f0 - x[2], 1f0 - x[1]), decompose_uv(rect))
-    uvm = GeometryBasics.Mesh(GeometryBasics.meta(ps; uv=uv), faces)
-    return GeometryBasics.normal_mesh(uvm)
+    uvm = GeometryBasics.Mesh(GeometryBasics.meta(ps; uv=uv), faces, )
```
SimonDanisch, line 953: Just do here:
```julia
return Mesh(meta(ps; uv=uv, normals=nan_aware_normals(ps, faces), faces)
```
And skip the whole `nan_aware_*(geometry_func)` business ;)

SimonDanisch (changes requested):

CairoMakie/src/primitives.jl
asinghvi17: Huh, didn't know you could do that...but should have guessed 😅 

Do we want this kind of behaviour for meshes as well though?  If so we can force recompute the normals in mesh as well.

SimonDanisch: That'd be a pretty big change... I'll just have the `nan_aware_normals` function in Makie, us it for a while and go from there ;)

asinghvi17: Sounds good.

asinghvi17:
CI is green now, we should be good to go!

src/utilities/utilities.jl
```diff
@@ -322,6 +322,67 @@ function surface_normals(x, y, z)
...
+"""
+function nan_aware_orthogonal_vector(v1, v2, v3)
+    centroid = Vec3f(((v1 .+ v2 .+ v3) ./ 3)...)
+    normal = [0.0, 0.0, 0.0]
+    # if the coord is NaN, then do not add.
+    (isnan(v1) | isnan(v2)) || (normal += cross(v2 .- centroid, v1 .- centroid))
+    (isnan(v2) | isnan(v3)) || (normal += cross(v3 .- centroid, v2 .- centroid))
+    (isnan(v3) | isnan(v1)) || (normal += cross(v1 .- centroid, v3 .- centroid))
+    return Vec3f(normal).*-1
+end
```
ffreyer, line 344: This produces a NaN vector with one NaN input because centroid cmbines all 3:
```julia
julia> Makie.nan_aware_orthogonal_vector(Point3f(0), Point3f(1,0,0), Point3f(NaN))
3-element Vec{3, Float32} with indices SOneTo(3):
 NaN
 NaN
 NaN
```
I'm not sure why this doesn't affect things down the line, but you can probably just do
```julia
any(isnan, (v1, v2, v3)) && return Vec3f(0)
return cross(v2 - v1, v3 - v1)
```
instead?

ffreyer:
This can look quite a bit different from GLMakie in extreme cases, e.g.

```julia
surface([0,1,1], [0, 1], [0 0; 1 0; 0 1], colormap = [:lightblue, :lightblue])
```
(CairoMakie | GLMakie)

Screenshot from 2023-01-17 13-52-16

but I'm not sure whether CairoMakie or GLMakie should be seen as the problem here.

In terms of NaN values this produces nicer results than GLMakie (no darkening) and it also happens to clean up the little artifact you get when rendering a sphere as a surface plot:

Screenshot from 2023-01-17 13-54-53

src/utilities/utilities.jl
SimonDanisch: `any(isnan, (v1, v2, v3))` this is in the drawing code, so I guess that's why it isn't a problem

ffreyer: Ah, no, it is affecting things down the line.

If we have two faces
```
a --- b
| \   |
|   \ |
c --- d
```
and only b is NaN, orthogonal_vector will produce NaN as the face normal of (a, b, d). This is getting added to the normal at vertex a, b and d via
```julia
for i in 1:length(F)
    fi = face[i]
    normals_result[fi] = normals_result[fi] + n
end
```
When we iterate the faces, (a, b, d) but (a, c, d) is not. It seems like the shading is doing something differently though, so it's not as obvious.

---

With a plain color you can see the effect
```julia
zs = rand(10, 10)
ns = copy(zs)
ns[4, 3:6] .= NaN
f, a, p = surface(1..10, 1..10, ns, colormap = [:lightblue, :lightblue])
scatter!(a, [Point3f(i, j, zs[i, j]) for i in 1:10 for j in 1:10], color = isnan.(ns'[:]), depth_shift = -1f-3)
wireframe!(a, CairoMakie.surface2mesh(to_value.(p.converted)...), depth_shift = -1f-3, color = :black)
f
```

Screenshot from 2023-01-17 16-13-12

If you check the number of NaN normals you get 15.

```julia
m = CairoMakie.surface2mesh(to_value.(p.converted)...)
isnan.(m.normals) |> sum
```

If you mark them you get this (not showing position with NaN values in z)

Screenshot from 2023-01-17 16-18-05

The bottom right corner is probably not in a triangle with a NaN position, so the normal is clean. The two on left - no idea... 

With
```julia
function nan_aware_orthogonal_vector(v1, v2, v3)
    (isnan(v1) || isnan(v2) || isnan(v3)) && return Vec3f(0)
    Vec3f(cross(v2 - v1, v3 - v1))
end
```
you get

Screenshot from 2023-01-17 16-26-41

in CairoMakie (with NaN normals still marked, new zs/ns)

Looking at this in more detail also brings up two more questions:
- Why is GLMakie extending the surface past the vertex positions around NaN?
- Should CairoMakie be iterating quad faces to avoid getting triangles in corners? (I think we could just filter faces containing a NaN coordinate for this in `surface2mesh`)

ffreyer (approved):
I pushed a fix for the NaN normals and removed the triangles by filtering. I now get

Screenshot from 2023-01-17 16-38-34

with
```julia
zs = rand(10, 10)
ns = copy(zs)
ns[4, 3:6] .= NaN
f, a, p = surface(1..10, 1..10, ns, colormap = [:lightblue, :lightblue])
m = CairoMakie.surface2mesh(to_value.(p.converted)...)
scatter!(a, m.position, color = isnan.(m.normals), depth_shift = -1f-3)
wireframe!(a, m, depth_shift = -1f-3, color = :black)
f
```

I think with that I'm happy. Now GLMakie needs to catch up.

src/utilities/utilities.jl
asinghvi17: I had tried quad faces for surface earlier, but file sizes ballooned so it turned out to be counterproductive.

ffreyer:
I guess it should be expected that either the GLMakie version pr CairoMakie version of the new refimg test fails. I'd say we should upload the CairoMakie version for now and only test that

ffreyer:
I started working on fixing this in GLMakie and WGLMakie. I can push this here once I'm done or in another pr, whatever is preferred. Currently got GLMakie working with the two extra triangles and WGLMakie just like CairoMakie.

ffreyer:
Ok I think I got it all cleaned up now. Eventually I figured out that the integer index -> uv index in GLMakie was wrong, and that with an `isnan` check in the normal generation was all that was needed there. I noticed a few notes about `isnan` not working consistently, but I'm not sure what else one would check. (I've seen something like `x <= 0 || x >= 0`, maybe that?)
For WGLMakie I moved the `surface2mesh` function to Makie and just reused that with some minor tweaks. I also cleaned up the refimg test some more (specifically also removing the `rand`). Now it's this and should work across all backends: (This is WGLMakie)

Screenshot from 2023-01-29 01-27-15

ffreyer:
Tests should pass after a refimg update

SimonDanisch:
Nice!!
This is GLMakie:
image
That's wrong right?

ffreyer:
Yea it is. On mine it's fine.😞 

I'd guess this a couple bits of difference in texture reads. Removing the `+ 1e-6` might help (though I guess it might also trigger these issues on other gpus) but using `texelFetch` instead of `texture` is probably the way to go. I'll push another commit later.

SimonDanisch:
Damn, this introduces a regression for ` heatmaps & surface`:
image
image
The recorded version also seems wrong, but I diffusely remember giving up debugging the CI GPU, since it seemed to be an obscure glsl issue:
image

SimonDanisch:
WGLMakie flakiness should be fixed by https://github.com/MakieOrg/Makie.jl/pull/2661

GLMakie/src/glshaders/surface.jl
```diff
@@ -48,7 +49,8 @@ function _position_calc(
     """
     int index1D = index + offseti.x + offseti.y * dims.x + (index/(dims.x-1));
     ivec2 index2D = ind2sub(dims, index1D);
-    vec2 index01 = vec2(index2D) / (vec2(dims)-1.0);
```
SimonDanisch, line 51: Hm, it's this change that makes the test fail.. It looks a bit like the colors get moved out of the surface:
PR:
image
With the old `vec2(index2D) / (vec2(dims)-1.0)`:
image

Considering, that the old behavior maps more directly to the displayed matrix, why was this changed?

ffreyer: With `vec2(index2D) / (vec2(dims)-1.0)` you get values between 0 and 1, so the first index will sample the left edge of a pixel in a texture and the last index will sample the right edge of a pixel.

With `vec2(index2D) / (vec2(dims))` you'd always sample the left edge of a pixel. With the `0.5` shift that becomes the center of a pixel. 

I think I made this change to fix the discrepancy between GLMakies generated mesh and the wireframe representation of CairoMakies mesh: (right)
img

ffreyer: I think is also an expected/desired change.

The color data we have here is:
```
 1.0  NaN        NaN        5.0
 2.5    2.66667    2.83333  3.0
```
In surface these should map directly to vertices and we should be interpolating between them. Assuming quad faces, this should generate all NaN's. 

With vertices marked as `+` and the positions where the texture is sampled without interpolation marked with pink dots, this should be the before:
Screenshot from 2023-03-30 00-29-11
And this the after:
Screenshot from 2023-03-30 00-29-21

SimonDanisch: Hm, but we shouldn't render the edges pretty much outside the visible field?
I think, if we do this, we'll need to move the interpolation center to the middle of the quad!

SimonDanisch: Maybe to get this branch merged, revert things for now, so that we dont have a regression?

ffreyer: My understanding of surface plots is that it's a mesh generating method where vertex positions follow from xs and ys and the colors/values passed should exactly map to those positions. That's not the case on master. That's a bug imo.

Moving the color to the center of a quad doesn't work because we have N colors for N vertices. We could move the full vertex to quad center but then we would need to generate an N+1 grid to keep quad sizes regular.

Maybe it would be useful to compare to some other plotting libraries? To me it should just be like my after example.

ffreyer: If we look at this with automatically chosen color values, i.e. z values, I would expect the first color in the colormap to directly correspond to the lowest z value, the last to the highest and anything in between  to be smoothly interpolated. That's not the case on master, but is the case in the pr. 

This example makes the difference very obvious:
```julia
scene = Scene()
surface!(scene, [1 2 3; 1 2 3; 1 2 3], colormap = [:black, :white, :black], shading = false)
cam3d!(scene, projectiontype = Makie.Orthographic)
center!(scene)
axis3d!(scene)
lines!(scene, Point3f[(3, 3, 1), (3, 3, 2), (3, 3, 3)], color = [:black, :white, :black], linewidth = 50)
update_cam!(scene, Vec3f(5.5, 5, 2), Vec3f(1.5, 1.5, 2))
scene
```

Screenshot 2023-04-05 154927

The line to the left reaches full black at the line start/end, i.e. at minimum and maximum z. The surface on the other hand at z values significantly larger than the minimum and significantly lower than the maximum.

On another note, why is `surface` offset like heatmap? As a continuous thing I would expect it to use `axes(zs, 1)` and `axes(zs, 2)`, not some larger range like heatmap.

ffreyer: > Maybe it would be useful to compare to some other plotting libraries? To me it should just be like my after example.

MatPlotLibs `surf` seems to be discrete so it's not comparable. 

## Plots with GR backend
gr

## Plots with Plotly backend
Screenshot from 2023-04-05 16-06-42

## Plots with PythonPlot backend
Screenshot from 2023-04-05 16-07-35

## Plots with Gaston backend
Screenshot from 2023-04-05 16-16-58

All of these are
```julia
Plots.surface([1 1 1; 2 2 2; 3 3 3], colormap = cgrad([:black, :white, :black], 3))
```
---

## This PR

Screenshot from 2023-04-05 16-27-42

Out of those, none do our `0.0:1.5:3.0` x and y range, half do flat face coloring, half do smooth coloring and all match min/max z to the first/last color in the colormap like in this pr.

ffreyer:
Docs failure should be unrelated. The code it highlights works fine locally

ffreyer:
Closing this since it's in the beta branch

PR #2599, opened 2023-01-16
Special-case multipolygons as well as polygons  (continue #2590)

SimonDanisch:
Cleaned up the color handling from #2590


asinghvi17 (approved):
Looks great to me - thanks @SimonDanisch!  I think we can probably merge this as is.  I tested and it seems to work for `color::Vector{<: Colorant}` also, which covers all use cases that I'm aware of.

jkrumbiegel (approved):

PR #2614, opened 2023-01-19
Call empty! in block deletion

ffreyer:
# Description

See https://discourse.julialang.org/t/removing-row-from-gridlayout/93196

Still shouldn't completely clean up a block, but remove plots and a decent chunk of observables.

## Type of change

### DIFF
--- a/src/makielayout/blocks.jl
+++ b/src/makielayout/blocks.jl
@@ -494,6 +494,9 @@ end
 
 function Base.delete!(block::Block)
     block.parent === nothing && return
+    
+    # detach plots, cameras, transformations, px_area
+    empty!(block.blockscene)
 
     s = get_topscene(block.parent)
     deleteat!(


jkrumbiegel (approved):

jkrumbiegel:
At some point we should add tests that check that a block and the relevant scenes, plots, etc. can be garbage collected once they're deleted. I assume unless the last observable connection is severed, that's not possible.

ffreyer:
The example from discourse still has a wild amount of listeners in `fig.scene.px_area` if you remove all blocks with this. https://github.com/MakieOrg/Makie.jl/blob/94b5ec5c057f2bbcdbc8df1eecb3c480bb608679/src/scenes.jl#L177-L182 might be partially to blame since `on` doesn't add to `obs.inputs`? Though most listeners seem to be maps from https://github.com/MakieOrg/Makie.jl/blob/94b5ec5c057f2bbcdbc8df1eecb3c480bb608679/src/scenes.jl#L240-L248 which should show up, I think, but don't in any `block.blockscene.px_area.inputs`?

jkrumbiegel:
Yes, `px_area` and `scene.camera.projectionview` are lifted all over the place.

ffreyer:
`scene.events` are also kind of a nightmare to clean up atm, since the events struct is shared between all scenes. So you can't just empty all listeners there and call it a day. You also can't have `child.events.x = map(identity, parent.events.x)` because it would mess up priority. 
Maybe we could add some wrapper that keeps track of where listeners have been added, but still adds them to the scene-global events? I.e.
```julia
struct LocalEvents
    global_events::Events
    local_obsfuncs::Vector{Function}
end
```
that adds to `local_obs_funcs` on map, on, etc? That could then do `off.(local_obsfuncs)` when `empty!(scene)` is called (for each child scene as well).

PR #2616, opened 2023-01-19
Add scrollspeed for Menu and increase the default for non apple system

ffreyer:
# Description

Fixes https://discourse.julialang.org/t/how-to-increase-a-scrolling-speed-in-makie-menu/93221

I also added a news entry for #2614 since I forgot to do it there and this is a small pr as well.

## Type of change

## Checklist


ffreyer:
We could also try normalizing with `sign(scroll)` but I think it's better to have varying speed multipliers depending on the OS so that the extra resolution is used when available.

ffreyer:
Test failure is unrelated

jkrumbiegel:
I actually don't think it's an apple thing, I think it's a trackpad vs mouse thing. A scroll wheel usually has a discrete scroll behavior and that seems to trigger a value of 1. And trackpad values come much more frequently but with smaller values.

jkrumbiegel:
For example, this is what I get when I scroll a little bit, like a centimeter downwards, on my macbook:

```
s = Scene()
on(println, s.events.scroll)
s
```

```
julia> (0.0, 0.1)
(0.4, 0.6000000000000001)
(0.4, 0.4)
(0.2, 0.4)
(0.2, 0.4)
(0.1, 0.1)
(0.1, 0.4)
(0.2, 0.4)
(0.4, 0.5)
(0.4, 0.4)
(0.30000000000000004, 0.30000000000000004)
(0.1, 0.1)
(0.1, 0.5)
(0.0, 0.1)
(0.1, 0.30000000000000004)
(0.30000000000000004, 0.4)
(0.5, 0.5)
(0.30000000000000004, 0.5)
(0.30000000000000004, 0.4)
(0.2, 0.4)
(0.2, 0.30000000000000004)
(0.5, 0.5)
(0.2, 0.30000000000000004)
(0.1, 0.1)
(0.30000000000000004, 0.30000000000000004)
(0.2, 0.4)
(0.0, 0.1)
(0.1, 0.30000000000000004)
(0.1, 0.0)
```

I don't know what the correct way is to reconcile that with a mouse that outputs `(0, 1)` for every discrete scroll event.

ffreyer:
Hmm, I thought it was kind of the inverse issue where ios had larger numbers. But if it's trackpads that have more scroll events and feel faster then I'm not sure what to do about it. Something like `scroll_y == sign(scroll_y) ? 10 * scroll_y : scroll_y` might work but seems prone to glitchyness to me.  Maybe we should just add the scroll speed attribute here and try to figure out a solution for trackpad vs mousewheel some other time?

jkrumbiegel:
One hacky workaround I thought about was to simply feed the scroll values through some logic that checks if there has ever been a value other than 0, 1 or -1. If yes, then apply a trackpad scaling behavior, otherwise scroll wheel.

mkoculak:
My scroll wheel generates whole values, but if I scroll fast enough, I can get 2, 3, or 4 returned also.
So maybe just whole vs fractional values?

ffreyer:
On my old laptop the trackpad is recognized as -1/0/+1 on ubuntu. On windows it's not working at all anymore. I guess my drivers got corrupted somehow...

I guess we could check something like `is_integer = is_integer && (scroll_y == sign(scroll_y) * abs(scroll_y))` though that would still be problematic when switching between mouse and trackpad. It doesn't look like we have a better choice though

src/makielayout/blocks/menu.jl
```diff
@@ -292,7 +292,9 @@ function initialize_block!(m::Menu; default = 1)
     on(menuscene.events.scroll, priority=61) do (x, y)
         if is_mouseinside(menuscene)
             t = translation(menuscene)[]
-            new_y = max(min(t[2] - y, 0), height(menuscene.px_area[]) - listheight[])
+            # Hack to differentiate mousewheel and trackpad scrolling
+            step = (isinteger(y) ? 15.0 : 1.0) * m.scroll_speed[] * y
+            new_y = max(min(t[2] - step, 0), height(menuscene.px_area[]) - listheight[])
```
ffreyer, line 297: I assume this will cause problems with trackpads but maybe it's worth trying?

jkrumbiegel:
I think for the heuristic we do need some value tracking, because a trackpad can also output integers:

```julia
s = Scene()
on(vals -> any(x -> x != 0 && isinteger(x), vals) && println(vals), s.events.scroll)
s
```

With this I get occasional output when scrolling around:

```
julia> (0.4, -2.0)
(1.1, 9.0)
(0.0, -5.0)
(0.8, 5.0)
(0.0, -3.0)
(0.0, -11.0)
(0.4, 1.0)
(-0.5, -2.0)
(0.0, 3.0)
(0.0, 2.0)
(1.0, 5.9)
(0.0, 9.0)
(0.0, 2.0)
(0.0, -15.0)
(0.8, 3.0)
(0.0, 5.0)
(-0.4, -3.0)
(0.0, -1.0)
(-0.30000000000000004, -1.0)
(0.8, 2.0)
(0.7000000000000001, 2.0)
(0.4, 1.0)
```

ffreyer:
Yea, that's what I expected. Since this would be generally useful, should we adjust the scroll event to be `(dx, dy, has_always_been_integer)`? In that case we may want to split into two mini pr's, one for scroll_speed and one being a breaking change in the Events struct?

jkrumbiegel:
Wouldn't it be easiest to modify the number sent out by the event struct already so that users get usably transformed values? Kind of doing what glfw is not doing for us. The values would change compared to what they are now but I wouldn't consider that change breaking.

In terms of the logic, I would set the default to assume mouse and change that to touchpad when a non-integer is encountered maybe? That would flip the trackpad correctly from the first non-integer sample, which given the accelerating motion of the fingers should happen pretty much immediately? I have no idea if there aren't mice with smoothly varying wheels that would break this scheme but I guess we'll find out?

Another question is if this decision can be made across scenes or if it's per scene?

ffreyer:
> Wouldn't it be easiest to modify the number sent out by the event struct already so that users get usably transformed values? Kind of doing what glfw is not doing for us. The values would change compared to what they are now but I wouldn't consider that change breaking.

I generally prefer making changes like that a bit later, just in case some use cases do not want the change. But we can do that as well.

> In terms of the logic, I would set the default to assume mouse and change that to touchpad when a non-integer is encountered maybe? That would flip the trackpad correctly from the first non-integer sample, which given the accelerating motion of the fingers should happen pretty much immediately? I have no idea if there aren't mice with smoothly varying wheels that would break this scheme but I guess we'll find out?

Yea sounds good. From what googling about this topic showed me it seems like apple mouses have fractional scrolling on macs at least, while 3rd party ones do not. I have two logitech mouses with free swinging scroll wheels which only produce integer scroll steps on ubuntu and windows.)

> Another question is if this decision can be made across scenes or if it's per scene?

If we do the scaling before `events.scroll` it would be global. With `(dx, dy, has_always_been_integer)` the decision would be pushed to individual listeners, so every interaction could choose what to do with it.

jkrumbiegel:
I should find the mouse I have lying around here somewhere and test it on mac.

ffreyer:
Looking at GLFW it seems like apple has fractional scrolling natively
https://github.com/glfw/glfw/blob/d299d9f78857e921b66bdab42c7ea27fe2e31810/src/cocoa_window.m#L598-L611

Windows and linux with x11 don't
https://github.com/glfw/glfw/blob/d299d9f78857e921b66bdab42c7ea27fe2e31810/src/win32_window.c#L954-L966
https://github.com/glfw/glfw/blob/d299d9f78857e921b66bdab42c7ea27fe2e31810/src/x11_window.c#L1368-L1376

And linux with wayland, does I guess, but I don't know if it would need rescaling:
https://github.com/glfw/glfw/blob/d299d9f78857e921b66bdab42c7ea27fe2e31810/src/wl_window.c#L1289-L1314

jkrumbiegel:
Weird that glfw's also using seemingly arbitrary constants for scaling. I guess for Mac it would matter whether the active device `hasPreciseScrollingDeltas` or not

jkrumbiegel (approved):

PR #2617, opened 2023-01-19
Add TrackedEvents for better event cleanup

ffreyer:
# Description

`Events` are hard to clean up because every child scene has the same event struct, i.e. the same observables, containing all callbacks.
This is an attempt to solve this with two wrappers `TrackedObservable` and `TrackedEvents`.  Each child scene gets its own `TrackedEvents(parent.events)` which mirrors `Events` except that every observable is wrapped in a `TrackedObservble`. When calling `on` etc on any of those, the registered callback is tracked and added to the shared event observable. This means that every child scenes keeps track of its "local" event callbacks, which can then be cleaned up on `empty!(scene)`.

I tested this with the same Menu deleting code from https://discourse.julialang.org/t/removing-row-from-gridlayout/93196 and it seems to do its job.

See also: #2614

## Type of change

- [x] Bug fix?

## Checklist

- [ ] Cleanup

### DIFF
--- a/GLMakie/test/unit_tests.jl
+++ b/GLMakie/test/unit_tests.jl
@@ -1,4 +1,5 @@
 using GLMakie.Makie: getscreen
+using GLMakie.Makie.Observables: listeners
 
 function project_sp(scene, point)
     point_px = Makie.project(scene, point)
@@ -49,7 +50,7 @@ end
         # assure we correctly close screen and remove it from plot
         @test getscreen(ax.scene) === nothing
         @test !events(ax.scene).window_open[]
-        @test isempty(events(ax.scene).window_open.listeners)
+        @test isempty(listeners(events(ax.scene).window_open))
 
         # Test singleton screen replacement
         fig, ax, p = scatter(1:4);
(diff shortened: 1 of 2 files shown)


src/events.jl
```diff
@@ -0,0 +1,302 @@
...
+#     if field === :mousebuttons
+#         error("`events.mousebuttons` is deprecated. Use `events.mousebutton` to react to `MouseButtonEvent`s instead.")
+#     elseif field === :keyboardbuttons
+#         error("`events.keyboardbuttons` is deprecated. Use `events.keyboardbutton` to react to `KeyEvent`s instead.")
+#     elseif field === :mousedrag
+#         error("`events.mousedrag` is deprecated. Use `events.mousebutton` or a mouse state machine (`addmouseevents!`) instead.")
+#     else
+#         return getfield(e, field)
+#     end
+# end
```
ffreyer, line 171: This was just there to give people a useful error message when updating from an older version of the `Events` struct. I think it's fine to delete this now?

ffreyer: 2 years since those fieldnames were changed, 8 months since the warning was turned into an error

jkrumbiegel:
I think this is not necessary anymore with the changes in #2731 @SimonDanisch? You introduce some kind of tracking there, too.

SimonDanisch:
I don't really like the dependence on a new Observable type (it's somewhat breaking because we haven't really used `AbstractObservable` anymore since I've been determined not to use too many subtypes), so I'll close this for now! @ffreyer, let me know if there's something missing in #2731, besides, that users have to take care of this by using the `on/map(f, scene/plot, ...)` functions... 
I think we should soon update our event api, to improve the decoupling from the scene even more anyways!
E.g.:
```julia
on(plot, :projection) do projection 
...
end
on(plot, :mouseposition) do position
end
```
Which will then gets only connected when displayed on a certain scene...

PR #2620, opened 2023-01-20
NFC remove deprecated `Vararg{<:Integer}`

oscardssmith:
this caused deprecation errors when running with `--depwarn=yes`

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)


jkrumbiegel (approved):
Ah that's where that warning came from, thanks

SimonDanisch (approved):

PR #2622, opened 2023-01-20
Allow scalar numbers as colors, looked up from colormaps

jkrumbiegel:
# Description

Fixes # (issue)

This makes it easier to quickly pick certain colors from colormaps, by setting the colorrange to `(1, n)` for example, the `i`th color can be chosen by setting `color = i`:

```julia
fig = Figure()
for j in 1:4
    ax = Axis(fig[fldmod1(j, 2)...])
    for i in 1:3j
        lines!(ax, cumsum(randn(100)), color = i, colorrange = (1, 20), colormap = :tab20)
    end
end
fig
```

grafik

## Type of change

## Checklist


jkrumbiegel:
@SimonDanisch this is ready, could you have a look over the WGLMakie / GLMakie changes? I never feel that comfortable with those code bases

SimonDanisch:
Is it possible to integrate this change here: https://github.com/MakieOrg/Makie.jl/blob/master/src/interfaces.jl#L15
So that this doesn't need a backend change?
I think we want to move more towards preprocessing all the different color options before passing the plot to the backend anyways!

jkrumbiegel:
if you look at the code, I did do that. But all the other changes still seemed necessary.

SimonDanisch:
>if you look at the code, I did do that.

You updated it to work with single numbers as color, but you didn't update it to already convert to a single color, so that the backend needs to do more work, even though this should work the same way across backends...

jkrumbiegel:
superseded by #2754

PR #2628, opened 2023-01-25
Fix #2627 (SSAO artifacts at scene boundary)

### LINKED ISSUE #2627: Strange bounding box using mesh! on Axis3 with SSAO enabled
Following discussion on Discord, here's a MWE which highlights a strange bounding box which appears on Axis3 with SSAO enabled.

```julia
using GLMakie
GLMakie.activate!(ssao=true)
fig = Figure(resolution=(400,400))
ax = Axis3(fig[1,1])		
mesh!(ax, brain)
fig
```

Screenshot 2023-01-25 at 11 48 03

ffreyer:
# Description

Fixes #2627

This is an issue with clear values. The buffer for (normal_vec, occlusion) is cleared to (0, 0, 0, 0),  where 0 means a pixel should be maximally darkened. The first stage of SSAO overwrites these values within a scenes area. The second stage then grabs a (- blur_radius, + blur radius) region around each pixel in the scene and averages the occlusion values. This means pixel outside the scene area are grabbed too, which may still be 0, causing darkening.

The change here is inverting the scale of occlusion, so that 0 is now "no darkening" and 1 is "maximally darkened". Alternatively we could add another clear to `(0,0,0,1)` but I think the current change should have slightly better performance.

## Type of change


jkrumbiegel:
Shouldn't the kernel rather not pick values outside of the scene area at all? Seems like one could use repetition at the border

ffreyer:
Ok, for the SSAO example from the docs it pretty clearly is. 

Mean Time for 1k frames:
- before [4b76f65](https://github.com/MakieOrg/Makie.jl/pull/2628/commits/4b76f652287e65e89b1835e045ca957a598edc69): 0.465s
- using float math instead of branching: 0.398s
- returning early if there is no object at the current pixel + bool math: 0.349s

If the scene is dense enough it might eventually not be worth it to return early, but I think most 3D scenes end up with a lot of white space (and anything transparent counts as white space here) so it should usually be an improvement.

PR #2633, opened 2023-01-26
adapt to new JSServe version

SimonDanisch:


SimonDanisch:
Ok, WGLMakie benchmark is expected to fail, since master is still broken and it compares against that...

PR #2602, opened 2023-01-17
Add Fontconfig font search

asinghvi17:
# Description

This PR introduces a method to use Fontconfig to find and create font sets.

Basically, what this does is introduce 3 main methods.

1. The ability to obtain all styles in a font family (`populate_font_family`) - returns a NamedTuple of Symbol `styles => FTFont (Makie.NativeFont)`.
2. The ability to search for a particular style within a font family, along with some basic fallbacks (regular can be roman or book, for example, if there is no style called `regular`) - `search_font(family, style)`
3. The ability to cache font families and regularize some names, so that `(:regular, :bold, :italic, :bold_italic)` are always available, even if the font has obliques instead of italics.  Individual styles can still be referred to by their original names.  This is the `FontFamily` interface, represented by the `FontFamily` struct, which the user can edit.  It also automatically searches for a style if it can't exactly match, making it a bit more dynamic than what we currently have.

This results in the following:

```julia
julia> FontFamily("TeX Gyre Heros Makie")
FontFamily("TeX Gyre Heros Makie", Attributes with 5 entries:
  bold => FTFont (family = TeX Gyre Heros Makie, style = Bold)
  bold italic => FTFont (family = TeX Gyre Heros Makie, style = Bold Italic)
  bold_italic => FTFont (family = TeX Gyre Heros Makie, style = Bold Italic)
  italic => FTFont (family = TeX Gyre Heros Makie, style = Italic)
  regular => FTFont (family = TeX Gyre Heros Makie, style = Regular))
```
and `fonts = "TeX Gyre Heros Makie"` is equivalent to `fonts = FontFamily("TeX Gyre Heros Makie")`, which is what has been provided in the default theme.  This way, people can access the exact font style (`Extracondensed Semibold` or what have you) as well as some basic shortcuts.

The user interface is still not set in stone, and this is very much a draft PR.

## Type of change

Delete options that do not apply:

- [x] New feature (non-breaking change which adds functionality)

## TODOs

- [ ] Finalize the FontFamily interface and try to make it more efficient.
- [ ] Properly document the FontFamily interface - people shouldn't be passing strings and expecting elements of the family.
- [x] Add an entry in NEWS.md (for new features and breaking changes)
- [x] Add or change relevant sections in the documentation
- [ ] Add unit tests for new algorithms, conversion methods, etc.
- [ ] Add reference image tests for new plotting functions, recipes, visual options, etc.

PR #2641, opened 2023-01-30
Add fontconfig-based font search

asinghvi17:
# Description

Continuation of #2602 - adds Fontconfig-based font search to Makie.  Users access this through the `font_family(family::String)` function, which returns Attributes and is documented.

Internally, this works in the following way:

When a font family is constructed, we first search the system using Fontconfig to find all fonts which are applicable.  Then, we try to make sure that we have some kind of regular, bold, italic and bold italic font - mainly by trying to use substitutes (roman and book for regular, oblique for italic, etc).  Then, we return a set of Attributes which is guaranteed to have the four basic font styles used in Makie, if they exist.

I also put this kind of mechanism into the current `to_font` method, and hooked up caching, such that:

```julia
julia> @time to_font("lato", "regular")
  0.002450 seconds (42 allocations: 2.195 KiB)
FTFont (family = Lato, style = Black)

julia> @time to_font("lato", "regular")
  0.000003 seconds (1 allocation: 32 bytes)
FTFont (family = Lato, style = Black)

julia> @time to_font("lato", "regular")
  0.000005 seconds (1 allocation: 32 bytes)
FTFont (family = Lato, style = Black)
```

and `to_font("helvetica bold")` yields `FTFont (family = Helvetica, style = Bold)`, which circumvents the limitations of the previous system.

## Type of change

## Checklist


asinghvi17:
I have no clue why CI failed here - this is the error message:
```juulia
┌ Warning: Could not find family Up and at 'em!, using TeX Gyre Heros Makie.
│ Additional parameters were Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}().
└ @ Makie ~/work/Makie.jl/Makie.jl/src/fonts.jl:148
Font search: Test Failed at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:671
  Expression: contains_warn(read(fname, String), $(Expr(:escape, "Could not find family Up and at 'em!")))
```

SimonDanisch:
As I said before, can we first concentrate on moving the Fontconfig functionality into Fontconfig? And actually turn it into a package we can depend/rely on?  Fontconfig needs at minimum decent cross platform tests for the functionality we want to rely on here and someone committing to maintain it.
Without that, we can never merge this PR and it will just go stale.

asinghvi17:
I just put Fontconfig through the Clang.jl generator process and it seems to have worked quite well, so it shouldn't be too difficult to wrap all its functionality a little more efficiently.  I also started the process of moving a lot of the code in this PR to Fontconfig, which seems like it shouldn't be terribly difficult.

Will put a PR up on Fontconfig with the results of the refactor.

SimonDanisch:
I'm closing this PR for now, so that it's clear that the work needs to happen in Fontconfig.jl first and foremost!
We can re-open and continue once Fontconfig is a well tested & maintained package.

PR #2645, opened 2023-02-01
Fixes for GeoMakie

SimonDanisch:
# Description

- Always applies the plot's transform func when rendering, not the Scene's
- Always applies transform_func in `data_limits`
- Applies inverse transform to obtained data limits in `getlimits(::Axis, dim)` by default, so that the returned limit is in input space.
- Fix `update_cam!` for Scene - it now uses `data_limits`, instead of `limits(scene)` which was deprecated. 

## Type of change

## Checklist


src/makielayout/blocks.jl
```diff
@@ -67,7 +67,7 @@ macro Block(name::Symbol, body::Expr = Expr(:block))
 
         function Makie.default_attribute_values(::Type{$(name)}, scene::Union{Scene, Nothing})
             sceneattrs = scene === nothing ? Attributes() : theme(scene)
-            curdeftheme = deepcopy(CURRENT_DEFAULT_THEME)
+            curdeftheme = deepcopy(Makie.CURRENT_DEFAULT_THEME)
```
asinghvi17, line 70:
```suggestion
            curdeftheme = deepcopy($(@__MODULE__).CURRENT_DEFAULT_THEME)
```
because it's not always guaranteed that the user will have loaded Makie directly

SimonDanisch: Ah I guess you're right! I went with `$(Makie)` to make it clearer that it's not some other module that should get picked up....

PR #2647, opened 2023-02-03
speedup heatmaps in WGLMakie

SimonDanisch:
...And other primitives, once we have a new JSServe release.


SimonDanisch:
- [x] clean up scenes + plots + three objects
- [x] make display more stable

PR #2650, opened 2023-02-03
Fix clearing in GLMakie

ffreyer:
# Description

An easy way to see the problem is to plot this:

```julia
fig = Figure()
ax = Axis3(fig[2, 1]) # either order of Axis3 and Menu
m = Menu(fig[1, 1], options = string.(1:3)) 
ax.elevation[] = -0.2
screen = display(fig)
```

On master:
Screenshot from 2023-02-03 16-28-05

On this branch (commit 4):
Screenshot from 2023-02-03 16-42-26

The reason this happens is that we don't have a strong scene order. Every plot gets drawn in order of insertion regardless of whether it's over or under a scene with `clear = true`. (Some stuff still renders below - I assume because of depth)

I tried two(-ish) fixes which both reintroduce the stencil buffer:
1. (Second commit) Go through all scenes, incrementing the stencil value  and writing to the stencil buffer whenever the scene clears. During rendering, discard all plots associated with a smaller stencil value. This solution is sensitive to the call order of `Menu` and `Axis3` and currently doesn't work with one of them.
2. (Fourth commit) Here each scene gets a stencil value based on its depth in the scene tree. A `clear = true` scene will discard plots from its parent(s) in this case. This is also kind of problematic, because one may want to add a `clear = true` scene to the root scene to draw over, for example, an `Axis3`. But because the axis is deeper in the scene tree, it will not be occluded by the new scene. Perhaps that could be remedied by making the stencil value a scene attribute?

## Type of change

## Checklist


ffreyer:
I think this needs some discussion on what the render order of scenes should actually be. We could have

### (1) Plain recursive ordering
```julia
function draw(scene)
    render(scene)
    draw.(scene.children)
end
```
This would render `A = root.children[1].children[1]` before `B = root.children[2]`. If A and B overlap, then A would be obstructed by B. This could be desirable (you could add an overlay to fig.scene) or not (A could be something like a dropdown Menu, B an Axis)

### (2) Depth sorted ordering
```julia
function marked_scenes(scene, depth = 0, list = Tuple{Int, Scene}[])
    push!(list, (depth, scene))
    foreach(scene -> marked_scenes(scene, depth = depth+1, list), scene.children)
    return list
end

function draw(root)
    scenelist = marked_scenes(root)
    sort!(scenelist, by = first)
    foreach(render, scenelist)
end
```
This would handle the case above in the opposite way. It would render B before A, meaning that A can obstruct B. 

### (3) Insertion order

I think this is more or less what GLMakie used to have before the stencil buffer removal (the "less" being that `destroy!` might be messing with that order, not sure). This would mean that whether A or B renders first depends on when the scenes are created/added, which sounds too arbitrary to me.

### (4) Explicit ordering

Another option would be to add a scene attribute to set the order explicitly. (I.e. with an integer.) That would give a lot of control over to Makie but I think would get confusing quickly. The attribute would naturally increment from scene to scene so it will be even harder to set reasonable values than with priority in observables.

As far as I can tell:
- CairoMakie does (1) on a plot basis, ignoring clear for nested scenes (always fails Menu example)
- GLMakie does (3), ignoring clear on master (always fails Menu example)
- WGLMakie does (1) with clearing (fails Menu depending on order)
- this pr does (2) but not cleanly

_Note that the ordering indices/values don't need to be UInt8 for GLMakie. It should be fine to only increment the stencil buffer value when we hit a new `clear = true` scene, so the number of `clear = false` scenes is arbitrary. I think technically it would even be fine to only increment the value when we draw over another cleared area with the same value._

jkrumbiegel:
I don't know if clear as an attribute makes so much sense, to me it was always more useful to just insert a rectangle at some depth as the background. Only the very last layer in glmakie needs a nontransparent color, all the others could in principle be transparent and just use background rectangles if that's wanted.

ffreyer:
Adding a background rectangle has the same issues though, plus the extra problem of having to pick fitting z values and being less efficient (extra draw call and can't skip rendering underneath).

jkrumbiegel:
The fitting z values yes, but I find it more logical than having a "clear" command that is effectively like an infinity z thing for everything below.

ffreyer:
Being reminded of this - I think the name `clear` is confusing but I don't think it's conceptually confusing.

Depending on how we choose things to work you can think of a Scene as a drawing surface. With `clear = false` it would be like a sheet of colored paper, with `clear = true` if would be more or less transparent foil. And these surfaces will get composed together when we draw a figure in some way.

Iirc CairoMakie considers every plot to be on a new transparent surface and draws them in that recursive order.

GLMakie, after removing the stencil buffer, first draw all opaque scene backgrounds and then pretends every plot is on the same surface, drawing them all based on depth.

The problem with z values is that there is technically speaking no z value you can pick that strictly above or below everything in another scene. There is nothing stopping me from translating my scatter plot to the max z value. And if I do that things will depend on float precision errors and call order. 

A more realistic example for this is anything 3D. In that case your plot can end up at all kinds of z values after transformation, including close to the minimum and maximum. Sure, it's unlikely to get to exactly the minimum/maximum depth, but you may also want to have some depth order in your other scene. Put a background a couple units behind your legend you have some space a 3d plot could creep into.

This is something we can fix by taking scene order seriously. (And it's also just inconsistent between backends now)

We could say for example:
- scenes within the same level share depth order
- a parent scene has an independent depth order from its children
- parent draw before children

Then you could still have multiple scenes with plots drawing in between each other (though I don't really see why you would want this), but also have a scene that is strictly a background to its children and scenes that are strictly on top of the others. 

I think conceptually we also have some fairly clear layers in Makie:
- Figure which isn't really used but if you wanted to add the windows xp background you'd probably do it here
- Blocks that don't overlay - Axis, Slider, etc (Menu base)
- overlay - axislegend, Menu dropdown (which probably should go into another layer to avoid fighting with axislegend)

ffreyer:
I'm planning to redo this in a different way

PR #2544, opened 2023-01-02
Scale windows and resolution for HiDPI displays

jmert:
# Description

On my HiDPI monitors [*], GLMakie's windows appear with a small default size and very small font size compared to other UI elements. Overriding themes to enlarge font sizes, default window size, default line thicknesses, etc is possible, but that makes sharing plotting scripts across machines difficult.

The approach I've taken here is to distinguish between a logical pixel and physical pixel, based on the fact that GLFW can report the device scaling factor in use. Therefore, the default window size, font sizes, marker/line sizes, etc can all be left alone (i.e. 800x600 window with 16pt fonts), and all lengths are dynamically rescaled by the device scaling factor during rendering.

With what I've hacked together, I can view images in GLMakie that have similar scale to those I'd get from saving an image to disk and loading in Firefox (which does similar HiDPI scaling from nominal "1x" CSS pixels to actual screen pixels):

```julia
julia> using CairoMakie; CairoMakie.activate!()

julia> x = range(-2π, 2π, 5_000); lines(x, sin.(2.1x))
FigureAxisPlot()

julia> save("/tmp/test.svg", current_figure())
CairoMakie.Screen{SVG}


julia> using GLMakie; GLMakie.activate!()

julia> x = range(-2π, 2π, 5_000); lines(x, sin.(2.1x))
```
image

**Important caveats**
- I removed the retina-specific code paths, since they seemed to do something very similar but in a way that doesn't work for my Linux desktop where window and framebuffer sizes are always equal. I don't have a way to test on Mac, so it was easier for me to rip out and just ignore it, but someone else will need to test on OSX.
- I don't actually know GLSL, so the rendering modifications were a big exercise in pattern matching and guess-and-check. My stupidly-simple sample cases seem to now work, but I won't be surprised if there are still missing modifications for recipes I haven't tried.
  - I also don't really understand the SSAO, FXAA, etc, so I've probably not updated those rendering steps correctly either.
- I've only tested the "on content scale change" on Windows, where it seemed to work moving a window from the laptop builtin display at 125% to an external display at 200% and back. (I haven't been able to test on Linux since (AFAIU) that requires GLFW using the Wayland backend, but it seems like GLFW_jll is going through XWayland instead.)

I've also discovered that #2529 now exists, which must have been created just after I'd starting working on this independently. I've borrowed the `px_per_unit` nomenclature and including that in the theme from that PR, but otherwise I think this has taken a different approach. That might be a clue as to differences between Linux and OSX...

[*] Machine details:
[...]

PR #2653, opened 2023-02-06
improve docs for GLMakie backend activation

jwahlstrand:
# Description

This is mostly wordsmithing in docs, docstrings, and comments. Also includes a format fix for an admonition and a description of the `visible` option that was added to ScreenConfig recently (#2587).

## Type of change

This only changes docs and fixes typos in a couple of comments.

## Checklist


GLMakie/src/screen.jl
```diff
@@ -11,29 +11,29 @@ function renderloop end
...
+* `float = false`: Whether the window should float above other windows.
+* `focus_on_show = false`: If `true`, focuses the window when newly opened.
+* `decorated = true`: Whether or not to show window decorations.
 * `title::String = "Makie"`: Sets the window title.
-* `fullscreen = false`: Starts the window in fullscreen.
-* `debugging = false`: Starts the GLFW.Window/OpenGL context with debug output.
-* `monitor::Union{Nothing, GLFW.Monitor} = nothing`: Sets the monitor on which the Window should be opened.
+* `fullscreen = false`: Whether to start the window in fullscreen mode.
+* `debugging = false`: If `true`, starts the GLFW.Window/OpenGL context with debug output.
+* `monitor::Union{Nothing, GLFW.Monitor} = nothing`: Sets the monitor on which the window should be opened. If set to `nothing`, GLFW will decide which monitor to use.
```
asinghvi17, line 29: Would be nice to add how you actually set the monitor here...I think there's some discussion of this in #2544.

GLMakie/src/screen.jl
```diff
@@ -629,8 +630,9 @@ function fast_color_data!(dest::Array{RGB{N0f8}, 2}, source::Texture{T, 2}) wher
 end
 
 """
-depthbuffer(screen::Screen)
-Gets the depth buffer of screen.
+    depthbuffer(screen::Screen)
+
+Gets the depth buffer of a screen.
```
asinghvi17, line 635:
```suggestion
Gets the depth buffer of `screen`.  Returns a `Matrix{Float32}` of the dimensions of the screen's `framebuffer`.  

A depth buffer is used to determine which plot's contents should be shown at each pixel.
```

asinghvi17 (commented):
Thanks for the PR!  I had a couple of minor comments, but it looks great, and I love when someone pays attention to the docs :)

Since #2544 is making some changes in this region as well, I'd like to wait to merge this until that is merged, so we don't face any annoying merge conflicts.

GLMakie/src/screen.jl
asinghvi17: Yeah, it looks like the current setting method doesn't do what you'd think it does (set the monitor on which the window should be displayed).

Since GLFW uses X directly, I guess the best way to do this would be to set `ENV["DISPLAY"]` before launching the window?!

SimonDanisch: No, I don't think that's a good idea. I'm not 100% sure if that even works, and even if it does on linux, it won't work on windows.
So, I think we should just fix it to allow passing a `GLFW.Monitor`

SimonDanisch:
Thank you! :)

Issue #718, opened 2020-10-09
Support for complex numbers?


clad26:
Makie.jl doesn't support complex numbers for data, yet?
Is there a way to plot a 100 x 2 data of complex numbers?

auberjine:
I'm not sure if plotting complex numbers actually makes sense, what one usually do is to plot either the real/imaginary parts, or the modulus/argument separately
What would you like to see when plotting complex numbers ?

clad26:
If I give a single vector of complex numbers as argument, I expect it to be plot on the complex plane(real part vs imaginary part) (with the axes getting displayed properly)
If I give more vectors/matrices as argument, I expect it to plot the modulus(behave like normally giving numbers)

THANKS FOR RESPONDING！

auberjine:
I think that's why complex number are not supported directly : depending on the application people expect different results ^^

In your first case you can do
`scatter(real(complex_array), imag(complex_array))`

And in your second
`heatmap(abs.(complex_matrix))`

clad26:
Figured that out
Although I think its better if people could do that with scatter(z) or scatter(z1,z2).
More appealing for the package🙂
Thanks anyway！

SimonDanisch:
Well, there isn't a problem, as long as the behavior depends on the type!

clad26:
why not use function overloading？
function plot(a::Complex)

end
somewhere in the Makie src or something like that
Works right？

SimonDanisch:
Yeah, that's an option:
http://makie.juliaplots.org/dev/recipes.html

Uroc327:
Is there more detailed documentation for `convert_arguments` available?
I had hoped I could implement this somehow using
```julia
convert_arguments(::Type{<:Scatter}, x::Complex) = (real(x), imag(x),)
```
but I wasn't able to get a call like `scatter([1+5im, 2+6im, 3+7im, 4+8im])` to work.

SimonDanisch:
Well, if you want to plot a vector of complex number, you need to overload the conversion for a vector of complex numbers.
This should work:
```julia
function convert_arguments(::Type{<:Scatter}, numbers::AbstractVector{<: Complex})
     return (map(x-> Point2f0(real(x), imag(x)), numbers),)
end
```

Uroc327:
Thanks so much! I assumed this was recursive.

Uroc327:
For further reference, I used 

```julia
[...]

PR #2654, opened 2023-02-06
Implement point-based converts for complex numbers

asinghvi17:
as discussed in #718.

# Description

Adds converts from:
`Vector{Complex{T}}` -> `Vector{Point2{T}}`
`Vector{Complex{T}}, Vector{Real}` -> `Vector{Point3{T}}` where the second input represents z-value.

## Type of change

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 
 - Fixed an issue where `poly` plots with `Vector{<: MultiPolygon}` inputs with per-polygon color were mistakenly rendered as meshes using CairoMakie. [#2590](https://github.com/MakieOrg/Makie.jl/pulls/2478)
 - Fixed a small typo which caused an error in the `Stepper` constructor. [#2600](https://github.com/MakieOrg/Makie.jl/pulls/2478)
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
Not sure if I like this, to be honest. Although I never really use complex numbers, so I don't suffer the manual conversion. To me, this feels a bit on the convenient side and not the clearly motivated side. For example, one could do the same with Dual numbers, which are also 2-element numbers, but for those it would feel even less appropriate. Not sure, maybe others can chime in.

asinghvi17:
I recently had to do complex numbers in a polar axis, for which I thought it was pretty neat.  Of course you can simply pass `real.(vec), imag.(vec)` but it's nice when plotting to quickly visualize results.

jkrumbiegel:
In Plots, it makes sense to me if the axis labels go to `real` and `imaginary` then, you'll know what happened. But this convert treats complex numbers basically the same as generic points, on the lowest level.

SimonDanisch:
Lets close this for now and maybe revisit once we can change axis labels in a plot!

PR #2656, opened 2023-02-07
More detail in docstring for `update_theme!`

KronosTheLate:
# Description

What it says on the box.


src/theming.jl
```diff
@@ -201,8 +201,23 @@ end
 """
     update_theme!(with_theme::Theme; kwargs...)
 
-Updates the current theme incrementally, that means only the keys given in `with_theme` or through keyword arguments are changed, the rest is left intact.
+Update the current theme incrementally. This means that only the keys given in `with_theme` or through keyword arguments are changed, 
+the rest is left intact.  
 Nested attributes are either also updated incrementally, or replaced if they are not attributes in the new theme.
+
+# Example
+To change de default colormap to `:greys`, you can pass that mapping as
```
asinghvi17, line 209:
```suggestion
To change the default colormap to `:greys`, you can pass that attribute as

File: src/theming.jl [hunk 2]
```diff
@@ -201,8 +201,23 @@ end
...
+
+# Example
+To change de default colormap to `:greys`, you can pass that mapping as 
+a keyword argument to `update_theme!` as demonstrated below.
+```
+update_theme!(colormap=:greys)
+```
+
+This can also be achieved by passing an object of types `Attributes` or `Theme` 
+as the first and only positional argument:
```
asinghvi17, line 216:
```suggestion
This can also be achieved by passing an object of types `Attributes`
as the first and only positional argument:
```
I think we should choose one of the two names (theme/attributes) and stick to it to avoid confusion, so I removed the reference to Theme.  It's up for discussion as to which we want to stick to, though.

src/theming.jl
```diff
@@ -201,8 +201,23 @@ end
...
+```
+update_theme!(Attributes(colormap=:greys))
+update_theme!(Theme(colormap=:greys))
```
asinghvi17, line 219:
```suggestion
update_theme!(Attributes(colormap=:greys))
```
similar to the last suggestion

src/theming.jl
KronosTheLate: Do you expect that discussion to conclude in a few days? If not, I suggest documenting both for completeness, and removing one once that discussion is settled.

asinghvi17 (approved):
Seems to be no consensus on this yet, so will merge and let's discuss at MakieCon!

PR #2658, opened 2023-02-07
reduce amount of objects compiled into sysimage

SimonDanisch:


jkrumbiegel:
I think the missing ref image here is because I updated the ref images from a PR that was a little older, which deleted one image that was newer already. So it has nothing to do with this PR

timholy:
Oh, yes, glad you're cleaning out items! I had to do some of the same work in ImageView recently.

PR #2688, opened 2023-02-16
Add scroll-to-pan

pbouffard:
# Description

Adds the ability to pan by using inputs from the scroll wheel. For trackpads (on Mac anyway) this is a two-finger drag which is used across many applications and for some users will feel more natural than holding the right button (trackpad: holding two-finger press) and dragging.

The default behavior isn't changed. The API for enabling the feature isn't the greatest but at least it's fairly simple. I updated the docs for how to do this in 2D with `register_interaction!` but don't see an appropriate equivalent place for describing how to enable it for 3D, which is something like:

```
cc.attributes.scroll_mod = Or(Makie.Keyboard.right_shift, Makie.Keyboard.left_shift)
```

## Type of change

## Checklist

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -297,10 +297,15 @@ function add_translation!(scene, cam::Camera3D)
     end
 
     on(camera(scene), scene.events.scroll) do scroll
-        if is_mouseinside(scene) && ispressed(scene, scroll_mod[])
+        if is_mouseinside(scene)
+          if ispressed(scene, scroll_mod[])
             zoom_step = (1f0 + 0.1f0 * zoomspeed[]) ^ -scroll[2]
             zoom!(scene, cam, zoom_step, shift_lookat[], cad[])
-            return Consume(true)
+          else
+            camspeed = compute_diff([scroll[1], scroll[2]])
+            translate_cam!(scene, cam, 10f0 * Vec3f(-camspeed[1], camspeed[2], 0f0))
+          end
+          return Consume(true)
         end
         return Consume(false)
     end
--- a/docs/examples/blocks/axis.md
+++ b/docs/examples/blocks/axis.md
@@ -919,6 +919,7 @@ You can zoom in an axis by scrolling in and out.
 If you press x or y while scrolling, the zoom movement is restricted to that dimension.
 These keys can be changed with the attributes `xzoomkey` and `yzoomkey`.
 You can also restrict the zoom dimensions all the time by setting the axis attributes `xzoomlock` or `yzoomlock` to `true`.
+Note that if scroll pan is enabled (see below), scroll zoom will require (left or right) `shift` to be pressed.
 
 ### Drag pan
 
@@ -926,6 +927,16 @@ You can pan around the axis by right-clicking and dragging.
 If you press x or y while panning, the pan movement is restricted to that dimension.
 These keys can be changed with the attributes `xpankey` and `ypankey`.
 You can also restrict the pan dimensions all the time by setting the axis attributes `xpanlock` or `ypanlock` to `true`.
+
+### Scroll pan
+
+Users of trackpads and mice with wheels (especially both horizontal and vertical) may like to pan using these as is typically possible in other programs. This is not enabled by default, but can be (on a per-axis basis) by calling `register_interaction`:
+
+```
+register_interaction!(ax, :scrollpan, GLMakie.Makie.ScrollPan(20.0, 0.2))
+```
+
+When this interaction is enabled, in order to not conflict, the scroll zoom will require (left or right) `shift` to be pressed.
 
 ### Limit reset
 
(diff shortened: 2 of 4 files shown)


src/camera/camera3d.jl
```diff
@@ -297,10 +297,15 @@ function add_translation!(scene, cam::Camera3D)
...
+        if is_mouseinside(scene)
+          if ispressed(scene, scroll_mod[])
             zoom_step = (1f0 + 0.1f0 * zoomspeed[]) ^ -scroll[2]
             zoom!(scene, cam, zoom_step, shift_lookat[], cad[])
-            return Consume(true)
+          else
+            camspeed = compute_diff([scroll[1], scroll[2]])
+            translate_cam!(scene, cam, 10f0 * Vec3f(-camspeed[1], camspeed[2], 0f0))
+          end
+          return Consume(true)
```
ffreyer, line 308: I think we should adjust this to be zoom without a scroll mod and translation with one. I think that makes more sense and matches up better with Axis.

pbouffard: Not sure I understand -- the way this is implemented if the scroll mod default of `true` is untouched then the `ispressed` call returns `true` and you get the existing behavior of zooming on scroll. Changing the scroll mod to something else for example as I described earlier is what enables "panning" (translating) with scrolling.

As far as matching with Axis goes, I'm not convinced that it always makes sense to have the behaviors match. Even in testing it out I'm not sure I'd always want to use scroll to translate in a 3D view even though I probably almost always would in a 2D view since the latter feels very natural. Still, it's nice to have the option either way.

pbouffard: I see -- so to put it in other terms this would basically be adding the ability to translate on scroll ("special behavior") by holding down a key, e.g. shift, in addition to the existing behavior of holding the right mouse button down to translate. Either way just using scroll without a modifier would work the same as it does today.

I have no issue with that, and honestly adding anything to the 3D side of things was a bit of an afterthought. 2D panning with scroll is really the main idea. So maybe it would be better to just remove the 3D part of the PR.

pbouffard: Update - I went ahead and removed the 3D camera changes. @ffreyer (and others) any further comments?

pbouffard:
bump .. any further thoughts on this? The main thing I'm not loving at this moment is how it is enabled. Personally I would always want scroll-to-pan to be the global default on a touchpad-equipped system but I'm not sure if at this point Makie has any mechanism for specifying such defaults, or if there are any plans in that regard.

Having said that, is the current state of the PR acceptable? It doesn't change any default behavior, at least AFAICT -- I don't really have a way to test it on systems other than my own which at present is just my touchpad Mac.

SimonDanisch (approved):
Works well for me, thank you :) @jkrumbiegel let me know if you want to review as well!

jkrumbiegel:
yes I need a bit of time, wasn't sure from reading if I agree with the implementation

jkrumbiegel:
Ok I think I agree with @ffreyer that the state of this PR is not good, no action by default for zoom (only with shift pressed).

Maybe it would be worth rethinking the whole thing. What are really the most common paradigms to use for interactions like this? While I think touchpads can be nice for panning, it doesn't map to mouse scrolling very well. I can't think of (m)any viewer-like apps right now that pan on scroll. That seems to be more of a browser / line-based thing. I think in viewer-like apps it's more common to pan using the cursor, often by keeping the space bar pressed at the same time (which then makes the cursor into a hand).

I actually don't like our current default with right-click dragging that much either (especially not two-finger click drag on a touchpad, that feels tedious). But I also don't like the solution in this PR 🤷 

What do you think about the space modifier? Are there other ideas in the running? Would be helpful to compare to other classic tools directly.

pbouffard:
So I would say with the trackpad, pan-to-scroll (without keyboard modifier) is by far the most common paradigm across apps, or at least on Mac systems (the only ones with really decent trackpads I've encountered, IMHO). So I'd definitely contend that it should be the default at least on those systems. Or at least a user preference if such a thing is possible, or failing that, something that I can enable in my apps.

Having to use a modifier to zoom, though, definitely doesn't feel natural. In the case of a trackpad that would generally be a pinch-zoom gesture; I'm not sure if that's possible in this case. But if so I would argue that (again with trackpad only and perhaps only a Mac) these mappings would be what is 'natural':

single click = select item (with modifier like shift or cmd, multi-select)
click (left button/single finger) and drag = box zoom (or with item(s) selected, move the item(s))
click (right button/two finger) = context (pop up) menu (app & context dependent)
click (right button/two finger) and drag = app dependent (perhaps for GLMakie pan to keep current behavior)
double click = drill down (in some sense; app & context dependent)
pinch zoom = zoom (with modifier, restrict to horizontal or vertical zoom)
two finger drag ("scroll") = pan (restrict to horizontal or vertical with modifier)

pbouffard:
Also, perfect-world comments aside, if I had to choose between a) right-click-dragging and b) keyboard modifier (preferably spacebar, easier to find) plus "scroll" to pan, I'd definitely prefer b). Maybe it doesn't have to be either-or. If b) doesn't do anything currently then you could just add that to the current behavior of a), which might be preferable for the existing user base.

PR #2691, opened 2023-02-17
Fix color conversion in scatterlines

ffreyer:
# Description

Fixes https://discourse.julialang.org/t/makie-markercolor-magic/94760

Just removes the strict typing to colortypes so we can also use numeric vectors or whatever else here. 

## Type of change

## Checklist


SimonDanisch:
We should use `RGBColors` for that, which contains `Vector{<: Number}` ... No need to change this to `Any`, since backends & friends will only accept `RGBColors` ;) 
I pushed the change directly to the branch!

PR #2703, opened 2023-02-21
Improve Block deletion

jkrumbiegel:
I tried to just compare the listeners of most (all?) relevant observables that a fresh `Scene` has with those after a `Block` object has been added and `delete!`d. This way I found all observables that needed to be cleaned up for each `Block`.

While it seems that this correctly cleans up such that new objects in the place of deleted ones work correctly, I have the feeling it has also introduced some instability. Or maybe just uncovered previously existing instability that didn't show because nobody `delete!`d stuff and continued interacting with the `Scene` after.

Fixes ...

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks.jl
+++ b/src/makielayout/blocks.jl
@@ -507,25 +507,28 @@ function Base.delete!(block::Block)
     for finalizer in block.finalizers
         finalizer()
     end
-
-    block.parent === nothing && return
     
     # detach plots, cameras, transformations, px_area
-    empty!(block.blockscene)
+    empty!(block.blockscene) # TODO this hangs
 
-    s = get_topscene(block.parent)
-    deleteat!(
-        s.children,
-        findfirst(x -> x === block.blockscene, s.children)
-    )
-    # TODO: what about the lift of the parent scene's
-    # `px_area`, should this be cleaned up as well?
+    s = block.blockscene.parent
+
+    if s !== nothing
+        deleteat!(
+            s.children,
+            findfirst(x -> x === block.blockscene, s.children)
+        )
+    end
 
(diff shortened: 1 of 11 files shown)


src/makielayout/blocks.jl
```diff
@@ -16,6 +16,7 @@ macro Block(name::Symbol, body::Expr = Expr(:block))
             parent::Union{Figure, Scene, Nothing}
             layoutobservables::Makie.LayoutObservables{GridLayout}
             blockscene::Scene
+            finalizers::Vector{Any}
```
SimonDanisch, line 19: In JSServe, I started having these signatures:
```
map(f, session, args...)
on(f, session, obs)
onany(f, session, obs)
```
I was wondering if we shouldn't have this as the main API to bind observables to the lifecycle of another object?
Not sure if we can make it an official Observables API, but i think we should definitely add:
```julia
map(f, scenelike, args...)
on(f, scenelike, obs)
onany(f, scenelike, obs)
```
And use that in place where you now have the `x = on(...); push!(ax.finalizers, x)`

jkrumbiegel: That depends on what you mean by `lifecycle`, because I don't think it should mean garbage collection. With all our cross-connections between objects, it's likely that a `delete!`d `Axis` for example does not get garbage collected for a long time, if ever. The behavior when calling `delete!` should be entirely predictable.

SimonDanisch:
Probably should also try this with https://github.com/MakieOrg/Makie.jl/pull/2682, where I try to improve clean up of scenes as well...

jkrumbiegel:
superseded by #2731

PR #2706, opened 2023-02-23
Interpolate Makie in the `@Block` macro

asinghvi17:
Allows `@Block` to work in any scenario regardless of whether Makie was imported or not (eg if someone only imported CairoMakie).  With this, defining a `@Block` in an external package works well.

## Type of change

### DIFF
--- a/src/makielayout/blocks.jl
+++ b/src/makielayout/blocks.jl
@@ -13,8 +13,8 @@ macro Block(name::Symbol, body::Expr = Expr(:block))
 
     structdef = quote
         mutable struct $name <: Makie.Block
-            parent::Union{Figure, Scene, Nothing}
-            layoutobservables::Makie.LayoutObservables{GridLayout}
+            parent::Union{$(Makie).Figure, $(Makie).Scene, Nothing}
+            layoutobservables::$(Makie).LayoutObservables{$(Makie).GridLayout}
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
This doesn't look correct to me, should work without interpolation. It's probably better to check the escaping logic, maybe too much is escaped that doesn't need to be. Not sure

asinghvi17:
Should the statement be run in the Makie module or the user's module?  The latter is probably better for stacktraces, but the former lends a bit more convenience.

jkrumbiegel:
wherever the macro is called, but its symbols should be default refer to the Makie module if they are not escaped, which they shouldn't be I guess

asinghvi17:
Well if we escape it then the stacktrace points to where the user wrote `@Block`, and if we don't then it points to Makie's definition of `@Block` instead.

The only scenario in which this wouldn't work is if someone loads only a backend package (GLMakie, etc) and not Makie.  Since we recommend this in the docs, I think it makes sense to interpolate Makie just in case.

SimonDanisch:
Is this still needed after https://github.com/MakieOrg/Makie.jl/pull/2645?

asinghvi17:
Will check and get back to you!

SimonDanisch:
I think most should be fixed in #2645

PR #2708, opened 2023-02-23
Let the "glyphcollection" observable be a Vector{Any} in text

asinghvi17:
# Description

Allows arbitrary text input types to be dispatched on.
With this, I should be able to re-integrate MakieTeX, with the caveat (as before) that texts have to be initialized as LaTeXStrings or a TeX type in order to work with MakieTeX.

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/text.jl
+++ b/src/basic_recipes/text.jl
@@ -1,7 +1,7 @@
 function plot!(plot::Text)
     positions = plot[1]
     # attach a function to any text that calculates the glyph layout and stores it
-    glyphcollections = Observable(GlyphCollection[])
+    glyphcollections = Observable(Any[])
     linesegs = Observable(Point2f[])
     linewidths = Observable(Float32[])
     linecolors = Observable(RGBAf[])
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
No I don't think that's a good way forward, you'd have to split up the paths earlier than that and continue separately. It doesn't make sense to just keep everything called GlyphCollection and pass something arbitrary through.

asinghvi17:
There is no way to split the paths earlier than this though?  This is the "lowest common denominator" for any text signature, and if a user (or a recipe, which is the primary issue) uses this plot command, you can't dispatch on the type of the text no matter what you do.

jkrumbiegel:
then a new level would have to be added above that, at least no matter which way it has to be cleaner than this

asinghvi17:
What do you think of the following:

1. All text decomposes to the signature which is currently used for layouting
2. That signature passes a `Vector{Tuple{Point, T}}` down to a lower level text signature which does the layouting
3. This lower level signature can then be extended arbitrarily; the Makie implementation would support `T <: Union{AbstractString, Makie.Rich}` for interoperability.

I don't know if this would work with the plot type system, but don't mind giving it a shot.

SimonDanisch:
Lets close this for now until we have a better idea for our text interface!

PR #2712, opened 2023-02-26
Add weight support to hexbin

asinghvi17:
# Description

Allows a `weights` keyword in hexbin, which supports `nothing` or `StatsBase.UnitWeights` (each observation gets a weight of one) or any `AbstractVector{<: Real}` or `StatsBase.AbstractWeights` which provide arbitrary weights per observation.  These values are stored as Float64 in the Dict.

exp-scaling
Weights with scaling

## Type of change

## Checklist


asinghvi17:
This should be good to go now.

PR #2716, opened 2023-02-27
Default `fig` to `current_figure()` for `resize!` and `resize_to_layo…

KronosTheLate:
# Description

What it says on the box.

## Type of change

## Checklist

### DIFF
--- a/src/figures.jl
+++ b/src/figures.jl
@@ -130,6 +130,7 @@ end
 
 """
     resize_to_layout!(fig::Figure)
+    resize_to_layout!()
 
 Resize `fig` so that it fits the current contents of its top `GridLayout`.
 If a `GridLayout` contains fixed-size content or aspect-constrained
@@ -138,6 +139,8 @@ differs from the size of the `Figure`. This can result in superfluous
 whitespace at the borders, or content clipping at the figure edges.
 Once resized, all content should fit the available space, including
 the `Figure`'s outer padding.
+
+If no figure is provided, `fig` defaults to `current_figure()`.
 """
 function resize_to_layout!(fig::Figure)
     # it is assumed that all plot objects have been added at this point,
@@ -150,7 +153,9 @@ function resize_to_layout!(fig::Figure)
     resize!(fig.scene, widths(bbox)...)
     new_size
 end
+resize_to_layout!() = resize_to_layout!(current_figure())
 
+# No default to `current_figure()` added, as `Base` owns this function
 function Base.empty!(fig::Figure)
     empty!(fig.scene)
     empty!(fig.scene.events)
@@ -165,7 +170,12 @@ end
 # Layouts are already hooked up to this, so it's very simple.
 """
     resize!(fig::Figure, width, height)
+    resize!(width, height)
+
 Resizes the given `Figure` to the resolution given by `width` and `height`.
 If you want to resize the figure to its current layout content, use `resize_to_layout!(fig)` instead.
+
+If no figure is provided, `fig` defaults to `current_figure()`.
 """
 Makie.resize!(figure::Figure, args...) = resize!(figure.scene, args...)
+Makie.resize!(args...) = resize!(current_figure().scene, args...)  # Less specific type, so definition above is used if figure is passed.


KronosTheLate:
I updated the branch to master in the second commit. Is that stupid?

src/figures.jl
```diff
@@ -130,6 +130,7 @@ end
 
 """
     resize_to_layout!(fig::Figure)
+    resize_to_layout!()
```
asinghvi17, line 133:
```suggestion
    resize_to_layout!(fig::Figure = current_figure())

File: src/figures.jl [hunk 2]
```diff
@@ -138,6 +139,8 @@ differs from the size of the `Figure`. This can result in superfluous
 whitespace at the borders, or content clipping at the figure edges.
 Once resized, all content should fit the available space, including
 the `Figure`'s outer padding.
+
+If no figure is provided, `fig` defaults to `current_figure()`.
 """
 function resize_to_layout!(fig::Figure)
```
asinghvi17, line 145:
```suggestion
function resize_to_layout!(fig::Figure = current_figure())
```

src/figures.jl
```diff
@@ -150,7 +153,9 @@ function resize_to_layout!(fig::Figure)
     resize!(fig.scene, widths(bbox)...)
     new_size
 end
+resize_to_layout!() = resize_to_layout!(current_figure())
```
asinghvi17, line 156:
```suggestion
```

so on and so forth.

src/figures.jl
```diff
@@ -165,7 +170,12 @@ end
...
     resize!(fig::Figure, width, height)
+    resize!(width, height)
+
 Resizes the given `Figure` to the resolution given by `width` and `height`.
 If you want to resize the figure to its current layout content, use `resize_to_layout!(fig)` instead.
+
+If no figure is provided, `fig` defaults to `current_figure()`.
 """
 Makie.resize!(figure::Figure, args...) = resize!(figure.scene, args...)
+Makie.resize!(args...) = resize!(current_figure().scene, args...)  # Less specific type, so definition above is used if figure is passed.
```
asinghvi17, line 181: This is actually `Base.resize!`, so providing this general definition is probably not a good idea.

src/makielayout/blocks/axis.jl
```diff
@@ -918,10 +918,14 @@ end
 
 """
     autolimits!(la::Axis)
+    autolimits!()
```
asinghvi17, line 921:
```suggestion
```

src/figures.jl
KronosTheLate: Huh. So this is actually type piracy? I'll remove the line.

KronosTheLate:
Thanks @asinghvi17 for the double-checking. It makes it easy to make PR's without perfecting them first, which makes me actually get around to making them ^_^

src/figures.jl
asinghvi17: Yeah, looks like it...

src/makielayout/blocks/axis.jl
```diff
@@ -919,9 +919,12 @@ end
 """
     autolimits!(la::Axis)
 
-Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the data limits of all plot objects in the axis, as well as the autolimit margins for x and y axis.
+Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the
```
jkrumbiegel, line 922: while we're at it, you can change `la` to `ax` here, `la` is still from `LAxis` times

src/makielayout/blocks/axis.jl
```diff
@@ -919,9 +919,12 @@ end
...
     autolimits!(la::Axis)
 
-Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the data limits of all plot objects in the axis, as well as the autolimit margins for x and y axis.
+Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the 
+data limits of all plot objects in the axis, as well as the autolimit margins for x and y axis.
+
+If no axis is provided, `la` defaults to `current_axis()`.
 """
-function autolimits!(ax::Axis)
+function autolimits!(ax::Axis=current_axis())
```
jkrumbiegel, line 927: I'm not sure if this makes sense, because `current_axis()` could potentially return an axis object that doesn't have the `autolimits!` function implemented. The definition of what can be `current_axis()` is very loose, it's just whatever defines `Makie.can_be_current_axis` to be `true`

src/figures.jl
```diff
@@ -165,6 +168,8 @@ end
 # Layouts are already hooked up to this, so it's very simple.
 """
     resize!(fig::Figure, width, height)
+    resize!(width, height)
```
jkrumbiegel, line 171: remove this part

jkrumbiegel:
Yeah actually just remove everything but the `resize_to_layout!` overload, please. I think for that one it's reasonable to default to the current figure, for the others I'm quite ok with specifying the input arguments. For `resize_to_layout!`, I guess it's ok because it will be common enough to just use at the end of some workflow.

src/figures.jl
```diff
@@ -165,6 +168,8 @@ end
...
+
```
asinghvi17, line 172:
```suggestion

**Comment by asinghvi17** [comment]
Merged master into this branch and also removed all non-`resize_to_layout!` changes.

File: src/makielayout/blocks/axis.jl [hunk 10]
```diff
@@ -918,21 +918,15 @@ function update_linked_limits!(block_limit_linking, xaxislinks, yaxislinks, tlim
...
 """
 function autolimits!(ax::Axis)
     ax.limits[] = (nothing, nothing)
     return
 end
-function autolimits!()
-    curr_ax = current_axis()
-    isnothing(curr_ax)  &&  throw(ArgumentError("Attempted to call `autolimits!` on `current_axis()`, but `current_axis()` returned nothing."))
-    autolimits!(curr_ax)
-end
```
asinghvi17, line 935:
```suggestion
function autolimits!()
    curr_ax = current_axis()
    isnothing(curr_ax)  &&  throw(ArgumentError("Attempted to call `autolimits!` on `current_axis()`, but `current_axis()` returned nothing."))
    autolimits!(curr_ax)
end
```

src/makielayout/blocks/axis.jl
```diff
@@ -918,21 +918,15 @@ function update_linked_limits!(block_limit_linking, xaxislinks, yaxislinks, tlim
 end
 
 """
-    autolimits!()
-    autolimits!(la::Axis)
+    autolimits!(ax::Axis)
```
asinghvi17, line 921:
```suggestion
    autolimits!()
    autolimits!(ax::Axis)

File: src/makielayout/blocks/axis.jl [hunk 12]
```diff
@@ -918,21 +918,15 @@ function update_linked_limits!(block_limit_linking, xaxislinks, yaxislinks, tlim
...
-Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the data limits of all plot objects in the axis, as well as the autolimit margins for x and y axis.
-The argument `la` defaults to `current_axis()`.
+Reset manually specified limits of `ax` to an automatically determined rectangle, that depends on the data limits of 
+all plot objects in the axis, as well as the autolimit margins for x and y axis.
 """
 function autolimits!(ax::Axis)
     ax.limits[] = (nothing, nothing)
     return
 end
-function autolimits!()
```
SimonDanisch, line 931: Am I missing something or is just deleting this function breaking?

PR #2725, opened 2023-03-04
Remove offset in `show_data` for `scatter`

### LINKED ISSUE #2724: Inspector: tooltip offset in `show_data` for `Scatter` assumes scalar `markersize`
Hovering over a point in the following scatter plot
```julia
p = scatter(Point.(rand(SVector{3,Float32}, 10)), markersize = [Vec2f(100*rand()) for _ in 1:10]); DataInspector(); p
```
errors with
```
Error in callback:
MethodError: no method matching +(::Vector{Vec2{Float64}}, ::Int64)

Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:578
  +(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}
   @ Base int.jl:87
  +(::Distributions.Cauchy, ::Real)
   @ Distributions ~/.julia/packages/Distributions/w3oji/src/univariate/continuous/cauchy.jl:104
  ...

Stacktrace:
 [1] show_data(inspector::DataInspector, plot::Scatter{Tuple{Vector{Point{2, Float32}}}}, idx::UInt32)
   @ Makie ~/.julia/packages/Makie/O4ktt/src/interaction/inspector.jl:504
```

The reason is that this line assumes that `plot.markersize[]` is a scalar, i.e. the same for all markers
https://github.com/MakieOrg/Makie.jl/blob/6e27645a476123f3e1a866cd4d453ac37a3f3de4/src/interaction/inspector.jl#L504

The purpose of that line is to shift the tooltip label a little bit away from the marker. 

In my opinion, this line should be simply removed. That would solve the present issue and would also makes the tooltip point more precisely to the hovered maker. For comparison, tooltips in `MeshScatter` do not apply such an offset to the tooltip. 

This is the tooltip with the current offset (using a scalar markersize):
Screenshot 2023-03-04 at 19 11 06

This is the tooltip without the offset (as proposed here)
Screenshot 2023-03-04 at 19 10 53

This is how `meshcatter` does it (no offset)
Screenshot 2023-03-04 at 19 11 34

pablosanjose:
# Description

Fixes #2724 

The tooltip offset in the `show_data` method for `scatter` plots (relevant when using `DataInspector`) is removed.
As it is in master, the offset assumes all markers to be of the same size, so it errors when that is not the case.
In contrast `meshscatter` uses no tooltip offset, and hence does not suffer from this issue. This PR brings `scatter` behavior in line with `meshscatter`. Arguably, the offset only makes the tooltip less precise spatially in busy plots, so this is also an improvement in tooltip clarity in my opinion.

## Type of change

## Checklist

As a bugfix with no significant change in behavior, no additional actions apply, correct?

### DIFF
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -509,8 +509,8 @@ function show_data(inspector::DataInspector, plot::Scatter, idx)
         tt.text[] = position2string(plot[1][][idx])
     end
     tt.offset[] = ifelse(
-        a.apply_tooltip_offset[], 
-        0.5 * sv_getindex(plot.markersize[], idx) + 2, 
+        a.apply_tooltip_offset[],
+        0.5 * maximum(sv_getindex(plot.markersize[], idx)) + 2,
         a.offset[]
     )
     tt.visible[] = true
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -314,6 +314,8 @@ returning a label. See Makie documentation for more detail.
 - `enable_indicators = true)`: Enables or disables indicators
 - `depth = 9e3`: Depth value of the tooltip. This should be high so that the
     tooltip is always in front.
+- `apply_tooltip_offset = true`: Enables or disables offsetting tooltips based 
+    on, for example, markersize.
 - and all attributes from `Tooltip`
 """
 function DataInspector(fig_or_block; kwargs...)
@@ -332,6 +334,7 @@ function DataInspector(scene::Scene; priority = 100, kwargs...)
         depth = pop!(attrib_dict, :depth, 9e3),
         enable_indicators = pop!(attrib_dict, :show_bbox_indicators, true),
         offset = get(attrib_dict, :offset, 10f0),
+        apply_tooltip_offset = pop!(attrib_dict, :apply_tooltip_offset, true),
 
         # Settings for indicators (plots that highlight the current selection)
         indicator_color = pop!(attrib_dict, :indicator_color, :red),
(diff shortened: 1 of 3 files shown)


ffreyer (approved):
I added that bit because I thought it was nicer to be able to see the full marker. (Especially if it's something like an image scatter.) It's harder to get the same behavior with meshscatter because of 3D, so I didn't bother there. But this is ultimately just styling, you can argue either way so I'm fine with this change.

SimonDanisch:
Should we make this configurable?

ffreyer:
I guess we could? Something like
```julia
tt.offset[] = ifelse(a.should_offset_tooltip[], 0.5 * _offset_at(plot.markersize[], idx) + 2, a.offset[])
```
to reuse code from text layouting?

pablosanjose:
@ffreyer, that could work, thanks for the suggestion. However, I'm not sufficiently familiar with the codebase to attempt adding a new `:should_offset_tooltip`, could you do this?

Note also that the approach used to deal with attribute vectors in some other places look like this:
https://github.com/MakieOrg/Makie.jl/blob/982a0b75702889b3a354c087551ca37aeb15567f/src/interaction/inspector.jl#L586

ffreyer:
`_offset_at` wasn't general enough so I added and replaced it with `sv_getindex`. We can merge this once ci passes

pablosanjose:
Marvellous, thank you @ffreyer!
How long would we need to wait, approx, to get a new release tag with this once it is merged?

pablosanjose:
Oh, wait, this still fails:
```
p = scatter(Point.(rand(SVector{2,Float32}, 10)), markersize = [Vec2f(100*rand()) for _ in 1:10]); DataInspector(); p
```

We need to admit `Vector{<:Vec2f}` for markersize too, right?

SimonDanisch:
>Oh, wait, this still fails:

Oh :D Can we fix this in a new PR?

pablosanjose:
No, don't worry, the last commit I pushed fixes that too. T'is all good. Just looking forward to a tag to use this in a dependent project :-)

PR #2740, opened 2023-03-09
Call normalize with qualified name

### LINKED ISSUE #2738: Name clash in `plot!` for arrow plots
When trying to normalize an arrow plot in Makie version `v0.19.2`, I get the error
```
ERROR: MethodError: objects of type Observable{Any} are not callable
```

Minimal steps to reproduce:
```julia
arrows!(ax12, xs´, ys´, u, v;
    normalize=true,
)
```

Looking at the history, I can see that at some point `StaticArrays.normalize` was changed to just `normalize` here:

https://github.com/MakieOrg/Makie.jl/blob/0cbae0437a60e5c1a41c1114aabf8a1440e324a9/src/basic_recipes/arrows.jl#L128

Which now makes the name clash with the name of the parameter.

This looks simple enough to fix, so I could try doing it myself. I'd be happy to contribute but I just started learning Julia two days ago so I might need some help.

steffenhaug:
# Description

Fixes #2738

Qualifies the call to `normalize` with `LinearAlgebra` to avoid name clash with the `@extract`ed parameter of the same name.

## Type of change

In the past it was qualified with `StaticArrays`, however that is no longer a dependency of the library apparently which is probably why it was changed to break in the first place. The `StaticArrays` one is just a reexport of the `LinearAlgebra` one anyways as far as I can tell, so this should be identical to the old behavior.

Do you usually make tests for things like this? I'm not familiar enough with the internals of Makie yet to know how to verify programmatically that the data on the plot has actually been normalized, however I tested the fix on my use case and it works both in CairoMakie and GLMakie. If you want a test I can probably look into it with some help.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 - Fixed issue with scatterlines only accepting concrete color types as `markercolor` [#2691](https://github.com/MakieOrg/Makie.jl/pull/2691)
 - Fixed an accidental issue where `LaTeXStrings` were not typeset correctly in `Axis3`. [#2558](https://github.com/MakieOrg/Makie.jl/pull/2588)
 - Fixed a bug where line segments in `text(lstr::LaTeXString)` were ignoring offsets. [#2668](https://github.com/MakieOrg/Makie.jl/pull/2668)
+- Fixed a bug where the `arrows` recipe accidentally called a `Bool` when `normalize = true`. [#2740](https://github.com/MakieOrg/Makie.jl/pull/2740)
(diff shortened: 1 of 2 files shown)


steffenhaug:
I don't really understand the details of this check failing. I ran the test suite locally with Julia 1.8 and have tested making normalized arrow plots with all three backends. Is this check testing with Julia 1.0, and is that expected to work?

I guess what I want to know is, is there anything I should do now to get this in shape to merge?

SimonDanisch:
No this failure isn't related to your change..

Issue #2349, opened 2022-10-20
Add option to close an `Axis3`'s outline box


fatteneder:
As requested on Discord, implement the `ax.BoxStyle = 'full';` option from `MATLAB` here: https://www.mathworks.com/help/matlab/ref/box.html

I think that might be a good first issue. Could somebody please add a tag?

lukashappe:
# Description
I started working on this. So far I can close the outlinebox after adding fields for the color of Spine 4 to the Axis3-type and modifying  add_gridlines_and_frames! to optinaly use either 6 or 8 Points to make the Framelines.  I controll this by setting and the field fullbox I also added to the Axis3-type. 
Now first this seems to work and clos the outlinebox. But when I put data inside the box the new spines are obviously still shown behind the data. 
## Question
Can someone give me a hint how tho achieve that the front spines are shown above the contents of the box?

## Changes in the code
```
#in types.jl
@Block Axis3 <: AbstractAxis begin
...
"Controls if the 4. Spines are created to close the outline box"
fullbox = false
"The color of x spine 4"
xspinecolor_4 = :black
"The color of y spine 4"
yspinecolor_4 = :black
"The color of z spine 4"
zspinecolor_4 = :black
"The x spine width"
...


#in axis3d.jl
function add_gridlines_and_frames!(topscene, scene, ax, dim::Int, limits, ticknode, miv, min1, min2, xreversed, yreversed, zreversed)
...
 framepoints = lift(limits, scene.camera.projectionview, scene.viewport, min1, min2, xreversed, yreversed, zreversed
            ) do lims, _, pxa, mi1, mi2, xrev, yrev, zrev
        o = pxa.origin

        rev1 = (xrev, yrev, zrev)[d1]
        rev2 = (xrev, yrev, zrev)[d2]

        mi1 = mi1 ⊻ rev1
        mi2 = mi2 ⊻ rev2

        f(mi) = mi ? minimum : maximum
        p1 = dpoint(minimum(lims)[dim], f(!mi1)(lims)[d1], f(mi2)(lims)[d2])
        p2 = dpoint(maximum(lims)[dim], f(!mi1)(lims)[d1], f(mi2)(lims)[d2])
        p3 = dpoint(minimum(lims)[dim], f(mi1)(lims)[d1], f(mi2)(lims)[d2])
        p4 = dpoint(maximum(lims)[dim], f(mi1)(lims)[d1], f(mi2)(lims)[d2])
[...]

PR #2742, opened 2023-03-10
Add option to close an axis3's outline box

Capri-Corni:
# Description

Fixes # (issue #2349)

Started implementing a 3D box around the graph. I'm stuck at the moment because I got the following Error message:

```
ERROR: LoadError: Buffer vertex does not have the same length as the other buffers.
        color has length 3
        vertex has length 8
```

## Type of change

- I located the right passage and removed the # before points p7 and p8.
```
p7 = dpoint(minimum(lims)[dim], f(!mi1)(lims)[d1], f(!mi2)(lims)[d2])
p8 = dpoint(maximum(lims)[dim], f(!mi1)(lims)[d1], f(!mi2)(lims)[d2])
```
- Additionally, I added: `to_topscene_z_2d.([p1, p2, p3, p4, p5, p6, p7, p8])`
- I tried implementing 3 new spine colors

Do you have any suggestions for my next steps?


jkrumbiegel:
If there are 3 colors and 8 vertices that means that the new color for the last bit was not passed to the `linesegments` plot object. Linesegments usually have twice as many vertices as colors, one color for each segment.

ffreyer:
Sorry for never getting to this. Someone else implemented this in #4305 which ended up getting merged

PR #2755, opened 2023-03-14
Add Dendrogram recipe

### LINKED ISSUE #398: Heatmaps (clustering), dendrograms/trees
I'd love to see heatmaps support Clustering.jl's Hierarchical Clustering, sort of something like R's own `heatmap.2` (https://www.rdocumentation.org/packages/gplots/versions/3.0.1.1/topics/heatmap.2) where the clustering can be plotted as dendrograms in addition to the heatmaps. I imagine something like this could be done with layouts and line primitives, but I'm only just getting started on Makie and am not sure if that's possible.

kescobo:
# Description

Fixes #398 

I've begun working on a [dendrogram](https://en.wikipedia.org/wiki/Dendrogram) recipe.
The primary goal is to be able to plot the output of hclust objects from Clustering.jl,
and I made a recipe for StatsPlots.jl that did that a few years ago, but here,
I thought I would try to make something a bit more generic.

## Design

My initial thought is that a dendrogram is essentially a graph structure - to represent it in a plot,
you need nodes with (x, y) coordinates, and links between them.
But a dendrogram is a special-case, since it is a directed acyclic graph with a single root node,
and nodes can have either 0 children (leaf nodes) or 2 children (all others).

This makes plotting relatively simple, since you can start with the root node,
and then recursively draw children until you get to the leaves.

This is still very much work-in-progress, its not even a recipe yet, but I just wanted to throw  it up to solicit feedback before I go too far down the path in the wrong direction.

My initial design is as follows:

- a struct `DNode` that has a node index, `x` and `y` coordinates, and a `Tuple` of child indexes (or `Nothing` if a leaf)
- a function `find_merge` that takes 2 `(x,y)` tuples or 2 `DNode`s, and finds the `x` and `y` coordinates where their parent node should be
- `get_tree_coordinates()` and `get_box_coordinates()` to get the shapes for lines connecting a series of nodes (see below). The idea here is that there may be different ways to connect nodes (in the future, one might one a radial tree for example, and hopefully this would be extensible)
- `recursive_draw_dendrogram!()` which takes a dictionary of nodes, and a starting node, and does the drawing

I also included an `hcl_nodes()` for a proof-of-concept on how one might pull the right structure from a `HClust` - I do not anticipate pulling in Clustering in as a dependency, but rather would add the recipe there, or else make a separate `MakieDendrograms` package or something.

### Showcase

<details>
<summary>Basic dendrograms</summary>

```julia
leaves = Point2f.([
    (1,0),
    (2,0.5),
    (3,1),
    (4,2),
    (5,0)
])

merges = [
    (1, 2), # 6
    (6, 3), # 7
    (4, 5), # 8
    (7, 8), # 9
]

dendrogram(leaves, merges)
dendrogram(leaves, merges; branch_shape = :box)
```

</details>

 

<details>
<summary>Converting from HClust</summary>

```julia
using Clustering
using Distances

n = 20

mat = zeros(Int, n, n)

for i in 1:n
   last = minimum([i+Int(floor(n/5)), n])
   for j in i:last
       mat[i,j] = 1
   end
end

mat = mat[:, randperm(n)]

dm = pairwise(Euclidean(), mat; dims=2)
hcl = hclust(dm; linkage=:average, branchorder=:optimal)

heatmap(mat[:,hcl.order]) # see https://github.com/JuliaStats/Clustering.jl/pull/170

nodes = get_hcl(hcl)
dendrogram(nodes)
dendrogram(nodes; branch_shape=:box)
```

</details>

 

## Planned features

- [ ] Specify heights of parent nodes. Right now, each parent is exactly 1 higher than the highest child
- [ ] Enable grouping. Because of the structure, it should be possible to specify color groups, and color branches if all leaves belong to the same group.

## Help / Feedback requested

- I think I need some guidance on how to plug into the recipe system. I started from the boxplot recipe because I thought I understood all of those components, but I really don't understand the internals here. I plan to do a bunch of reading this week to see if I can get a further handle on it, but any pointers would be appreciated.
- Does this make sense to people? A different idea I had was to do something like use the `z` coordinate as a node index, but this seems like abusing the idea of `z`, and one would need to specify connections separately anyway.
- Would it be better to just try a PR to `GraphMakie.jl` that provides a conversion from `HClust` to a `Graphs.jl` graph? Part of me things that a primitive representation would be useful, but another part is worried I'm inventing a chiseled-rock wheel when Firestone tires exist.

## Type of change

## Checklist

### DIFF
--- a/src/stats/dendrogram.jl
+++ b/src/stats/dendrogram.jl
@@ -70,7 +70,7 @@ function get_box_connectors(parent, child1, child2)
     return [(x1, child1.y), (x1, yp), (x2, yp), (x2, child2.y)]
 end
 
-function recurssive_draw!(ax, node, nodes; branch_shape=:tree)
+function recursive_draw_dendrogram!(ax, node, nodes; branch_shape=:tree)
     isnothing(node.children) && return nothing
     child1 = nodes[node.children[1]]
     child2 = nodes[node.children[2]]
@@ -80,8 +80,8 @@ function recurssive_draw!(ax, node, nodes; branch_shape=:tree)
     
     lines!(ax, l)
 
-    recurssive_draw!(ax, child1, nodes; branch_shape)
-    recurssive_draw!(ax, child2, nodes; branch_shape)
+    recursive_draw_dendrogram!(ax, child1, nodes; branch_shape)
+    recursive_draw_dendrogram!(ax, child2, nodes; branch_shape)
     return nothing
 end
 
@@ -101,7 +101,7 @@ end
 
 function dendrogram(nodes::Dict{Int, DNode}; branch_shape=:tree)
     ax = Axis(Figure()[1,1])    
-    recurssive_draw!(ax, nodes[maximum(keys(nodes))], nodes; branch_shape)
+    recursive_draw_dendrogram!(ax, nodes[maximum(keys(nodes))], nodes; branch_shape)
     current_figure()
 end
 
--- a/src/stats/dendrogram.jl
+++ b/src/stats/dendrogram.jl
@@ -1,0 +1,123 @@
+"""
+    dendrogram(x, y; kwargs...)
+
+Draw a [dendrogram](https://en.wikipedia.org/wiki/Dendrogram),
+with leaf nodes specified by `x` and `y` coordinates,
+and parent nodes identified by `merges`.
+
+# Arguments
+- `x`: x positions of leaf nodes
+- `y`: y positions of leaf nodes (default = 0)
+# Keywords
+- `merges`: specifies connections between nodes (see below)
+- `treestyle`: one of `:??`, `:??` 
+"""
+@recipe(Dendrogram, x, y) do scene
+    Theme(
+        weights = automatic,
+        color = theme(scene, :patchcolor),
+        colormap = theme(scene, :colormap),
+        colorrange = automatic,
+        orientation = :vertical,
+        strokecolor = theme(scene, :patchstrokecolor),
+        strokewidth = theme(scene, :patchstrokewidth),
+        cycle = [:color => :patchcolor],
+        inspectable = theme(scene, :inspectable)
+    )
+end
+
+conversion_trait(x::Type{<:Dendrogram}) = SampleBased() #??
+
+function Makie.plot!(plot::Dendrogram)
+    args = @extract plot (weights, width, range, show_outliers, whiskerwidth, show_notch, orientation, gap, dodge, n_dodge, dodge_gap)
+
+end
+
+##############
+# Playground #
+##############
+
+struct DNode
+    idx::Int
+    x::Float32
+    y::Float32
+    children::Union{Tuple{Int,Int}, Nothing}
+end
+
+function find_merge(n1, n2; height=1)
+    newx = min(n1[1], n2[1]) + abs((n1[1] - n2[1])) / 2
+    newy = max(n1[2], n2[2]) + height
+    return Point2f(newx, newy)
+end
+
+function find_merge(n1::DNode, n2::DNode; height=1, index=max(n1.idx, n2.idx)+1)
+    newx = min(n1.x, n2.x) + abs((n1.x - n2.x)) / 2
+    newy = max(n1.y, n2.y) + height
+
+    return DNode(index, Point2f(newx, newy)..., (n1.idx, n2.idx))
+end
+
+
+function get_tree_connectors(parent, child1, child2)
+    return [(child1.x, child1.y), (parent.x, parent.y), (child2.x, child2.y)]
+end
+
+function get_box_connectors(parent, child1, child2)
+    yp = parent.y
+    x1 = child1.x
+    x2 = child2.x
+
+    return [(x1, child1.y), (x1, yp), (x2, yp), (x2, child2.y)]
+end
+
+function recurssive_draw!(ax, node, nodes; branch_shape=:tree)
+    isnothing(node.children) && return nothing
+    child1 = nodes[node.children[1]]
+    child2 = nodes[node.children[2]]
+   
+    l = branch_shape == :tree ? get_tree_connectors(node, child1, child2) :
+        branch_shape == :box  ? get_box_connectors(node, child1, child2) : error()
+    
+    lines!(ax, l)
+
+    recurssive_draw!(ax, child1, nodes; branch_shape)
+    recurssive_draw!(ax, child2, nodes; branch_shape)
+    return nothing
+end
+
+
+function dendrogram(leaves, merges; branch_shape=:tree)
+    nodes = Dict(i => DNode(i, n[1], n[2], nothing) for (i,n) in enumerate(leaves))
+    nm = maximum(keys(nodes))
+
+    for m in merges
+        nm += 1
+        nodes[nm] = find_merge(nodes[m[1]], nodes[m[2]]; index = nm)
+    end
+
+    dendrogram(nodes; branch_shape)
+end
+
+
+function dendrogram(nodes::Dict{Int, DNode}; branch_shape=:tree)
+    ax = Axis(Figure()[1,1])    
+    recurssive_draw!(ax, nodes[maximum(keys(nodes))], nodes; branch_shape)
+    current_figure()
+end
+
+
+function hcl_nodes(hcl; useheight=false)
+    nleaves = length(hcl.order)
+    nodes = Dict(i => DNode(i, x, 0, nothing) for (i,x) in enumerate(invperm(hcl.order)))
+    nm = maximum(keys(nodes))
+
+    for (m1, m2) in eachrow(hcl.merges)
+        nm += 1
+        
+        m1 = m1 < 0 ? -m1 : m1 + nleaves
+        m2 = m2 < 0 ? -m2 : m2 + nleaves
+        nodes[nm] = find_merge(nodes[m1], nodes[m2]; index=nm)            
+    end
+
+    return nodes
+end


src/stats/dendrogram.jl
```diff
@@ -0,0 +1,123 @@
...
+        colorrange = automatic,
+        orientation = :vertical,
+        strokecolor = theme(scene, :patchstrokecolor),
+        strokewidth = theme(scene, :patchstrokewidth),
+        cycle = [:color => :patchcolor],
+        inspectable = theme(scene, :inspectable)
+    )
+end
+
+conversion_trait(x::Type{<:Dendrogram}) = SampleBased() #??
```
asinghvi17, line 29: I don't think we have this kind of trait - if you want to make one, you can define it as `struct SampleBased <: MakieCore.ConversionTrait end`.

You don't _have_ to define this - what kind of arguments could `dendrogram` expect?  Could there be other recipes to which this would be common?

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,123 @@
...
+with leaf nodes specified by `x` and `y` coordinates,
+and parent nodes identified by `merges`.
+
+# Arguments
+- `x`: x positions of leaf nodes
+- `y`: y positions of leaf nodes (default = 0)
+# Keywords
+- `merges`: specifies connections between nodes (see below)
+- `treestyle`: one of `:??`, `:??` 
+"""
```
asinghvi17, line 14:
```suggestion

# Extended help

$(ATTRIBUTES)
"""
```

when you enter `??` at the REPL, it shows the default values of the attributes.

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,123 @@
...
+end
+
+##############
+# Playground #
+##############
+
+struct DNode
+    idx::Int
+    x::Float32
+    y::Float32
```
asinghvi17, line 43: should this be a `Point`?  What about 3D dendrograms (if those are a thing)?

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,123 @@
...
+
+function recursive_draw_dendrogram!(ax, node, nodes; branch_shape=:tree)
+    isnothing(node.children) && return nothing
+    child1 = nodes[node.children[1]]
+    child2 = nodes[node.children[2]]
+   
+    l = branch_shape == :tree ? get_tree_connectors(node, child1, child2) :
+        branch_shape == :box  ? get_box_connectors(node, child1, child2) : error()
+    
+    lines!(ax, l)
```
asinghvi17, line 81: Since Makie works on observables, every recipe has to be written with the idea that the underlying data may be updated.  `recursive_draw_dendrogram!` should return an array of points, and potentially an array of colors, then those would be plotted in a single `lines!` call in the recipe.

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,123 @@
...
+        weights = automatic,
+        color = theme(scene, :patchcolor),
+        colormap = theme(scene, :colormap),
+        colorrange = automatic,
+        orientation = :vertical,
+        strokecolor = theme(scene, :patchstrokecolor),
+        strokewidth = theme(scene, :patchstrokewidth),
+        cycle = [:color => :patchcolor],
+        inspectable = theme(scene, :inspectable)
+    )
```
asinghvi17, line 26: `theme` is a bit misleading, it only provides the current default theme IIRC.  Better to use `inherit` here.

```suggestion
    Theme(
        weights = automatic,
        color = Makie.inherit(scene, :color, :black),
        colormap = Makie.inherit(scene, :colormap, :viridis),
        colorrange = automatic,
        orientation = :vertical,
        strokecolor = Makie.inherit(scene, :strokecolor, :black),
        strokewidth = Makie.inherit(scene, :strokewidth, 1.0),
        cycle = [:color => :patchcolor],
        inspectable = Makie.inherit(scene, :inspectable)
    )
```

asinghvi17:
Sorry for the deluge of comments - here's the last one!

I couldn't help myself and refactored the code to fit the Makie schema a bit more.  Hopefully it makes sense.

(This is basically the sum of all my comments + an interface to allow extending the branch shape + compatibility with Observables).  It would be nice to give `DNode` some sort of `color` attribute, so we could potentially specialize on that as well.

```julia
"""
    dendrogram(leaves, merges; kwargs...)
Draw a [dendrogram](https://en.wikipedia.org/wiki/Dendrogram),
with leaf nodes specified by `x` and `y` coordinates,
and parent nodes identified by `merges`.
# Arguments
- `x`: x positions of leaf nodes
- `y`: y positions of leaf nodes (default = 0)
# Keywords
- `merges`: specifies connections between nodes (see below)
- `treestyle`: one of `:??`, `:??`.  Overload `dendrogram_connectors(::Val{:mystyle}, parent, child1, child2)` to define a new style.
"""
@recipe(Dendrogram, nodes) do scene
    Theme(
        weights = Makie.automatic,
        branch_shape = :box,
        linewidth = Makie.inherit(scene, :linewidth, 1.0),
        color = Makie.inherit(scene, :color, :black),
        colormap = Makie.inherit(scene, :colormap, :viridis),
        colorrange = Makie.automatic,
        orientation = :vertical,
        cycle = [:color => :patchcolor],
        inspectable = Makie.inherit(scene, :inspectable, false)
    )
end

function recursive_dendrogram_points(node, nodes, ret_points = Point2f[], ret_colors = []; branch_shape=:tree)
    isnothing(node.children) && return nothing
    child1 = nodes[node.children[1]]
    child2 = nodes[node.children[2]]
   
    l = dendrogram_connectors(Val(branch_shape), node, child1, child2)
    
    # even if the inputs are 2d, the outputs should be 3d - this is what `to_ndim` does.
    append!(ret_points, Makie.to_ndim.(Point3f, l, 0))
    push!(ret_points, Point3f(NaN)) # separate segments
    append!(ret_colors, [Float32(node.idx) for _ in 1:length(l)])
    push!(ret_colors, NaN32) # separate segments

    recursive_dendrogram_points(child1, nodes, ret_points, ret_colors; branch_shape)
    recursive_dendrogram_points(child2, nodes, ret_points, ret_colors; branch_shape)
    return ret_points, ret_colors
end


function Makie.plot!(plot::Dendrogram{<: Tuple{<: Dict{<: Integer, <: Union{DNode{2}, DNode{3}}}}})
    # args = @extract plot (weights, width, range, show_outliers, whiskerwidth, show_notch, orientation, gap, dodge, n_dodge, dodge_gap)

    points_vec = Observable{Any}()
    colors_vec = Observable{Any}()

    lift(plot[1], plot.branch_shape) do nodes, branch_shape
        # this pattern is basically first updating the values of the observables,
        points_vec.val, colors_vec.val = recursive_dendrogram_points(nodes[maximum(keys(nodes))], nodes; branch_shape)
        # then propagating the signal, so that there is no error with differing lengths.
        notify(points_vec); notify(colors_vec)
    end

    lines!(plot, points_vec; color = colors_vec, colormap = plot.colormap, colorrange = plot.colorrange, linewidth = plot.linewidth, inspectable = plot.inspectable)

end

# branching styles

struct DNode{N}
    idx::Int
    position::Point{N, Float32}
    children::Union{Tuple{Int,Int}, Nothing}
end

function DNode(idx::Int, point::Point{N}, children::Union{Tuple{Int,Int}, Nothing}) where N
    return DNode{N}(idx, point, children)
end

function dendrogram_connectors(::Val{:tree}, parent, child1, child2)
    return [child1.position, parent.position, child2.position]
end

function dendrogram_connectors(::Val{:box}, parent::DNode{2}, child1::DNode{2}, child2::DNode{2})
    yp = parent.position[2]
    x1 = child1.position[1]
    x2 = child2.position[1]

    return Point2f[(x1, child1.position[2]), (x1, yp), (x2, yp), (x2, child2.position[2])]
end

function dendrogram_connectors(::Val{:box}, parent::DNode{3}, child1::DNode{3}, child2::DNode{3})
    yp = parent.position[2]
    x1 = child1.position[1]
    x2 = child2.position[1]

    return Point3f[
        (x1, child1.position[2], child1.position[3]), 
        (x1, yp, (parent.position[3] + child1.position[3])./2), 
        (x2, yp, (parent.position[3] + child2.position[3])./2), 
        (x2, child2.position[2], child2.position[3])
    ]
end


# convert utils

function find_merge(n1, n2; height=1)
    newx = min(n1[1], n2[1]) + abs((n1[1] - n2[1])) / 2
    newy = max(n1[2], n2[2]) + height
    return Point2f(newx, newy)
end

function find_merge(n1::DNode{2}, n2::DNode{2}; height=1, index=max(n1.idx, n2.idx)+1)
    newx = min(n1.position[1], n2.position[1]) + abs((n1.position[1] - n2.position[1])) / 2
    newy = max(n1.position[2], n2.position[2]) + height

    return DNode{2}(index, Point2f(newx, newy), (n1.idx, n2.idx))
end

function find_merge(n1::DNode{3}, n2::DNode{3}; height=1, index=max(n1.idx, n2.idx)+1)
    newx = min(n1.position[1], n2.position[1]) + abs((n1.position[1] - n2.position[1])) / 2
    newy = max(n1.position[2], n2.position[2]) + height
    newz = min(n1.position[3], n2.position[3]) + abs((n1.position[3] - n2.position[3])) / 2

    return DNode{3}(index, Point3f(newx, newy, newz), (n1.idx, n2.idx))
end

function Makie.convert_arguments(::Type{<: Dendrogram}, leaves::Vector{<: Point}, merges::Vector{<: Tuple{<: Integer, <: Integer}})
    nodes = Dict(i => DNode(i, n, nothing) for (i,n) in enumerate(leaves))
    nm = maximum(keys(nodes))

    for m in merges
        nm += 1
        nodes[nm] = find_merge(nodes[m[1]], nodes[m[2]]; index = nm)
    end
    return (nodes,)
end


function hcl_nodes(hcl; useheight=false)
    nleaves = length(hcl.order)
    nodes = Dict(i => DNode(i, x, 0, nothing) for (i,x) in enumerate(invperm(hcl.order)))
    nm = maximum(keys(nodes))

    for (m1, m2) in eachrow(hcl.merges)
        nm += 1
        
        m1 = m1 < 0 ? -m1 : m1 + nleaves
        m2 = m2 < 0 ? -m2 : m2 + nleaves
        nodes[nm] = find_merge(nodes[m1], nodes[m2]; index=nm)            
    end

    return nodes
end


```

mkborregaard:
This looks great! Just wondering if it would make sense to have this in GraphMakie? I guess a dendrogram is really a NetworkLayout.buchheimtree with custom line drawing?

src/stats/dendrogram.jl
kescobo: I copied that from BarPlot - no idea what it does 😅

src/stats/dendrogram.jl
kescobo: Oh neat, was just putting that in as a placeholder while I think of what to name then.

kescobo:
> This looks great! Just wondering if it would make sense to have this in GraphMakie? I guess a dendrogram is really a NetworkLayout.buchheimtree with custom line drawing?

See my bullet 3 under help/feedback ☺️. I agree that that might make the most sense. 

One consideration (with the caveat that I'm not very familiar with GraphMakie, so it might be moot) is that it did not seem super obvious to me how to fix the points of the leaves. Since my primary use case at the moment is attaching dendrograms to heatmaps, that's a critical bit.

@asinghvi17 holy crap, thanks! I will need some time to study this, but it looks great! 🙇

src/stats/dendrogram.jl
asinghvi17: Oh, my bad - was looking at MakieCore.  It does seem to exist :D.

`conversion_trait` is basically a trait system for plots, where e.g. `lines`, `scatter`, and `poly` all share the `PointBased` trait - this enables people to write converts for their data types dispatching on PointBased, and thereby hook that in to the conversion pipeline without having to specify a specific plot type.

kescobo:
So, this is working a bit now - I've been trying to sort out how to deal with color. I figured out that I can replace `colors_vec.val` with a vector of eg `:black`. But I'm trying to get grouping working to add different colors, and I'm a bit stuck. I created `recursive_leaf_groups()`, which enables checking the groups of all leaves from a node. For example:

```julia
leaves = Point2f.([
    (1,0),
    (2,0.5),
    (3,1),
    (4,2),
    (5,0)
    ])

    merges = [
    (1, 2), # 6
    (6, 3), # 7
    (4, 5), # 8
    (7, 8), # 9
    ]
    
nodes, = Makie.convert_arguments(Dendrogram, leaves, merges)
groups = [1,1,2,3,3]

recursive_leaf_groups(nodes[9], nodes, groups) # [1,1,2,3,3]
recursive_leaf_groups(nodes[6], nodes, groups) # [1,1]
recursive_leaf_groups(nodes[7], nodes, groups) # [1,1,2]
recursive_leaf_groups(nodes[8], nodes, groups) # [3,3]
```
 
The idea here is that, if all of the children of a given node are in the same group, that node and all its descendants can have the same color. I'd like to be able to, for example, do

```
dendrogram(leaves, merges; groups = [1,1,2,3,3], color=[:blue, :orange, :purple], fallback_color=:gray)
```

What I have here *sort of* works for just assigning numbers, eg `dendrogram(leaves, merges; groups=[1,1,2,3,3])` gets me

image


while  `dendrogram(leaves, merges; groups=[1,1,1,3,3]) ` gets me

image

Gonna take a little break and come back to it.

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,180 @@
...
+end
+
+
+function Makie.plot!(plot::Dendrogram{<: Tuple{<: Dict{<: Integer, <: Union{DNode{2}, DNode{3}}}}})
+    args = @extract plot (color, groups)
+
+    points_vec = Observable{Any}()
+    colors_vec = Observable{Any}()
+    group_colors = Observable{Any}() # placeholder - I don't really understand this
+    group_colors = Dict(1=> :black, 2=> :blue, 3=> :orange) # want to get this automatically, but don't know how
```
asinghvi17, line 78: Could make the default color map categorical, and rely on that instead for the colors?

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,180 @@
...
+    return ret_points, ret_colors
+end
+
+
+function Makie.plot!(plot::Dendrogram{<: Tuple{<: Dict{<: Integer, <: Union{DNode{2}, DNode{3}}}}})
+    args = @extract plot (color, groups)
+
+    points_vec = Observable{Any}()
+    colors_vec = Observable{Any}()
+    group_colors = Observable{Any}() # placeholder - I don't really understand this
```
asinghvi17, line 77: I don't think this needs to be an observable - you can just create a `groups = []` inside the lift.  If you want to make it really efficient, create it outside the lift, and `empty!(groups)` on lift...

asinghvi17 (commented):
Looks good!  Added a couple of comments, hope they make sense :).  Otherwise, it looks pretty close to completion! Just needs some examples and tests now.

@kescobo were you planning to add labels to the leaf nodes?  That should be pretty easy to do :)

kescobo:
Sorry I've stalled on this a bit - other work stuff came up :-(

kescobo:
OK, so I'm making progress, I think, but not sure what to do next. I made a function that can find whether a node belongs to a group, so eg:

```
leaves = Point2f.([
    (1,0),
    (2,0.5),
    (3,1),
    (4,2),
    (5,0)
])

merges = [
    (1, 2), # 6
    (6, 3), # 7
    (4, 5), # 8
    (7, 8), # 9
]

dendrogram(leaves, merges; groups = [1,1,1,2,2])
```
gives
image
 but 

```
dendrogram(leaves, merges; groups = [1,1,3,2,2])
```

gives

image


But I'm not sure how to

1. Set a default color - should be black. At the moment, if I don't pass a `groups` argument, I get `ERROR: Can't interpolate in a range where cmin == cmax. This can happen, for example, if a colorrange is set automatically but there's only one unique value present.`

asinghvi17:
> 1. Set a default color - should be black. At the moment, if I don't pass a groups argument, I get ERROR: Can't interpolate in a range where cmin == cmax. This can happen, for example, if a colorrange is set automatically but there's only one unique value present.

An idea comes to mind, and this should help with (2) as well: let `groups` accept either a vector of colors, or a vector of reals.  If reals, then do the colormap stuff.  If colors, simply plot using `group_colors = to_color.(groups)`.  Multiple dispatch based on `eltype(groups)` should help here.

> 2. Allow passing a vector of colors (eg [:red, :blue, :orange]

As above.

> 3. Right now, group 3 in the second example above ends up with the default color (0) because it has only 1 member. Not sure how to set it as a group if groups only have 1 member.

Not sure about this one - will take some thought, I think.

SimonDanisch:
This looks great! :) Could you add a file `dentrogram.md` e.g. like this one: https://github.com/MakieOrg/Makie.jl/blob/master/docs/examples/plotting_functions/arc.md (same folder etc, it should get picked up by the plot function example page).
Also a reference test like this one would be nice:
https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/short_tests.jl#L195-L197
Best with thick lines so that problems are quickly visible, and multiple plots in one figure to test different attributes would be nice :) 
I guess adding the plots you posted on this PR to docs/tests would be all we need!

kescobo:
Yeah, totally my bad. I actually think it needs more than just docs and tests - my last iteration wasn't able to pass a vector of colors, but I definitely need to come back to this.

Also, FWIW, I'm fine if someone else wants to take up the mantel. I just had a baby and even I do back to work I'll be swamped, and I don't want to be a cookie licker.

bjarthur:
so i'm trying to get this to work.  saved the dendrogram.jl file in this PR locally, and then `include`d it.  get a bunch of errors when running the most recent code block above:

```
julia> using GLMakie

julia> include("src/dendrogram.jl")
recursive_leaf_groups (generic function with 1 method)

julia> leaves = Point2f.([
           (1,0),
           (2,0.5),
           (3,1),
           (4,2),
           (5,0)
       ])
5-element Vector{Point{2, Float32}}:
 [1.0, 0.0]
 [2.0, 0.5]
 [3.0, 1.0]
 [4.0, 2.0]
 [5.0, 0.0]

julia> merges = [
           (1, 2), # 6
           (6, 3), # 7
           (4, 5), # 8
           (7, 8), # 9
       ]
4-element Vector{Tuple{Int64, Int64}}:
 (1, 2)
 (6, 3)
 (4, 5)
 (7, 8)

julia> dendrogram(leaves, merges; groups = [1,1,1,2,2])
1   : #version 410
2   : 
3   : 
4   : 
5   : struct Nothing{ //Nothing type, to encode if some variable doesn't contain any data
6   :     bool _; //empty structs are not allowed
7   : };
8   : 
9   : #define FAST_PATH
10  : 
11  : in vec2 vertex;
12  : 
13  : in float lastlen;
14  : in float valid_vertex;
15  : 
16  : in float      color;
17  : uniform Nothing  color_map;
18  : uniform Nothing  intensity;
19  : uniform Nothing color_norm;
20  : uniform float  thickness;
21  : 
22  : vec4 _color(vec3 color, Nothing intensity, Nothing color_map, Nothing color_norm, int index, int len);
23  : vec4 _color(vec4 color, Nothing intensity, Nothing color_map, Nothing color_norm, int index, int len);
24  : vec4 _color(Nothing color, float intensity, sampler1D color_map, vec2 color_norm, int index, int len);
25  : vec4 _color(Nothing color, sampler1D intensity, sampler1D color_map, vec2 color_norm, int index, int len);
26  : 
27  : uniform mat4 projectionview, model;
28  : uniform uint objectid;
29  : uniform int total_length;
30  : 
31  : out uvec2 g_id;
32  : out vec4 g_color;
33  : out float g_lastlen;
34  : out int g_valid_vertex;
35  : out float g_thickness;
36  : 
37  : vec4 getindex(sampler2D tex, int index);
38  : vec4 getindex(sampler1D tex, int index);
39  : 
40  : vec4 to_vec4(vec3 v){return vec4(v, 1);}
41  : vec4 to_vec4(vec2 v){return vec4(v, 0, 1);}
42  : 
43  : int get_valid_vertex(float se){return int(se);}
44  : int get_valid_vertex(Nothing se){return 1;}
45  : 
46  : uniform float depth_shift;
47  : 
48  : void main()
49  : {
50  :     g_lastlen = lastlen;
51  :     int index = gl_VertexID;
52  :     g_id = uvec2(objectid, index+1);
53  :     g_valid_vertex = get_valid_vertex(valid_vertex);
54  :     g_thickness = thickness;
55  : 
56  :     g_color = _color(color, intensity, color_map, color_norm, index, total_length);
57  :     #ifdef FAST_PATH
58  :         gl_Position = projectionview * model * to_vec4(vertex);
59  :     #else
60  :         gl_Position = to_vec4(vertex);
61  :     #endif
62  :     gl_Position.z += gl_Position.w * depth_shift;
63  : }
64  : 
┌ Warning: shader /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/lines.vert didn't compile. 
│ ERROR: 0:56: No matching function for call to _color(float, Nothing, Nothing, Nothing, int, int)
└ @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLShader.jl:129
1   : #version 410
2   : 
3   : 
4   : 
5   : struct Nothing{ //Nothing type, to encode if some variable doesn't contain any data
6   :     bool _; //empty structs are not allowed
7   : };
8   : 
9   : #define FAST_PATH
10  : 
11  : in vec2 vertex;
12  : 
13  : in float lastlen;
14  : in float valid_vertex;
15  : 
16  : in float      color;
17  : uniform Nothing  color_map;
18  : uniform Nothing  intensity;
19  : uniform Nothing color_norm;
20  : uniform float  thickness;
21  : 
22  : vec4 _color(vec3 color, Nothing intensity, Nothing color_map, Nothing color_norm, int index, int len);
23  : vec4 _color(vec4 color, Nothing intensity, Nothing color_map, Nothing color_norm, int index, int len);
24  : vec4 _color(Nothing color, float intensity, sampler1D color_map, vec2 color_norm, int index, int len);
25  : vec4 _color(Nothing color, sampler1D intensity, sampler1D color_map, vec2 color_norm, int index, int len);
26  : 
27  : uniform mat4 projectionview, model;
28  : uniform uint objectid;
29  : uniform int total_length;
30  : 
31  : out uvec2 g_id;
32  : out vec4 g_color;
33  : out float g_lastlen;
34  : out int g_valid_vertex;
35  : out float g_thickness;
36  : 
37  : vec4 getindex(sampler2D tex, int index);
38  : vec4 getindex(sampler1D tex, int index);
39  : 
40  : vec4 to_vec4(vec3 v){return vec4(v, 1);}
41  : vec4 to_vec4(vec2 v){return vec4(v, 0, 1);}
42  : 
43  : int get_valid_vertex(float se){return int(se);}
44  : int get_valid_vertex(Nothing se){return 1;}
45  : 
46  : uniform float depth_shift;
47  : 
48  : void main()
49  : {
50  :     g_lastlen = lastlen;
51  :     int index = gl_VertexID;
52  :     g_id = uvec2(objectid, index+1);
53  :     g_valid_vertex = get_valid_vertex(valid_vertex);
54  :     g_thickness = thickness;
55  : 
56  :     g_color = _color(color, intensity, color_map, color_norm, index, total_length);
57  :     #ifdef FAST_PATH
58  :         gl_Position = projectionview * model * to_vec4(vertex);
59  :     #else
60  :         gl_Position = to_vec4(vertex);
61  :     #endif
62  :     gl_Position.z += gl_Position.w * depth_shift;
63  : }
64  : 
┌ Warning: shader /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/lines.vert didn't compile. 
│ ERROR: 0:56: No matching function for call to _color(float, Nothing, Nothing, Nothing, int, int)
└ @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLShader.jl:129
ERROR: program 47 not linked. Error in: 
/Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/fragment_output.frag or /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/util.vert or /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/lines.vert or /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/lines.geom or /Users/arthurb/.julia/packages/GLMakie/45A8G/assets/shader/lines.frag
ERROR: One or more attached shaders not successfully compiled

Stacktrace:
  [1] error(::String, ::String, ::String, ::String)
    @ Base ./error.jl:44
  [2] compile_program(shaders::Vector{GLMakie.GLAbstraction.Shader}, fragdatalocation::Vector{Tuple{Int64, String}})
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLShader.jl:188
  [3] (::GLMakie.GLAbstraction.var"#70#75"{GLMakie.GLAbstraction.ShaderCache, Vector{Vector{String}}, Vector{Vector{String}}, NTuple{5, String}})()
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLShader.jl:261
  [4] get!(default::GLMakie.GLAbstraction.var"#70#75"{GLMakie.GLAbstraction.ShaderCache, Vector{Vector{String}}, Vector{Vector{String}}, NTuple{5, String}}, h::Dict{Any, GLMakie.GLAbstraction.GLProgram}, key::Tuple{NTuple{5, String}, Vector{Vector{String}}})
    @ Base ./dict.jl:468
  [5] gl_convert(cache::GLMakie.GLAbstraction.ShaderCache, lazyshader::GLMakie.GLVisualizeShader, data::Dict{Symbol, Any})
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLShader.jl:252
  [6] gl_convert
    @ ~/.julia/packages/GLMakie/45A8G/src/glshaders/visualize_interface.jl:88 [inlined]
  [7] GLMakie.GLAbstraction.RenderObject(data::Dict{Symbol, Any}, program::GLMakie.GLVisualizeShader, pre::GLMakie.GLAbstraction.StandardPrerender, post::GLFW.Window, context::GLFW.Window)
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLTypes.jl:458
  [8] GLMakie.GLAbstraction.RenderObject(data::Dict{Symbol, Any}, program::GLMakie.GLVisualizeShader, pre::GLMakie.GLAbstraction.StandardPrerender, post::GLFW.Window)
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/45A8G/src/GLAbstraction/GLTypes.jl:404
  [9] assemble_shader(data::Dict{Symbol, Any})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/glshaders/visualize_interface.jl:108
 [10] draw_lines(screen::Any, position::Observable{Vector{Point{2, Float32}}}, data::Dict)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/glshaders/lines.jl:119
 [11] (::GLMakie.var"#203#206"{GLMakie.Screen{GLFW.Window}, Lines{Tuple{Vector{Point{2, Float32}}}}})(gl_attributes::Dict{Symbol, Any})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:309
 [12] (::GLMakie.var"#177#180"{GLMakie.var"#203#206"{GLMakie.Screen{GLFW.Window}, Lines{Tuple{Vector{Point{2, Float32}}}}}, GLMakie.Screen{GLFW.Window}, Scene, Lines{Tuple{Vector{Point{2, Float32}}}}})()
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:135
 [13] get!(default::GLMakie.var"#177#180"{GLMakie.var"#203#206"{GLMakie.Screen{GLFW.Window}, Lines{Tuple{Vector{Point{2, Float32}}}}}, GLMakie.Screen{GLFW.Window}, Scene, Lines{Tuple{Vector{Point{2, Float32}}}}}, h::Dict{UInt64, GLMakie.GLAbstraction.RenderObject}, key::UInt64)
    @ Base ./dict.jl:468
 [14] cached_robj!(robj_func::GLMakie.var"#203#206"{GLMakie.Screen{GLFW.Window}, Lines{Tuple{Vector{Point{2, Float32}}}}}, screen::GLMakie.Screen{GLFW.Window}, scene::Scene, x::Lines{Tuple{Vector{Point{2, Float32}}}})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:103
 [15] draw_atomic
    @ ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:278 [inlined]
 [16] insert!(screen::GLMakie.Screen{GLFW.Window}, scene::Scene, x::Lines{Tuple{Vector{Point{2, Float32}}}})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:151
 [17] (::GLMakie.var"#183#184"{GLMakie.Screen{GLFW.Window}, Scene})(x::Lines{Tuple{Vector{Point{2, Float32}}}})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:156
 [18] foreach(f::GLMakie.var"#183#184"{GLMakie.Screen{GLFW.Window}, Scene}, itr::Vector{AbstractPlot})
    @ Base ./abstractarray.jl:3075
 [19] insert!(screen::GLMakie.Screen{GLFW.Window}, scene::Scene, x::Combined{dendrogram, Tuple{Dict{Int64, DNode{2}}}})
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/drawing_primitives.jl:153
 [20] insertplots!(screen::GLMakie.Screen{GLFW.Window}, scene::Scene)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/screen.jl:447
 [21] insertplots!(screen::GLMakie.Screen{GLFW.Window}, scene::Scene) (repeats 2 times)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/screen.jl:450
 [22] display_scene!(screen::GLMakie.Screen{GLFW.Window}, scene::Scene)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/screen.jl:379
 [23] GLMakie.Screen(scene::Scene, config::GLMakie.ScreenConfig; visible::Nothing, start_renderloop::Bool)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/screen.jl:396
 [24] GLMakie.Screen(scene::Scene, config::GLMakie.ScreenConfig)
    @ GLMakie ~/.julia/packages/GLMakie/45A8G/src/screen.jl:392
 [25] getscreen(::Module, ::Scene; screen_config::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/.julia/packages/Makie/1hq9u/src/display.jl:408
 [26] getscreen
    @ ~/.julia/packages/Makie/1hq9u/src/display.jl:391 [inlined]
 [27] display(figlike::Makie.FigureAxisPlot; backend::Module, inline::Bool, update::Bool, screen_config::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie ~/.julia/packages/Makie/1hq9u/src/display.jl:161
 [28] display(figlike::Makie.FigureAxisPlot)
    @ Makie ~/.julia/packages/Makie/1hq9u/src/display.jl:131
 [29] print_response(errio::IO, response::Any, show_value::Bool, have_color::Bool, specialdisplay::Union{Nothing, AbstractDisplay})
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:0
 [30] (::REPL.var"#57#58"{REPL.LineEditREPL, Pair{Any, Bool}, Bool, Bool})(io::Any)
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:287
 [31] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:557
 [32] print_response(repl::REPL.AbstractREPL, response::Any, show_value::Bool, have_color::Bool)
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:285
 [33] (::REPL.var"#do_respond#80"{Bool, Bool, REPL.var"#93#103"{REPL.LineEditREPL, REPL.REPLHistoryProvider}, REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::Any, ok::Bool)
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:899
 [34] #invokelatest#2
    @ ./essentials.jl:819 [inlined]
 [35] invokelatest
    @ ./essentials.jl:816 [inlined]
 [36] run_interface(terminal::REPL.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/LineEdit.jl:2647
 [37] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL ~/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/REPL/src/REPL.jl:1300
 [38] (::REPL.var"#62#68"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL ./task.jl:514
```

this is with julia 1.9.3, GLMakie 0.8.10, on an apple M2.

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+
+function Makie.plot!(plot::Dendrogram{<: Tuple{<: Dict{<: Integer, <: Union{DNode{2}, DNode{3}}}}})
+    args = @extract plot (color, groups)
+
+    points_vec = Observable{Any}()
+    colors_vec = Observable{Any}()
+    
+    lift(plot[1], plot.branch_shape, plot[:color]) do nodes, branch_shape, color
```
bjarthur, line 78:
```suggestion
    points_vec = Observable{Vector{GeometryBasics.Point{2, Float32}}}([[0,0]])
    colors_vec = Observable{Any}([NaN])

    length(plot[1][])>1 && lift(plot[1], plot.branch_shape, plot[:color]) do nodes, branch_shape, color
```

this suggested change plots an empty graph should there be only 0 or 1 node.

kescobo:
@bjarthur I just invited you to my fork, so you can push directly to my `dendrogram` branch to work on this, if you'd like.

bjarthur:
great, thanks!  will push directly going forward.  for now though, does the suggested change above make sense to you?  if so, easiest to probably just merge it.

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+
+Draw a [dendrogram](https://en.wikipedia.org/wiki/Dendrogram),
+with leaf nodes specified by `x` and `y` coordinates,
+and parent nodes identified by `merges`.
+# Arguments
+- `x`: x positions of leaf nodes
+- `y`: y positions of leaf nodes (default = 0)
+# Keywords
+- `merges`: specifies connections between nodes (see below)
+- `treestyle`: one of `:tree`, `:box`.  Overload `dendrogram_connectors(::Val{:mystyle}, parent, child1, child2)` to define a new style.
```
asinghvi17, line 23: These need to be updated, and I don't see an equivalent to `merges` anywhere here?

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+function recursive_dendrogram_points(node, nodes, ret_points = Point2f[], ret_colors = []; color=:black, branch_shape=:tree, groups=nothing)
+    isnothing(node.children) && return nothing
+    child1 = nodes[node.children[1]]
+    child2 = nodes[node.children[2]]
+   
+    l = dendrogram_connectors(Val(branch_shape), node, child1, child2)
+    
+    # even if the inputs are 2d, the outputs should be 3d - this is what `to_ndim` does.
+    append!(ret_points, Makie.to_ndim.(Point3f, l, 0))
+    push!(ret_points, Point3f(NaN)) # separate segments
```
asinghvi17, line 51: These can be `Point3d` to use Float64 precision

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+
+    if isnothing(groups)
+        cgroup = 0
+    else
+        gs = recursive_leaf_groups(node, nodes, groups)
+        @debug gs
```
asinghvi17, line 57: Should be removed before merging if not required

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+        cgroup = length(unique(gs)) == 1 ? first(gs) : 0
+    end
+
+    @debug cgroup maxlog=2
```
asinghvi17, line 61: Samae as above

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+
+    append!(ret_colors, [cgroup for _ in 1:length(l)])
+    push!(ret_colors, NaN32) # separate segments
```
asinghvi17, line 64: Colors can be Float64 too

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+    recursive_dendrogram_points(child1, nodes, ret_points, ret_colors; branch_shape, groups)
+    recursive_dendrogram_points(child2, nodes, ret_points, ret_colors; branch_shape, groups)
+    return ret_points, ret_colors
+end
+
+
+function Makie.plot!(plot::Dendrogram{<: Tuple{<: Dict{<: Integer, <: Union{DNode{2}, DNode{3}}}}})
+    args = @extract plot (color, groups)
+
+    points_vec = Observable{Vector{GeometryBasics.Point{2, Float32}}}([[0,0]])
```
asinghvi17, line 75: Should the starting point be an attribute of the plot?

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,178 @@
...
+
+function find_merge(n1::DNode{3}, n2::DNode{3}; height=1, index=max(n1.idx, n2.idx)+1)
+    newx = min(n1.position[1], n2.position[1]) + abs((n1.position[1] - n2.position[1])) / 2
+    newy = max(n1.position[2], n2.position[2]) + height
+    newz = min(n1.position[3], n2.position[3]) + abs((n1.position[3] - n2.position[3])) / 2
+
+    return DNode{3}(index, Point3f(newx, newy, newz), (n1.idx, n2.idx))
+end
+
+function Makie.convert_arguments(::Type{<: Dendrogram}, leaves::Vector{<: Point}, merges::Vector{<: Tuple{<: Integer, <: Integer}})
```
asinghvi17, line 141:
```suggestion
function convert_arguments(::Type{<: Dendrogram}, leaves::AbstractVector{<: VecTypes}, merges::AbstractVector{<: VecTypes{2, <: Integer}})
```
```julia
julia> Makie.VecTypes{2, Float16}
Union{Tuple{Float16, Float16}, StaticArraysCore.StaticArray{Tuple{2}, Float16, 1}}
```

src/stats/dendrogram.jl
kescobo: `merges` should come from the Hclust object itself I think - it's a vector of tuples specifying the order of leaves

https://juliastats.org/Clustering.jl/stable/hclust.html#Clustering.Hclust

src/stats/dendrogram.jl
bjarthur: grep'ing Makie/src i find an equal number of uses of Point3d and Point3f.  i had thought Float32 was preferred as that is what most consumer grade GPUs use.  is there a preference though for Float64 in the Makie code base?

bjarthur:
@kescobo for months now i have been using a cut & paste of this code in production.  as noted above, colors don't work, so i simply hacked it by replacing

```
lines!(plot, points_vec; color = colors_vec, colormap = plot.colormap, colorrange = plot.colorrange ...)
```

with

```
lines!(plot, points_vec; color = :black ...)
```

personally i'd be happy if that change were permanent, so from my standpoint all that's needed is docs and tests.

i ofc am swamped too :). all i've done so far to this PR is to squash all the commits into one.

src/stats/dendrogram.jl
asinghvi17: We recently switched to allow Float64 all the way through Makie, only converting when transferring data to the GPU in v0.21, just so we can keep the extra precision.  Especially if plotting over a small interval but with a high offset (e.g. `10^10:(10^10+10)` or something), you do end up needing Float64 precision.

src/stats/dendrogram.jl
```diff
@@ -65,19 +66,22 @@ function recursive_dendrogram_points(node, nodes, ret_points = Point2f[], ret_co
...
     args = @extract plot (color, groups)
 
-    points_vec = Observable{Vector{GeometryBasics.Point{2, Float32}}}([[0,0]])
-    colors_vec = Observable{Any}([NaN])
+    points_vec = Observable{Vector{GeometryBasics.Point{2, Float32}}}([])
+    colors_vec = Observable{Any}([])
 
     length(plot[1][])>1 && lift(plot[1], plot.branch_shape, plot[:color]) do nodes, branch_shape, color
+        points_vec[] = []
+        colors_vec[] = []
```
asinghvi17, line 80:
```suggestion
        points_vec.val = []
        colors_vec.val = []
        notify(points_vec)
        notify(colors_vec)
```

Otherwise this will throw an error, since the length of the color vector is not the same as the length of the point vector.

dcelisgarza:
Any upates regarding this PR?

kescobo:
Nothing substative, AFAIK, but it's one of many dangling open source things I'm **hoping** to come back to in the next couple of months. That said, I'd be perfectly happy if someone else wants to take the reins - there are a lot of things in that category and I don't know how many I'm going to get to

kescobo:
Finally coming back to this after a long absence! Just merged current master and updated the precision in a few places.

When running the code in the very first comment, I'm getting the error:

```
julia> dendrogram(leaves, merges)
ERROR: MethodError: no method matching red(::Float32)
The function `red` exists, but no method is defined for this combination of argument types.
```

But I don't see anywhere in my code that I'm using the `red` function, and the default color should be black, so I'm at a bit of a loss. I get a slightly different error with GLMakie, but it also appears to be color-related.


<details>
<summary>Full Error</summary>

```
julia> show(err)
1-element ExceptionStack:
MethodError: no method matching red(::Float32)
The function `red` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  red(::ColorTypes.ARGB32)
   @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/traits.jl:18
  red(::ColorTypes.RGB24)
   @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/traits.jl:17
  red(::ColorTypes.TransparentColor{C, T, 4} where {C<:ColorTypes.AbstractRGB, T})
   @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/traits.jl:16
  ...

Stacktrace:
  [1] draw_multi(primitive::Lines{Tuple{Vector{Point{2, Float64}}}}, ctx::Cairo.CairoContext, positions::Vector{Vec{2, Float32}}, colors::Vector{Float32}, linewidths::Float32, dash::Nothing)
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/primitives.jl:233
  [2] draw_atomic(scene::Scene, screen::CairoMakie.Screen{CairoMakie.IMAGE}, primitive::Union{Lines, LineSegments})
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/primitives.jl:68
  [3] draw_plot(scene::Scene, screen::CairoMakie.Screen{CairoMakie.IMAGE}, primitive::Lines{Tuple{Vector{Point{2, Float64}}}})
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/infrastructure.jl:135
  [4] cairo_draw(screen::CairoMakie.Screen{CairoMakie.IMAGE}, scene::Scene)
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/infrastructure.jl:49
  [5] display(screen::CairoMakie.Screen{CairoMakie.IMAGE}, scene::Scene; connect::Bool, screen_config::@Kwargs{})
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/display.jl:45
  [6] display(screen::CairoMakie.Screen{CairoMakie.IMAGE}, scene::Scene)
    @ CairoMakie ~/Repos/makiedev/Makie.jl/CairoMakie/src/display.jl:40
  [7] display(figlike::Makie.FigureAxisPlot; backend::Module, inline::MakieCore.Automatic, update::Bool, screen_config::@Kwargs{})
    @ Makie ~/Repos/makiedev/Makie.jl/src/display.jl:166
  [8] display(figlike::Makie.FigureAxisPlot)
    @ Makie ~/Repos/makiedev/Makie.jl/src/display.jl:130
  [9] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
 [10] invokelatest
    @ ./essentials.jl:1052 [inlined]
 [11] print_response(errio::IO, response::Any, show_value::Bool, have_color::Bool, specialdisplay::Union{Nothing, AbstractDisplay})
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:409
 [12] (::REPL.var"#70#71"{REPL.LineEditREPL, Pair{Any, Bool}, Bool, Bool})(io::Any)
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:378
 [13] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:661
 [14] print_response(repl::REPL.AbstractREPL, response::Any, show_value::Bool, have_color::Bool)
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:376
 [15] (::REPL.var"#do_respond#96"{Bool, Bool, REPL.var"#112#130"{REPL.LineEditREPL, REPL.REPLHistoryProvider}, REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::Any, ok::Bool)
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:1003
 [16] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
 [17] invokelatest
    @ ./essentials.jl:1052 [inlined]
 [18] (::REPL.var"#122#140"{Regex, Regex, Int64, Int64, REPL.LineEdit.Prompt, REPL.LineEdit.Prompt, REPL.LineEdit.Prompt})(::REPL.LineEdit.MIState, ::Any, ::Vararg{Any})
    @ REPL ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/REPL.jl:1410
 [19] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
 [20] invokelatest
    @ ./essentials.jl:1052 [inlined]
 [21] (::REPL.LineEdit.var"#30#31"{REPL.var"#122#140"{Regex, Regex, Int64, Int64, REPL.LineEdit.Prompt, REPL.LineEdit.Prompt, REPL.LineEdit.Prompt}, String})(s::Any, p::Any)
    @ REPL.LineEdit ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/LineEdit.jl:1711
 [22] macro expansion
    @ ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/LineEdit.jl:2861 [inlined]
 [23] macro expansion
    @ ./lock.jl:273 [inlined]
 [24] (::REPL.LineEdit.var"#282#284"{REPL.Terminals.TTYTerminal, REPL.LineEdit.ModalInterface, REPL.LineEdit.MIState, ReentrantLock, REPL.LineEdit.Prompt})()
    @ REPL.LineEdit ~/Dotfiles/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/share/julia/stdlib/v1.11/REPL/src/LineEdit.jl:2851
```

</details>

ffreyer:
Looks like either `color::Vector{Float32}` passes through without doing colormapping or the colormap is a `Vector{Float32}` rather than `Vector{RGBAf}`

cyanescent:
Not sure it is related: the color vector input `colors_vec` in `lines!(plot, points_vec; color = colors_vec, colormap = plot.colormap, colorrange = plot.colorrange, [...])` (l. 90 of dendrogram.jl) looks strange:
`Any[0, 0, 0, 0, NaN32, 0, 0, 0, 0, NaN32, 0, 0, 0, 0, NaN32, 0, 0, 0, 0, NaN32]`
Why this format? It triggers this error:
```
ERROR: program 61 not linked. Error in: 
/home/alexandre/.julia/packages/GLMakie/87u59/assets/shader/fragment_output.frag or /home/alexandre/.julia/packages/GLMakie/87u59/assets/shader/lines.vert or /home/alexandre/.julia/packages/GLMakie/87u59/assets/shader/lines.geom or /home/alexandre/.julia/packages/GLMakie/87u59/assets/shader/lines.frag
error: linking with uncompiled/unspecialized shader
```

kescobo:
@cyanescent That's a bit of a hack - a way to "break" the lines so that they're not connected. Not sure why it's `Any` though - should be a Float type. 
```
julia> using CairoMakie

julia> x = [1,2,3, NaN, 4,5,6]; y = [3,2,1,NaN,1,2,3];

julia> lines(x,y)
```
image

That shouldn't be causing the error you're seeing

dcelisgarza:
Thank you guys for looking into this, this is the only thing keeping me from switching all my plotting to makie.

src/stats/dendrogram.jl
cyanescent: Its works when I add `Makie.convert_arguments` (at least in my simple testing setup).

ffreyer:
Alright I updated quite a bit.

I fixed the errors around colors. That was happening because colors were initialized to `Observable{Vector{Any}}`, which break dispatch further down the line.

I reworked a bunch of things:
- `nodes` are now represented as a Vector instead of Dict. As far as I can tell this shouldn't change anything
- grouping now happens once for all nodes, creating an array of labels in the same order as `nodes`. If a parent node merges different groups, NaN is used as the label
- as a result `nan_color` currently acts as the "ungrouped" or "default" color with grouping
- colormap controls the colors picked by grouping
- per-node color and value vectors can now be processed
- `dendrogram_connectors` got an `!` and now writes points directly instead of producing an intermediate array
- attribute handling should now include all lines attributes with passthrough

Some TODOs/Questions:
- `hcl_nodes()` is currently a dangling function. I assume it's a meant to be a `convert_arguments()` for Clustering.jl? Maybe this is something that could be implemented with package extensions?
- `weights` is an unused attribute. What's it for?
- `orientation` is an unused attribute. Me being me, I would argue we could drop it and just advertise rotate!() and translate!() but that's probably not as user friendly...
- `groups` and `merges` need documentation
- Is `nan_color` as the default, ungrouped color strange? Should we add a dedicated attribute instead?

I'm probably not a good person to finish this pr because I don't have a good idea of where it should go. I'll at least need some strong feedback.

kescobo:
@ffreyer Amazing! Thank you for pushing this forward! I will take a look in more detail next week, though I'm also not sure where it goes. To answer your specific questions:

1. Yes, `hcl_nodes()` is a conversion for `Clustering.jl`, and should be an extension either here (probably) or in `Clustering.jl` itself. I'm happy to take that part on, it should be pretty straightforward :tm: 
2. I don't recall what the `weights` attribute was supposed to be for :thinking: 
3. `orientation` I think was supposed to be for leaves facing down (default) or leaves facing right a la @cyanescent 's [comment](https://github.com/MakieOrg/Makie.jl/pull/2755#issuecomment-2679141693). If there's a generic way to handle this, I think that would be fine, but it's a very common need - I think we should either support it or have very clear documentation for how to achieve it
4. Regarding `nan_color` - I don't know, it could make sense as the default, but having a more accessible keyword that provides the "default" color would be my preference.

ffreyer:
> Amazing! Thank you for pushing this forward! I will take a look in more detail next week, though I'm also not sure where it goes. To answer your specific questions:
> 
> 1. Yes, `hcl_nodes()` is a conversion for `Clustering.jl`, and should be an extension either here (probably) or in `Clustering.jl` itself. I'm happy to take that part on, it should be pretty straightforward ™️

Ok, sounds good.

> 2. I don't recall what the `weights` attribute was supposed to be for 🤔

Maybe you started out from a recipe that had it. I removed it.

> 3. `orientation` I think was supposed to be for leaves facing down (default) or leaves facing right a la @cyanescent 's [comment](https://github.com/MakieOrg/Makie.jl/pull/2755#issuecomment-2679141693). If there's a generic way to handle this, I think that would be fine, but it's a very common need - I think we should either support it or have very clear documentation for how to achieve it

I think with all the other plots that have orientation (or synonyms) it doesn't make sense to push for a more complicated generic way here. I connected it now. I also added `origin`, though I'm not sure if setting the root position to (0, 0) is a good default.
 
> 4. Regarding `nan_color` - I don't know, it could make sense as the default, but having a more accessible keyword that provides the "default" color would be my preference.

Yea, I added `ungrouped_color` instead.

bjarthur:
re. `orientation`, it would be good to support four values, corresponding to having the leaves facing down, right, up, and left, so that folks could put a dendrogram on all sides of a plot.

alternatively, if `rotate!` and `translate!` can be used with the code as is, we could just document how to use them to put the dendrogram in other positions.  even tilted at an angle.

thanks so much for your work on this @ffreyer !  dendrograms are quite common in biological research.

dcelisgarza:
> re. `orientation`, it would be good to support four values, corresponding to having the leaves facing down, right, up, and left, so that folks could put a dendrogram on all sides of a plot.
> 
> alternatively, if `rotate!` and `translate!` can be used with the code as is, we could just document how to use them to put the dendrogram in other positions. even tilted at an angle.
> 
> thanks so much for your work on this @ffreyer ! dendrograms are quite common in biological research.

Yes, one of the common uses of dendrograms is to display them together with distance/correlation matrices. The lack of dendrograms is the only thing keeping me from switching to makie.

ffreyer:
Do 3D dendrograms make sense or was that just an attempt to generalize? If they do make sense, does orientation make sense with them?

kescobo:
I think that was just @asinghvi17 's attempt to generalize - that adds a ton of complexity that is hard for me to wrap my head around.

asinghvi17:
We can remove it

ffreyer:
Alright, I removed the 3D bits and swapped out `orientation` for `rotation`. That also accepts <: Real so you can place the dendrogram at odd angles too.

```julia
leaves = Point2f.([
    (1,0),
    (2,0.5),
    (3,1),
    (4,2),
    (5,0)
    ])

merges = [
    (1, 2), # 6
    (6, 3), # 7
    (4, 5), # 8
    (7, 8), # 9
]
f = Figure(size = (500, 500))
a = Axis(f[1, 1], aspect = DataAspect())
dendrogram!(leaves, merges; origin = Point2f( 0, -2), rotation = :down,  ungrouped_color = :gray,      groups = [1,1,2,3,3], colormap=[:blue, :orange, :purple])
dendrogram!(leaves, merges; origin = Point2f( 2,  0), rotation = :right, ungrouped_color = :red,       groups = [1,1,2,3,3], colormap=[:blue, :orange, :purple])
dendrogram!(leaves, merges; origin = Point2f( 0,  2), rotation = :up,    ungrouped_color = :green,     groups = [1,1,2,3,3], colormap=[:blue, :orange, :purple])
dendrogram!(leaves, merges; origin = Point2f(-2,  0), rotation = :left,  ungrouped_color = :lightblue, groups = [1,1,2,3,3], colormap=[:blue, :orange, :purple])
dendrogram!(leaves, merges; origin = Point2f( 4,  4), rotation = 3pi/4,  ungrouped_color = :orange,    groups = [1,1,2,3,3], colormap=[:blue, :orange, :purple])
f
```

image

asinghvi17:
I was thinking it would be cool to have this on polar axes, but we should probably have some scaling parameters as attributes (`xscale` and `yscale` maybe) that can be numbers or functions, in order to allow rescaling the points pre-transformation.

I could whip up a PR to this branch tomorrow ish if that makes sense?  Or maybe there's another interface that fits better.

mkborregaard:
This functionality would also be useful for plotting phylogenies https://github.com/EcoJulia/Phylo.jl/pull/15#issue-359044367

bjarthur:
> This functionality would also be useful for plotting phylogenies [EcoJulia/Phylo.jl#15 (comment)](https://github.com/EcoJulia/Phylo.jl/pull/15#issue-359044367)

i like the radial dendrograms you have there!  is that possible with this PR?  maybe by plotting into a polar axis??

jkrumbiegel:
I guess the polar version could just work, if there's a parameter that increases resolution of the horizontal segments

ffreyer:
Did some prototyping for Polar dendrogram:
image

> I was thinking it would be cool to have this on polar axes, but we should probably have some scaling parameters as attributes (`xscale` and `yscale` maybe) that can be numbers or functions, in order to allow rescaling the points pre-transformation.

Yes, this will be necessary. xscale and yscale sound like bad names choices though, since `rotation` will swap the meaning and they don't fit polar coordinates. What would these be called for trees? width and depth?

asinghvi17:
width and depth seem good to me!

src/stats/dendrogram.jl
```diff
@@ -116,11 +124,55 @@ function plot!(plot::Dendrogram)
...
         # move root to (0, 0), rotate, then to origin
         root_pos = nodes[end].position
-        for (i, p) in enumerate(points_vec[])
-            points_vec[][i] = R * (p - root_pos) + origin
+        for (i, p) in enumerate(ps)
+            ps[i] = R * (p - root_pos) + origin
         end
 
-        notify(points_vec)
+        if tf isa Polar
```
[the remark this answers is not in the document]
ffreyer: I thought about making that a function but I noticed that I also had to remap colors. Because of that it seems pretty specialized to me.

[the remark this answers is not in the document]
asinghvi17: Any transformation that subsamples would require that though, right?

kescobo:
I'm not 100% sure what you mean with width, depth, and height.

In the two examples I'm most familiar with:

- For an hclust, only the links between leaves and nodes are specified, even the position of leaves is indeterminant. That is, rotation about the nodes has no impact on the correctness of the tree (though in practice, the order matters a lot visually, so you want to be able to set this.
- For a phylogenetic tree, the relationship between the leaves/nodes, and also the distance between one leaf / node and the next (vertically if the leaves are at the bottom) are also specified.

Practically speaking though, I'd think that you'd want to specify where the leaves should be, and then rotatate around some axis to get the final position.

ffreyer:
> I'm not 100% sure what you mean with width, depth, and height.

224875975-377db23c-07ff-4c5c-a44b-449c77b73fdd
In this plot the width would be the x-distance between leaf nodes and depth or hight the y distance between the leaf nodes and the root node. 

---

image
In this width would be the theta distance and height/depth the radius.

Maybe it would be best if we finished up the pr as it is now and fix problems when they come up in issues

kescobo:
@cyanescent I think we'd want to do something like a recipe / extension in Phylo.jl or something

asinghvi17:
Yeah, a `convert_arguments` in an extension there would be good

ffreyer:
Updated docs with a bunch more examples to cover the attributes. The example plots are now:

| Screenshot 2025-05-14 175336 | Screenshot 2025-05-14 175345 |
| --- | --- |
| Screenshot 2025-05-14 175355 | Screenshot 2025-05-14 175405 |

I also added `Makie.dendrogram_node_position(plot)` which (re)calculates where the nodes go, i.e. including origin, rotation, width and depth. That seemed useful to me for e.g. text/textlabel on leaf nodes or scatter for marking joins.

Also tweaked the refimg test to cover more attribute combinations. Dendrogram is relatively slim so I think one test with a couple different plots is already fine.

I think from my side this is ready for review

src/stats/dendrogram.jl
```diff
@@ -0,0 +1,356 @@
...
+    nodes = [DNode(i, Point2d(x, 0), nothing) for (i,x) in enumerate(invperm(hcl.order))]
+
+    for (m1, m2) in eachrow(hcl.merges)
+        m1 = ifelse(m1 < 0, -m1, m1 + nleaves)
+        m2 = ifelse(m2 < 0, -m2, m2 + nleaves)
+        push!(nodes, find_merge(nodes[m1], nodes[m2]; index = length(nodes)+1))
+    end
+
+    return nodes
+end
```
kescobo, line 286: Yes - this should be a package extension. In the meantime, it may be best to add this to the documentation page. I can do that in a separate PR when this is merged.

kescobo (commented):
Thank you for dragging this towards the finish line! I definitely could not have gotten this much done on my own

dcelisgarza:
Do you guys reckon it's safe to use this branch in a project yet? Or would it be better to wait until it's merged?

dcelisgarza:
Do you guys reckon it's safe to use this branch or would it be best to wait for it to be merged?

PR #2765, opened 2023-03-14
don't convert if not necessary

SimonDanisch:


SimonDanisch:
The benchmark CI seems really flaky right now, I'm 99% sure that the CairoMakie regressions can't be real, especially if GLMakie doesn't mirror them ^^

PR #2766, opened 2023-03-14
Add `space=:transformed`

asinghvi17:
# Description

Basically data space, without applying `apply_transform`.  Works pretty well.

MWE:
```julia
fig, ax, plt = lines(1:10, 1:10; axis = (; xscale = log10))
lines!(ax, LinRange(0, 10, 10), 1:10; space = :transformed)
```
download-2

This is useful if you want to control when the plot's transform_func is applied, especially in a recipe!  

I'm working on a potential solution to plot Tyler tiles in geographic projections by plotting a pre-transformed mesh, and texturing it by the image.  The WIP code is [in this gist](https://gist.github.com/asinghvi17/896b7851f6f91fae29242edfa6e08579), and the relevant issue is https://github.com/MakieOrg/Tyler.jl/issues/25.

More generally, we could also use this `transformed` space in `poly`, and provide the mesh triangulation points in transformed space!

Needs tests and docs.

## Type of change

## Checklist


ffreyer:
I got a bit confused that this doesn't touch the transformation code, but there we are already only allowing `space = :data`
https://github.com/MakieOrg/Makie.jl/blob/7ff2df7adb8596ab0c4fac2211e0726350606830/src/layouting/transformation.jl#L203

I don't like the name that much but I can see the point of this.

Mixing in the opengl terminology from [leanropengl](https://learnopengl.com/Getting-started/Coordinate-Systems) we'd have 

- `:data -> (this) -> world -> view (camera) -> :clip -> (opengl internal screen)`
- `:relative -> :clip -> (opengl internal screen)`
- `:pixel -> :clip -> (opengl internal screen)` 
 
where `data -> (this)` applies `transform_func` and `(this) -> world` applies the model matrix (i.e. transformation.scale, etc). Since everything beyond `transform_func` is a linear transformation, maybe `linear_data` would be a more precise name for this?

asinghvi17:
Hmm, I see what you're saying but I'm not sure `:linear_data` is the most obvious name.  Although `:transformed` isn't good either, since it could be seen as related to e.g. `plot.transformation`.  

Another possible name might be `:transformed_data`, I wanted it to evoke `apply_transform` and `transform_func` in some way...

ffreyer:
Maybe it's better to leave it as is now and adjust things later if we find a better name.

This stuff may need some adjustments as well:

https://github.com/MakieOrg/Makie.jl/blob/2116627b5caed21ddf0e3bbd2103f55d5679d881/src/camera/projection_math.jl#L317-L345

It's used for scatter and text. I think it applies after transformations, so it's probably fine to consider `:transformed` as `:data` here?

ffreyer:
Oh yea.

Another question is whether this should affect limits. Since it applies before "world" space I'd say it probably should?

asinghvi17:
What would happen if the transformation doesn't have an inverse defined?

We have a nice method to deal with bounding box transformations for those transformations which are not linearly separable, but have an inverse defined.  So that's there, at least.

I guess we could check `applicable(Makie.inverse_transform(trans))`?

asinghvi17:
Unfortunately, it's not :( - at least, not when a plot is nested within a recipe.  If you try putting a `bracket` inside a recipe, the bounding box is computed from the raw points (which are in pixel space)...

src/layouting/data_limits.jl
```diff
@@ -144,6 +144,9 @@ function iterate_transformed(plot)
 end
 
 function iterate_transformed(points, model, space, trans_func)
+    if space != :data
+        (,)
```
asinghvi17, line 148:
```suggestion
        tuple()
```
I always make this mistake...

asinghvi17:
I patched `data_limits` to skip all plots whose space is not in data space.  We could extend this to transformed space as well, but would need to somehow transform those limits back to data space.   What do you think?

src/layouting/data_limits.jl
```diff
@@ -144,6 +144,9 @@ function iterate_transformed(plot)
 end
 
 function iterate_transformed(points, model, space, trans_func)
+    if space != :data
+        tuple()
+    end
```
ffreyer, line 149: This should apply from the outside via `exclude`

https://github.com/MakieOrg/Makie.jl/blob/ffc31a0e7d21584082443daf21c59227d95ed13e/src/layouting/data_limits.jl#L194-L202

starting from

https://github.com/MakieOrg/Makie.jl/blob/ffc31a0e7d21584082443daf21c59227d95ed13e/src/makielayout/blocks/axis.jl#L859-L882

Scene should already be fine

https://github.com/MakieOrg/Makie.jl/blob/ffc31a0e7d21584082443daf21c59227d95ed13e/src/scenes.jl#L571-L583

ffreyer: Oh actually axis should be fine as is too, since it has `is_data_space(to_value(get(plot, :space, :data))) || return true `.

ffreyer:
```julia
fig, ax, plt = lines(1:10, 1:10; axis = (; xscale = log10))
lines!(ax, LinRange(0, 10, 10), 1:10; space = :transformed)
```
This produces 
Screenshot from 2023-06-14 21-22-07
because the `10^1.0` is `1.0` in world space, not 10. To get the result from the initial post you need `LinRange(0, 1, 10)`.

ffreyer:
> I tried to patch `data_limits` to skip all plots whose space is not in data space. It doesn't seem to work, though.
> 
> ```julia
> @recipe(Test1) do scene
>        Attributes()
> end
> 
> # brackets were having bbox issues in recipes earlier, this should ideally look like the docs example
> function Makie.plot!(p::Test1)
>        bracket!(p, pi/2, 1, 5pi/2, 1, offset = 5, text = "Period length", style = :square)
> 
>        bracket!(p, pi/2, 1, pi/2, -1, text = "Amplitude", orientation = :down,
>            linestyle = :dash, rotation = 0, align = (:right, :center), textoffset = 4, linewidth = 2, color = :red, textcolor = :red)
> 
>        bracket!(p, 2.3, sin(2.3), 4.0, sin(4.0),
>            text = "Falling", offset = 10, orientation = :up, color = :purple, textcolor = :purple)
> 
>        bracket!(p, Point(5.5, sin(5.5)), Point(7.0, sin(7.0)),
>            text = "Rising", offset = 10, orientation = :down, color = :orange, textcolor = :orange,
>            fontsize = 30, textoffset = 30, width = 50)
> end
> 
> fig = Figure()
> test1(fig[1. 1])
> fig
> ```
> 
> iTerm2 WBdMrA

I would consider this a bounding box/limits problem. `test!` generates no `space` attribute and most places that need it call `get(plot, :space, :data)`. Axis limits do that too, so it assumes everything here to be in data space

asinghvi17:
Hey @SimonDanisch, could we add this to v0.21 as well potentially?

asinghvi17:
This PR should still work if merged, it's pretty light on changes :D.  Could we get this in soon?

ffreyer:
Are Axis limits and Float64 conversion working fine with this after 0.21?

asinghvi17:
Will have to check that - might just open a new PR then!

ffreyer:
I'm closing this as not planned.

Previous prs already changed transformation inheritance to be more careful and only trigger if the parent (scene or plot) and child (plot) are compatible. #4792 now removed space from `apply_transform` so that transformations are always applied. It also added shortcuts for controlling transformation inheritance. (Specifically you can choose between inheriting fully, just transform_func, just model or neither.) You can have `space = :transformed` by doing `transformation = :inherit_model, space = :data` now.

PR #2776, opened 2023-03-15
Details for (de)activating default interactions

stillyslalom:
# Description

Add additional detail to documentation of interactions, namely the location of an axis's interactions and the names of the default interactions.

## Type of change
Documentation

## Checklist


asinghvi17 (commented):
Looks good to me!

asinghvi17 (approved):

jkrumbiegel:
I think ax.interactions isn't api, if says there

> You can check which interactions are currently active by calling `interactions(ax)`

Although Makie is admittedly not very good about avoiding field accesses in its api. I would still remove it as it's technically internal implementation detail.

SimonDanisch:
I moved it to where `interaction(ax)` is mentioned, could merge it like that?

asinghvi17 (approved):

PR #2783, opened 2023-03-17
Add clipping options to GLMakie

ffreyer:
# Description

This adds the option to specify a world space clipping box which excludes anything outside it from rendering.

```julia
f, a, p = mesh(
    Sphere(Point3f(0), 1.2f0), color = :yellow, 
    clip_planes = Rect3f(Vec3f(-1), Vec3f(2))
)
```

Screenshot from 2023-03-17 23-05-33

Currently this is a per-plot attribute that defaults to what the scene defines. Can be `nothing`, a `Rect3` or `WorldAxisLimits`.

Working so far
- [x] mesh
- [x] meshscatter
- [x] scatter, text
- [x] volume
- [x] surface
- [x] heatmap, image
- [x] lines, linesegments

lines, linesegements, scatter and text may need to work differently since they transform to along the way.

## Type of change

## Checklist


ffreyer:
I tried two options for lines - one uploading two buffers, one with world space positions and one with screen space positions; and one that only uploads world space positions, i.e. lets the screen space positions calculated for last_len go to waste. In the end they both perform about equal for me, but the latter saves some memory which I assume is valuable for some. It's also a bit less code, so that's what I went with.

ffreyer:
https://stackoverflow.com/questions/22628186/glclipplane-is-there-an-equivalent-in-webgl
I guess for WGLMakie we could use the three.js implementation for clipping planes?

jkrumbiegel:
we also need polygons for clipping, for example for nonlinear axes. is that something that would fit the scope of this pr or not at all?

ffreyer:
Short summary of how this works: 
Each vertex has a `ClipDistance` which can hold some number of float distances. These values each get interpolated when going to the fragment shader, and if any of them fall below 0 the fragment will be discarded. For a plane you can get inside vs outside by doing something like `distance = plane_distance_from_zero - dot(plane_normal, position)`. 

If you have a complex polygon or mesh to check against you'll need lots of clip planes, and you can only pass so many `ClipDistance`s. (Minimum 8) So you'll probably run into issues quickly like this. A circle would be very easy to do though, since that's just one distance.

If the clipping shapes are 2D you can also go through stencil buffers. Conceptually it shouldn't be too difficult to render an arbitrary poylgon to the stencil buffer and then use that as a mask for clipping. In practice our renderloop needs to be more adjustable for this, I think.

jkrumbiegel:
> If the clipping shapes are 2D you can also go through stencil buffers. Conceptually it shouldn't be too difficult to render an arbitrary poylgon to the stencil buffer and then use that as a mask for clipping. In practice our renderloop needs to be more adjustable for this, I think.

Yes this would have to be the mechanism for weird axis shapes, I think.

ffreyer:
Closing this in favor of #3958

PR #2786, opened 2023-03-19
Updates for Northstar & v3

SimonDanisch:


asinghvi17:
Nice!  How could we make the color update dynamically?  I'd like to try and add that kind of linkage for all atomic plots, to make animations really easy (if slow:D)

PR #2787, opened 2023-03-20
Minor fixes to documentation

DuncanBloemers:
# Description

Minor fixes to documentation, previously stated that coloured text is not possible, which is no longer true, now points to rich text functionality. Also standardised some use of the British spelling of colour to American (unfortunately).

## Checklist

### DIFF
--- a/docs/documentation/events.md
+++ b/docs/documentation/events.md
@@ -378,7 +378,7 @@ Furthermore you can wrap any of the above in `Exclusively` to discard matches wh
 
 ## Interactive Widgets
 
-Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can read about in the \myreflink{Blocks} section.
+Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can learn about in the [blocks and widgets](https://docs.makie.org/stable/examples/blocks/) section.
 
 ## Recording Animations with Interactions
 
--- a/docs/documentation/events.md
+++ b/docs/documentation/events.md
@@ -378,7 +378,7 @@ Furthermore you can wrap any of the above in `Exclusively` to discard matches wh
 
 ## Interactive Widgets
 
-Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can learn about in the [blocks and widgets](https://docs.makie.org/stable/examples/blocks/) section.
+Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can learn about in the \myreflink{Blocks} section.
 
 ## Recording Animations with Interactions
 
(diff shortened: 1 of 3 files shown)


docs/documentation/fonts.md
```diff
@@ -53,6 +53,8 @@ f
 
 ## Emoji and color fonts
 
-Currently, Makie does not have the ability to draw emoji or other color fonts.
+Currently, Makie does not have the ability to draw emoji.
```
asinghvi17, line 56: Color fonts are actually https://www.colorfonts.wtf/



that aren't rendered because of the way we render text in Makie.

I guess that colored text would be the correct name for what we're thinking of...

DuncanBloemers: Oh interesting I didn't know that I assumed it was synonymous with coloured text. IMO it would still be useful to have a pointer to the ability to change text colour since I imagine that's what most people would looking for in this context.

asinghvi17: Oh definitely - we'll just have to disambiguate the terms somewhere.

asinghvi17:
```suggestion
Currently, Makie does not have the ability to draw emoji or other color fonts.  

!!! note
    Color fonts refer specifically to fonts with color embedded.  Makie does support arbitrarily coloring text using the `rich` text API.

Comment (ffreyer) [reply docs/documentation/fonts.md:56]: Maybe it would be better to be more direct and write that Makie doesn't support multi-color characters with emojis and colorfonts being examples?

Comment (DuncanBloemers) [reply docs/documentation/fonts.md:56]: Updated to link that site as well as coloured text support.

File: docs/documentation/fonts.md [hunk 2]
```diff
@@ -53,6 +53,9 @@ f
...
 This is due to the implementation of text drawing in GLMakie and WGLMakie, which relies on signed distance fields that can only be used to render monochrome glyphs, but not arbitrary bitmaps.
+
+Multicolored fonts such as those in the context of the [color fonts](https://www.colorfonts.wtf/) project are currently not supported.
```
asinghvi17, line 59:
```suggestion
Multicolored _fonts_, like those in the [color fonts](https://www.colorfonts.wtf/) project, are currently not supported.

File: docs/documentation/events.md [hunk 3]
```diff
@@ -378,7 +378,7 @@ Furthermore you can wrap any of the above in `Exclusively` to discard matches wh
 
 ## Interactive Widgets
 
-Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can read about in the \myreflink{Blocks} section.
+Makie has a couple of useful interactive widgets like sliders, buttons and menus, which you can learn about in the [blocks and widgets](https://docs.makie.org/stable/examples/blocks/) section.
```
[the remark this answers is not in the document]
asinghvi17: This locks the version in correctly, which is useful.

PR #2795, opened 2023-03-21
fix lines on nvidia

SimonDanisch:
I'm not 100% sure if this is the correct fix! I think @ffreyer was intending this to be used differently.
But we do need to repeat first and last indices with `lines_adjecency`, see comment:
https://github.com/MakieOrg/Makie.jl/blob/master/GLMakie/src/glshaders/lines.jl#L82-L84
```
       # Duplicate the vertex indices on the ends of the line, as our geometry
       # shader in `layout(lines_adjacency)` mode requires each rendered
       # segment to have neighbouring vertices.
```
Not sure I understand the comment correctly in the shader, since we still do duplicate these indices, and from memory I do think that was necessary, since `lines_adjecency` needs to be always multiple of 4... 
I'm surprised though, that this seems to be an nvidia only problem, so there still seems to be more to it...

### DIFF
--- a/GLMakie/assets/shader/lines.geom
+++ b/GLMakie/assets/shader/lines.geom
@@ -525,10 +525,10 @@ void main(void)
     // CPU side by repeating the first and last points via the index buffer. It
     // just requires a little care further down to avoid degenerate normals.
     bool isvalid[4] = bool[](
-        g_valid_vertex[0] == 1 && g_id[0].y != g_id[1].y,
+        g_valid_vertex[0] == 1,
         g_valid_vertex[1] == 1,
         g_valid_vertex[2] == 1,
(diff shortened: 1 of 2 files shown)


ffreyer:
That's something I kept from the previous version:
https://github.com/MakieOrg/Makie.jl/blob/156e651ffd1b20eed3f511f925c1e8ac4eed1474/GLMakie/assets/shader/lines.geom#L60-L65

Not having this would break the anti-aliasing bits for line start/end I added here:
https://github.com/MakieOrg/Makie.jl/blob/cc7ef1015e97f57c5a6cdfb2606531d7c6bdc082/GLMakie/assets/shader/lines.geom#L484-L492

Maybe that's also what's causing problems?

ffreyer:
Yea this breaks the line start/end AA
Screenshot from 2023-03-21 14-06-29

PR #2799, opened 2023-03-22
add cheatsheet and use it for precompile

SimonDanisch:


src/cheatsheet.jl
```diff
@@ -0,0 +1,413 @@
+using LinearAlgebra, GeometryBasics
+using Makie.SparseArrays
+
+function cheatsheet_3d(rn, testimage)
```
asinghvi17, line 4:
```suggestion
function cheatsheet_3d(rn = randn(10), testimage = Makie.logo)
```
so that you can just call `Makie.cheatsheet_3d()`

SimonDanisch: yeah, i have to run and quickly wanted to spin up CI for the benchmarks, so this isn't the final version ;)

SimonDanisch:
Closing for now, will add more to cheatsheets when improving the docs

PR #2801, opened 2023-03-23
General improvements to RPRMakie`

asinghvi17:
# Description

The aim of this PR is to try and get RPRMakie up to par with the other backends in terms of basic functionality (saving, recording animations, reacting to all observables, etc.)

I also noticed some gaps in Makie's volume interface (it's not hooked up to `color_and_colormap!`, so `colorrange` is not automatic but `(0, 1)` unless specifically overridden).  Fixed some of those issues as well.

Additionally, a new example using [WaterLily.jl](https://github.com/weymouth/WaterLily.jl), a Julia-based fluid dynamics simulator, was added to the RPRMakie examples!

## Type of change

- [debatable] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
-using RPRMakie
+using RPRMakie, ProgressMeter
 using Makie, NIfTI, FileIO
 using GLMakie
 r = LinRange(-1, 1, 100)
```
lazarusA, line 4:
```suggestion
r = range(-1, 1, 100)
```

RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
...
 cube = [(x .^ 2 + y .^ 2 + z .^ 2) for x = r, y = r, z = r]
 
+RPRMakie.activate!(
+    iterations = 32,
```
lazarusA, line 8: 32 seems to much, if this is suppose to be interactive for the user, as for the example maybe even more than 100 will be ok, since is not running in CI anyway. So, I would say iterations=1, and a comment to the user to increase that number if he wants a nice render.

RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
...
+vp.colormap[] = :turbo
+
+scr = RPRMakie.Screen(ax.scene; iterations = 10)
+
+record(ax.scene, "volume_cube.mp4"; backend = RPRMakie, iterations = 15, framerate = 15) do io
+    @showprogress for a in LinRange(0, 3, 30)[2:end]
+        vp.absorption[] = Float32(a)
+        recordframe!(io)
+    end
+end
```
lazarusA, line 42: should we put this in a `begin end` block? so that the end user doesn't run this line by line.

RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
...
+lights = Makie.AbstractLight[
+    EnvironmentLight(1.0, load(RPR.assetpath("studio026.exr"))),
+    PointLight(Vec3f(10), RGBf(radiance, radiance, radiance * 1.1))
+]
+
+fig = Figure(; resolution=(1000, 1000));
+ax = LScene(fig[1, 1]; show_axis=false, scenekw=(lights=lights,));
+vp = Makie.volume!(
+    ax, 
+    #=0..3, 0..3.78, 0..3.18,
```
lazarusA, line 24:
```suggestion
```

RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
...
+    brain=#
```
lazarusA, line 25:
```suggestion
```

RPRMakie/examples/volume.jl
```diff
@@ -1,16 +1,44 @@
...
+    absorption=1f0, 
+    colormap = Reverse(:plasma)
+)
+vp.absorption[] = 0.5f0
+vp.colormap[] = :turbo
+
+scr = RPRMakie.Screen(ax.scene; iterations = 10)
+
+record(ax.scene, "volume_cube.mp4"; backend = RPRMakie, iterations = 15, framerate = 15) do io
+    @showprogress for a in LinRange(0, 3, 30)[2:end]
```
lazarusA, line 38:
```suggestion
    @showprogress for a in range(0, 3, 30)[2:end]
```

lazarusA (commented):
that's just some comments, I will need to actually test the branch and see what I do not like :D in order to give a full assessment.

RPRMakie/examples/volume.jl
asinghvi17: It's fine even if run line-by-line - I was using it that way on VSCode.

RPRMakie/examples/volume.jl
asinghvi17: Wanted to keep that there for the brain example when I fix it.  Might split that out into a separate file though.

RPRMakie/examples/volume.jl
asinghvi17: any particular reason for the change in syntax?  LinRange always seemed more expressive to me

RPRMakie/examples/volume.jl
asinghvi17: Fair enough.  5 should be fine in that case (without that, there's way too much noise in the render).
```suggestion
RPRMakie.activate!(
    iterations = 5, # increase this if you want a better render!
```

RPRMakie/examples/volume.jl
lazarusA: just easier and faster to write.

RPRMakie/src/meshes.jl
```diff
@@ -142,13 +142,23 @@ function to_rpr_object(context, matsys, scene, plot::Makie.Surface)
...
+            z
+        end
+    end
+    
+    lift(color_obs) do color
+        material = mesh_material(context, matsys, plot, color)
+        set!(rpr_mesh, material)
+    end
     return rpr_mesh
 end
```
lazarusA, line 164: could we have also observables hook up here, so that surface also works with `record`? as in your volume example.

RPRMakie/examples/waterlily.jl
```diff
@@ -0,0 +1,62 @@
...
+
+# Set up viz data and figure
+dat = Observable(data_func(sim))
+
+fig, ax, plt = Makie.volume(
+    dat,
+    colorrange = (π,4π),
+    algorithm = :absorption,
+    absorption = 1f0,
+    axis = (type = LScene, scenekw = (lights = lights,))
```
lazarusA, line 43: I don't see the definition of lights within this file.

RPRMakie/examples/volume.jl
asinghvi17: Rather be expressive about range IMO, will probably keep this as LinRange for now and maybe switch later.

RPRMakie/src/meshes.jl
asinghvi17: Does it not already?

RPRMakie/examples/waterlily.jl
asinghvi17: Good point, will add that.  Thanks!

PR #2804, opened 2023-03-24
Add weights to `hexbin`

asinghvi17:
# Description

Allows a `weights` keyword in hexbin, which supports `nothing` or `StatsBase.UnitWeights` (each observation gets a weight of one) or any `AbstractVector{<: Real}` or `StatsBase.AbstractWeights` which provide arbitrary weights per observation.  These values are stored as Float64 in the Dict.

exp-scaling
Weights with scaling

## Type of change

## Checklist

This was previously #2712.


asinghvi17:
Bump - CI passes, a doc example was added, and internal testing has also worked well.

src/stats/hexbin.jl
```diff
@@ -141,7 +150,7 @@ function Makie.plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Point2}}})
                 for ix in 0:_nx-1
                     _x = xoff + 2 * ix * xspacing + (isodd(iy) * xspacing)
                     _y = yoff + iy * yspacing
-                    c = get(d, (ix, iy), 0)
+                    c = get(d, (ix, iy), 0e0)
```
SimonDanisch, line 153: Why this instead of `0.0`?

PR #2809, opened 2023-03-26
add a current_axis() method to handle case when there's no current figure

jwahlstrand:
# Description

In GLMakie or CairoMakie, when you call current_axis() before creating a plot, an error is thrown:
```
julia> using GLMakie
julia> current_axis()

ERROR: MethodError: no method matching current_axis(::Nothing)
Closest candidates are:
  current_axis() at ~/.julia/packages/Makie/z1gJs/src/figures.jl:37
  current_axis(::Figure) at ~/.julia/packages/Makie/z1gJs/src/figures.jl:38
Stacktrace:
 [1] current_axis()
   @ Makie ~/.julia/packages/Makie/z1gJs/src/figures.jl:37
 [2] top-level scope
   @ REPL[2]:1
```

This PR adds a method so that `current_axis()` returns `nothing` if you call it before creating a figure, like `current_figure()`. Alternatively, a more helpful error could be thrown. I also expanded the docstrings for the `current_axis` and `current_figure` functions a bit. 

## Type of change

## Checklist


src/figures.jl
```diff
@@ -35,6 +35,7 @@ current_figure!(fig) = (_current_figure[] = fig)
 
 "Returns the current active axis (or the last axis that got created)"
 current_axis() = current_axis(current_figure())
+current_axis(::Nothing) = nothing
```
asinghvi17, line 38: This should probably just be a more descriptive error, something along the lines of "No axis exists yet, please create one first" or something.

Thanks for the PR!

jkrumbiegel:
The error comes from `current_figure()` returning `nothing`, right? So the error should be `No figure exists, yet` or the return value should just be `nothing` like it is here. Whichever you prefer. Maybe returning `nothing` is fine as well, if `current_figure()` can return that, too.

jkrumbiegel:
The error comes from `current_figure()` returning `nothing`, right? So the error should be `No figure exists, yet` or the return value should just be `nothing` like it is here. Whichever you prefer. Maybe returning `nothing` is fine as well, if `current_figure()` can return that, too.

jwahlstrand:
Yeah, I picked the "return `nothing`" option because that's what `current_figure()` does in the same situation. I think a REPL user will figure out what's going on from seeing `nothing` is returned. But I can see the benefit of having this throw an error too. I'm new here -- is there a typical practice for Makie?

jkrumbiegel:
I think returning `nothing` can sometimes be nicer because `!== nothing` is better than a try catch block, which an erroring function forces you to use. So let's just stick with that.

jkrumbiegel (approved):

PR #2810, opened 2023-03-27
Skip limit reset on plot insertion when autolimits disabled

ffreyer:
# Description

Currently when a plot is added to an axis the limit resetting machinery always runs. This means that if you either set limits yourself (xlims!, ylims!) or any one of the inserted plots has valid limits, axis limits will reset. If you are dynamically inserting plots this can be rather annoying. (E.g. if you insert plots on click you'll end up resetting limits with every click.)

I adjusted things to skip limit resets when `xautolimits = false, yautolimits = false` in the inserted plot. Alternatively we could also add a new attribute for this. (Maybe one that is removed before the plot is constructed?)

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -801,7 +801,10 @@ function Makie.plot!(
     # adjust the limit margins in those cases automatically.
     needs_tight_limits(plot) && tightlimits!(la)
 
-    if is_open_or_any_parent(la.scene)
+    xauto = to_value(get(allattrs, :xautolimits, true))
+    yauto = to_value(get(allattrs, :yautolimits, true))
+
+    if is_open_or_any_parent(la.scene) && (xauto || yauto)
         reset_limits!(la)
     end
     plot


src/makielayout/blocks/axis.jl
```diff
@@ -801,7 +801,10 @@ function Makie.plot!(
     # adjust the limit margins in those cases automatically.
     needs_tight_limits(plot) && tightlimits!(la)
```
asinghvi17, line 802: Should this also be moved into the same block that `reset_limits!` is in?

jkrumbiegel:
> Currently when a plot is added to an axis the limit resetting machinery always runs

Does it? I think it only runs if the figure is being displayed right now (because you'd expect some plot you're viewing to react). In other cases, the update should be deferred until display. That's what that function `update_before_display!` or whatever it's called is for. I don't particularly like it but it seemed to make common scripts more performant by skipping these updates.

jkrumbiegel:
This could also be a setting on the axis instead? Currently, if manual limits are set, these will not be touched when adding new plots. So avoiding a reset is equivalent to manually setting the current limits, which could be done by some function `fix_limits!(ax)` instead of `limits!(ax, ax.finallimits[])`

ffreyer:
> This could also be a setting on the axis instead?

Having the option per plot gives you a bit more control. You could have some that do reset limits and some that do not. Maybe that can also be useful?



> That's the difference between `reset_limits!` and `autolimits!` (the first one only updates automatically set limits or those you get by interacting with the axis I think). So avoiding a reset is equivalent to manually setting the current limits, which could be done by some function `fix_limits!(ax)` instead of `limits!(ax, ax.finallimits[])`

That would work too, though `plot!(..., reset_limit = false)` seems a bit easier to understand to me. `fix_limits` would also make you unable to jump back to previously set limits, while skipping the `reset_limits!` call would still allow that.

jkrumbiegel:
Yes it could be useful. I still don't really like that we have no "process" behind adding attributes to plot objects, that's my main reason for hesitating.

ffreyer:
Here's an example for what I want to adjust:

```julia
fig = Figure()
ax = Axis(fig[1,1])
xlims!(ax, -3, 3)
ylims!(ax, -3, 3)
on(events(fig).mousebutton, priority = 1000) do event
    if (event.button == Mouse.right) && (event.action == Mouse.press)
        scatter!(ax, randn(50), randn(50), reset_limits = false)
    end
end
fig
```

With the current solution this leaves limits and the current view alone when right clicking and still allows jumping back to the originally set limits.

> Yes it could be useful. I still don't really like that we have no "process" behind adding attributes to plot objects, that's my main reason for hesitating.

I don't know if it makes it better or worse but I'm removing it on plot insertion. So it doesn't end up in `plot.attributes`.

jkrumbiegel:
> I don't know if it makes it better or worse but I'm removing it on plot insertion

A little better I'd say? Still, these attributes have a couple undesirable properties right now. There's no error if you pass an unsupported one and "supported" can mean, acts like a keyword or acts like a real observable attribute.

jkrumbiegel:
We have `xautolimits = false` etc. but those were added pretty ad-hoc and should be refactored into something integrated better with nested boundingboxes etc. (They are not at all currently)

ffreyer:
After a quick health check this still seems to be a partially valid problem. Specifically, there is no way to avoid resetting the plot with the second `scatter!` here:
```julia
f, a, p = scatter(Rect2f(0,0,1,1))
display(f)
# zoom
scatter!(a, Point2f(0.5) xautolimits = false, yautolimits = false) 
```
If we add `x/ylims!()` the plot resets to the set limits instead.

The cause for this issue was a  plot that is inserted on-click. Iirc the clickd position acted as a seed to some code generating a bunch of lines, like an initial value for a chaotic differential equation. 

@jkrumbiegel Have your thoughts on this changed?

SimonDanisch:
I actually thought quite a bit about it, and decided that the easiest API for this might just be to do:
`scatter!(ax.scene, ...)`

ffreyer:
Seems hacky to me and I also dream of limits becoming a scene thing so that e.g. Camera3D can use them too. If that happens this problem would reappear.

PR #2813, opened 2023-03-28
Make scene.visible trigger render updates

ffreyer:
# Description

Reported on discord.

We should probably also have `scene.visible` propagate to other child scenes because right now
```julia
f, a, p = scatter(rand(10))
a.blockscene.visible[] = false
a.scene.parent === a.blockscene
# true
```
results in

Screenshot from 2023-03-28 14-18-25

I also noticed we're not using `Observables.inputs` (for map!/map at least)
https://github.com/JuliaGizmos/Observables.jl/blob/28bcea62fc74d34f132dd481804ba2a0ba0efcbe/src/Observables.jl#L515
Don't we want that for better bookkeeping?

## Type of change


SimonDanisch:
Oh, I never merged & tagged https://github.com/JuliaGizmos/Observables.jl/pull/105...That should bring back the book keeping!

Issue #1907, opened 2022-05-11
Menu appears in bottom left corner


tecosaur:
Hi,

With GLMakie 0.6.0 / Makie 0.17.0 (and also having tried `HEAD`), I'm seeing Menu text/dropdowns all appear (overlapping) in the bottom right corner.

![image](https://user-images.githubusercontent.com/20903656/167753947-2ea6a405-586b-432c-8082-17a4f9596f22.png)

I've got no idea what's going on here, other than that this seems to be occurring consistently.

jkrumbiegel:
Hm I thought I had fixed that on master a while ago. Blocks like Menu have a different Scene structure now and this changed some coordinate calculations for where things should be. @SimonDanisch did anything in that regard change recently? I definitely tested all Blocks to work with GLMakie before I merged that PR.

SimonDanisch:
Oh, I may have incorrectly resolved some merge conflicts, e.g. here:
https://github.com/JuliaPlots/Makie.jl/pull/1818/files#diff-d8b9416e400a74c92f67c1c5d3b0685bc4a39b153f4b66b7890311ab1b38421e
![image](https://user-images.githubusercontent.com/1010467/167815458-320fc45d-be35-40ef-9574-71dfe356a361.png)

I made a PR, and as far as I can tell, things work again!

PR #2815, opened 2023-03-28
Don't move scenes to the origin (fix #2814)

### LINKED ISSUE #2814: Textbox contents appear in the wrong place
When running the text box example as used in the [docs](https://docs.makie.org/stable/examples/blocks/textbox/#textbox)
```julia
using GLMakie

f = Figure()
Textbox(f[1, 1], placeholder = "Enter a string...")
Textbox(f[2, 1], width = 300)

f
```
the content of the text boxes is displayed in the lower left-hand corner of the figure, not in the boxes. 
image

This problem seems to be reproducable in the documentation build environment as the same symptoms can be seen in the example image.

I'm on GLMakie v0.8.3 and have experienced this problem on Julia 1.8.5 and 1.9.0-rc1

ffreyer:
# Description

This has been in Makie since day 1 and I don't know why. Setting the origin to zero means moving the scene to the bottom left corner. Why?

I checked 

```julia
fig = Figure()
Axis(fig[1, 1])
p = scatter!(rand(10))
Axis3(fig[1, 2])
LScene(fig[1, 3])
Box(fig[2, 1])
Button(fig[2, 2])
Colorbar(fig[2, 3], tellwidth=false)
IntervalSlider(fig[3, 1])
Label(fig[3, 2])
Slider(fig[3, 3])
Legend(fig[4, 1], [p], ["test"])
Menu(fig[4, 2], options = string.(1:3))
Textbox(fig[5, 1])
Toggle(fig[5, 2])
Menu(fig[1, 4][1, 1])
fig
```

And everything seems fine here. Same with the example from https://docs.makie.org/stable/tutorials/layout-tutorial/

I may have intentionally removed this in MakieLayout in #1818 which caused a similar problem with #1907. I think this was before `blockscene` which probably resulted in less consistent scene sizes. Maybe there `zero_origin` made sure to place a full size scene at the origin and not an offset. But I don't think this is a problem anymore, and if it is I think it should be handled in makielayout.

## Type of change

Bug fix if we consider this internal, breaking if not.

## Checklist

### DIFF
--- a/src/scenes.jl
+++ b/src/scenes.jl
@@ -326,7 +326,7 @@ function Scene(
         kw...
     )
     if isnothing(px_area)
-        map!(zero_origin, child, child_px_area, parent.px_area)
+        map!(identity, child, child_px_area, parent.px_area)
     elseif !(px_area isa Observable) # observables are assumed to be already corrected against the parent to avoid double updates
         a = Rect2i(px_area)
         on(child, pixelarea(parent)) do p


SimonDanisch:
>Why?

It has been a long time, but I think it's because any subscene is treated, looking at it alone, like a parent window with now offset.

The scene tutorial does not look right: https://docs.makie.org/previews/PR2815/tutorials/scenes

Maybe we need to fix this in the blocks?

SimonDanisch:
Maybe just here https://github.com/MakieOrg/Makie.jl/blob/2116627b5caed21ddf0e3bbd2103f55d5679d881/src/makielayout/blocks/textbox.jl#L17 for  #2814 ?

ffreyer:
> > Why?
> 
> It has been a long time, but I think it's because any subscene is treated, looking at it alone, like a parent window with now offset.

If you just need the width and height you can easily get that either way. Getting relative offsets from absolute ones is also easier than the other way around.
Mouse positions are based on the root window offsets to the root window are useful here. I think glviewport needs absolute offsets too.
The only reason I see for having zero_origin would be relative scene offsets, but where would we need them?

I'd say camrelative being broken without zero_origin is just something that needs to be fixed rather than an argument for keeping it.

PR #2822, opened 2023-03-30
Allow `to_cairo_image` to take any abstract matrix of real numbers

asinghvi17:
# Description

This was a bug I found in a recipe which was defined for custom abstract arrays with e.g. UInt8 or UInt16 eltypes (i.e., real but not float).

## Type of change

## Checklist

Needs tests.


asinghvi17:
Should we also remove the UInt32 dispatch?  I'm not sure where it's used, but we could always reinterpret the array instead.

SimonDanisch:
To `to_rgba_image` got removed in  https://github.com/MakieOrg/Makie.jl/pull/2900 and UInt16 works in that PR with CairoMakie + heatmap (and other primitives).

PR #2824, opened 2023-03-30
Don't record states for now

SimonDanisch:
Since it seems to take too long on CI right now with the new JSServe version


SimonDanisch:
fixed in #2661

PR #2840, opened 2023-04-04
make Makie.jl threadsafe

SimonDanisch:
This makes makie threadsafe (not necessarily the backends, though).
We still need to figure out how to better handle current figure (currently there is no way to create any figure, without making it globally the current figure).

### DIFF
--- a/CairoMakie/src/CairoMakie.jl
+++ b/CairoMakie/src/CairoMakie.jl
@@ -34,6 +34,6 @@ function __init__()
     activate!()
 end
 
-include("precompiles.jl")
+# include("precompiles.jl")
 
 end
--- a/CairoMakie/src/CairoMakie.jl
+++ b/CairoMakie/src/CairoMakie.jl
@@ -34,6 +34,6 @@ function __init__()
     activate!()
 end
 
-# include("precompiles.jl")
+include("precompiles.jl")
 
 end
--- a/GLMakie/src/precompiles.jl
+++ b/GLMakie/src/precompiles.jl
@@ -25,7 +25,7 @@ let
                 display(plot(x); visible=false)
             catch
             end
-            Makie._current_figure[] = nothing
+            Makie.CURRENT_FIGURE[] = nothing
             empty!(atlas_texture_cache)
             closeall()
             @assert isempty(SCREEN_REUSE_POOL)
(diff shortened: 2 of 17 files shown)


asinghvi17:
Huh, was it only fonts which needed handling?  We should probably also make sure FreeType doesn't need per-thread contexts like some other libraries...@SimonDanisch is this the case?

With separate screens per thread, CairoMakie might be (mostly) fine.

SimonDanisch:
Yeah, this is with: https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/pull/78

I wonder how we should change figure @jkrumbiegel. I think we have two options: 
`Figure(; no_current_figure=true)` or something like that, or don't make `Figure()` set the current default figure, and only set it when created by e.g. `scatter(1:4)`.

jkrumbiegel:
> Yeah, this is with: [JuliaGraphics/FreeTypeAbstraction.jl#78](https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/pull/78)
> 
> I wonder how we should change figure @jkrumbiegel. I think we have two options: `Figure(; no_current_figure=true)` or something like that, or don't make `Figure()` set the current default figure, and only set it when created by e.g. `scatter(1:4)`.

I think in a threaded context people should not use the implicit figure / axis mode? Because that can easily be mutated by some other thread, even under lock. I think it is too useful for a normal interactive plotting workflow to remove, though.

SimonDanisch:
>I think in a threaded context people should not use the implicit figure / axis mode

Yes, but currently there is currently no way to not use it, since the most basic Figure constructor sets it ;)

jkrumbiegel:
It can set it, but if the threaded code never makes use of the implicit figure arg then it should still work right?

jkrumbiegel:
Hm or can we make `current_figure` etc a vector with as many entries as there are threads?

jkrumbiegel:
Maybe these functions would need a way to specify the thread ID then? And you can mutate across threads if you choose?

SimonDanisch:
I think we should rather have `thread_local_set_theme!`? Usually one doesnt want to set the theme for a specific thread, but just the one youre working in?

asinghvi17:
Another edge case:

```julia
with_theme(...) do
@threads for i in ...
    fig = Figure(...)
    save("...", fig)
end
end
``` 

I think @SimonDanisch 's option makes sense here, where you would have to use threadsafe functions for maximum performance.  The current functions would then mutate across all threads, and we could have either thread safe versions or dispatches.

SimonDanisch:
I'm going to merge this, since its a solid improvement for multithreading.
Will follow up a PR, to clean up current_figure + setting the theme.

PR #2859, opened 2023-04-14
Docs community videos

### LINKED ISSUE #2853: Links to videos and presentations on (or using) Makie
It would be neat if there were a page in the docs which links to Makie tutorials and videos out in the wild!  Right now I think you have to use Google to find many of them - there's no link from e.g. the docs to the Julia Data Science tutorial which was quite neat.

bperard:
# Description

Fixes #2853

Added communityvideos.md page to docs to create a hub for links to community content within the documentation.

## Type of change

## Checklist

### DIFF
--- a/docs/communityvideos.md
+++ b/docs/communityvideos.md
@@ -1,0 +1,77 @@
+@def order = 4
+
+# Community Videos
+
+## JuliaCon
+
+@@box-container
+  @@box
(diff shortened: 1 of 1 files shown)


jkrumbiegel (commented):
Thank you for the PR! What about just calling the page `Videos`, I think it might read a bit nicer in the navigation and then we can have subsections there if we want. But already the JuliaCon and In the Wild sections look good :)

PR #2860, opened 2023-04-17
Change "x-limits" to "$(dimsym)-limits".

KronosTheLate:
# Description
I quite often get the error `Found invalid x-limits (0.0f0, 1.331f0) for scale log10 which is defined on the interval 0.0..Inf (open)` when setting the **y-scale** of plots. Below is the plot which produced that message when doing `ax.yscale=log10`, despite having run `ylims!(1e-100, nothing)`. Note that the limits in the error are in fact y-limits.
image

I believe this should fix the erroneous text (warning about x-axis). However, no error should be raised, as I explicitly set legal y-limits. I do not know what needs to be changed to fix that.

## Type of change


asinghvi17 (approved):
This PR looks good to me!  The underlying issue seems to be something else though, but we can merge this as is.

PR #2868, opened 2023-04-18
[WIP] Dynamic `PlotList` recipe which plots arbitrary arrays of PlotSpec

asinghvi17:
# Description

As discussed in MakieCon 2023 Day 1...will add an example in a bit.

This currently takes the form of a recipe, but we could make it a normal function which is just passed an Observable list of PlotSpec.

```julia

fig = Figure()
ax = Axis(fig[1, 1])

pl = plotlist!(
    ax,
    [PlotSpec{Lines}(0..1, sin.(0:0.01:1); color = :blue), PlotSpec{Heatmap}(0..1, 0..1, Makie.peaks(); transformation = (; translation = Vec3f(0, 0, -1)))]
)

fig
```
plot_1029

```julia
pl[1][] = [PlotSpec{Lines}(0..1, sin.(0:0.01:1); color = :red), PlotSpec{Heatmap}(0..1, 0..1, Makie.peaks(); transformation = (; translation = Vec3f(0, 0, -1)))]

fig
```
plot_1030

```julia
pl[1][] = [
    PlotSpec{Lines}(0..1, sin.(0:0.01:1); color = Makie.resample_cmap(:viridis, 101)), 
    PlotSpec{Surface}(0..1, 0..1, Makie.peaks(); transformation = (; translation = Vec3f(0, 0, -1))),
    PlotSpec{Poly}(Rect2f(0.45, 0.45, 0.1, 0.1)),
]

fig
```
plot_1031

```julia
pl[1][] = [
    PlotSpec{Surface}(0..1, 0..1, Makie.peaks(); colormap = :viridis, transformation = (; translation = Vec3f(0, 0, -1))),
]

fig
```
plot_1032

## Type of change

## Checklist


asinghvi17:
This is functionally complete, but please try whatever you think might break it and let me know!

I think we might also want to add eg palette support but that would need a refactor of the axis code.

pepijndevos:
> This currently takes the form of a recipe, but we could make it a normal function which is just passed an Observable list of PlotSpec.

I tried this out and immediately ran into problems where things like color cycling don't work because it's a recipe. The below example produces only black lines, instead of cycling through colors if you'd call `lines!` twice.

`plotlist([PlotSpec{Lines}(0:5, 0:5), PlotSpec{Lines}(0:5, ones(6))])`
image

pepijndevos:
I tried this some more, because I really need this functionality.

First I tried to make it not a recipe, but got weird errors.

Then I tried just using the recipe with explicit attributes, which "works" in that it doesn't error, but rather leaves dangling plot elements (that don't resolve themselves if you change back the Observed) so once you remove a plot it's just broken.

It seems there was also an off-by-one error in `length(old_plotspec_types):-1:(length(plotspec_types)+1)` on line 64

My recipe is something like

```julia
function Makie.plot!(plt::SignalPlot{<:Tuple{<:AbstractArray}})
    Main.plt = plt
    ms = plt[1]
    specs = lift(ms, values(plt.attributes)...) do ms, attrs...
        pairs = zip(keys(plt.attributes), attrs)
        [PlotSpec{SignalPlot}(item; pairs...) for item in ms]
    end
    plotlist!(plt, specs)
end
```

staticfloat:
I tried re-working the fundamental machinery in here to use a `Dict` instead of an array, so that we don't have to track order of plots (since as far as I can tell, order doesn't matter):

```
function update_plot_with_plotspec!(cached_plot::AbstractPlot, spec::PlotSpec)
    # Update args in cached `input_args` list
    for arg_index in eachindex(spec.args)
        cached_plot.input_args[arg_index].val = spec.args[arg_index]
    end
    # Update attributes
    for (attribute, new_value) in pairs(spec.kwargs)
        update_attributes_inplace!(cached_plot, attribute, new_value)
    end
    
    # notify all args, without broadcasting
    map(notify, cached_plot.input_args)
    # notify all accessed kwargs
    map(_notify!, (getindex(cached_plot, attr) for attr in keys(spec.kwargs)))
end

function Makie.plot!(p::PlotList{<: Tuple{<: AbstractArray{<: PlotSpec}}})
    # Cache plots here so that we aren't re-creating plots every time;
    # if a plot still exists from last time, update it accordingly.
    # If the plot is removed from `plotspecs`, we'll delete it from here
    # and re-create it if it ever returns.
    cached_plots = IdDict{Any,AbstractPlot}()

    lift(p[1]) do plotspecs
        Main.p = p
        println()
        println("----------------------------------------")
        plots_to_delete = Base.Set(collect(keys(cached_plots)))

        # First, scan through `plotspecs`, creating or updating all plots
        # specified by `plotspecs`, marking previously-extant plots as not
        # to be deleted via `plots_to_delete`
        for plotspec in plotspecs
            if !haskey(cached_plots, plotspec.args[1])
                # Create new plot, store it into our `cached_plots` dictionary
                cached_plots[plotspec] = plot!(p,
                    typeof(plotspec).parameters[1],
                    Attributes(plotspec.kwargs),
                    plotspec.args...,
                )
            else
                update_plot_with_plotspec!(cached_plots[plotspec], plotspec)
            end

            # Because this plotspec appeared in `plotspecs`, it should not
            # be deleted, so remove it from our deletion queue.
            delete!(plots_to_delete, plotspec)
        end

        # Next, delete all plots still extant in in `plots_to_delete`
        for plotspec in plots_to_delete
            idx = findfirst(==(cached_plots[plotspec]), p.plots)
            if idx !== nothing
                deleteat!(p.plots, idx)
            end
            delete!(cached_plots, plotspec)
        end
    end
end
```

But for some reason I can't figure out, updates don't flow through the system; my plots remain static.

SimonDanisch:
While reviewing & figuring out your approaches, I've rewritten the implementation.
Caching needs to be based on the argument + attribute types of the plot spec, because that's what needs to be the same to be "updatable". Also, I now plot directly into the scene and remove plots from that scene, because if we add the plots nested inside the recipe plot type, it gets a bit more complicated to `delete!`.

I also simplified the updating mechanism for now, removing nested attributes, which aren't that well supported within Makie yet anyways and shouldn't be needed for most plots.
The transformation argument is maybe the biggest exception, but that seems to be broken right now?

Anyways, I also played around with a simple API to make it a bit nicer to use:
```julia
import Makie.PlotspecApi as P
plots[] = [
    P.image(0 .. 1, 0 .. 1, Makie.peaks()),
    P.poly(Rect2f(0.45, 0.45, 0.1, 0.1)),
    P.lines(0 .. 1, sin.(0:0.01:1); linewidth=10, color=Makie.resample_cmap(:viridis, 101)),
]
```

SimonDanisch:
@pepijndevos not sure what to do about the cycling, since that needs the axis.
I've been wanting to make cycling part of the scene, which should fix this... Otherwise, I guess we need to manually re-implement the cycling inside the recipe, not sure how to do this correctly otherwise.

staticfloat:
Yes, I agree. The `PlotList` functionality feels much more "top-level" than most recipes; I would expect other recipes to fit inside of it, so having the `PlotList` plotting directly interact with axes makes sense to me.

SimonDanisch:
This isn't 100% ready, but since sd/beta-20 had some problems with the PlotSpec implementation which are cleaned up in here, I'm going to merge this and polish the rest in sd/beta-20.
For sd/beta-20, the remaining work is also just polish and testing, so I might just do it in one go.

PR #2872, opened 2023-04-19
Scatterjitter

behinger:
# Description

Fixes https://github.com/MakieOrg/AlgebraOfGraphics.jl/issues/253

Adds weighted-jitter along the density of the distribution

## Type of change
image

## Checklist

### DIFF
--- a/docs/examples/plotting_functions/scatterjitter.md
+++ b/docs/examples/plotting_functions/scatterjitter.md
@@ -1,0 +1,29 @@
+# scatterlines
+
+{{doc scatterlines}}
+## Examples
+
+\begin{examplefigure}{svg = true}
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+
+
+y = vcat([rand(k.*50).*5*k.+k for k = 1:5]...)
+
+f = Figure()
+	for (i,type) = enumerate([:uniform,:pseudorandom,:quasirandom])
+        for (j,clamp) = enumerate([0.,0.3])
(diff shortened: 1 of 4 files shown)


ffreyer:
I think it would be better to have this as function you can apply to your data rather than a new plot type. E.g. `scatter(jitter(xs), ys)`. That could then be reused with lines and other plotting primitives without the need for extra recipes.

behinger:
could such a function work with AoG?

PS: I made something related with Dodge, e.g. `Dodge(Scatter,...)` - but the other way round ;)
https://github.com/MakieOrg/AlgebraOfGraphics.jl/issues/393#issuecomment-1417789588

behinger:
further, jitter would need access to both `x` & `y`, x to jitter, y to calculate the kde

ffreyer:
`jitter(xs, ys)` and/or `jitter(ps)` would be fine too

behinger:
It is still unclear to me how this would work with AoG.jl

behinger:
ok- then I would have to rewrite it though, to follow the AoG syntax. This is unlikely to happen in the next months, pretty swamped.

Here is an example:
https://github.com/MakieOrg/AlgebraOfGraphics.jl/blob/master/src/transformations/smooth.jl

I could also move the whole recipe to AoG

asinghvi17:
This is implemented in https://github.com/asinghvi17/SwarmMakie.jl as an extra algorithm for beeswarm plots :)

PR #2878, opened 2023-04-20
Allow `nothing` to be passed to high/lowclip

asinghvi17:
in `color_and_colormap!`

# Description

Loosens type constraints on high/lowclip.

Needs test.  This doesn't work with [W]/GLMakie, but works for eg Cairo.

## Type of change

## Checklist

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -16,10 +16,10 @@ function color_and_colormap!(plot, intensity = plot[:color])
     if intensity[] isa Union{Number, AbstractArray{<: Number}}
         @converted_attribute plot (colormap,)
         replace_automatic!(plot, :highclip) do
-            lift(last, plot, colormap)
+            map!(last, plot, Observable{Any}(), colormap)
         end
         replace_automatic!(plot, :lowclip) do
-            lift(first, plot, colormap)
+            map!(first, plot, Observable{Any}(), colormap)
         end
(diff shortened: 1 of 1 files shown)


SimonDanisch:
That's not fixing the root cause and also, nothing shouldn't be the sentinel for "do this automatically" ;) 
I'll make sure it works in https://github.com/MakieOrg/Makie.jl/pull/2900

jkrumbiegel:
It wasn't the sentinel for `automatic` though, it meant "don't use a custom high-clipping color, just use the highest value of the colormap as default behavior

SimonDanisch:
> "don't use a custom high-clipping color, just use the highest value of the colormap as default behavior"

Yes, that's what I meant. 
I've been considering that case in the PR, which is why now high/low-clip is `Union{Automatic, RGBAf}`:
https://github.com/MakieOrg/Makie.jl/blob/daf7c7b07d3198b184261048393f7574b5ed6bc0/src/colorsampler.jl#L185-L186
Colorbar should be much easier to implement now, by looking into `calculated_color` and getting a definitive answer about the kind of color the plot is mapped to.

SimonDanisch:
Thinking more about it, I think one can argue for both `nothing` and `automatic` to be the right sentinel... 
From the backends view, we actively "calculate" a color when going out of `colorrange`.
From a users view, nothing might be make more sense, although it could also indicate a transparent color I think.

Issue #473, opened 2019-04-16
Stroke doesn't work for rectangles


c42f:
As noted by @SimonDanisch on slack, stroke is broken for rectangles.

This is due to a workaround which was added to make histograms look nice: basically antialiasing is disabled by always setting the signed distance to 1 in that case. See https://github.com/JuliaPlots/GLMakie.jl/pull/30 and https://github.com/JuliaPlots/GLMakie.jl/blob/9f4f6253d7c394140ea11ef0420d357f1cdfae5d/src/GLVisualize/assets/shader/sprites.geom#L158-L161

There's a solution for rectangles, which is to decompose the SDF into a pair of SDFs for the parallel sides, do the anisotropic transform on those, and recompose them with a `min` function after the transform.

Having done this, we then hit another problem: the blending of histogram boxes should result in uniform looking gaps between the boxes. Currently this is not the case for some reason.

---
Side note: 

A few weeks ago I started trying to fix this problem in general and quickly ran into trouble.  See the cjf/aniso-sdf-shapes for some initial work (https://github.com/JuliaPlots/GLMakie.jl/commit/7e5089157e8f4a51d9ce47ae372eb5b60c40420d). The problem is that the gradient of an SDF is not continuous and transforming it by an anisotropic transform is kind of nonlocal / nonsmooth.

I think the problem is unsolvable in general without ditching SDFs completely, but there might be approximations which get us close enough for thin strokes. Particularly adapting the approach of ["Improved Corners with Multi‐Channel Signed Distance Fields" by Chlumský, Sloup and Šimeček (2018)](https://doi.org/10.1111/cgf.13265) to anisotropic transformations could work out.

PR #2882, opened 2023-04-21
Fix stroke for rect markers

### LINKED ISSUE #2880: marker stroke broken for `marker=Rect`
```julia
using GLMakie

fig = Figure()

scatter(fig[1,1], rand(5), marker=Rect)
scatter(fig[1,2], rand(5), marker=Rect, strokewidth=5)
scatter(fig[2,1], rand(5), marker=:rect)
scatter(fig[2,2], rand(5), marker=:rect, strokewidth=5)
```

image

Note that the marker for `Rect` gets bigger as `strokewidth` increases. `strokecolor` has no effect in this case.

(`:circle` and `Circle` both work as expected)

ffreyer:
# Description

Fixes #2880, #473

## Type of change

## Checklist


GLMakie/assets/shader/distance_shape.frag
```diff
@@ -138,7 +138,7 @@ void main(){
     else if(shape == ROUNDED_RECTANGLE)
         signed_distance = rounded_rectangle(f_uv, vec2(0.2), vec2(0.8));
     else if(shape == RECTANGLE)
-        signed_distance = 1.0; // rectangle(f_uv);
```
SimonDanisch, line 141: Hm, I remember I had all kind of artifacts with `rectangle(f_uv)` so I gave up and just used 1.0...
Will need to look at the reference images and try it out to see how bad it actually is.

GLMakie/assets/shader/sprites.geom
```diff
@@ -164,11 +164,7 @@ void main(void)
     // Compute required amount of buffering
     float sprite_from_viewport_scale = 1.0 / viewport_from_sprite_scale;
     float bbox_buf = sprite_from_viewport_scale *
-                     (// Hack!! antialiasing is disabled for RECTANGLE==1 for now
-                      // because it's used for boxplots where the sprites are
-                      // long and skinny (violating assumption 1 above)
```
ffreyer, line 169: I don't see a difference between master and this: (master | pr)
Screenshot from 2023-04-30 10-19-31Screenshot from 2023-04-30 10-14-30

ffreyer: This is coming from https://github.com/JuliaPlots/GLMakie.jl/pull/30, but without images I don't know what this actually fixed.

ffreyer:
That's a general issue I think. The default marker and `Circle` have this issue too: 
Screenshot from 2023-05-05 18-11-46
Probably a result of `f_viewport_from_u_scale` being a single value rather than an xy(z?) scale

PR #2883, opened 2023-04-21
Datashader

jkrumbiegel:
Copied the prototype from https://github.com/cjdoris/ShadeYourData.jl over and modified it into a normal recipe. It's pretty slow when trackpad zooming (too many updates at once), but ok with the rectangle zoom.

```julia
using GLMakie
using DelimitedFiles

GLMakie.activate!(float = true)

airports = Point2f.(eachrow(readdlm(assetpath("airportlocations.csv"))))

fig, ax, ds = datashader(airports,
    colormap = [:black; Makie.to_colormap(:batlow)],
    binfactor = 2,
    post = mat -> log10.(mat .+ 1f0),
    axis = (; aspect = DataAspect())
)
```

datashader


asinghvi17:
It would be interesting to have a "datashader axis", which has the capability to datashade any plot plotted into it.  It could theoretically do this by just consuming the atomic plot elements (lines + linesegments + scatter) and rasterizing them.  Surfacelikes could potentially be disallowed.

This might be useful to visualize densely connected graph clusters and similar strategies.  We could then also

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+    cdf = cumsum(h_eq.weights)
+    cdf = cdf / cdf[end]
+    edg = h_eq.edges[1]
+    # TODO is this the correct linear interpolation?
+    return Makie.interpolated_getindex.((cdf,), matrix, (Vec2f(first(edg), last(edg)),))
+end
+
+"""
+    datashader(points::AbstractVector{<: Point})
+
```
felixcremer, line 203:
```suggestion

Plot a rasterization of the scatter markers for the `points ` which is refined on zooming in.
```
This should give a rough idea what is going to be plotted.

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+end
+
+
+export AggAny, AggCount, AggMean, AggSum, AggSerial, AggThreads
+
+end
+
+using ..PixelAggregation
+
+function equalize_histogram(matrix; nbins=256 * 256)
```
jkrumbiegel, line 191: I'm not sure why `nbins=256 * 256` was chosen here. Doesn't each bin get one step in the final colormap? I think `nbins = 256` might be more appropriate then, as in 8bit color depth as is usually available.

src/basic_recipes/datashader.jl
jkrumbiegel: Not sure if `of the scatter markers` is accurate, it's just pixel binning (at least in its current state, one could rasterize lines or other stuff later)

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+    edg = h_eq.edges[1]
+    # TODO is this the correct linear interpolation?
+    return Makie.interpolated_getindex.((cdf,), matrix, (Vec2f(first(edg), last(edg)),))
+end
+
+"""
+    datashader(points::AbstractVector{<: Point})
+
+Points can be any array type supporting iteration & getindex, including memory mapped arrays.
+If you have x + y coordinates seperated and want to avoid conversion + copy, consider using:
```
jkrumbiegel, line 205: If you have separate arrays for x and y coordinates and want to avoid conversion and copy

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+```Julia
+using Makie.StructArrays
+points = StructArray{Point2f}((x, y))
+datashader(points)
+```
+Do pay attention though, that if x and y don't have a fast iteration/getindex implemented, this might be slower then just copying it into a new array.
```
jkrumbiegel, line 211: attention, though

slower than

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+
+For best performance, use `method=Makie.AggThreads()` and make sure to start julia with `julia -t8` or have the environment variable `JULIA_NUM_THREADS` set to the number of cores you have.
```
jkrumbiegel, line 213: `-t8` seems like a recipe for problems as general advice :) Doesn't Julia have auto-threads now?

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+    ```
+
+- `method = AggThreads()` can be `AggThreads()` or `AggSerial()`.
+- `async_latest::Bool = true` will calculate aggregation in a task, and skip any zoom/pan updates while busy. Great for interaction, but must be disabled for saving to e.g. png or when inlining in documenter.
+
+- `global_post::Function = Makie.equalize_histogram` function which gets called on the whole aggregation array before display (`global_post(final_aggregation_result)`).
+- `local_post::Function = identity` function which gets call on each element after aggregation (`map!(x-> local_post(x), final_aggregation_result)`).
+
+- `point_func::Function = identity` function which gets applied to every point before aggregating it.
+- `binfactor::Number = 1` factor defining how many bins one wants per screen pixel. Set to n > 1 if you want a corser image.
```
jkrumbiegel, line 238: I named it `binfactor` but I'm not sure if I like the name. Maybe `binsize` is more accurate?

Also "coarser"

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+        Makie.PixelAggregation.null(::MyAgg{T}) where {T} = zero(T)
+        Makie.PixelAggregation.embed(::MyAgg{T}, x) where {T} = convert(T, x)
+        Makie.PixelAggregation.merge(::MyAgg{T}, x::T, y::T) where {T} = x + y
+        Makie.PixelAggregation.value(::MyAgg{T}, x::T) where {T} = x
+    ```
+
+- `method = AggThreads()` can be `AggThreads()` or `AggSerial()`.
+- `async_latest::Bool = true` will calculate aggregation in a task, and skip any zoom/pan updates while busy. Great for interaction, but must be disabled for saving to e.g. png or when inlining in documenter.
+
+- `global_post::Function = Makie.equalize_histogram` function which gets called on the whole aggregation array before display (`global_post(final_aggregation_result)`).
```
jkrumbiegel, line 234: this is not true if the default is `automatic`

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+        elseif _global_post === identity && p.colorrange[] isa Automatic
+            colorrange[] = Vec2f(mini_maxi)
+        end
+        elapsed = time() - tstart
+        if p.show_timings[]
+            println("aggregation took $(round(elapsed; digits=5))s")
+        end
+        return
+    end
+    @show p.async_latest[]
```
jkrumbiegel, line 374: remove `@show`

src/basic_recipes/datashader.jl
```diff
@@ -0,0 +1,388 @@
...
+        # just set to automatic, so that if one sets local_post, one doesn't do equalize_histogram on top of things.
+        global_post = automatic,
+        local_post = identity,
+
+        point_func = identity,
+        binfactor = 1,
+        show_timings = false,
+
+        colormap = theme(scene, :colormap),
+        colorrange = automatic
```
asinghvi17, line 274: IMO the colorrange calculation should be slightly changed here, so that we don't have to fiddle with colormaps.  Let the colorrange be (1, max) instead of (0, max), then set `lowclip = :transparent`, or `:black`, or what have you.  This should also make the colormap a lot more exact!

docs/examples/plotting_functions/datashader.md
```diff
@@ -0,0 +1,169 @@
...
+        # For a big dataset its interesting to see how long each aggregation takes
+        show_timings = true,
+        # Use a local operation which is faster to calculate and looks good!
+        local_post=x-> log10(x + 1),
+        #=
+            in the code we used to save the binary, we had the points in the wrong order.
+            A good chance to demonstrate the `point_func` argument,
+            Which gets applied to every point before aggregating it
+        =#
+        point_func=reverse,
```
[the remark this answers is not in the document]
asinghvi17: It may even be worth it to remove this kwarg and pick up `plot.transformation.transform_func` instead.

SimonDanisch:
@jkrumbiegel I'm going to merge this when this goes green... It's still not perfect, but I cant invest much more time into this.
I added a warning, that its still experimental!

jkrumbiegel:
Yeah I'm ok with that. Not much more movement on it, maybe once people can actually use it.

jkrumbiegel:
Should this work together with the new CategoricalColormap?

jkrumbiegel:
What's the legend element with the gradient? Did you add a new legend element kind?

SimonDanisch:
It's a poly marker with per vertex color ;)

jkrumbiegel:
> It's a poly marker with per vertex color ;)

While it's kind of cool, I think I'd still change those to normal solid colors. It's the color that tells you the category, not the alpha.

Issue #2612, opened 2023-01-18
`linesegments` with arrow, or `arrows` method that takes a pair of points


ParadaCarleton:
Sometimes I just want to draw an arrow going from point A to point B; in that case, `linesegments` doesn't work (since it's a line, not an arrow), and `arrows` doesn't work (it's really meant to plot vector fields, rather than a single vector).

This could be fixed either by adding an extra argument to `linesegments` or to `arrows`.

BambOoxX:
This is achievable by providing `arrows` with the right origin point and the right arrow component lengths
Somehting like 
```julia
origin_points = rand(Point3,10)
target_points = rand(Point3,10)
arrow_components = (target_points .- origin_points) 
f = Figure()
l = LScene(f[1,1])
scatter!(l,origin_points,color=:red)
scatter!(l,target_points,color=:green)
arrows!(l,origin_points,arrow_components,color = RGBAf(0,0,255,0.5))
``` 
![image](https://user-images.githubusercontent.com/42067365/214352958-a4e33ac6-8876-446e-a3c3-34a26819caf3.png)

Pay attention to the arrowsize correction since an arrow head currently does not "point" at the target not, just its body see https://github.com/MakieOrg/Makie.jl/issues/2607 or https://github.com/MakieOrg/Makie.jl/issues/1206, https://github.com/MakieOrg/Makie.jl/issues/2472

briochemc:
What's nice about @aplavin's MakieExtra.jl solution is that the pointy arrow precisely points to the end point (instead of a overshooting by the arrow head size).

ffreyer:
arrows (both variants) have an option to provide endpoints instead of a direction now

Issue #2607, opened 2023-01-17
How to make 2D arrowhead points point to endpoint instead of being centered at endpoint?


khiner:
I am having trouble accounting for the fact that Makie's `arrows!` function _centers_ arrowheads at the vector endpoints instead of _pointing_ to the endpoint.

Here is my code with default behavior:

```julia
arrowsize=40
# For each arrow...
arrows!([origin.x], [origin.y], [val.x], [val.y]; arrowsize=arrowsize, ...)
```
<img width="692" alt="Screen Shot 2023-01-17 at 11 01 07 AM" src="https://user-images.githubusercontent.com/685394/212950921-70707150-0fd4-4daf-9fbd-1b8aab4af7a4.png">

And here is my naive attempt at adjusting for arrowhead size:

```julia
arrowsize = 40 # In pixel space
pix_scale_2d = camera(Scene()).pixel_space[][1:2,1:2]
# For each arrow...
# Convert scene value to pix value.
val_pixels = (pix_scale_2d ^ -1) * val
# Subtract half the arrow in pixel space, in the direction of the value.
val_pixels -= normalize(val) * (arrowsize/2)
# Convert value back to pixel space.
val = pix_scale_2d * val_pixels
arrows!([origin.x], [origin.y], [val[1]], [val[2]]; arrowsize=arrowsize, ...)
```
<img width="667" alt="Screen Shot 2023-01-17 at 12 27 43 PM" src="https://user-images.githubusercontent.com/685394/212969779-0c2830b0-76ba-4119-9f51-cc4a5a0d0ce8.png">

If I subtract  `normalize(val) * (arrowsize*2)` instead of `normalize(val) * (arrowsize/2)` (which in my head is twice the arrow size instead of half, but is clearly not in actuality), I get:

<img width="674" alt="Screen Shot 2023-01-17 at 12 27 57 PM" src="https://user-images.githubusercontent.com/685394/212970096-71883728-5e69-4eb8-b835-f1773eeb8fc1.png">

It seems like my naive attempt is pretty close, but most likely for the wrong reasons. What am I doing wrong here?

Thanks for the help!

khiner:
Also, I would very much appreciate an `arrow` option to make this arrow end at the vector point. Ideally, this would be the default, like it is in the Plots package. I would think most use cases would prefer arrows ending exactly where the vector ends rather than placing an arrow marker centered around the vector end point.

jkrumbiegel:
[...]

Issue #1688, opened 2022-02-23
transparent arrows


JonasIsensee:
Hi,
I was trying to plot a semi-transparent arrow (over a color gradient) and found the following:

![image](https://user-images.githubusercontent.com/33990604/155284807-8abf74f7-bf12-4a68-ae84-594a32867b1a.png)

PR #2884, opened 2023-04-21
fix[arrows]: `align` option (need help, draft)

kunzaatko:
- chore(format): arrows.jl
- fix(arrows): alignment with :center

# Description

The issue that this solves is not yet filed.

However, as you can see from the image below:
before_arrows_alignment

the alignment argument to `arrows` previously did nothing except the `:origin` value.

I fixed the `:center` value and the figure now looks like this:
center_arrows_alignment

I would like to fix the other options and the 3D variant as well, but I need help with getting the arrow head size
generally to fix it even when it is determined automatically...

Note that the second commit makes the relevant changes. The first is only formatting.

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/arrows.jl
+++ b/src/basic_recipes/arrows.jl
@@ -121,9 +121,14 @@ function plot!(arrowplot::Arrows{<:Tuple{AbstractVector{<:Point{N}},V}}) where {
                          lengthscale) do points, dirs, n, align, s
             map(points, dirs) do p1, dir
                 dir = n ? LinearAlgebra.normalize(dir) : dir
-                if align in (:head, :lineend, :tailend, :headstart, :center)
+                if align == :headcenter
                     shift = s .* dir
-                else
+                elseif align == :center
+                    shift = (s .* dir) ./ 2
+                    # TODO: :head, :lineend, :tailend, :headstart, :center
+                elseif align in (:head, :lineend, :tailend, :headstart, :center)
+                    shift = s .* dir
+                else # align == :origin
                     shift = Vec2f(0)
                 end
                 return Point2f(p1 .- shift) => Point2f(p1 .- shift .+ (dir .* s))
--- a/src/basic_recipes/arrows.jl
+++ b/src/basic_recipes/arrows.jl
@@ -38,10 +38,8 @@ function arrow_head(N, marker::Automatic, quality)
     if N == 2
         return :utriangle
     else
-        merge([
-           _circle(Point3f(0), 0.5f0, Vec3f(0,0,-1), quality),
-           _mantle(Point3f(0), Point3f(0,0,1), 0.5f0, 0f0, quality)
-        ])
+        merge([_circle(Point3f(0), 0.5f0, Vec3f(0, 0, -1), quality),
+               _mantle(Point3f(0), Point3f(0, 0, 1), 0.5f0, 0.0f0, quality)])
     end
 end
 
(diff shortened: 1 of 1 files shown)


kunzaatko:
I will complete the documentation and news checklist, when the PR is complete

ffreyer:
The alignment options were only meant to be 3 cases:
- have the arrow tail start at the given positions
- have the arrow center be at the given positions
- have the arrow tip be at the given positions

Back when I added those options I didn't know what would be some good names so I ended up adding some alternative names. I wouldn't mind deprecating those and just moving to one name for each option. Or maybe changing this option to be a float where 0 is the tail and 1 is the tip so you have more freedom to pick.

I also wouldn't worry about the tip alignment being aligned to the center of the tip rather than the tip of the tip. That's another issue, see #2612, #2607, ... For 3D that can be fixed in the arrows recipe. For 2D, I think we should add an (internal) attribute to end lines earlier and tweak the scatter and line placement that way.

kunzaatko:
What do you mean by
> end lines earlier

?

ffreyer:
There are already internal attributes for offsetting markers from their position in pixel space. For scatter those are used to center markers, for text they position the characters. I'm thinking of something like that for lines, so that you could tell a line to end, for example, 10px earlier. With those attributes you could overwrite the centering for scatter in arrows, so that the tip actually ends at the end position of an arrow, and then shift the line back by one markersize.

kunzaatko:
On a similar note, If we were to be changing these internal attributes I also think that it would be a good idea to consider, whether the arrows shouldn't end at the ends of the vectors... I am not sure If this is what you meant in the previous message.

ffreyer:
Yea, that's the issue that would solve. That and the overdraw we currently have, e.g. #1688

kunzaatko:
Great! I will look into it. Do you know what recipe uses internal attributes, so I can see how it would be done?

ffreyer:
This is a more involved change to make. If you want to try you could look at `marker_offset` from Scatter as an example to trace. It's first set here:
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/MakieCore/src/basic_plots.jl#L422
Though you wouldn't need to define line offsets at that point. You could also define them in the backends/use `get(attributes, name, default)`.

---

For **CairoMakie** it moves fairly directly into the Scatter draw functions:
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/CairoMakie/src/primitives.jl#L207

---

For **GLMakie** it gets renamed:
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/GLMakie/src/drawing_primitives.jl#L40

I think this is some type normalization: (the function is probably in conversions.jl)
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/GLMakie/src/drawing_primitives.jl#L233

Here it gets converted to an OpenGL object. 
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/GLMakie/src/glshaders/particles.jl#L178
Assuming we only want/need one offset for all line starts and one for all line ends we wouldn't want this to be a GLBuffer. I think those always end up in the vertex array, which would be one offset for each point. If we just want to values this should end up as a Vec2f instead.

And then finally you get into the shaders, which probably aren't that useful to look at since the input is part of vertexarray, not a uniform. 

For line offsets, you'd need to add `uniform vec2 <name from gen_default!>;` and then tweak the positions of the line segment accordingly. I.e. nudge p0 and p1 in v0 direction around here
https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/GLMakie/assets/shader/line_segment.geom#L58-L68
Note that you'll also need to adjust the `l`, otherwise linestyles/patterns will be off. p0 and p1 are in pixel coordinates here. (This file only affects linesegments. We could also do these changes for lines but that shouldn't be necessary for arrows.)

---

For **WGLMakie** I think you need to pass the attribute to uniforms in https://github.com/MakieOrg/Makie.jl/blob/master/WGLMakie/src/lines.jl, which would give you a `get_<name in uniforms>()` function in the shader, which you can use to adjust points in https://github.com/MakieOrg/Makie.jl/blob/master/WGLMakie/assets/line_segments.vert. From the look of it you'll need to adjust the points in clip space, so do the inverse of `screen_space`to the offsets.

PR #2886, opened 2023-04-21
removed specific function names

timoleistner:
removed specific function names, otherwise `@js_str, Session, App` would also be needed for both examples.

# Description
Without the above specified function imports the examples throw an UndefVarError.

Fixes # (issue)

Removed import of specific funtion names and import the whole JSServe package.

## Type of change

## Checklist

Now sure how to add the entry to NEWS.md as the current NEWS.md file is only untill version v0.19.4 but v0.20.0 is already released.


SimonDanisch:
Thanks!

PR #2891, opened 2023-04-22
Support for GIF looping in recorded video output

ztangent:
# Description

Adds the `loop` option to the `VideoStreamOptions` struct, and adds `loop` as a keyword argument to `VideoStream`, which is passed on to FFMPEG. This allows control over the number of times a GIF is looped for videos produced by `record`.

## Type of change

## Checklist


src/ffmpeg-util.jl
```diff
@@ -58,11 +62,16 @@ struct VideoStreamOptions
         if format in ("mp4", "webm")
             (compression === nothing) && (compression = 20)
         end
+        
+        if format == "gif"
+            (loop === nothing) && (loop = 0)
```
asinghvi17, line 67:
```suggestion
            isnothing(loop) && (loop = 0)

asinghvi17 (commented):
This looks pretty cool!  

Is there a way to test that this works in tests?  For example, if I try to set `loop=2` and load it using ffmpeg, would I get 2x the frames and time?

src/ffmpeg-util.jl
ztangent: I was trying to maintain consistency with the earlier lines of code by using `===` instead of `isnothing`, but if you prefer it this way, happy to switch!

ztangent:
@asinghvi17 Unfortunately I don't think there's an easy way to check based on the number of frames, or length of the video, alone -- it looks like the loop count for animated GIFs is stored somewhere in the header of the file (see https://en.wikipedia.org/wiki/GIF#Animated_GIF) and I'm not sure how to use FFMPEG to look up that number. Maybe it's possible to just look at that byte number (31D) directly?

SimonDanisch:
Thanks!

PR #2892, opened 2023-04-23
Let some kwargs vary with k_ideal in WilkinsonTicks

asinghvi17:
# Description

Specifically, `k_min = k_ideal - 4`, `k_max = k_ideal + 4`.  Users can still override this as necessary.

## Type of change

Arguably a bug fix in the UX for WilkinsonTicks.

## Checklist


jkrumbiegel:
With -4 you could go to 1, 0, or even negative, all of which are bad. Maybe a better heuristic would be +- 30% or so, with an absolute minimum of 2 because only then can you even tell what the limits are.

So `max(2, round(Int, k_ideal * 0.7))` e.g.

Datseris:
Thanks!

SimonDanisch:
@jkrumbiegel can we make this mergeable? I think it would be nice to have it easier to construct the WilkinsonTicks

PR #2894, opened 2023-04-23
Block attribute docs

jkrumbiegel:
TODO
- [x] control which backend and with what args renders an example in the docs
- [ ] API to show specific example of an attribute


SimonDanisch:
This seems super close to what I wanted initially for Makie's docs! Still need to try it out though and see how it works in detail!

jkrumbiegel:
Sure I'll wait until you've had time :) There's not too much to try out right now, it's just `?Axis.xticks` in the repl plus the changed docs style.

SimonDanisch (approved):

PR #2895, opened 2023-04-24
Set visible=true for interactive record

### LINKED ISSUE #2879: Interactive record closes figure window
The recommended way to record an interactive session is:
https://docs.juliahub.com/MakieGallery/Ql23q/0.2.17/interaction.html#Recording-your-interactions

```julia
fig, ax, p = lines(rand(10))
fps = 10
record(fig, "test.mp4"; framerate = fps) do io
    for i = 1:100
        sleep(1/fps)
        recordframe!(io)
    end
end
```

but `record` causes the figure window to close so no interactions can be recorded

using:
GLMakie v0.8.4
Julia v"1.8.2"
macos 13.3.1 (22E261)

alex-s-gardner:
# Description

Fixes #2879 

Sets default figure  behavior  to `visible=true` for interactive record

## Type of change

### DIFF
--- a/src/recording.jl
+++ b/src/recording.jl
@@ -139,7 +139,7 @@ end
 """
 function record(func, figlike::FigureLike, path::AbstractString; kw_args...)
     format = lstrip(splitext(path)[2], '.')
-    io = Record(func, figlike; format=format, kw_args...)
+    io = Record(func, figlike; format=format, visible=true, kw_args...)
     save(path, io)
 end
 


src/recording.jl
```diff
@@ -139,7 +139,7 @@ end
 """
 function record(func, figlike::FigureLike, path::AbstractString; kw_args...)
     format = lstrip(splitext(path)[2], '.')
-    io = Record(func, figlike; format=format, kw_args...)
+    io = Record(func, figlike; format=format, visible=true, kw_args...)
```
SimonDanisch, line 142: Thank you! I think it should still be customizable though! So should be part of the keywords :)

SimonDanisch:
Thanks, you're right! One can still pass the attribute :)

PR #2914, opened 2023-05-04
switch to precompile tools

SimonDanisch:


timholy:
Wait, did you not get an automatic PR? If so, that's a little weird; I wonder how many other packages might have been missed.

SimonDanisch:
Hm, nope, not that I've seen ;)

timholy:
Hmm. I double-checked and it *was* on the list, but maybe there was some error that omitted it. Sorry about that, hope it wasn't too much effort to do manually.

BTW, the new `@recompile_invalidations` was added to PrecompileTools because I was sad that loading lots of other packages and Makie raised the TTFX from ~1s to ~8s. Now creating a "Startup" package with Makie + the rest fixes that.

PR #2911, opened 2023-05-03
allow zero length geometries

rafaqz:
# Description

Fixes #2819 

Checks for zero length everywhere (I could find so far) that `geom[1]` etc is used.

Should I put tests in conversions.jl?

## Type of change

Delete options that do not apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] Added an entry in NEWS.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #2917, opened 2023-05-05
Rafaqz allow zero length

SimonDanisch:
Clean up of #2911 ...
@jkrumbiegel, did we have any reason to close polygons, if they were already closed?
I now added a check here:
image

So we don't already close closed polygons... Weirdly enough, it looks like we specifically tested for this


rafaqz (approved):
LGTM

jkrumbiegel:
Not sure, could have been something Cairo related at the time. Let's just watch out for visual regressions.

SimonDanisch:
I did check the reference image tests explicitly, to make sure there are no changes below threshold.

PR #2918, opened 2023-05-05
reverse interior for Cairo hole drawing

SimonDanisch:
`GeometryBasics.Polygon` doesn't care about winding order for holes and so doesn't GLMakie.
But Cairo needs winding directions for holes to be counter-clockwise, so this PR makes sure interiors are always drawn counter-clockwise. 
Funnily enough, we have a test for this, which has been "failing" for cairomakie without going above test failure threshold -.-

Test before PR:
image

With PR:
image


asinghvi17:
Should we make this change in GeometryBasics?  We could also change the fill rule. I think, which might achieve the same result.

Geometries from other packages are reversed in order, so this might cause complications there.  I'll PR some tests from ArchGDAL geometries to check this.

jkrumbiegel:
the holes don't need to be counter-clockwise I think, they need to be opposite to the direction of the exterior. unless you change the fill rule, I don't remember the edge cases right now

jkrumbiegel:
yeah no it shouldn't be in geometrybasics, just cairomakie should make sure to draw it correctly

jkrumbiegel:
according to this it would need to be `evenodd` https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule and is probably `nonzero` right now

SimonDanisch:
Yeah, it's basically just a [sum of neighboring cross products](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/master/src/triangulation.jl#L49), so reasonably cheap ;)
So for a polygon with 10^7 points it takes ~0.008s 
Also I removed the allocations from `interior[2:end]` which should _almost_ make up for it^^

jkrumbiegel:
yeah it's `CAIRO_FILL_RULE_EVEN_ODD` https://www.cairographics.org/manual/cairo-cairo-t.html#CAIRO-FILL-RULE-WINDING:CAPS

```
cairo_set_fill_rule ([cairo_t](https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-t) *cr,
                     [cairo_fill_rule_t](https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-fill-rule-t) fill_rule);
```

This would just need to be called before drawing a polygon, then the order doesn't matter anymore.

jkrumbiegel:
Just to add, it wouldn't matter anymore for "normal" polygons. Where edges don't weirdly intersect, like in the svg example with the star. But I assume the polygons we care about are not weird, I think geo packages try not to generate these strange shapes but instead split into multiple polys if needed.

Issue #2805, opened 2023-03-25
Inline plots stopped appearing


kbarros:
Hi,

It looks like https://github.com/MakieOrg/Makie.jl/pull/2731 changed the behavior of `showable` (via `_backend_showable`) with this new line: https://github.com/MakieOrg/Makie.jl/blob/2116627b5caed21ddf0e3bbd2103f55d5679d881/src/display.jl#L149 

This caused surprising behavior for a [Literate.jl](https://github.com/fredrikekre/Literate.jl) workflow that would execute a source file and generate Markdown. Prior to this change, Literate would generate a png file for each inline image. After this change, Literate will just generate the text `Figure()`, thinking that the Figure is not showable as `image/png`.

~~I observed this surprising behavior on Github actions, but could not repeat it locally on my Mac. I'm not sure why yet.~~ Reproduces on Github actions and locally on Mac. I can work on creating a minimum example if it would be helpful.

The problem can be fully resolved by including `Makie.inline!(true)`, but it seems better to restore the old behavior (automatically detect that the figure is in fact "showable").

kbarros:
To reproduce, please create a Julia file 'makie_test.jl' containing some simple GLMakie commands:

```julia
# Execute this file using the command:
#   import Literate; Literate.markdown("makie_test.jl", execute=true)

using GLMakie

# After executing within Literate, this example either:
#  (a) outputs an inline SVG into the output Markdown
#  (b) outputs the text "Figure()" into the output Markdown

f = Figure()
ax = Axis(f[1, 1])
x = range(0, 10, length=100)
y = sin.(x)
lines!(ax, x, y)
f
```

Then from the Julia REPL, execute `import Literate; Literate.markdown("makie_test.jl", execute=true)`. 

When I first tried this, I got SVG output inlined into the output Markdown. Then I played around with some GLMakie commands, and somehow the result changed so that only the text `Figure()` was inlined into the output Markdown. So, unfortunately, it seems the precise behavior I'm observing may be hard to reproduce.

If `Makie.inline!(true)` is included at the top of the script above, then the `Literate.markdown` command will output a .png which gets linked to the output Markdown.

SimonDanisch:
Ah sorry.. I tried to finally make it work, that vscode doesn't display anything in the plot pane if inline! Is false... And didn't realize that it destroys the default for documenter / literate and friends

asinghvi17:
[...]

Issue #2901, opened 2023-04-26
GLMakie update to 0.8.3  breaks show method for Pluto notebooks.


j-fu:
In GLMakie 0.8.3 and 0.8.4, the show method for GLMakie plots appears to be broken.

MWE: Pluto notebook with just one cell:
```
begin
	using Pkg
	Pkg.activate(mktempdir())
	Pkg.add(name="GLMakie",version=v"0.8.3")
	using GLMakie
	X=0:0.01:20
	lines(X,sin.(X))
end
```
The error shown is
```
Failed to show value:

MethodError: no method matching show(::IOContext{IOBuffer}, ::Nothing, ::Makie.FigureAxisPlot)

Closest candidates are:

show(::IO, ::Nothing)

@ Base show.jl:1140

show(::IO, !Matched::MIME{Symbol("text/plain")}, ::Makie.FigureAxisPlot)

@ Makie ~/.julia/packages/Makie/z1gJs/src/figureplotting.jl:7

show(::IO, !Matched::MIME, ::Union{Makie.Figure, Makie.FigureAxisPlot, Makie.Scene})

@ Makie ~/.julia/packages/Makie/z1gJs/src/display.jl:181
```
With 0.8.2 and newest CairoMakie (0.10.4) and WGLMakie (0.8.8) it is ok.

SimonDanisch:
https://github.com/MakieOrg/Makie.jl/pull/2919 should close this!

PR #2919, opened 2023-05-05
Fix Makie.inline!(...) for most cases

SimonDanisch:
This should fix https://github.com/MakieOrg/Makie.jl/issues/2805 and https://github.com/MakieOrg/Makie.jl/issues/2901.
It's still not perfect, but I think at this point only a breaking change can improve upon this.

### DIFF
--- a/CairoMakie/src/screen.jl
+++ b/CairoMakie/src/screen.jl
@@ -98,7 +98,7 @@ function device_scaling_factor(surface::Cairo.CairoSurface, sc::ScreenConfig)
     return is_vector_backend(surface) ? sc.pt_per_unit : sc.px_per_unit
 end
 
-const LAST_INLINE = Ref(true)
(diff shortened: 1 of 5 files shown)


SimonDanisch:
I've tested this on a few platforms with Pluto, IJulia, VSCode and REPL and things seem to work mostly as expected besides a few expected issues with WGLMakie, because the display system via JSServe and the browser is a bit more complex.
E.g. `inline!(false)` doesn't help to open a browser window with WGLMakie when in VSCode with active plotpane, because it would need to communicate this to JSServe which does the actual displaying.
The good news is, disabling the plotpane now at least starts opening a browser window with WGLMakie, and the same is true for the REPL.
I hope this won't introduce any other weird issues with E.g. documenter, but I think most corner cases should work fine now (but are hard to test).

Issue #2926, opened 2023-05-06
`@inherit` calls include path in docstrings


KronosTheLate:
From `julia> Base.Docs.getdoc(subtypes(Makie.Block)[1])` in a REPL, i.e. docstring for `Axis` on my system:
```
    •  yticklabelsize: The font size of the yticklabels. Default: #=
       /home/dennishb/.julia/packages/Makie/Iqcri/src/makielayout/types.jl:273
       =# @inherit :fontsize 16.0f0
```

This is not good for offline documentation though docstrings. Why does this happen?

jkrumbiegel:
Ah yes that has annoyed me as well, it's just the default printing of macro expressions which prints out LineNumberNodes

PR #2932, opened 2023-05-08
Strip line info from default exprs

jkrumbiegel:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/2926


src/makielayout/blocks.jl
```diff
@@ -94,7 +94,7 @@ macro Block(name::Symbol, body::Expr = Expr(:block))
     esc(q)
 end
 
-_defaultstring(x) = string(x)
+_defaultstring(x) = string(MacroTools.striplines(x))
```
SimonDanisch, line 97: Could it work to do this without MacroTools, by just having:
```julia
striplines(@nospecialize(expr)) = expr
function striplines(expr::Expr)
    args = [striplines(elem) for elem in expr.args if !(elem isa LineNumberNode)]
    return Expr(expr.head, args...)
end

```
For a simple test, this seems to work fine!

jkrumbiegel: yeah sure, was just to lazy to copy paste and thought MacroTools is always an indirect dep anyway

Issue #2898, opened 2023-04-25
hexbin fails when there are less than ~25 points


sefffal:
The `hexbin` plot command fails when the input size is small.

For example, the following command works as expected:
```julia
hexbin(randn(100), randn(100))
```

But this does not (see error message below):
```julia
hexbin(randn(10), randn(10))
```

It seems to me that hexbins should be well defined with arbitrarily few points.

X-ref: https://github.com/sefffal/PairPlots.jl/issues/16

Tested with latest stable CairoMakie and Julia 1.9-rc2.

Error message:
```
Failed to show value:

Can't interpolate in a range where cmin == cmax. This can happen, for example, if a colorrange is set automatically but there's only one unique value present.

error(::String)@error.jl:35
interpolated_getindex@colorsampler.jl:36[inlined]
#70@colorsampler.jl:164[inlined]
iterate@generator.jl:47[inlined]
_collect(::Vector{Float32}, ::Base.Generator{Vector{Float32}, Makie.var"#70#71"{Vector{ColorTypes.RGBA{Float32}}, ColorTypes.RGBA{Float32}, ColorTypes.RGBA{Float32}, ColorTypes.RGBA{Float32}, Float32, Float32}}, ::Base.EltypeUnknown, ::Base.HasShape{1})@array.jl:802
collect_similar@array.jl:711[inlined]
map@abstractarray.jl:3261[inlined]
numbers_to_colors(::Vector{Float32}, ::Vector{ColorTypes.RGBA{Float32}}, ::GeometryBasics.Vec{2, Float32}, ::ColorTypes.RGBA{Float32}, ::ColorTypes.RGBA{Float32}, ::ColorTypes.RGBA{Float32})@colorsampler.jl:156
numbers_to_colors(::Vector{Float32}, ::MakieCore.Scatter{Tuple{Vector{GeometryBasics.Point{2, Float32}}}})@colorsampler.jl:147
to_cairo_color(::Vector{Float32}, ::MakieCore.Scatter{Tuple{Vector{GeometryBasics.Point{2, Float32}}}})@utils.jl:125
draw_atomic(::Makie.Scene, ::CairoMakie.Screen{CairoMakie.IMAGE}, ::MakieCore.Scatter)@primitives.jl:222
[...]

PR #2933, opened 2023-05-08
Fix hexbin with singular colorrange

jkrumbiegel:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/2898

The colorrange became singular valued sometimes, mostly when there were only few values so all existing cells had a count of `1`.

### DIFF
--- a/src/stats/hexbin.jl
+++ b/src/stats/hexbin.jl
@@ -170,7 +170,19 @@ function Makie.plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Point2}}})
         if isempty(count_hex[])
             (0, 1)
         else
-            (minimum(count_hex[]), maximum(count_hex[]))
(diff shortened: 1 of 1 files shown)


src/stats/hexbin.jl
```diff
@@ -170,7 +170,19 @@ function Makie.plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Point2}}})
         if isempty(count_hex[])
             (0, 1)
         else
-            (minimum(count_hex[]), maximum(count_hex[]))
+            mi, ma = extrema(count_hex[])
+            # if we have only one unique value (usually happens) when there are very few points
+            # and every cell has only 1 entry, then we set the minimum to 0 so we do not get
+            # a singular colorrange error down the line.
+            if mi == ma
```
SimonDanisch, line 177: Could `distinct_extrema_nan` work here?

PR #2936, opened 2023-05-08
Mention GLMakie-CUDA interop in the docs

asinghvi17:


jkrumbiegel:
I'm not for adding untested experimental stuff to the docs, makes it look unpolished. Especially with language like "easily" which these things usually are not. There should be a place to direct people to this kind of information though, I just don't like presenting it almost like "it just works" if it really doesn't. The link is broken on top of that

asinghvi17:
Hmm, fair enough.  I guess we could have an experimental section in the docs where we could put this kind of stuff.

jkrumbiegel:
Yes I would support something like that, especially because a lot of Makie users are also into hacking and tweaking it, which is nice in a visualization library anyway. It should just be clearly separated from what we consider core functionality I think.

PR #2939, opened 2023-05-09
[WIP] Add `extra_options` passed directly to FFMPEG in VideoStream

### LINKED ISSUE #2938: Allow for increased customizability for VideoStreamOptions
Currently all the options for video are quite restrictive. Personally I would like to use hardware acceleration for example. Would it be possible/easy/wanted to have an option to specify raw ffmpeg flags to the VideoStreamOptions (and record function)? This way it would be possible to tweak the encoding options much more control.

asinghvi17:
# Description

Adds an extra field to `VideoStreamOptions`, which is a `Cmd` to be inserted after the computed FFMPEG options.

Needs to be tested.

Fixes #2938 

## Type of change

## Checklist


MarcusTL12:
It might be useful to have a way of disabling the `ffmpeg_options` being included as currently it would be impossible to have an mp4 and specify the codec to be `-c:v libx265` for example (or any other codec), as it is already specified as `-c:v libx264` in the `ffmpeg_options`. Currently I would get around this by using the mkv format as there are no options added there, but that is probably not the optimal solution.

jkrumbiegel:
I think it's not such a good idea to pass through arbitrary command pieces while the rest of the command is relatively opaque to the user (unless they piece it together from the docs or from the source code). To me it seems like once you want to go the way of custom ffmpeg flags, you might as well go all the way and specify the full thing. So I would rather think in the direction of parameterizing videostream on the type of options you pass, and then we just add `CustomOptions` or so?

SimonDanisch:
Ok, lets close this for now then...

PR #2746, opened 2023-03-12
Camera3D Improvements

ffreyer:
# Description

Some general camera improvements/changes

#### Breaking

- revert the switch to fov based zoom. `Camera3D` now uses the same zoom style as `old_cam3d!` again, i.e. translating `eyeposition`. 
    - to keep clipping issues (#602, #941) from occurring  near and far are now relative to `norm(eyeposition - lookat)`, i.e. the zoom level. 
    - because of this stretch/contract view hotkeys have been removed as they are equivalent to zoom in/out. The old zoom in/out is now available as increase/decrease fov.
- Separate attributes into `cam.controls` for hotkeys and `cam.settings` for everything else

#### Non-breaking

- make near & far relative to `norm(eyeposition - lookat)`, i.e. zoom level
- scale translations with zoom level and fov. This normalizes the perceived translation speed 
- scale camera centering on bounding boxes with fov. This makes the full bbox visible at a larger range of fovs.
- add orbital `update_cam!(scene, cam, phi, theta[, radius, center])
- add deprecation warning for `old_cam3d!` etc. `Camera3D` can now reproduce this, though not exactly due to QoL tweaks.
- improve default hotkey choices
- add functionality to focus camera on a point on a plot

Fixes #2721, #1600, #1236 (y/z swap, documentation)

#### DataInspector fixes

I moved a bunch of code around that's also used by `DataInspector` so I did some testing and fixing on that too. Most of these fixes are probably not related to this pr.
- fixed error related to `getindex_sv(::Range, idx)`
- fixed blinking in `image` inspection
- fixed typos in arrows (closes #2764) & lines(egments) 
- fixed incorrect values when inspecting volumeslices
- fixed incorrect indicator placement in lines and added workaround to avoid multiple indicators

## Type of change

Breaking change

## Checklist

- [x] Added an entry in NEWS.md (for new features and breaking changes)
- [x] Added or changed relevant sections in the documentation
- [x] Check DataInspector 
- [x] Document changes to `ispressed`

Issue #2624, opened 2023-01-24
Reference frame like in CAD software ?


BambOoxX:
It is common in 3D CAD software to have a reference frame displayed in the picture 

e.g. this in CATIA
![image](https://user-images.githubusercontent.com/42067365/214353617-fdfb7cad-bc7b-4563-981c-97303811d373.png)
Solidworks
![image](https://user-images.githubusercontent.com/42067365/214353720-724008ad-324a-4bc3-9c84-0e7b15efd053.png)

It's fairly obvious that one can reproduce that in GLMakie with arrows, however, it can't figure how to have the arrow origin remain fixed in pixel space. Is there a way to achieve this. I'm guessing something quite equivalent to the `:pixe` of `text` would be nice. 
The arrow should however rotate according to the `:data` space and the camera angles

ffreyer:
I think the way to go about this is to create a separate Scene for the indicator and update its view matrix based on the parent plot.

With an `LScene` (using some 3d camera) this involves reading out eyeposition, lookat and upvector, moving lookat to 0 and adjusting eyeposition to a comfortable zoom level:

```julia
# Your plot
fig = Figure()
lscene = LScene(fig[1,1])
p = scatter!(rand(Point3f, 10))
display(fig)

scene = Scene(
    fig.scene, px_area = Rect2f(20, 20, 100, 100), 
    # backgroundcolor = RGBAf(1,1,0), clear = true # to see scene region
);
linesegments!(scene, 
    Point3f[(0,0,0), (1,0,0), (0,0,0), (0,1,0), (0,0,0), (0,0,1)],
    color = [:red, :red, :green, :green, :blue, :blue],
    linewidth = 10
)

cam = cameracontrols(lscene.scene)
scene.camera.projection[] = Makie.perspectiveprojection(45f0, 1f0, 0.01f0, 100f0)
onany(cam.lookat, cam.eyeposition, cam.upvector) do lookat, eyepos, up 
    viewdir = 4f0 * normalize(eyepos - lookat)
    scene.camera.view[] = Makie.lookat(viewdir, Vec3f(0), up)
    return
end
```


With an Axis3 you need to read out the relevant angles and calculate eyeposition from that:

```julia
fig = Figure()
ax = Axis3(fig[1,1])
p = scatter!(ax, rand(Point3f, 10))
display(fig)

scene = Scene(
    fig.scene, px_area = Rect2f(20, 20, 100, 100), 
[...]

PR #2944, opened 2023-05-11
3D Camera Orientation Widget

ffreyer:
# Description

Implements the Widget from #2624, which allows you to orient a 3D axis by clicking on different faces of the mesh, or rotating that mesh with mouse interactions. This will need some adjustments after #2746.

Screenshot from 2023-05-11 21-01-18

I think this looks pretty bad atm so I'm open to suggestions. 

Blender has this with only 6 directions, which is a bit too coarse imo. 

Maybe this should be similar to the Navball in Kerbal Space Program, which could also double as an (external) axis type thing. To orient the plot we could allow different angular step size, which we could round to on click. (Ignore all the UI around the ball)

Navball

TODO:
- [ ] more attributes for fine tuning things
- [ ] make a prototype for the Navball version
- [ ] allow this to be detached from layouting
- [ ] cleanup debug printing

## Type of change

## Checklist

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -234,13 +234,14 @@ cam3d_cad!(scene; cad = true, zoom_shift_lookat = false, fixed_axis = false, kwa
     Camera3D(scene, cad = cad, zoom_shift_lookat = zoom_shift_lookat, fixed_axis = fixed_axis; kwargs...)
 
 function deselect_all_cameras!(scene)
-    cam = cameracontrols(scene)
-    cam isa Camera3D && (cam.attributes.selected[] = false)
+    deselect_camera!(cameracontrols(scene))
     for child in scene.children
         deselect_all_cameras!(child)
     end
     nothing
 end
+deselect_camera!(cam::Camera3D) = cam.attributes.selected[] = false
+deselect_camera!(::AbstractCamera) = nothing
 
 
 function add_translation!(scene, cam::Camera3D)
--- a/src/makielayout/MakieLayout.jl
+++ b/src/makielayout/MakieLayout.jl
@@ -33,6 +33,7 @@ include("blocks/legend.jl")
 include("blocks/scene.jl")
 include("blocks/menu.jl")
 include("blocks/textbox.jl")
+include("blocks/viewport.jl")
 
 export Axis
 export Axis3
--- a/src/makielayout/blocks/viewport.jl
+++ b/src/makielayout/blocks/viewport.jl
@@ -1,0 +1,362 @@
+# Mesh generation
+function Rhombicuboctahedron(; center = Point3f(0), radius = 1f0)
+    o = 1 / (1 + sqrt(2))
+    ps = radius .* Point3f[
+        (-o, -o, -1), (-o, o, -1), (o, o, -1), (o, -o, -1),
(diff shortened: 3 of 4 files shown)


asinghvi17:
There are some nice textures we could use for a KSP like navball [here](https://github.com/linuxgurugamer/NavBallTextureChanger/tree/master/GameData%2FNavBallTextureChanger%2FPluginData%2FSkins) - all under a FOSS license

ffreyer:
Oh, cool. I think as a default we'd want a texture with x/y/z marked rather than north, south, etc, but I could try to be compatible with those textures.

ffreyer:
Navball prototype with a texture from the repo:
Screenshot from 2023-05-12 11-42-12
I think with a (linear) gauge showing the step width rather than the plain text this would be pretty nice.

asinghvi17:
I could do a spherical axis instead, that might actually be better since it's more customizable.  Will send the code when I can.

ffreyer:
I was planning to generate a texture with Makie code. That could be customized, though you probably want to do it in CairoMakie because fxaa makes this kinda of ugly in GLMakie.

ffreyer:
This is how it currently looks:
Screenshot from 2023-05-16 18-53-53

The texture is generated with the two prs I referenced. The white region marks where you can snap to on click, and it reacts to hovers. The degree indicator appears when you scroll and disappears after 3s

src/layouting/layouting.jl
```diff
@@ -287,14 +261,6 @@ function padded_vcat(arrs::AbstractVector{T}, fillvalue) where T <: AbstractVect
     arr
 end
 
-function alignment2num(x::Symbol)
-    (x === :center) && return 0.5f0
-    (x in (:left, :bottom)) && return 0.0f0
-    (x in (:right, :top)) && return 1.0f0
-    return 0.0f0 # 0 default, or better to error?
-end
```
ffreyer, line 295: I added `halign2num` and `valign2num` instead of this. Those functions also accept float values and you can optionally add an error message for when the conversion fails. I could also go back and extend this function instead, if that's preferred.

src/makielayout/blocks/ballcontroller.jl
```diff
@@ -0,0 +1,706 @@
...
+            texture[] = controller.texture_high[]
+        end
+    end
+
+    # Generate sphere mesh
+    m = uv_normal_mesh(Tesselation(Sphere(Point3f(0), 1f0), 50))
+    mp = mesh!(
+        scene, m, color = map(x -> ShaderAbstractions.Sampler(x, anisotropic = 16f0), texture),
+        transparency = false, fxaa=!false
+    )
```
ffreyer, line 563: It seems to me that ShaderAbstractions.jl was meant to allow `Sampler(::Observable)` but that fails to update in GLMakie. Perhaps GLAbstractions needs to `import ShaderAbstraction: update!`?

ffreyer: That code doesn't work, ignores updates

ffreyer:
TODO:
- [ ] Updating `img` with `Sampler(img)` does seem to change `Sampler.data` but creates the same texture as pre-update if that update happens before displaying.
- [ ] Anisotropic filtering doesn't work anymore?

PR #2948, opened 2023-05-12
Mention margin in figure_padding docs

asinghvi17:
Quick doc change from https://discourse.julialang.org/t/how-to-remove-margin-in-single-makie-plot/98169


jkrumbiegel (approved):
In the future, this should just be under `figure_padding` like the other attribute docs are going to be.

PR #2949, opened 2023-05-13
Color blending for datashader

asinghvi17:
ation
As seen in Datashader!

This method basically performs an online color mean, using the `z` coordinate in Point3f input as an index to a color palette.  This could potentially be extended to color gradients as well.

The palette is constructed as a struct with two fields: `palette`, which is a vector of colors, and `null_color` which is a single color which serves as the initial color, e.g. black, white or transparent.
Here are the docs:
```
    AggColorPalette(palette::Vector{ColorType} = Makie.wong_colors(), null_color = zero(ColorType))

Aggregation operation which computes the colorimetric mean of values in a bin.

By default, this operates in RGB space using the Wong color palette.  However, 
the intrepid user can specify their own color space by converting the input palette
into e.g. `Lab` or `XYZ` space.

We personally recommend using `Lab` (CIELAB) space, since addition there is actual color composition.
```

For a quick example with two neighbouring normal distributions, 2 million points,
```julia
normaldist = randn(Point2f, 1_000_000)
ds1 = to_ndim.(Point3f, normaldist .+ (Point2f(-1, 0),), 1)
ds2 = to_ndim.(Point3f, normaldist .+ (Point2f(1, 0),), 2)
datashader(vcat(ds1, ds2); agg = Makie.PixelAggregation.AggColorPalette(), global_post = identity, async_latest = false)
```
download-12

## LAB vs RGBA color mixing
white background compositing


jkrumbiegel:
This looks cool :) I wonder about the implementation, you're currently taking a weighted color mean in every step. Can that be inaccurate somehow, or does that aspect not matter? My first thought was one would use an `NTuple` with `k` `UInt32` elements (one entry could keep track of 4 billion elements, I would assume that's enough?) and just count each category separately, then in the end do one color conversion (also because I assume addition would be faster, even though it takes more memory). The benefit of that method might be that we can still keep an auxiliary array around that contains the counts for each category, thereby enabling the DataInspector to tell the user how many entries of each category each pixel had.

jkrumbiegel:
Yeah in principle one doesn't need ntuples, but it might make the approach easier because you just change the element type and not the dimensionality of the array. On the other hand, you'd have to compile for each `k` (and that in turn could theoretically give some speed benefit if it helps the compiler, no idea).

PR #2950, opened 2023-05-13
Adjust stroke/glow/AA scaling in GLMakie

ffreyer:
# Summary

This pr changes how signed distance fields are scaled when the scale has different x and y scalings. For Circle, Rectangle and RoundedRectangle is adjusted to keep stroke, glow and anti-aliasing width uniform under scaling. For everything else (char markers, BezierPath markers, text) non-uniform scaling now always results in stretching, rather than sometimes stretching and sometimes squishing. (Depended on whether the y scale is larger or the x scale is larger.)

# Description

While messing around with generating a texture for #2944 I noticed that uneven scaling (e.g. `fontsize = Vec2f(80, 20)` in text) results in a reduced strokewidth (and also AA width, though I didn't pay much attention to it.)

texture_before
Screenshot from 2023-05-13 12-11-47

This happens because the scaling of the signed distance field is based on `abs(o_w.z)`, which is the x value of fontsize/markersize. With an uneven size this will effectively squish or stretch the strokewidth in y direction by scale[2]/scale[1], squishing it in this case.

The change I'm proposing here would always result in a stretched strokewidth (and AA region, glow width). For the texture I'm generating that's more useful since the texture mapping will squish the stretched strokewidth back down to (roughly) the given value. In a more general use case I think both options aren't really what we want. It's either "stroke with at most strokewidth" or "stroke with at least strokewidth", when you probably want "stroke with exactly strokewidth regardless of scaling" which I think would require recomputation of signed distance fields since the closest point can change under scaling.

Texture after change:
texture_after

## Type of change

- [x] Minor rendering change

## Checklist


ffreyer:
The shader generated signed distance fields for rectangle and rounded rectangle should be scaling correctly now. Circle seems to be a difficult case, as it becomes an ellipse under uneven scaling which has a complicated coordinate system.

ffreyer:
I found https://iquilezles.org/articles/ellipsedist/ which has a couple of ways to deal with ellipses. I added one for scaled circles.

```julia
begin
    scene = Scene(resolution = (350, 250))
    for (x, marker) in zip([-0.75, -0.25, 0.25, 0.75], [Circle, Rect, :circle, 'a'])
        scatter!(
            scene, [x, x, x], [-0.8, -0.2, 0.5], marker = marker, 
            markersize = Vec2f[(80, 20), (40, 40), (20, 80)]
        )
    end
    scene
end
```

Before this pr:
Screenshot from 2023-05-14 13-00-31

After this pr:
Screenshot from 2023-05-14 13-19-30

CairoMakie (unchanged, for reference):
scaled_marker_cairo

SimonDanisch (approved):

PR #2953, opened 2023-05-14
Fix AA with large linewidth differences

ffreyer:
# Description

This is another fix coming out of #2944. I added per position linewidths with the fixes for line patterns but I didn't think carefully enough about the padding for those cases. Thus we find no anti-aliasing when the change in linewidth from one point to the next is (much) larger than the distance between those points:

```julia
scene = Scene(resolution = (400, 400))
campixel!(scene)
lines!(scene, fill(200, 4), [0, 190, 210, 400], linewidth = [2, 2, 202, 202], color = :black)
scene
```

Screenshot from 2023-05-14 19-56-47

The pr adds scaling so that this is fixed:

Screenshot from 2023-05-14 19-58-38

I didn't bother with patterned lines since, well, they do their own weird thing....
Screenshot from 2023-05-14 19-49-19

## Type of change

## Checklist

### DIFF
--- a/GLMakie/assets/shader/lines.geom
+++ b/GLMakie/assets/shader/lines.geom
@@ -639,9 +639,8 @@ void draw_solid_line(bool isvalid[4])
     vec2 n2 = vec2(-v2.y, v2.x);
 
     // determine stretching of AA border due to linewidth change
-    vec2 edge_v = segment_length * v1.xy + (g_thickness[2] - g_thickness[1]) * n1;
-    vec2 edge_n = vec2(-edge_v.y, edge_v.x);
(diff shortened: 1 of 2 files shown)


ffreyer:
Here's a sketch for the scaling factor
makie_sketch
v and n match the names in the shader, describing the line direction and normal. AA is AA_THICKNESS. The factor 1/2 in the linewidth difference comes from that difference existing symmetrically on both sides of the line segment. y is the final (geometric) padding we apply, which is still representative of a padding by AA_THICKNESS in `uv.y` causing that to have a reduced linewidth.

PR #2954, opened 2023-05-14
Convert kwargs early and nospecialize

jkrumbiegel:
This PR attempts to reduce as much recompilation for slightly different plotting calls as possible by removing all the forwared `kwargs...` which cause excess specialization via `NamedTuple` types, and by compiling certain `plot` methods only once no matter which plot type or `args`.


jkrumbiegel:
I tried a little to get the top-level attributes conversion to work but I got into dispatch hell pretty soon. I think it would be most practical if you transfer only that from your breaking PR over and we merge that first, later we can work on other aspects. The breaking PR is way too big to merge anyway.

SimonDanisch:
Lets close this in favor of https://github.com/MakieOrg/Makie.jl/pull/3082.
Although we should still check if there are useful ideas in here that we should move to #3082 (e.g. the ` Core.kwcall(@nospecialize(nt::NamedTuple), ::Type{Attributes})`)

PR #2959, opened 2023-05-16
Fix 3D heatmaps in CairoMakie

### LINKED ISSUE #2889: CairoMakie plots heatmaps incorrectly in 3D
We need to figure out how to scale the element in 2d to achieve the 3d effect.

```julia
f, a, p = heatmap(-2..2, -1..1, rand(100, 100); axis = (; type = LScene))
```
download-3

ffreyer:
# Description

Fixes #2889, though the heatmap tiles aren't exactly aligned and image still doesn't work

## Type of change

## Checklist


SimonDanisch:
I think this is ready, no? :)

ffreyer:
It fixes the heatmap issue but the same issue still exists for image. I'm not sure what problems I ran into trying to fix it. I think I had issue getting the image to transform like the rect it's drawn on?

ffreyer:
I tried used a poylgon defined with `line_to` + image surface transformed by a matrix. That doesn't work because of perspective projection:
Screenshot from 2023-07-14 15-55-00

ffreyer:
I added the example from the issue as a refimg test. Maybe we should also add one for image and disable it in CairoMakie for now?

PR #2964, opened 2023-05-19
Allow indexing a figure with a CartesianIndex

tecosaur:
# Description

I think it's rather convenient for creating a grid of plots to use a `CartesianIndex`, e.g.

```
for pos in setdiff(CartesianIndex(1,1):CartesianIndex(3,3), (CartesianIndex(2,2),))
    ax = Axis(fig[pos], ...)
    ...
end
```

One can currently use (in the above example) say `[(i,j) for i in 1:3, j in 1:3]` and then splat `pos`, but I think it's fair to call this less idiomatic.

It's also worth noting if somebody had a matrix of objects to be plotted, `CartesianIndicies(matrix)` would be a quite natural way of iterating through them.

## Type of change

A `getindex(::Figure, ::CartesianIndex{2})` method is added so the above example works.

## Checklist

### DIFF
--- a/src/figures.jl
+++ b/src/figures.jl
@@ -99,6 +99,10 @@ function Base.getindex(fig::Figure, rows, cols, side = GridLayoutBase.Inner())
     fig.layout[rows, cols, side]
 end
 
+function Base.getindex(fig::Figure, index::CartesianIndex{2}, side = GridLayoutBase.Inner())
+    fig.layout[Tuple(index)[1], Tuple(index)[2], side]
+end
+
 function Base.setindex!(fig::Figure, obj, rows, cols, side = GridLayoutBase.Inner())
     fig.layout[rows, cols, side] = obj
     obj
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## master
 
+- Add `CartesianIndex`-based indexing of `Figure`s [#2964](https://github.com/MakieOrg/Makie.jl/pull/2964)
+
 ## v0.19.5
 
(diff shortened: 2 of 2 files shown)


src/figures.jl
```diff
@@ -99,6 +99,10 @@ function Base.getindex(fig::Figure, rows, cols, side = GridLayoutBase.Inner())
     fig.layout[rows, cols, side]
 end
 
+function Base.getindex(fig::Figure, index::CartesianIndex{2}, side = GridLayoutBase.Inner())
+    fig.layout[Tuple(index)[1], Tuple(index)[2], side]
+end
```
asinghvi17, line 104: We might also want to implement this logic in [GridLayoutBase](https://github.com/jkrumbiegel/GridLayoutBase.jl), so that indexing for the following things retains the same behaviour:
- `Figure`
- `GridLayout`
- `GridPosition`
- (maybe) `GridSubposition`

jkrumbiegel:
Yeah Anshul is correct, this would need to live in GridLayoutBase and should automatically apply in Makie, as I just forward indexing over to GridLayoutBase here.

I'm not super sure if I like it or not, as I don't see that much additional use personally. The indices are actually ranges usually, and cartesianindices are not meant for that, only for the scalar case. But I also don't mind that much 🤷‍♂️

SimonDanisch:
Ok, then lets close this for now! Thanks @tecosaur for quickly creating a PR, much appreciated, but I guess we should move this to GridLayoutBase.
I'd like this change too, since I'm often also iterating over CartesianIndices for creating gridded plots... Not the worst to do `i, j = Tuple(idx)` but certainly not that nice, especially since CartesianIndex doesn't support indexing, and the `Tuple` is pretty confusing I think.

PR #2966, opened 2023-05-19
Error on unknown plot attributes

jkrumbiegel:
# Description

It's quite annoying that plot functions silently accept misspelled keyword arguments. The reason is that unlike `Block`s like `Axis` etc., plot objects like `Scatter` do not have a completely well-defined list of attributes that they accept. Instead, they have dynamically generated default themes that might merge in themes of other plot types etc.

Also, there are a couple of extra keyword arguments in use that are not meant for the plot objects themselves but serve as metadata for other functionality, for example if `xautolimits = false` is encountered by an `Axis`, it will ignore the x-dimension of that plot for the limit calculation. Because this system has grown in a rather ad-hoc way where keywords were added by different people for different reasons, it's a bit messy and unclear what's allowed and what's not.

Still I believe an allowlist that is too broad is better than having none at all, especially because misspelled keyword arguments are the main thing we want to filter out.

So here I've implemented a very simple method, I just crosscheck the input plot attributes against the keys in the dynamically generated default theme, and against a (for now static and general) allowlist. If it's not in there it's likely to be wrong. Helpfully, a list of all allowed keywords is printed out for the user to quickly find the correct spelling, for example.

It looks like this:

image

I made some simple plots in the REPL which worked after making a short allowlist, however I'm sure CI will uncover quite a few more that need to be added.


src/stats/hist.jl
```diff
@@ -180,6 +181,11 @@ function Makie.plot!(plot::Hist)
         x === :values ? :y : x
     end
     # plot the values, not the observables, to be in control of updating
+
+    for key in [:bins, :normalization, :over_background_color, :over_bar_color, :scale_to, :weights]
```
asinghvi17, line 185: Any particular reason why?  Users could still play around with these, no?

src/interfaces.jl
```diff
@@ -218,13 +218,16 @@ function (PlotType::Type{<: AbstractPlot{Typ}})(scene::SceneLike, attributes::At
...
 
     FinalType = Combined{Typ, ArgTyp}
+
+    trans = pop!(attributes, :transformation, automatic)
+
     plot_attributes = merged_get!(
         ()-> default_theme(scene, FinalType),
-        plotsym(FinalType), scene, attributes
+        plotsym(FinalType), scene, attributes;
+        allowlist = Set([:text, :xautolimits, :yautolimits, :zautolimits, :interpolate_in_fragment_shader, :label, :fxaa, :rasterize, :model])
```
asinghvi17, line 227: `inspectable`, `material`, `texture`, `transformation`, are also things which should be added, plus all the GLMakie lighting stuff (specular, diffuse, etc)

It would also be nice to have a signature which does not perform this check, for easy programmatic passing of attributes inherited from the parent plot!

src/stats/hist.jl
jkrumbiegel: ah yeah meant to delete these from a copy

src/interfaces.jl
jkrumbiegel: `transformation` is already removed elsewhere prior to this check

Wouldn't it be better to group all backend related keywords under one umbrella rather than just letting them potentially collide with the rest?

I think I would like it better if an RPRMakie plot had `meshscatter(data; RPRMakie = (; material = ...)`

jkrumbiegel: same for GLMakie

ffreyer: If we want both name checking and the ability to add arbitrary attributes to a plot we need to separate them somehow. I was thinking we could have one whitelisted attribute Dict of core attributes, which may become fields of specific plot object (`struct Scatter ... end` etc) if we want that, and one Dict with everything else.

Having `RPRMakie = (; ...)` etc is similar to that I guess. Though I would say attributes that influence other objects (inspectable, label, ...) and overwrites (model, view, ...) should also move out of the directly defined attributes. I guess the latter could also be part of the backend groups. The former is kind of the "random stuff that you want to attach to a plot" group, which I think is useful to have in a dynamic way.

jkrumbiegel: we could also add a `misc` allowed key or something which should only ever be set to an `Attributes` instance with further keys. Then we could disallow such a key for any other recipe so we know it will never conflict, and people can use it to carry arbitrary stuff forward. I want to avoid a situation where people make recipes and we break them by introducing some new add-on keyword arg.

jariji:
> misspelled keyword arguments are the main thing we want to filter out.

In my case I pass kwargs to the wrong function. Eg I pass something to `scatter` that I'm supposed to pass to `Axis` or `Figure`. Maybe it could hint "this keyword doesn't apply here, it's supposed to go in `Axis`" or something.

jkrumbiegel:
This has not moved forward in a while, I think the main blocker was that we do like to add certain special attributes to plots here and there, and there was no way to detect them.

Now I had the idea, what if we solve this by convention and say, any keyword starting with `_` is ignored? Of course people can then still misspell a bunch of things starting with `_`, but at least for the normal keywords, this would solve the problem. I think we already kind of follow the convention to name "secret" attributes with `_`. And we could further formalize this by hiding such attributes from tab-completion.

PR #2990, opened 2023-05-30
Continuation of polar axis

ffreyer:
# Description

Continuation of #2014

I rewrote the axis decorations to use a polar transformedscene to avoid most of the explicit polar space -> pixel space transformation. I'm not sure if this is something we want in the end so I'm using my own branch for this.

TODO:
- [x] add polar space overlay scene for axis
- [x] convert axis decorations to polar space
- camera/interactivity:
  - [x] add axis preserving zoom
  - [ ] add reset key
  - [ ] maybe add rotation of axis
  - [ ] maybe add translation of everything (with modifier key)
  - [ ] maybe add zoom of everything (with modifier key, restricted zoom out)
  - [ ] maybe move code to a camera controller
- [x] allow non-square scenes
- [x] figure out label padding
- [x] reserve space for labels
- [x] get clip working again
- [x] replace $\theta$ (currently theta instead)
- [ ] maybe add interaction interface

Reminders on cleanup/visual improvements:
- cleanup debug colors, comments, dead code, tick frequency
- add stroke to r-ticks for visibility
- hide minor grid
- reduce frequency of theta ticks
- switch radians to degrees

## Type of change

## Checklist


src/makielayout/types.jl
```diff
@@ -1589,3 +1589,129 @@ end
...
+        "The width of the outline of `r` ticks. Setting this to 0 will remove the outline."
+        rticklabelstrokewidth = 1.0
+        "The color of the outline of `r` ticks. By default this uses the background color."
+        rticklabelstrokecolor = automatic
+        "Controls if the `r` ticks are visible."
+        rticklabelsvisible = inherit(scene, (:Axis, :xticklabelsvisible), true)
+        "The angle in radians along which the `r` ticks are printed."
+        rtickangle = π/8
+        "The specifier for the angular (`theta`) ticks, similar to `yticks` for a normal Axis."
+        thetaticks = ((0:45:315) .* pi/180, ["$(x)°" for x in 0:45:315])
```
ffreyer, line 1657: `MultiplesTicks` doesn't like to increment by 45° so I switched to manual ticks here.

lazarusA:
I want to use this. Is there like a minimal self-independent recipe for this? i.e. manual transformation of data and manual setting of ticks and labels. It will be useful, until this is merge.

ffreyer:
I would just collect the additions into a file, import the things that are missing and add `Makie.` until it works. If you want a very minimal version you can just copy the Polar transform (i.e. what's in transformations.jl here) and build things from an empty scene, but I think you'll probably end up rewriting half the things in here.

On another note we should discuss what the goal for this pr actually is, i.e. what functionality we want for a first version of a polar axis. As far as I remember this was pretty usable already and I wasn't sure what else was actually important to add.

lazarusA:
For a user perspective it looks like that PR already contains all things that we might need. All further additions could be done afterwards with vanilla Makie :D

CairoMakie/src/overrides.jl
```diff
@@ -115,7 +115,7 @@ draw_poly(scene::Scene, screen::Screen, poly, circle::Circle) = draw_poly(scene,
 function draw_poly(scene::Scene, screen::Screen, poly, polygons::AbstractArray{<:Polygon})
     model = poly.model[]
     space = to_value(get(poly, :space, :data))
-    projected_polys = project_polygon.(Ref(scene), space, polygons, Ref(model))
+    projected_polys = project_polygon.(Ref(poly), space, polygons, Ref(model))
```
ffreyer, line 118: The `project_x` functions in CairoMakie are currently build to take `transform_func` from passed scenes, which isn't correct but usually irrelevant. However for `PolarAxis` I added a poly without transformations to fill the outer background which is plotted incorrectly with this approach, so I did a quick and dirty fix here (and in utils.jl). Proper fix/cleanup is in #3040.

ffreyer:
I added an blocks example page for `PolarAxis` (do I need to add it in some list to include it?), some refimg tests, some tests for the polar transform and fixed a bunch of smaller issues I found. The ref images I added are also in the documentation:

Screenshot from 2023-07-15 17-16-49

Screenshot from 2023-07-15 17-16-57
^ This may fail in WGLMakie due to dashed lines

Screenshot from 2023-07-15 17-17-35
^ This may fail in CairoMakie due to different text stroke handling (GLMakie expands, CairoMakie doesn't) and in WGLMakie due to missing stroke.

src/makielayout/blocks/polaraxis.jl
```diff
@@ -0,0 +1,462 @@
...
+
+
+function Makie.plot!(
+    po::PolarAxis, P::Makie.PlotFunc,
+    attributes::Makie.Attributes, args...;
+    kw_attributes...)
+
+    allattrs = merge(attributes, Attributes(kw_attributes))
+
+    # cycle = get_cycle_for_plottype(allattrs, P)
```
SimonDanisch, line 358: Ah guess we should try to add those back?

SimonDanisch (approved):
Akright, I added cycler support, so I think this is good to go!

jkrumbiegel:
This doesn't yet play well with themes:

grafik

jkrumbiegel:
Other than that, so far I think it looks pretty great!

SimonDanisch:
I think we can leave that for later ...

ffreyer:
Fixed the background color, now it's just the spine that's off. But it's kind of weird to inherit that from e.g. just the top spine of an Axis... We should probably just have a `PolarAxis` section in the theme instead?
Screenshot 2023-07-18 122932

jkrumbiegel:
Ah actually I think the spine color of `Axis` already inherits `linecolor`, so that could be done for PolarAxis as well

SimonDanisch (approved):

Issue #1455, opened 2021-11-18
Error bars scale with x-axis scaling


Klafyvel:
When using `errorbars` with `whiskerwidth` ≠ 0 on an axis with `scale=log10`, the whiskers scale with the axis.

```julia
using CairoMakie

f = Figure()
ax = Axis(f[1,1], xscale=log10)

xs = [1, 1/10, 1/100]
ys = randn(3)
δys = randn(3) .* 0.1

errorbars!(ax, xs, ys, δys, whiskerwidth = 3)

f
```

![image](https://user-images.githubusercontent.com/6209071/142390867-cc1640b7-a913-4ca4-9b9e-cd6057ba81d9.png)

Edit : package version

```julia
  [13f3f980] CairoMakie v0.6.6
```

KronosTheLate:
I came here to raise the same issue:

<img width="620" alt="image" src="https://user-images.githubusercontent.com/61620837/170044593-00d4a454-5806-417a-80f0-de22eae9b3ec.png">

Also, the value of whiskerwidth in my plot is 10_000. Perhaps it could be possible to set the whiskerwidth in terms of pixels, for non-identity scaling?

jkrumbiegel:
It is supposed to be in screen units, it's just that I forgot to include scene transformations at the time. But now that we have space settings it should be easier to do this anyway. Right now it's implemented by projecting once into screen space and then back into data space because at the time there was only data space to use.

ffreyer:
This was part of my motivation for #1730. Even with that it'll be annoying though, since whiskers are centered at a data space position with a pixel width and thickness. It might actually be easier/better to implement it with scatter for GLMakie. (I.e. make a specialized line character, center-aligned and scaled with a `Vec2f(width, height)`)

chelate:
Had this problem with boxplots as well....

I know this may be an unpopular opinion, but I believe one should be able to change the axis labels to log space and then transform the input functions as one pleases. It's much easier to manipulate things this way without worrying about yet another coordinate. A broadcasted log10.(y) is much more readable than another keyword.

KronosTheLate:
You could also make the xlabel something like log(x), or manually set the xticks by passing (i believe it is) a touple og vectors, the first with the linear ticks and the second with the derired display strings. Otherwise, I personally really like not having to transform the data, and have it happen automatically to match the actual axis scale.

Klafyvel:
Amazing, thanks.

PR #3012, opened 2023-06-16
Fix errorbar and bracket interaction with transform functions

### LINKED ISSUE #3005: Whisker width on error bars does not work as expected on log scale plot.
I would like to display error bars on my graph which has the x axis plotted on a log scale. Choosing a large enough whisker width does allow them to show up, however, all whisker widths are of different sizes for each point, shown in below MWE (generated using CairoMakie):
```julia
f = Figure()
ax = Axis(f[1,1], xlabel="n", xscale=log10)
ns = 2 .^ 1:8
ys = sin.(ns)
errors = abs.(ys) ./ 10

errorbars!(ax, ns, ys, errors; whiskerwidth=100)
scatter!(ax, ns, ys)
display(f)
```
Result:
temp

How can one get constant width whiskers for the errorbars?

## Version Info

Makie: `[13f3f980] CairoMakie v0.10.6`
Julia: `v1.9.1`
OS: Windows 11

ffreyer:
# Description

Fixes #3005 and #1455

I replaced `scene_to_screen` and `screen_to_scene` with `plot_to_screen` and `screen_to_plot` here, which includes applying the model matrix, transform function and considers `plot.space`. I changed the name because these functions are more closely related to what a plot is doing, and a plot can have different transformations (and coordinate space) than a scene.I'm also hoping it trips up CI if I missed some calls to the old functions.

This now also fixes some issues with bracket. On master it doesn't react to window resizing and it doesn't include transformations when calculating its positions.

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/bracket.jl
+++ b/src/basic_recipes/bracket.jl
@@ -84,6 +84,7 @@ function Makie.plot!(pl::Bracket)
         end
 
         notify(bp)
+        notify(textpoints)
     end
 
     notify(points)
--- a/src/basic_recipes/bracket.jl
+++ b/src/basic_recipes/bracket.jl
@@ -54,16 +54,17 @@ function Makie.plot!(pl::Bracket)
         return to === automatic ? Float32.(0.75 .* fs) : Float32.(to)
     end
 
-    onany(pl, points, scene.camera.projectionview, pl.offset, pl.width, pl.orientation, realtextoffset,
-          pl.style) do points, pv, offset, width, orientation, textoff, style
+    onany(pl, points, scene.camera.projectionview, pl.model, transform_func(pl), 
+          scene.px_area, pl.offset, pl.width, pl.orientation, realtextoffset,
+          pl.style) do points, _, _, _, _, offset, width, orientation, textoff, style
 
         empty!(bp[])
         empty!(textoffset_vec[])
         empty!(textpoints[])
 
         broadcast_foreach(points, offset, width, orientation, textoff, style) do (_p1, _p2), offset, width, orientation, textoff, style
-            p1 = scene_to_screen(_p1, scene)
-            p2 = scene_to_screen(_p2, scene)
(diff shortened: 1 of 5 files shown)


src/basic_recipes/bracket.jl
```diff
@@ -105,10 +107,12 @@ function Makie.plot!(pl::Bracket)
...
     end
 
-    series!(pl, bp; space = :pixel, solid_color = pl.color, linewidth = pl.linewidth, linestyle = pl.linestyle)
+    # Avoid scale!() / translate!() / rotate!() to affect these
+    series!(pl, bp; space = :pixel, solid_color = pl.color, linewidth = pl.linewidth, 
+        linestyle = pl.linestyle, transformation = Transformation())
     text!(pl, textpoints, text = texts, space = :pixel, align = pl.align, offset = textoffset_vec,
         fontsize = pl.fontsize, font = pl.font, rotation = autorotations, color = pl.textcolor,
-        justification = pl.justification)
+        justification = pl.justification, model = Mat4f(I))
```
ffreyer, line 115: Currently `space = :pixel` still includes model transformation (at least in GLMakie). Since `plot_to_screen` deals with them now, they'd be included twice. To prevent that I'm overwriting model/transformation. I'm using transformation for series because model doesn't get passed down.

src/basic_recipes/error_and_rangebars.jl
```diff
@@ -187,19 +187,18 @@ function _plot_bars!(plot, linesegpairs, is_in_y_direction)
 
     scene = parent_scene(plot)
 
-    whiskers = lift(plot, linesegpairs, scene.camera.projectionview,
-        scene.camera.pixel_space, whiskerwidth) do pairs, _, _, whiskerwidth
```
ffreyer, line 191: With `camera.pixel_space` this sometimes fails to update correctly when zooming or resizing. `scene.px_area` works fine.

jkrumbiegel (approved):

Issue #3010, opened 2023-06-15
GLMakie Compilation Error: AssertionError: status == GL_FRAMEBUFFER_COMPLETE 


OmarAshkar:
Hi,

I had previous issue and I had to update my OpenGL using the following variable:

```
export MESA_GL_VERSION_OVERRIDE=4.6
```

Now I am getting the following error when I run `using GLMakie`:

```
ERROR: LoadError: AssertionError: status == GL_FRAMEBUFFER_COMPLETE
```

Any help is appreciated.

Thank you!

ffreyer:
We should update that assert to show what the status actually is...

This should be a relatively minimal example for setting up a framebuffer and showing its status. Maybe that reproduces the error
```julia
using GLFW
using ModernGL

begin
    window = GLFW.Window()

    # Avoid the OS complaining about non-responsive window
    @async let
        while !GLFW.WindowShouldClose(window)
            GLFW.PollEvents()
            GLFW.SwapBuffers(window)
            sleep(1/30)
        end
        GLFW.DestroyWindow(window)
    end

    # color buffer texture
    cb_id = let
        ref = UInt32[0]
        glGenTextures(1, ref)
        ref[]
    end
    glBindTexture(GL_TEXTURE_2D, cb_id)
    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, 400, 300, 0, GL_RGBA, GL_UNSIGNED_BYTE, C_NULL)

    # framebuffer
    fb_id = let
        ref = UInt32[0]
        glGenFramebuffers(1, ref)
        ref[]
    end
    glBindFramebuffer(GL_FRAMEBUFFER, fb_id)
    glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, cb_id, 0)

    statuses = Dict(
        GL_FRAMEBUFFER_COMPLETE => "GL_FRAMEBUFFER_COMPLETE",
        GL_FRAMEBUFFER_UNDEFINED => "GL_FRAMEBUFFER_UNDEFINED",
        GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT => "GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT",
        GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT => "GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT",
[...]

PR #3014, opened 2023-06-16
Better error message when the GL status is not `GL_FRAMEBUFFER_COMPLETE`

asinghvi17:
# Description

Quick fix to get a better error message.

For full effect, this should probably use a lookup table/Dict.

Ref #3010

## Type of change


ffreyer:
I think we should add a lookup to this before merging. Getting the raw numbers doesn't seem very useful to me.

PR #3020, opened 2023-06-21
Clean up list of available formats

jkrumbiegel:
Removes formats we don't support (anymore).


SimonDanisch:
We still support JPG though!

jkrumbiegel:
Really? I remember removing conversions for that a long time ago in CairoMakie

SimonDanisch:
Yup, we have a fallback in Makie.jl :D

SimonDanisch:
Should we only remove bmp and leave jpg for now?

jkrumbiegel:
Maybe I'm pedantic in that way, but jpg is just not a good format for thin lines etc., we all know what the block artifacts look like. So I'm not really for keeping it. bmp is ok but just makes huge files.

SimonDanisch:
Well, we don't only offer thin line plots, and it's basically free to offer jpg... So I don't see the harm ;) 
But I guess anyone can still do `save("plot.jpg", colorbuffer(fig))`, so, i guess it could be ok to remove it entirely...

jkrumbiegel:
Yeah, either way, it doesn't matter that much

SimonDanisch (approved):

Issue #794, opened 2020-12-24
Slider updates not reflected graphically


yha:
When an `LSlider`'s `value` is updated programatically, it doesn't show. Here you can see the text updating when the button is clicked, but not the slider:
```
scene, layout = layoutscene()
btn = layout[1,1] = LButton(scene)
sl = layout[2,1] = LSlider(scene; width=500)
txt = layout[3,1] = LText(scene)
on(btn.clicks) do _
    sl.value[] = 10
end
on(sl.value) do n
    txt.text[] = string(n)
end

scene
```

yha:
Looks like the same happens with text boxes:
```

scene,layout = layoutscene()
btn = layout[1,1] = LButton(scene)
tb = layout[2,1] = LTextbox(scene; width=500)
txt = layout[3,1] = LText(scene)
on(btn.clicks) do _
    tb.stored_string[] = "foo"
end
on(tb.stored_string) do n
    txt.text[] = string(n)
end

scene
```

jkrumbiegel:
The problem is that for some observables that are outputs and inputs semantically, it's really hard to set them up that way. `value` is an output of the slider, it doesn't affect the internal state of the slider if you change it because it's not backwards connected. Same with the textbox. The slider has the method `set_close_to!` to set the value (problem with setting exactly is that the slider's range might often not contain the exact value specified). For textboxes it's `set!(textbox, string)` currently https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/makielayout/lobjects/ltextbox.jl#L308-L320

yha:
I see, that does make sense. Perhaps the attribute docs should mention how to set them. I'll try putting together a docs PR.

> Same principle with the textbox. For textboxes it's `set!(textbox, string)` currently https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/makielayout/lobjects/ltextbox.jl#L308-L320

I see that this is not exported, which makes sense for a very general name like `set!`. Maybe it would make sense to add a `setindex!` method for that, so the text can be set as `textbox[] = string`?

jkrumbiegel:
Yeah `set!` is too generic while not being a Base method, maybe there's a better possibility there. I didn't finish that part of the LTextbox interface yet. I'll think about setindex, although I usually prefer more obvious function interfaces

bjarthur:
maybe just add "see also set_close_to!" in the docstring for Slider?

bjarthur:
[...]

PR #3035, opened 2023-06-29
add function signature to set_close_to! docstring

bjarthur:
see https://github.com/MakieOrg/Makie.jl/issues/794#issuecomment-1613872576


jkrumbiegel:
Thanks!

Issue #2881, opened 2023-04-21
`boundingbox` cleanup


ffreyer:
In short I think we should

1. Clarify what `data_limits` and `boundingbox` does
2. Include the extent of, for example, a scattered marker in boundingbox
3. Add a `space` option in `boundingbox`
4. Add per-point `boundingboxes` and other intermediate stages

---

Currently we have both `data_limits` and `boundingbox` which for most plots only differ by an application of the model matrix

https://github.com/MakieOrg/Makie.jl/blob/bc13edb84e293ea26d143cb17789c480b4400129/src/layouting/boundingbox.jl#L1-L8

which often times is just an identity transform. The exception to this is `text` which returns a pixel space boundingbox for the scattered string(s). `text` also includes the string boundingbox depending on markerspace in `data_limits`

https://github.com/MakieOrg/Makie.jl/blob/84a1cdf49d718f19b6fee36978c265fa1ad3ab03/src/layouting/data_limits.jl#L43-L53

which goes against my understanding of what `data_limits` is supposed to be.

I think `data_limits` should only care about the positional arguments passed to a plot, as it does in most cases now. `boundingbox` on the other hand should be a tight boundingbox around the full plot, including markersize, linewidth, etc. Including those will require some back and forth transformation, e.g. space -> markerspace, add markersize, markerspace -> space for scatter. (1, 2)

In general it would be good to be able to specify the output space of a boundingbox. That way we could unify text with the other bounding boxes, and have some methods that skip the final transformation. (3) 

Making different stages of the boundingbox computation available would also be useful to skip some of steps which might be unnecessary in some use cases. The pipeline here might be: (4)
1. Per position `rawboundingboxes(plot)` without the position argument applied (probably always in markerspace, as positions may influence the result in other spaces)
2. Per position `boundingboxes(plot, space = plot.markerspace)` with positions applied (transform positions from space -> markerspace here, user defined output space)
3. The combined `boundingbox(plot, space = plot.space)` which merges all the individual bounding boxes from the previous step

These steps don't make sense for every (primitive) plot, but I think it's quite obvious when a plot has them and when it doesn't. Though it might be a bit more difficult to generalize this to recipe. Maybe we could just use `boundingboxes` on the child plots that implement them, or simply make that function only work on the primitive types.

ffreyer:
[...]

PR #3040, opened 2023-07-02
Rework projection code

### LINKED ISSUE #3013: Cleanup projection code
We have quite a bit of code that deals with projecting positional data from one space to another. I think there is a fair amount of duplicate code in this, which would be good to remove/unify. The `project` methods also need to be clearer about what they do I think. 

In terms of clarity it might be good to keep `project(cam, input_space, output_space, pos)` and add more `space` options to exhaust all the options. It might also be nice to allow `input_space => output_space` as an argument here.

---

In https://github.com/MakieOrg/Makie.jl/blob/master/src/camera/projection_math.jl we have `to_world` and various `project` methods which aren't all that clear. These all ignore `model` and `transform_func`.

In https://github.com/MakieOrg/Makie.jl/blob/master/src/basic_recipes/error_and_rangebars.jl we have `scene_to_screen` and `screen_to_scene`  which functionally match `project`. After #3012 these will be changed to `plot_to_screen` and `screen_to_plot` and include model and `transform_func`.

GLMakie has some code doing a full plot space -> screen space projection without using `project` at all in https://github.com/MakieOrg/Makie.jl/blob/master/GLMakie/src/drawing_primitives.jl (lines).

CairoMakie defines its own project functions in https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/utils.jl which are used instead of Makie's `project`.

ffreyer:
Since this pr is shaping out to be breaking I'm planning on handling #3013, #2881 and #1730 here.

# Description

This pr has two main goals. The first is to reduce the amount of (more or less) redundant code dealing with projecting positions. The second is to add missing `space` options. The pipeline of coordinate spaces/transformations we have looks like this

spaces

where transformed, world and eye space are not accessible through `space`. (See also #2766) These are added here.

Something I'm not sure about yet is whether pixel and relative space should ignore `Transformations`. It might be a bit weird that things like `scale!()` or polar transforms don't work with them. If we do allow them to affect pixel and relative space it might make sense to separate them completely from `space`. Otherwise naming will get confusing I think. We should also disallow transform propagation if spaces don't match.

I'm also considering restarting #1730 here, as it may change how some things need to work.

Closes #3013
Closes #2766 since this also implements transformed space.

TODO
- [x] Rework `project` to be more useful. It now includes `model` and `transform_func` depending on input space and can deal with vector inputs. 
- [x] Add `projection_obs` to simplify projecting data in observables
- [x] add:transformed, :world and :eye space to make every coordinate space accessible
- [x] Remove redundant code from Makie
- [x] Update GLMakie
- [x] Rework CairoMakies projection code
- [x] Update WGLMakie
- [ ] Update RPRMakie
- [ ] fix axis scale problem (bracket causes error due to late update?)
- [ ] figure out how to deal with transformations vs camera projections (both controlled via space or just projections)
- [ ] split spaces
- [ ] update bounding boxes with new project code
- [ ] consider doing #2881 here
- [ ] Documentation
    - new page
    - docstrings, e.g. hvlines, project functions, ... 
- [ ] Cleanup of old code and notes

Other changes
- fixed duplicate update issues with projectionview
- allow `transform_func` to be passed directly as a plot kwarg
- remove `transformation` and `transform_func` from plot attributes on construction

## Type of change

While `project` may or may not be internal, I think it is fairly likely to be used outside in similar situations as `errorbars`, `h/vlines`, `h/vspan`, `bracket`, etc. (I.e. whether part of or the full plot needs to be in pixel space.)

## Checklist


src/utilities/utilities.jl
```diff
@@ -242,15 +244,31 @@ function same_length_array(arr, value::Vector)
...
         i > N2 && return ET(fillval)
         @inbounds return vec[i]
     end)
 end
 
+function to_ndim(trg::VT, src::VecTypes{N2}) where {N, N2, VT <: VecTypes{N}}
+    VT(ntuple(Val(N)) do i
+        @inbounds i > N2 ? trg[i] : src[i]
+    end)
+end
```
ffreyer, line 261: I did some benchmarking on `to_ndim` while working on this pr. It's relatively slow if the type isn't known at compile time. You can see this by benchmarking the function directly:

```julia
T = Point4f
p2 = Point2f(1, 2)
fillval = 1f0
@benchmark Makie.to_ndim($T, $p2, $fillval) # ~74ns
```

And comparing with a wrapped version where the type is hardcoded:

```julia
f2(p, v) = Makie.to_ndim(Point4f, p, v)
@benchmark f2($p2, $fillval) # 1.4ns
```

Using a function barrier does not improve performance:

```julia
f1(T, p, v) = Makie.to_ndim(T, p, v)
@benchmark f1($T, $p2, $fillval) # ~74ns
```

However if you use a instantiated object rather than passing a type things become fast with a "dynamic" type:

```julia
p4 = Point4f(1)
@benchmark Makie.to_ndim($p4, $p2) # 1.4ns
```

This also has a real effect on `Makie.project`. With 1000 Point2f's it cuts down 150µs to about 9.6µs for me.

src/camera/camera.jl
```diff
@@ -89,11 +91,13 @@ function Camera(px_area)
...
-    # But nobody should do that, right?
-    # GLMakie uses map on view
+    # Both view and projection trigger an update of projectionview which can 
+    # have rather expensive listeners. To avoid running those twice we block 
+    # them when updating view. 
+    lock = Observables.listeners(camera.projectionview)[1][2]::ObservableLocker
+    lock!(lock)
     camera.view[] = view
+    unlock!(lock)
     camera.projection[] = projection
```
ffreyer, line 101: Alternatively pull observables or https://github.com/JuliaGizmos/Observables.jl/pull/109

ffreyer:
I've been thinking about whether `Transformation`s should be part of what `space` controls or not, and I'm leaning towards separating them.

Conceptually I see `Transformation`, i.e. the transform_func, translate!, rotate! and scale! (or the model matrix) as transformations of the input data, and transformations derived from :world, :eye, :pixel, :relative and :clip space as transformations of the "world". In other words `Transformation` describes how data is placed in the world and the latter group describes the coordinate system of the world (or how the world maps to something displayable).

In practice it seems reasonable to me that any plot can be scaled, translated, rotated or even have a nonllinear transform_func like a polar transform before being placed into a world, even if that world is in pixel space or clip space. (This is also the way it works if we use a different camera rather than adjusting space.)

Detaching `space` and `Transformations` would allow you to define a pixel space box for example, and use `translate!()` it to where it needs to be without directly updating the box. This might be useful when updating that box involves expensive computations. Another example would a polar transform in pixel or relative space. You could effectively change the units of the radius with this.

Changes I'm thinking about making:
- add `transform_func` as a kwarg to `plot!` (which gets moved to `Transformation`, not attributes)
- maybe model decomposition into rotation, translation, scale so that it can be absorbed into `Transformation` (only static model matrices)
- `transform = :full / :linear / :nonlinear / :none` as an attribute for controlling transform_func and model application (transformation is used for `Transformation`)
- maybe detach `Transformation` from parents if spaces do not match

ffreyer:
With higher test accuracy this fails
https://github.com/MakieOrg/Makie.jl/blob/83fc003e775c9468aa362a15b45b9e8d675234c6/ReferenceTests/src/tests/examples3d.jl#L461-L472
because the axis limits have become a bit larger, changing the generated grid. I assume this happens because the data_limits of the contour plot changed to include the z-translation it has. Not sure what caused this specifically, but I think this is something we probably want anyway.

ffreyer:
I think this pr was trying to do too much at once. I'll close this and hopefully work on it piece by piece

PR #3041, opened 2023-07-03
Fix automatically generated normals

jkrumbiegel:
grafik


SimonDanisch:
We should also fix that no normals are generated as default... That was always intended, but I think it has been bitrotting!

jkrumbiegel:
So not merge this?

SimonDanisch:
Hm the example is nice & the comment about order dependence...
But would be nice to fix the normal generation as well... Can take a look tomorrow!

SimonDanisch (approved):

SimonDanisch:
Ok, the example didn't make that much sense anymore, after fixing the automatic normal generation.
I removed it for now, and hope to add it back better later!

PR #3042, opened 2023-07-03
[WGLMakie] Add automatic figure resizing

### LINKED ISSUE #1318: WGLMakie Resize Feature
I was wondering if it would be possible to auto-resize a WGLMakie figure?  For example, when the figure is displayed in a VSCode Plot Pane window, would it be possible to:

1. Fill the plot pane window based on the window initial size
2. Auto-resize the figure based on a change to the plot pane size?  

Basically, I'm looking for similar functionality as the PlotlyJS.jl package, which fills and resizes based on the Blink window or the VSCode Plot Pane size.

staticfloat:
# Description
This implements basic resizing support for WGLMakie figures.  This has primarily been tested in the VSCode plot pane and the JSServe auto-launched browser window (using Chrome and Safari).

The implementation creates a new `resize` event that the javascript frontend can send back to the Julia backend, with a computed size (in pixels) that the backend should resize to.  Additionally, the javascript `throttle()` function is altered to no longer drop the last few events when throttled, it instead delays the last event until the end of the throttling period, which is important when resizing quickly so that the final size is faithfully recorded.

Fixes #1318

## Type of change

## Checklist

- [X] Added an entry in NEWS.md (for new features and breaking changes)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 - Adjusted scaling of scatter/text stroke, glow and anti-aliasing width under non-uniform 2D scaling (Vec2f markersize/fontsize) in GLMakie [#2950](https://github.com/MakieOrg/Makie.jl/pull/2950).
 - Scaled `errorbar` whiskers and `bracket` correctly with transformations [#3012](https://github.com/MakieOrg/Makie.jl/pull/3012).
 - Updated `bracket` when the screen is resized or transformations change [#3012](https://github.com/MakieOrg/Makie.jl/pull/3012).
+- Added auto-resizing functionality to WGLMakie plot figures [#3042](https://github.com/MakieOrg/Makie.jl/pull/3042)
 
 
 ## v0.19.6
--- a/WGLMakie/src/events.jl
+++ b/WGLMakie/src/events.jl
@@ -102,6 +102,9 @@ function connect_scene_events!(scene::Scene, comm::Observable)
                     e.keyboardbutton[] = KeyEvent(code_to_keyboard(keyup), Keyboard.release)
                 end
             end
+            @handle msg.resize begin
+                resize!(scene, tuple(resize...))
+            end
         catch err
             @warn "Error in window event callback" exception=(err, Base.catch_backtrace())
         end
--- a/WGLMakie/src/three_plot.jl
+++ b/WGLMakie/src/three_plot.jl
@@ -45,8 +45,8 @@ function three_display(session::Session, scene::Scene; screen_config...)
     window_open = scene.events.window_open
     width, height = size(scene)
     canvas_width = lift(x -> [round.(Int, widths(x))...], pixelarea(scene))
-    canvas = DOM.um("canvas"; tabindex="0")
-    wrapper = DOM.div(canvas)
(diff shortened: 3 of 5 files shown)


staticfloat:
NOTE: I've run `deno bundle` in order to be able to use this branch on projects, but the diff to the bundled source seems pretty large; perhaps I'm using a different version of `deno` or something?

SimonDanisch:
Cool! :) Will need to test this, to make sure it doesn't behave weirdly on pluto etc...

>but the diff to the bundled source seems pretty large; perhaps I'm using a different version of deno or something?

I think is bound to happen when checking in something automatically generated. 
I still need to get around to make this nicer, but I had the problem that I needed JSServe + WGLMakie to run on systems that have no deno binaries, so not having to install deno on the target platform was needed and therefore checking in the bundles.

Ideally, the bundles would be artifacts, I think @fonsp did something like that.

SimonDanisch:
Welp, pluto doesn't like it ^^ 
image
And IJulia neither.
If nobody has a smart way to make this work in IJulia/Pluto, while allowing to auto resize in the right contexts (or to the right size?), I'd say we make this optional for now via `WGLMakie.activate!(resize_to_parent=true)` or `set_theme!(WGLMakie=(;  resize_to_parent=true))`.
Open to a better name, but `window` / `screen` seems wrong, since it's not really resizing to a window, but just to the html parent it's inserted in.

staticfloat:
@SimonDanisch as an alternative to #3044, what do you think of this latest commit; it simplifies things somewhat by forcing all elements between the `canvas` and the `html` tag to have `height: 100%` applied, which allows the resizing calculation to be done much more simply.  The most dubious thing about this is that it's basically triggering off of something to figure out if we're in a VSCode plot panel or not.  I gave up trying to make it work with `JSServe`, as there are too many pathways where we may or may not get elements without a `height: 100%` applied.

SimonDanisch:
Hm, I think I'd rather have more complicated height/width calculation then having to mess with css/html that we don't own (and relying on finding the VSCode div).

kbarros:
It looks like the aspect ratio of the rendered image changes with the panel dimensions. See attached two images.  Is there a way to specify a fixed aspect ratio to avoid this distortion? Tested on Mac/VSCode.

image
image

SimonDanisch:
What axis type is this?

kbarros:
> What axis type is this?

LScene: https://github.com/SunnySuite/Sunny.jl/blob/main/src/Plotting.jl#L5

SimonDanisch:
What happens if you do the same with GLMakie? I thought LScene shouldn't change the aspect ratio.
But since resizing should work pretty similar between GLMakie and this PR, I'd be surprised if the aspect ratio change is due to this PR.

kbarros:
If I try with GLMakie then the aspect ratio is preserved.

kbarros:
Similar distortion happens inside Jupyter notebook (see attached image). However, it's exciting to see that this PR is so close to working!

image

SimonDanisch:
Merged in https://github.com/MakieOrg/Makie.jl/pull/3044

PR #3043, opened 2023-07-04
Add nothing case for theme

pepijndevos:
# Description

When calling `theme(scene, :key)` in a recipe, if scene is nothing it seems to look up the default theme.
But `theme(scene)` has no such fallback.

## Type of change

## Checklist

### DIFF
--- a/src/theming.jl
+++ b/src/theming.jl
@@ -196,6 +196,7 @@ function with_theme(f, theme = Theme(); kwargs...)
     end
 end
 
+theme(::Nothing) = CURRENT_DEFAULT_THEME
 theme(::Nothing, key::Symbol) = theme(key)
 function theme(key::Symbol; default=nothing)
     if haskey(CURRENT_DEFAULT_THEME, key)


SimonDanisch:
Hm, the thing is, `CURRENT_DEFAULT_THEME` shouldn't be directly handed to the user for mutation.
Although, `set_theme!` also mutates it, but I'm not sure I want more ways to mutate the default_theme.
What exactly do you want to do with `theme()`?

pepijndevos:
Get an attribute that may or may not exist. Apparently a recipe gets called with a nothing scene sometimes, so if you write `get(theme(scene), key, default) instead of `theme(scene, key)` it'll error

SimonDanisch:
Ah maybe we should add `theme(scene, key; default)` for more signatures then...

pepijndevos:
That would be another option yes. For reference, this is a line of code I currently have:
```julia
        xgridcolor = get(something(get(theme(scene), :Axis, nothing), Attributes()), :xgridcolor, RGBAf(0, 0, 0, 0.12)),
```

SimonDanisch:
Do you want to implement `theme(scene, key; default)` in this PR?

pepijndevos:
Sure after I'm back from holiday :)

pepijndevos:
Superseded by #3214 since this one is like an one line change that's now out of date

PR #3044, opened 2023-07-04
Sf/vscode resizing

SimonDanisch:
Continue #3042 and make it optional!


staticfloat:
FTR, I think this is approach is fine.  I just put `WGLMakie.activate!(;resize_to_body=true)` at the top of my script, and it all works.  Thanks!

kbarros:
I can confirm the aspect ratio bug has been fixed and with this branch, resizing now works as intended! 

If I may request a future feature: Can resizing take into account the width of the panel? See for example the two attached VSCode images. Changing the width of the panel will not enlarge or shrink the image size.

image
image 

This behavior is shared by GLMakie pop-up windows. In GLMakie, it's fine. However, "tall and narrow" panels seem to appear by default in VSCode.  The image below shows the behaviors of a Plots.jl image that resizes according to the panel width inside VSCode.

image

Interestingly, the Plots.jl image ignores the panel height (so it's the converse of Makie).

staticfloat:
> I wonder if an ideal behavior is to use something like min(width, height) so scroll bars don't appear? It seems like a tricky design decision.

I agree that locking to the minimum width is likely the ideal solution, but should probably be implemented in a follow-on PR.

SimonDanisch:
Actually, I accidently achieved the plots.jl behaviour by using the wrappers div `offsetWidth`/`offsetHeight` attributes, instead of using the exact body dimensions.
If that's more desired, we could use that?
That would also simplify the `resize_callback` code a lot:
```julia
  function resize_callback() {
      comm.notify({ resize: [wrapper.offsetWidth, wrapper.offsetHeight] });
  }
```

kbarros:
Interesting. I'd be happy to test that change in VSCode and Jupyter if you put it on a branch.

SimonDanisch:
@kbarros: https://github.com/MakieOrg/Makie.jl/pull/3056

PR #3056, opened 2023-07-10
resizing to width

SimonDanisch:
PR for https://github.com/MakieOrg/Makie.jl/pull/3044#issuecomment-1629250398


kbarros:
For me, this one isn't working well. The rendered image is not expanding with the width _or_ height of the panel. Below shows a widened panel with no enlargement of the WGL viewport.

image

Also, there is a bug where if I shrink the height of the panel and reexpand, the image totally disappears.

PR #3057, opened 2023-07-10
display fixes

SimonDanisch:
So, GLMakie not opening a window with activated Plotpane really was not a great idea...
Lots of complaints, and I also tripped over this quite a bit.
This PR changes this back to opening a window by default for GLMakie!
But this means, that one now needs to call `GLMakie.activate!(inline=true)` to get GLMakie back into the plotpane.

This PR also changes a few other things:

* `can_show_inline(backend)` was weirdly buggy with VSCode, a.k.a most of the time it worked that e.g. WGLMakie would show in the plotpane if the plotpane is enabled, but sometimes it just stopped working...Not sure why, but looking at the VSCode code, it should never work the way `can_show_inline` is implemented (not checking for `MIME"juliavscode/html"`) - but it definitely worked correctly most of the time 🤷 
* now one can do `display(plot; inline=false)` to force a window even if inline is true. Of course this doesn't work the other way around, since VSCode somehow doesn't get into the show path anymore with calling `display(fig; inline=true)` directly... Not sure why...

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -106,7 +106,7 @@ mutable struct ScreenConfig
     end
 end
 
-const LAST_INLINE = Ref{Union{Makie.Automatic, Bool}}(Makie.automatic)
+const LAST_INLINE = Ref{Union{Makie.Automatic, Bool}}(false)
 
 """
     GLMakie.activate!(; screen_config...)
--- a/docs/documentation/fonts.md
+++ b/docs/documentation/fonts.md
@@ -42,7 +42,6 @@ Here's an example:
 ```julia
 using CairoMakie
 CairoMakie.activate!() # hide
-Makie.inline!(true) # hide
 
 f = Figure(fontsize = 24, fonts = (; regular = "Dejavu", weird = "Blackchancery"))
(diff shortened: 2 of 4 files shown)


jkrumbiegel:
What about changing `display(plot; inline=false)` to `display(plot; window = true)` instead? I never found `inline` that clear to be honest. Although, does opening an os window or a browser tab count as a "window"? At least, if I think about it, the default in Julia is to display however the display stack is currently configured. If I instead always want to force a window, shouldn't that be positively affirmed using a `true` option instead of negating the opposite?

We could also think about having a slightly more "convenient" syntax for opening new windows with this. Instead of `window = true` one could maybe pass `window = :new` or so. Just a suggestion. But I felt so far users didn't find the `display(GLMakie.Screen(), fig) ` super intuitive.

ffreyer:
I would also suggest adding `inline` as a theme argument to make it more accessible. `display` feels kind of niche to me in Makie and it also only applies to this one figure.

SimonDanisch:
Yeah I think those are good ideas, but right now I just want to make a small bug fix PR.
We can target bigger API changes in the beta 0.20 branch!
The `display(fig; inline=true)` thing is basically also just a bug fix, considering that I intended all `activate!` kwargs to be usable in `display` too.

Issue #3050, opened 2023-07-06
WGLMakie plots "interleave" traces when updated chaotically


staticfloat:
- [X] are you running newest version (version from docs) ?
- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [X] What platform + GPU are you on?

Reproducible on macOS and Linux.

### Issue description

When dragging sliders around crazily, one occasionally sees a flicker in the plots that can look like garbage.  Careful examination shows that it is two traces that have their vertex data interleaved with eachother.  [Here's a video](https://youtu.be/YHu-kcuIl4k) showing the problem, but a single screen grab I think gives the most information:

<img width="978" alt="Screenshot 2023-07-06 at 09 38 13" src="https://github.com/MakieOrg/Makie.jl/assets/130920/56cecdfa-4d6e-41ea-8137-3b3018508087">

Here you can clearly see that the y coordinates seem to be alternating between two different sinusoids one with a higher frequency and one with a lower frequency.  I believe this is because somehow data from one update (with frequency A) and data from another update (with frequency B) have somehow become "interleaved", and that is what is being displayed.  It is likely that the X coordinates suffer the same fate (as they're all part of the same update) but because those values are always the same, it doesn't matter much.

I briefly looked into logging the data being transferred, but I'm unsure what the best way to do this is; I imagine it is possible the issue is in Julia code, in Javascript code, or even is a race condition when updating the WGL buffers.  I'm open to suggestions on how to track this down.

### Code reproducer

```julia
using Revise
using WGLMakie, JSServe
using WGLMakie.Makie.StructArrays

function slider_test()
    fig = Figure()
    ax = Axis(fig[1, 1:2])

    # Create two sliders
    sg = SliderGrid(
        fig[2, 1],
        ((;label = name,range = LinRange(0, 2π, 50),) for name in ("frequency", "phase"))...,
    )
    sliderobservables = [s.value for s in sg.sliders]
    data = lift(sliderobservables...) do idxs...
        t = collect((0:99)/100)
        y = sin.(t .* (4*idxs[1] + 2π) .+ idxs[2])
        sleep(rand()*1.0)
        return StructArray{Point2{Float64}}((t, y))
    end

    lines!(ax, data)
    ylims!(ax, (-1, 1))
    fig
end

slider_test()
```

SimonDanisch:
Fixed in #3113

Issue #2977, opened 2023-05-23
[Roadmap] WGLMakie


SimonDanisch:
# Serverless visualization

JSServe + WGLMakie already work pretty well with a running Julia server, but this doesn't scale very well to serve interactive demos for thousands of people (e.g. blogpost/articles).  For that we need to Move more work into the browser on the clients computer and make it easier to remove any communication with Julia.

## Move more to JS

* Rewrite 2D camera & axis in JS [and others](https://github.com/MakieOrg/Makie.jl/issues/2326)
* Make it easier to overload event callbacks and recipes in the backend (ties into [new api for plotobjects](https://github.com/MakieOrg/Makie.jl/issues/725))
* Move Tyler to JS (needs better event + plotobject api)
* Datashader in JS + WebGL
* More of the drawing & updating API in JS
* Datainspector in JS

## WASM integration

* more of a future pathway, but [toy examples](https://tshort.github.io/Lorenz-WebAssembly-Model.jl/?p1=11) are already possible
* make sure JSServe is ready for it, by e.g. enabling:

```julia

statically_compilable_julia_func(args...) = ...

js"""
js_args = [...] // from julia or directly js
const result = $(statically_compilable_julia_func)(args)
update_plot(plot, result)
"""
```

Or another possible API:

```julia
using WGLMakie, JSServe,
obs = WASMObservable(data)
data = lift(statically_compilable_julia_func, obs)
plot(data)
```

# Compsoable GUI elements

* playbutton for anything playable
* drawing over every plot
* drag & drop
* More predefined UI elements for dashboards (especially layouts)
* complex recipes including sliders and interactions and composable UI elements

# Declarative API for Pluto & AoG and easier animation

Right now one animates in Makie like this:

```julia
arg1 = Observable(...)
arg2 = Observable(...)
attribute1 = Observable(...)

# only call plot one time!
plot(arg1, arg2; attribute1=attribute1)
```

This is highly performant, but is difficult for complex dashboards where data can change the plot types or attribute passed (e.g. like in AoG).
This kind of mutating API using observables and just one plot call is also working against Plutos way of animating, which executes the complete cell defining the plot on every change.
So we need a declarative API like this:

```julia
arg1 = Observable(...)
arg2 = Observable(...)
attribute1 = Observable(...)
plot_spec = lift(arg1, arg2, attribute1) do arg1, arg2, attribute1
    if attribute1 == something
        return Scatter(arg1, arg2, marker=attribute1)
    else
[...]

PR #3062, opened 2023-07-13
Rework WGLMakie lines

SimonDanisch:
- [x] More efficient implementation of lines purely in JS
- [x] implement updating
- [x] fix https://github.com/MakieOrg/Makie.jl/issues/3050
- [ ] line patterns
- [x] 3d lines
- [x] colormaps + clean up implementation

This should also be a template for how we want to generate the plot objects in JS in the future, which should lead to a large TTFP improvement. 
If I find the time to do the same treatment for all plot types in WGLMakie, it may just get the fastest TTFP of all backends, since it will basically just create dictionaries from plots.

This should also bring us closer to https://github.com/MakieOrg/Makie.jl/issues/2977, with a more comprehensive plotting API directly in JS.


staticfloat:
@SimonDanisch I just tried out the current state of the branch (b9e2bad9d84a10aa82d2b4fde9ccf22cb910745d) and it seems to have completely solved #3050!  I can no longer trigger that kind of issue.

WGLMakie/src/wglmakie.js
```diff
@@ -129,6 +128,9 @@ function threejs_module(canvas, comm, width, height, resize_to_body) {
         // we return nothing which will be handled by caller
         return;
     }
+    const pixel_ratio = window.devicePixelRatio || 1.0;
```
staticfloat, line 131: Because you renamed this to `pixel_ratio` from `pixelRatio` there are a few references to `pixelRatio` that need to be updated in this file, see for instance the definition of `resize_callback`

PR #3068, opened 2023-07-14
Add keyword docs to save

felixcremer:
This adds short documentations about the keywords for the `save` function. Feel free to change the wording especially for the update keyword I am not so sure what it fully entails. 

Could the possibility to use update=false to keep the interactive changes of a figure added somewhere to the documentation apart from the changes to the docstring? 
I skimmed the docs but didn't find a good place to add this.


SimonDanisch:
Thanks! :)

PR #3073, opened 2023-07-18
Try to fix marker size

SimonDanisch:
https://github.com/MakieOrg/Makie.jl/pull/2990  introduced a problem in the exact marker size:

marker-alignment

### DIFF
--- a/ReferenceTests/src/tests/primitives.jl
+++ b/ReferenceTests/src/tests/primitives.jl
@@ -248,6 +248,17 @@ end
     for (i, marker) in enumerate(markers)
         scatter!(Point2f.(1:5, i), marker = marker, markersize = range(10, 30, length = 5), color = :black)
         scatter!(Point2f.(1:5, i), markersize = 4, color = :white)
+
+        # # Debug - show bbox outline
+        # if !(marker isa Char)
+        #     scene = Makie.get_scene(ax)
+        #     bb = Makie.bbox(Makie.DEFAULT_MARKER_MAP[marker])
+        #     w, h = widths(bb)
+        #     ox, oy = origin(bb)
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Hmpf, this _does_ fix the above issue, but circle polys are now cropped:
image

image

ffreyer:
Yea that's kind of what I expected. You probably want even dividers for of the pixelsize for the aspect ratio.

ffreyer:
I tried counteracting the distortion from rounding to full pixels by scaling `widths(bb)` (origin is fine regardless) in the offset and scale calculations but didn't really get anywhere...
From just experimenting with factors directly `Vec2f(1.01838, 1)` seems to be about where it needs to be for hex markers. That doesn't match up well with anything I can get out of `ceil.(Int, x * aspect) ./ (x * aspect)` where x is some pixel size between 10 and 256*64.

ffreyer:
I don't understand what I did yet but I think I got it working...
Screenshot from 2023-07-19 14-29-40
Bezier markers with bboxes:
Screenshot from 2023-07-19 14-28-14

ffreyer:
I reformulated/simplified the math here and added a comment that explains what's happening. I think this should be good to merge now. (Checked the two tests from above and a triangle with large offset)

ffreyer:
I think the reason for the scaling factor being so large (0.018 rather than 1/256 = 0.0039 or even 1/64 = 0.015625) is that the integer rounding errors compound during downsampling. So I made another change that effectively adds padding to avoid this. I don't really see things moving though, it just removes a bit of blur on hline (compared to last commit) or vline (compared to refimgs).

### Refimg
BezierPath markers

### No pixelsize matching (before commit)
BazierPath markers new

### With pixelsize matching (after commit)
BazierPath markers new fractional size

ffreyer:
Oh yea, latest commit will need a texture atlas update

ffreyer (approved):

PR #3077, opened 2023-07-20
Fix invalid range logscale

SimonDanisch:
Continuation of https://github.com/MakieOrg/Makie.jl/pull/1671


SimonDanisch:
Ah, this still fails (should fail), since it changes the limits for plots with only one point, e.g. `text(Point2f(0,0); text="hi")`

SimonDanisch:
Ok, i think this is the wrong approach! 
Plot limits from e.g. `text(Point2f(0,0); text="hi")` already get expanded...
The question is, should we error in the limit setting code for `dmin==dmax`, or just also expand it, like its done for a plot with only one point?
Will close this PR for now, since a fix will look pretty different!

PR #3079, opened 2023-07-20
prepare 0.19.7 release

SimonDanisch:
- [x] update project tomls
- [x] update NEWS.md
- [x] merge some more PRs!?


SimonDanisch:
Damn, didn't realize that we now have something of a circular dependency via DelaunayTriangulation depending on MakieCore and Makie depending on DelaunayTriangulation, making it hard to release a breaking MakieCore version....
I guess we can act like it's not breaking, since annyone using MakieCore, will need to install a matching Makie version, that has `=MakieCoreVersion`...

DanielVandH:
> Damn, didn't realize that we now have something of a circular dependency via DelaunayTriangulation depending on MakieCore and Makie depending on DelaunayTriangulation, making it hard to release a breaking MakieCore version....
I guess we can act like it's not breaking, since annyone using MakieCore, will need to install a matching Makie version, that has =MakieCoreVersion...

@SimonDanisch: Is this a significant issue? Currently, [MakieCore is a package extension inside DelaunayTriangulation](https://github.com/DanielVandH/DelaunayTriangulation.jl/blob/main/ext/DelaunayTriangulationMakieCoreExt.jl), though that doesn't do anything <1.9. Is there a way I can change some of the code up inside DelaunayTriangulation to help you out here?

SimonDanisch:
I think now that makie depends on delaunaytriangulation, we could just move the extension code to Makie?

DanielVandH:
That would be a good idea. Once I have some spare time, I will look into doing that.

PR #2220, opened 2022-08-15
breaking-release combined

SimonDanisch:
Combines the changes from #1918 #2056 and #2046 to put a stop to rebase hell.
Will replace #1918, but doesn't yet as long as tests aren't passing.

PR #3082, opened 2023-07-22
early attribute conversion + no proxy scene anymore

### LINKED ISSUE #2820: Empty vectors of 2d Polygon plot 3d
```julia
poly = Polygon([Point2(1, 2), Point2(3, 4), Point2(4, 5), Point2(1, 2)])
# 2d
Makie.poly([poly])
# 3d
Makie.poly(typeof(poly)[])
```

SimonDanisch:
This extracts useful bits from #2220, which aren't as breaking as #2220:

- [x] convert plot attributes as early as possible, to not pass namedtuples through lots of functions, therefore reducing compilation costs
- [x] uses `plot_preferred_axis` to figure out the default axis, instead of using `boundingbox(plot!(Scene(), args...))`, which is expensive and kind of hacky

closes #2820


src/basic_recipes/axis.jl
```diff
@@ -317,8 +317,8 @@ function draw_axis3d(textbuffer, linebuffer, scale, limits, ranges_labels, fonts
     return
 end
 
-function plot!(scene::SceneLike, ::Type{<: Axis3D}, attributes::Attributes, args...)
-    axis = Axis3D(scene, attributes, args)
+function plot!(axis::Axis3D)
+    scene = parent_scene(axis)
```
ffreyer, line 321: `parent_scene` gives me the expectation that the axis is contained in some scene that is returned/used here. Maybe it would be better to use `get_scene` to avoid that?

SimonDanisch:
Merged in #3113

PR #2900, opened 2023-04-26
Clean up colormap handling and implement colorscale + alpha kw

SimonDanisch:
Continuation of #colorscale https://github.com/MakieOrg/Makie.jl/pull/2493

Issue #2835, opened 2023-04-02
Colorrange not forwarded in arrows recipe


brendanwallison:
See MWE below, where setting colorrange has no effect on the arrow colors

```
using CairoMakie

f = Makie.Figure(resolution = (1600, 800))
ax = Makie.Axis(f[1, 1])
colorrange=(0, 100)
ar = Makie.arrows!(ax, 1:0.25:2, 1:0.25:2, rand(5,5), rand(5,5), arrowsize = 10, lengthscale = 0.1, color=rand(25), colormap=:viridis, colorrange=colorrange)
Makie.Colorbar(f[1,2], ar)
f
```

ffreyer:
Fixed

Issue #2655, opened 2023-02-06
arrows() doesn't have a colorrange attribute that controls linecolor/arrowcolor in CairoMakie


Zou-Bo:
As the title says, I cannot control the color range when drawing an arrows plot.

I try the colorrange=(a,b) attribute, but no matter what values I put for a and b, the color of the arrows doesn't change.

Further, if I make a colorbar of it, the range of colorbar corresponds with the colorange=(a,b) attribute, and doesn't correspond to the actual colors of the arrows.

Issue #1778, opened 2022-03-22
`Colorbar` fails with `getindex` on a `poly` but not a `mesh`


BambOoxX:
While testing the examples for [`mesh`](https://makie.juliaplots.org/stable/examples/plotting_functions/mesh/)  I stumbled upon a weird error when replacing `mesh` with `poly`.

In fact, evaluating this example works fine 
```
using FileIO
using GLMakie

brain = load(assetpath("brain.stl"))

f, ax, m = mesh(
    brain,
    color = [tri[1][2] for tri in brain for i in 1:3],
    colormap = Reverse(:Spectral),
    figure = (resolution = (1000, 1000),)
)

Colorbar(f[1,2],m)
```
and expectedly returns the brain image with a colorbar on its right
![image](https://user-images.githubusercontent.com/42067365/159449043-e00804a2-239b-4efa-bac5-a11477d2a11d.png)

However, just replacing the `mesh` with a `poly` fails with 
```
ERROR: MethodError: no method matching getindex(::MakieCore.Automatic, ::Int64)
Stacktrace:
  [1] scaled_steps(steps::Vector{Float64}, scale::Function, lims::MakieCore.Automatic)
    @ Makie.MakieLayout D:\bambooxx\.julia\packages\Makie\f11Ut\src\makielayout\layoutables\colorbar.jl:383
  [2] (::Makie.MakieLayout.var"#477#507")(bbox::GeometryBasics.HyperRectangle{2, Float32}, v::Bool, steps::Vector{Float64}, gradient::PlotUtils.ContinuousColorGradient, scale::Function, lims::MakieCore.Automatic)
    @ Makie.MakieLayout D:\bambooxx\.julia\packages\Makie\f11Ut\src\makielayout\layoutables\colorbar.jl:174
  [3] lift(::Function, ::Observable{GeometryBasics.HyperRectangle{2, Float32}}, ::Observable{Any}, ::Vararg{Any}; kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Makie D:\bambooxx\.julia\packages\Makie\f11Ut\src\interaction\observables.jl:30
  [4] lift(::Function, ::Observable{GeometryBasics.HyperRectangle{2, Float32}}, ::Observable{Any}, ::Observable{Vector{Float64}}, ::Vararg{Any})
[...]

PR #3090, opened 2023-07-24
improve colorbar errors for plots that don't use a colormap

SimonDanisch:
* refactors Colorbar to use image + heatmap instead of the heavier poly + image. If it wasn't for CairoMakie, we could just use heatmap alone. Once CairoMakie supports irregular gridded heatmaps with interpolation, we can get rid of the image call
* Uses now `ColorMap` (TODO, rename to `ColorMapping`), internally in Colorbar, to have a more consistent interface
* Forwards `cgrad` correctly to Colorbar now, which wasn't the case anymore after #2900 
* Introduces `extract_colormap` (TODO, also rename to `extract_colormapping`), which Colorbar now uses to extract the colormapping from any plot type. Has a fallback for any recipe, which recursively searches the child plots. One can also directly overload it for a recipe, where this isn't a good default. Will error if multiple colormappings are found in a recipe.
* throws better errors
* introduces experimental `colormap=Categorical(colormap)`, which does a true categorical colormap:
```julia
f, ax, pl = barplot(1:3, color=1:3, colormap=Categorical(:viridis))
Colorbar(f[1, 2], pl)
f
```
image
We may remove this, since its not very elegant to pass `categorical=true` to Colorbar and cgrad, but `cgrad` doesn't define a true categorical colormap, and we don't have any alternative to define a truly categorical colormap.
But, since I already implemented the support for it, I wanted a way to expose it. I think the most elegant way would be to do `barplot(...; color=Categorical(1:3))` since that's where the categories actually are, and where we also need to know that things are categorical. To not balloon this PR even more, I refrained from implementing this for now.

With the new `extract_colormapping` we also get colormaps for recipes that didn't have colormap support before:
image

Closes https://github.com/MakieOrg/Makie.jl/issues/3188, https://github.com/MakieOrg/Makie.jl/issues/3119, https://github.com/MakieOrg/Makie.jl/issues/2835, https://github.com/MakieOrg/Makie.jl/issues/2655, https://github.com/MakieOrg/Makie.jl/issues/1778


src/makielayout/blocks/colorbar.jl
```diff
@@ -27,11 +27,18 @@ end
...
     if haskey(plot, :calculated_colors) && plot.calculated_colors[] isa ColorMap
         cmap = plot.calculated_colors[]
         scale = cmap.scale
     else
+        # Atomic plots should always have calculated_colors set, if they actually use a colormap
+        if is_atomic_plot(plot)
+            error("Plot `$(plotfunc(plot))` has no values that are mapped to colors, so it cannot be used to create a colorbar.")
+        end
+        # So this is for recipes
+        if !all(x-> haskey(plot, x), (:colormap, :colorrange, :highclip, :lowclip))
```
jkrumbiegel, line 39: Would be better if it said which ones were there and which ones werent.

src/makielayout/blocks/colorbar.jl
```diff
@@ -27,11 +27,18 @@ end
...
 function Colorbar(fig_or_scene, plot::AbstractPlot; kwargs...)
     colorbar_check((:colormap, :limits, :highclip, :lowclip), keys(kwargs))
-
     if haskey(plot, :calculated_colors) && plot.calculated_colors[] isa ColorMap
         cmap = plot.calculated_colors[]
         scale = cmap.scale
     else
+        # Atomic plots should always have calculated_colors set, if they actually use a colormap
+        if is_atomic_plot(plot)
+            error("Plot `$(plotfunc(plot))` has no values that are mapped to colors, so it cannot be used to create a colorbar.")
```
jkrumbiegel, line 36: This is worded a bit confusingly maybe? "has no values that are mapped to colors" might not be clear to a casual user.

src/makielayout/blocks/colorbar.jl
```diff
@@ -27,11 +27,28 @@ end
...
+        # So this is for recipes
+        if !all(x-> haskey(plot, x), (:colormap, :colorrange, :highclip, :lowclip))
+            attributes = filter(x-> haskey(plot, x), [:colormap, :colorrange, :highclip, :lowclip])
+            error("Plot $(func) doesn't have all colormap attributes. Attributes needed: colormap, colorrange, highclip, lowclip. Found: $(attributes)")
+        end
+        if to_value(plot.colorrange) isa Automatic || (to_value(get(plot, :color, nothing)) isa Union{AbstractVector{<: Colorant}, Colorant})
+            error("""
+            Plot $(func) doesn't have a color attribute that uses a colormap to map to colors, so it cannot be used to create a colorbar.
+            Please create the colorbar manually e.g. via `Colorbar(f[1, 2], colorrange=the_range, colormap=the_colormap)`.
+            This could also mean, that you're using a recipe, which itself doesn't calculate the colorrange, and leaves it to its children.
```
jkrumbiegel, line 48: Did this not work before (the barplot example)? Would seem to be a big regression if so.

SimonDanisch: It was indeed able to create one, but it created the _wrong_ Colorbar before, since it didn't connect the colorrange, which only happens for the meshplot inside.

jkrumbiegel: hm one way or another I think it should work, it's very weird having to access the nested plots, no? maybe an internal function specifying the child plot with the color that gets called by the `Colorbar` constructor?

SimonDanisch: I mean we could search for subplots that do have a complete colormap... Now with `calculated_colors`, it should be pretty clear, that such a plot _does_ have numbers that are mapped to colors... The question is, how do we deal with multiple subplots? 
We could check if they use the same `calculated_colors` and if not error...and if it's just one it should quite clearly be the plot to be used for Colorbar

SimonDanisch: Maybe a combination, like with data_limits? Some plots need to overload `data_limits` but there is a fallback for all recipes, that usually does the correct thing by calculating `data_limits` of the children

SimonDanisch:
I'm going to merge this since I need to update the reference images in other PRs and need to base at least the datashader PR against this to make the colorbar integration easier.
@t-bltg, would be still nice to review & test this - we can just fix whatever bugs come up in a new PR!

t-bltg:
Maybe we can add the example of https://github.com/MakieOrg/Makie.jl/pull/2493 to the test suite (comparison against Python so we know it's right), if we didn't have it already in the ref imgs (on my phone ;)).

t-bltg:
@SimonDanisch I've ran several examples, and I couldn't identify a regression related to this PR, so we can probably remove the commented test.

Issue #3089, opened 2023-07-24
How to define custom function for axis scaling?


johnomotani:
Thanks to @jkrumbiegel's hint, here is a correct example of how to define a custom function for an axis scale
```
module ScaleTest

export mylog

import Makie

function mylog(x)
    return log10(x)
end

function Makie.inverse_transform(::typeof(mylog))
    return x -> 10.0^x
end

Makie.defaultlimits(::typeof(mylog)) = (1.0, 1000.0)

Makie.defined_interval(::typeof(mylog)) = Makie.OpenInterval(0.0, Inf)

end

using CairoMakie
using .ScaleTest

x = collect(1.0:10.0)
y = rand(10)

fig, ax = lines(x,y)
ax.yscale = mylog

save("foo.pdf", fig)
```

=========================================================
Original post follows:
- [x] are you running newest version (version from docs) ? Yes, v0.19.7
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) Yes
- [x] What platform + GPU are you on? Linux Mint 21.1, Julia-1.8.5

In the docs (https://docs.makie.org/stable/examples/blocks/axis/#yscale) it says
> Can be any invertible function, some predefined options are identity, log, log2, log10, sqrt, logit, Makie.pseudolog10 and Makie.Symlog10. To use a custom function, you have to define appropriate methods for Makie.inverse_transform, Makie.defaultlimits and Makie.defined_interval.

How can those 'appropriate methods' be defined? My best guess is something like the following
```
module ScaleTest

export mylog

import Makie: inverse_transform, defaultlimits, defined_interval

function mylog(x)
    return log10(x)
end

function inverse_transform(::typeof(mylog))
    return x -> 10.0^x
end

defaultlimits(::typeof(mylog)) = (1.0, 1000.0)

defined_interval(::typeof(mylog)) = (0.0, Inf)

end

using CairoMakie
using .ScaleTest

x = collect(1.0:10.0)
y = rand(10)

fig, ax = lines(x,y)
ax.yscale = mylog

save("foo.pdf", fig)
```
But if I run that code, I get an error
```
ERROR: LoadError: Found invalid y-limits (0.013601727f0, 0.87287176f0) for scale mylog which is defined on the interval (0.0, Inf)
```

It would be nice if there was some explanation in the docs both of how to define those functions and of what they have to do.

[...]

Issue #3088, opened 2023-07-24
custom `colorscale` support


t-bltg:
https://github.com/MakieOrg/Makie.jl/pull/2900 / https://github.com/MakieOrg/Makie.jl/pull/2493 which landed in `0.19.7` introduced some subtle bugs, ~~where it seems that the colorscale is applied multiple times in the pipeline (resulting in multiple values overlapping on the colorbar)~~.

EDIT: this is related to the inverse transform of custom `colorscale`s.

I still have to come up with an mwe, because my example is complicated.

![overlap](https://github.com/MakieOrg/Makie.jl/assets/13423344/c4e95345-b699-45b0-be37-a24e5d3a9d8d)

Currently observed on `CairoMakie` (other backends un-tested).

- [x] are you running newest version (version from docs) ?

t-bltg:
Here is a mwe, occurs when using a custom colorscale:
```julia
using CairoMakie

main() = begin
  xs = range(0, 2π, length=100)
  ys = range(0, 2π, length=100)
  zs = [sin(x*y) for x ∈ xs, y ∈ ys]

  fig = Figure()
  ax = Axis(
    fig[1, 1];
    aspect=DataAspect(), xlabel="x", ylabel="y", ylabelrotation=0,
    xgridvisible=false, ygridvisible=false
  )
  hm = heatmap!(
    ax, xs, ys, zs; colorscale=x -> log10(1 + x)
  )
  Colorbar(fig[2, 1], hm; vertical=false, flipaxis=false)

  display(fig)
  save("colorscale.png", fig; px_per_unit=2)
  return
end

main()
```

SimonDanisch:
I think we're re-using parts of the Axis infrastructure, so maybe this helps https://github.com/MakieOrg/Makie.jl/issues/3089?

t-bltg:
Thanks for the reference.

I think the solution implemented in https://github.com/MakieOrg/Makie.jl/pull/3095 is more generic (works for axis scale, but also for colorscale) ?

PR #3095, opened 2023-07-25
add and export `ReversibleScale` type - simplify `LogFunctions`

t-bltg:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/3088, continuation of https://github.com/MakieOrg/Makie.jl/pull/2493 / https://github.com/MakieOrg/Makie.jl/pull/2900.

Add a new `ReversibleScale` type for custom scale functions (adds genericity in damping functions):

<details>
<summary>old examples</summary>

Example:
```julia
using CairoMakie

x = 10.0.^(1:0.1:4);
y = 1.0:0.1:5.0;
z = broadcast((x, y) -> x, x, y');

# forward and reverse scale function, with optional pseudo-log ticks
scale = ReversibleScale(x -> log10(x + 2), x -> exp10(x) - 2; logbase = "10");
axis = (; xscale = scale, xminorticksvisible = true, xminorticks = IntervalsBetween(9));

fig, ax, hm = heatmap(x, y, z; colorscale = scale, axis);
Colorbar(fig[1, 2], hm; minorticksvisible = true, minorticks = IntervalsBetween(9));

display(fig)
save("colorscale.png", fig)
```
colorscale

</details>

Fix wrong algorithm in `scaled_steps`, which lead to `log` `DomainError`s on the following example (on `0.19.7`), added a new reference image:
```julia
using CairoMakie

x = 10.0.^(1:0.1:4);
y = 1.0:0.1:5.0;
z = broadcast((x, y) -> x, x, y');

axis = (; xscale = log10, xminorticksvisible = true, xminorticks = IntervalsBetween(9));
colorscale = ReversibleScale(x -> log10(x + 2), x -> exp10(x) - 2)

fig, ax, hm = heatmap(x, y, z; colorscale, axis);
Colorbar(fig[1, 2], hm; minorticksvisible = true, minorticks = IntervalsBetween(9));

display(fig)
save("colorscale-wrong.png", fig)
```

Without the `ReversibleScale` type, one would fail to determine the inverse transform function, hence leading to unoptimal colorbar ticks (root cause of the ticks overlap observed in https://github.com/MakieOrg/Makie.jl/issues/3088):

colorscale-wrong

Simplify log functions declaration.

Fix https://github.com/MakieOrg/Makie.jl/issues/3089:
```julia
using CairoMakie

x = collect(1.0:10.0);
y = rand(10);

scale = ReversibleScale(log10; limits = (1, 10), interval = (0, Inf));
fig, ax, p = lines(x, y; axis = (; yscale = scale));

display(fig)
save("foo.png", fig)
```
foo

## Type of change

## Checklist


src/makielayout/lineaxis.jl
```diff
@@ -581,18 +581,20 @@ function get_ticks(tickfunction::Function, any_scale, formatter, vmin, vmax)
     return tickvalues, ticklabels
 end
 
+_logbase(s::ReversibleScale) = s.logbase
 _logbase(::typeof(log10)) = "10"
 _logbase(::typeof(log2)) = "2"
 _logbase(::typeof(log)) = "e"
+_logbase(::Any) = ""
```
jkrumbiegel, line 588: this should probably be `nothing`, `get_ticks` will be specialized anyway

src/makielayout/blocks/colorbar.jl
```diff
@@ -68,7 +68,12 @@ function Colorbar(fig_or_scene, contourf::Union{Contourf, Tricontourf}; kwargs..
 
 end
 
-colorbar_range(start, stop, length, _) = LinRange(start, stop, length)  # noop
+colorbar_range(start, stop, length, _::typeof(identity)) = LinRange(start, stop, length)
+function colorbar_range(start, stop, length, _)
+    @warn "cannot determine inverse_transform"
```
jkrumbiegel, line 73: Not sure what this does, but I'm almost never fond of warnings. I'd rather have a clear error if we hit edge cases

src/makielayout/lineaxis.jl
```diff
@@ -581,18 +581,20 @@ function get_ticks(tickfunction::Function, any_scale, formatter, vmin, vmax)
...
 
-
-function get_ticks(::Automatic, scale::Union{typeof(log10), typeof(log2), typeof(log)},
-        any_formatter, vmin, vmax)
-    get_ticks(LogTicks(WilkinsonTicks(5, k_min = 3)), scale, any_formatter, vmin, vmax)
+function get_ticks(::Automatic, scale::Union{LogFunctions,ReversibleScale}, any_formatter, vmin, vmax)
+    wt = WilkinsonTicks(5, k_min = 3)
+    ticks = isempty(_logbase(scale)) ? wt : LogTicks(wt)
```
jkrumbiegel, line 592: are you using `_logbase` for nothing else than this `isempty` check? I don't think we really need it, log functions should just dispatch to LogTicks and all others don't

jkrumbiegel (changes requested):

src/makielayout/lineaxis.jl
t-bltg: Right, this was in order to simplify the `ReversibleScale` `logbase` type which was a `String` and now has to be a `Union{Nothing,String}`.

src/makielayout/blocks/colorbar.jl
t-bltg: This is when an automatic scaling has been applied, but we cannot determine the inverse transform for the colorbar ticks, hence falling back to identity, which leads to un-optimal ticks. I got tricked by this, and I think a warning is thus useful. Maybe it should advise to use `Makie.ReversibleScale` instead ...

src/makielayout/lineaxis.jl
t-bltg: No, I'm not. I will change it to a `nothing` check instead then.

SimonDanisch:
Alright, CI is green! @jkrumbiegel do we want to merge this now? :)

src/makielayout/blocks/colorbar.jl
SimonDanisch: Yeah, I think this is a good place for a warning, _but_ only with a good advice on how to resolve the warning ;)

t-bltg: So, is the `maybe use Makie.ReverseibleScale instead ?` hint added in https://github.com/MakieOrg/Makie.jl/pull/3095/files#diff-b49aa74de270f83f7d96f661d23e3e1044de710184415a64e5db70890a671841R95 sufficient ?

SimonDanisch: Maybe:

`@warn "Cannot determine inverse transform: you can use colorscale=Makie.ReversibleScale($(_), reverse($(_))) instead"`
I love warnings that give context + and basically just need a copy and paste to be fixed.
Also, shouldn't be this somewhere else in the stack?
So that we get a similar warning for ` inverse_transform(scale)` ? Maybe to complicated for now, so just musing ^^
I guess one could do:
```julia
function colorbar_range(start, stop, length, scale)
    rev = inverse_transform(scale)
    if isnothing(rev)
          @warn "Cannot determine inverse transform: you can use `colorscale=Makie.ReversibleScale($(_), reverse($(_)))` instead. Falling back to identity, which leads to sub optimal ticks!"`
          rev = identity
    end
    rev.(range(start, stop; length))
end
```
This would also keep out the union type from `colorbar_range` which _may be_ desirable for compilation (likely not a big issue).

t-bltg: Oh yeah, inverting the logic here also allows us to get rid on the `REVERSIBLE_SCALES` global `Union`, better !

I've implemented the changes in the latest commit.

jkrumbiegel:
I have to take another look tomorrow, I just saw that the warning is still in there and I have to double check the logbase thing

jkrumbiegel:
With this I get an empty colorbar (plus the warning): 
```julia
f = Figure()
Colorbar(f[1, 1], limits = (1, 100), scale = (x -> log10(x)))
Colorbar(f[1, 2], limits = (1, 100), scale = log10)
f
```
image

image

I think it should probably not be possible to use a scale function without the correct inverse defined. I don't like this warning and I'm not sure the current design is what we really want.

Let's take a step back maybe and try to motivate again what we want, and why the current state of things is not enough. Is it the added convenience of the `ReversibleScale` struct over making a function plus the correct overloads? I've only seen the example with a log + offset function, are there other typical ones we don't cover?
If we want generality, why do we need the `logbase` field? This specializes the implementation on log-like scales, while we might rather want to have the ability to pass a tick formatter hint object, which might be used for other cases like sqrt or whatever there is out there.

I don't really like that it's so easy to pass a wrong log base to the scale, I'd rather want to have that determined either automatically (log2, log, log10) or done with a struct like `LogTransform(some_value)` where we automatically use some value, and the backwards transform is also automatic. We could even give `LogTransform` an offset value if that's commonly required.

You can also easily pass a mismatched pair of transforms, while that's also a "user error" thing I'm not sure if it's good.

```
f = Figure()
Colorbar(f[1, 1], limits = (1, 100), scale = Makie.ReversibleScale(sqrt, x->x^2))
Colorbar(f[1, 2], limits = (1, 100), scale = Makie.ReversibleScale(sqrt, exp10))
f
```

image

These things are hard to see in a final figure, so we should be careful and take things slow to get the implementation right.

t-bltg:
> I think it should probably not be possible to use a scale function without the correct inverse defined. 

I don't see neither in which case one would use this, so let's hard error instead.

> You can also easily pass a mismatched pair of transforms, while that's also a "user error" thing I'm not sure if it's good.

Since the given limits define a valid range, It's fairly easy to add a check cover this invalid usage: done in the latest commits.

> If we want generality, why do we need the logbase field?

I agree that the `logbase` thing is a bit hacky, but I don't find an easy way to reproduce the first figure in https://github.com/MakieOrg/Makie.jl/pull/3095#issue-1820811797 without it.

> where we automatically use some value, and the backwards transform is also automatic

Implemented and tested.

jkrumbiegel:
This looks better, I like the new tests. I still don't like `logbase` :) In a way you're just using that to have a shortcut to log-style formatting, by `Union`ing `ReversibleScale` into the signature. Instead, I would give `ReversibleScale` its own entrypoint and use a different property called something along the lines of `style` or `format` or so, which can then be used to dispatch on in `get_ticks`.

Actually, now I think the previous logbase logic you added is buggy. If I have scale `log10(x + 1)` and I compute linear ticks in that transformed space with, then transform them back via `exp10(x) - 1` I cannot write `10 ^ scaled_value`, I need to write `(10 ^ scaled_value) - 1` no? You can't just stick the `10` in there.

t-bltg:
Right, I couldn't come up with a clean solution, so let's drop using `LogTicks` for now.

`Makie.Symlog10` and `Makie.pseudolog10` is now implemented using `Makie.ReversibleScale` ([`Symlog10` would previously error](https://github.com/MakieOrg/Makie.jl/blob/b9f25b912f3a03dd63dc8074236f12e34e86466b/src/colorsampler.jl#L182)):
```julia
ERROR: MethodError: Cannot `convert` an object of type Makie.Symlog10 to an object of type Function
```

SimonDanisch:
@jkrumbiegel is this now ready to merge? :)

SimonDanisch:
I'd say yes to both!

t-bltg:
Also, I don't have a strong opinion on the `ReversibleScale` name, so maybe there are better naming alternatives ?

jkrumbiegel:
Ah yeah similar to the previous method, this one also doesn't work for PRs because those users are not allowed to access the token needed. I'm not sure what to do to be honest, we can't comment, we can't status change, but we need to know if refimages are missing

jkrumbiegel:
The behavior changed, now it sets a "success" status if no missing reference images were found, and that doesn't work here. We could remove that but then it still wouldn't work in cases with refimages.

t-bltg:
I like to see these green statuses in the morning, awesome @jkrumbiegel !

green

jkrumbiegel:
This run hopefully gives the status message. The last one was green but was missing the status (because getting the right commit SHA is not trivial in github actions..), but now I made changes on master that should resolve it correctly. We should see a "No new reference images" status here once the GLMakie CI runs finish.

jkrumbiegel:
It worked 🎉

docs/reference/plots/heatmap.md
```diff
@@ -116,3 +116,25 @@ Colorbar(fig[:, end+1], colorrange = joint_limits)  # equivalent
...
+```julia
+using CairoMakie
+CairoMakie.activate!() # hide
+
+x = 10.0.^(1:0.1:4)
+y = 1.0:0.1:5.0
+z = broadcast((x, y) -> x, x, y')
+
+scale = ReversibleScale(log10)
+fig, ax, hm = heatmap(x, y, z; colorscale = scale, axis = (; xscale = scale))
```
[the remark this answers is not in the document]
jkrumbiegel: Yeah but then I don't think the example is good, because why would you wrap `log10` if you can pass it on its own (like shown in other places in the docs as well). We don't want to teach the complicated version if it's not better, no?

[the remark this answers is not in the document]
t-bltg: Alright, I've tweaked the example.

jkrumbiegel (approved):

jkrumbiegel:
Ok let's get this one in

SimonDanisch:
Thank you all :)

Issue #748, opened 2020-11-03
Heatmap / Image / Surface API clarification and nonlinear grids


jkrumbiegel:
We had multiple issues about nonlinear grids for heatmaps, but also about the bin centering of heatmaps and images (basically that pixel [1, 1] is not centered on point [1, 1]).

These discussions have mostly looked at implementing things in the backends, but I think it's an AbstractPlotting abstraction issue. Right now, there is a `SurfaceLike` trait which heatmap, image and surface use to convert their arguments. I think that this is incorrect, because they are different, even if only in a subtle way.

The difference is that a surface colors the vertices of a grid, while image / heatmap colors the cells of a grid. I therefore suggest to remove `SurfaceLike` and instead add `VertexSurfaceLike` and `CellSurfaceLike` or something else to that effect. Maybe there are better words for this.

Then the question is how the coordinates of these grids should be defined. Let's say our color data are always in the form of a matrix, then there are multiple different ways of defining the coordinates of vertices / cells.

Here's what I would suggest: 

| x & y coordinates example | type | interpretation `VertexSurfaceLike` | interpretation `CellSurfaceLike` |
| :-- | :-- | :-- | :-- |
| `1.0..10.0` | ClosedInterval | n linspaced vertices from 1 to 10 | n equal-sized cells with centers from 1 to 10 |
| 1:10 | Range | 10 linspaced vertices from 1 to 10 (must match n), basically like vector | 10 equal-sized cells with centers from 1 to 10 (must match n), basically like vector |
| [1, 2, 3] | Vector| 3 vertices (should be sorted?), can be nonlinear, must match n | 3 cell centers (should be sorted?), cells always extend halfway to neighbor (mirror outward at edges), must match n |

Matrices would then just be extensions of the vector case.

One thing that this doesn't cover is a plot where the coordinates given are vertices, but the colors are for the cells. One could dispatch to that method by checking if the length of heatmap coordinates are n+1, but that might be a bit brittle?

jkrumbiegel:
cf https://github.com/JuliaPlots/CairoMakie.jl/pull/111, https://github.com/JuliaPlots/GLMakie.jl/pull/137, https://github.com/JuliaPlots/Makie.jl/issues/661, https://github.com/JuliaPlots/Makie.jl/issues/674

jkrumbiegel:
2D contour plots would be `VertexSurfaceLike` by this logic (important for the isoband PR that I want to merge soon as well)

[...]

PR #3102, opened 2023-07-28
Add triplot and voronoiplot from DelaunayTriangulation.jl

DanielVandH:
# Description

This PR adds `triplot` and `voronoiplot` from DelaunayTriangulation.jl ([see here for the original code](https://github.com/DanielVandH/DelaunayTriangulation.jl/blob/main/ext/DelaunayTriangulationMakieCoreExt.jl)) to fix the circular dependency issue mentioned from @SimonDanisch [here](https://github.com/MakieOrg/Makie.jl/pull/3079#issuecomment-1644332878). Just for reference, here's an example of what it looks like:

```julia
fig = Figure()
ax1 = Axis(fig[1, 1],width=400,height=400)
ax2 = Axis(fig[1, 2],width=400,height=400)
pts = tuple.(rand(100), rand(100))
tri = triangulate(pts)
refine!(tri; max_area = 1e-3)
vorn = voronoi(tri, true)
triplot!(ax1, tri)
voronoiplot!(ax2, vorn)
resize_to_layout!(fig)
```

image

Once this is in (and after adding any other suggestions / changes to these recipes that might be made), I'll remove the corresponding code from DelaunayTriangulation.jl so that it is Makie-free.

## Type of change

## Checklist


DanielVandH:
Worth adding anything from this to the documentation? I figured not since it is pretty DelaunayTriangulation.jl specific.

Two other comments:

1. While adding this I tried fixing some of the Observables issues, e.g. I want this to work:

```julia
tri = triangulate([rand(2) for _ in 1:50])
_tri = Observable(tri)
fig, ax, sc = triplot(_tri)
map(_tri) do tri 
	refine!(tri, max_area=1e-4get_total_area(tri))
end # Plot should now be updated 
```

Is my understanding of how this should work incorrect?

2. I want to be able to truncate the ghost edges properly when they are plotted. I thought `xautolimits` and `yautolimits` should work, as I've done in `triplot`, but it doesn't seem to do anything:

```julia
tri = triangulate(rand(2, 50))
triplot(tri, show_ghost_edges = true)
```

image

ffreyer:
> I want to be able to truncate the ghost edges properly when they are plotted. I thought xautolimits and yautolimits should work, as I've done in triplot, but it doesn't seem to do anything

`autolimits` doesn't really truncate the edge anyway. It just tells axis whether to consider this plot for limit calculations. You can still zoom in and out after that. To actually remove these lines you'd need to remove the data points responsible for them. Maybe you can check for points outside a bounding box or check line lengths for that?

ffreyer:
Oh are they supposed to fill out the rectangle of the Axis? In that case I would suggest you create your own rectangle in the recipe and just stop them there. Then you can define `needs_tight_limits(::PlotType) = true` so that the axis behaves like it does for a heatmap plot

DanielVandH:
Didn't know about `needs_tight_limits`, that seems useful. I'll try and come up with a reasonable default for a rectangle size and then implement that, then.

DanielVandH:
Works a treat! Thanks for that suggestion.

```julia
triplot(triangulate(rand(2, 50)), show_ghost_edges = true)
```

image

jkrumbiegel:
> needs_tight_limits

@ffreyer I was actually hoping we could remove this at some point. It's kind of an axis setting set by a plotting function (setting `xautolimitmargin` and `y...` to (0, 0)` and I'm not sure if it's the wrong level of abstraction. I just had to add it at the time so that making heatmaps and images isn't as annoying..

SimonDanisch:
Maybe we should release a breaking DelaunayTriangulation.jl release without Makie and rely on that here?

DanielVandH:
That would be good. Let me release one tomorrow and change the compat in
this PR.

Do you know when a new release for Makie might be released if I do that? Is
it a schedule type release? Just thinking in case a new user gets the
newest release for DT.jl but has to know to downgrade to get 'triplot' in
the interim.

DanielVandH:
I've released the new version now (and then realised I have a bunch of edge cases to fix with the plots :) ). The test failures seem to just be from old reference images after I updated the recipes a bit.

DanielVandH:
Added doc pages for `triplot` and `voronoiplot`. It is sufficient to just add a new file inside the plotting functions folder as I've done?

jkrumbiegel:
should be I think, they are picked up just for being in that folder

DanielVandH:
Great! I don't think there's anything else I want to commit further, then

SimonDanisch:
I can't believe this actually fails the tests:
voronoy
While he have a pretty low score for this pretty serious looking regression:
delaunay
Still need to see if the latter is from the PR or the previous Delaunay PR

SimonDanisch:
I can't believe this actually fails our CI:
voronoy
While this pretty serious looking regression doesn't fail the CI:
delaunay
Still need to see if that's a regression from this PR, or the previous Delaunay PR!

SimonDanisch:
I think it's the last PR!

DanielVandH:
Weird... Do you want me to take a look into it? The one shown on the docs looks okay good though:

image

DanielVandH:
I assume the reference image would be the one generated from MiniQhull? (If not, maybe this comment is irrelevant): If you look at the two triangulations:

```julia
tri = triangulate([x'; y'])
orig_tri = Triangulation([x'; y'])
for T in eachcol(Int64.(MiniQhull.delaunay([x'; y'])))
    add_triangle!(orig_tri, T)
end
fig, ax1, sc = triplot(tri)
triplot!(Axis(fig[1, 2]), orig_tri) # MiniQhull
fig
```

image

They're very different (Delaunay triangulations aren't unique when points are cocircular). I imagine the different triangles here lead to very different contours. I don't know how `tricontourf` actually makes the contours, is it piecewise linear interpolation over each element? If so, this might also contribute to the differences since I imagine the triangles in mine are more appropriate. (This is also why something like [NaturalNeighbours.jl](https://github.com/DanielVandH/NaturalNeighbours.jl) is useful for interpolation ;) ) Here are the comparisons for the contours:

```julia
fig, ax, pl = tricontourf(tri, z, strokewidth=1, strokecolor=:black)
tricontourf!(Axis(fig[1,2]), orig_tri, z, strokewidth=1, strokecolor=:black)
fig
```

image

DanielVandH:
If none of that is related, then I'm not sure it would be related to DelaunayTriangulation now being there

ffreyer:
For me the marker stroke is visible in GLMakie and CairoMakie if I run the example manually. :shrug:

src/basic_recipes/voronoiplot.jl
```diff
@@ -0,0 +1,171 @@
...
+        push!(colors, (a + b) / 2)
+    end
+    return colors
+end
+
+
+Makie.conversion_trait(::Voronoiplot) = PointBased()
+Makie.convert_arguments(::Voronoiplot, x::DelTri.VoronoiTessellation) = (x,)
+
+function Makie.plot!(p::Voronoiplot{<: Tuple{<: Vector{<: Point2f}}})
```
ffreyer, line 96: These three function are there so `voronoiplot` also works with PointBased data, so `Vector, Vector` or `Vector{Point2f}`. I guess we would want this for triplot too?

src/basic_recipes/voronoiplot.jl
```diff
@@ -0,0 +1,166 @@
...
+    attr = copy(p.attributes)
+    smooth = pop!(attr, :smooth)
+
+    # Handle transform_func early so tessellation is in cartesian space.
+    vorn = map(p.transformation.transform_func, p[1], smooth) do tf, ps, smooth
+        transformed = Makie.apply_transform(tf, ps)
+
+        # TODO: Make this work with Point2f directly
+        tri = DelTri.triangulate(Point2{Float64}.(transformed))
```
ffreyer, line 105: I noticed that with Float32 this errors somewhere down the line. Would be good to fix this

ffreyer (commented):
I also set up the `voronoiplot(points)` method to apply transform functions before the tessellation. I think that's necessary to get clean tessellations with nonlinear transforms. For example in a PolarAxis:

```julia
f = Figure()
ax = PolarAxis(f[1, 1])
points = Point2f[(r, phi) for r in 1:10 for phi in range(0, 2pi, length=36)[1:35]]
tr = voronoiplot!(ax, points, smooth = false, show_generators = false)
ax = PolarAxis(f[1, 2])
tr = voronoiplot!(ax, points, smooth = true, show_generators = false)
f
```
Screenshot from 2023-08-05 16-15-23

With this `voronoiplot` should fairly usable as a nonlinear/unstructured heatmaps as discussed in #748. I think we should figure out a better boundary for the outer cells though. Perhaps we can do something with the average distance of voronoi vertices to the central point?

src/basic_recipes/voronoiplot.jl
DanielVandH: Can you share the stacktrace you saw? I thought I had fixed this previously. e.g.

```
julia> triangulate(rand(Float32, 2, 64))
Delaunay Triangulation.
    Constrained: false
    Has ghost triangles: false
    Number of points: 64
    Number of triangles: 116
    Number of edges: 189
```

```
julia> triangulate(Point2f[rand(2) for _ in 1:50])
Delaunay Triangulation.
    Constrained: false
    Has ghost triangles: false
    Number of points: 50
    Number of triangles: 83
    Number of edges: 147
```

src/basic_recipes/voronoiplot.jl
DanielVandH: Yes, I'll put those into `triplot` on Monday. Basically anything that works for `voronoiplot` should work for `triplot`.

DanielVandH:
That `PolarAxis` is very cool! Yes the boundary cells have always been a problem. I tried to find something satisfactory for them when I implemented the smoothing algorithm, but I couldn't come up with anything good yet. I'm not sure what can be done there.

src/basic_recipes/voronoiplot.jl
ffreyer: I thought I could get away without converting to a matrix but `centroidal_smooth` seems to error with points

src/basic_recipes/voronoiplot.jl
ffreyer: I remembered how this works incorrectly but it should work now

ffreyer:
Hmm, yea I guess without having a target shape it's hard to know how many vertices to add. Maybe it's better to make that a future goal.

src/basic_recipes/voronoiplot.jl
DanielVandH: Ah. You are right:

```
julia> centroidal_smooth(voronoi(triangulate(pts)))
ERROR: MethodError: no method matching set_point!(::Vector{Point{2, Float32}}, ::Int64, ::Float32, ::Float32)

Closest candidates are:
  set_point!(::Any, ::Any, ::Any)
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\interfaces\points.jl:290
  set_point!(::Triangulation, ::Any, ::Any, ::Any)
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\data_structures\triangulation\points.jl:78
  set_point!(::AbstractMatrix{T}, ::Any, ::Any, ::Any) where T
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\interfaces\points.jl:286
  ...

Stacktrace:
 [1] move_generator_to_centroid!(points::Vector{Point{…}}, vorn::VoronoiTessellation{Triangulation{…}, Tuple{…}, Int64, Tuple{…}, Set{…}, Tuple{…}}, generator::Int64)
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\voronoi\lloyd.jl:7
 [2] _centroidal_smooth_itr(vorn::VoronoiTessellation{…}, set_boundary_nodes::Set{…}, points::Vector{…}, edges::Nothing, boundary_nodes::Vector{…}, I::Type, E::Type, V::Type, Es::Type, Ts::Type, F::Type, rng::Random.TaskLocalRNG; kwargs::@Kwargs{})
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\voronoi\lloyd.jl:22
 [3] _centroidal_smooth_itr(vorn::VoronoiTessellation{…}, set_boundary_nodes::Set{…}, points::Vector{…}, edges::Nothing, boundary_nodes::Vector{…}, I::Type, E::Type, V::Type, Es::Type, Ts::Type, F::Type, rng::Random.TaskLocalRNG)
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\voronoi\lloyd.jl:17
 [4] centroidal_smooth(vorn::VoronoiTessellation{Triangulation{…}, Tuple{…}, Int64, Tuple{…}, Set{…}, Tuple{…}}; maxiters::Int64, tol::Float64, rng::Random.TaskLocalRNG, kwargs::@Kwargs{})
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\voronoi\main.jl:77
 [5] centroidal_smooth(vorn::VoronoiTessellation{Triangulation{…}, Tuple{…}, Int64, Tuple{…}, Set{…}, Tuple{…}})
   @ DelaunayTriangulation C:\Users\User\.julia\packages\DelaunayTriangulation\Hlp1Y\src\voronoi\main.jl:59
 [6] top-level scope
   @ REPL[15]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

I've made the function `set_point!` too specific:

```julia
function set_point!(points::AbstractVector{T}, i, x, y) where {F,T<:NTuple{2,F}}
    points[i] = (F(x), F(y))
end
```

Probably needs to be something else like

```
function set_point!(points::AbstractVector{T}, i, x, y) where {T<:AbstractVector}
    points[i] = T(x, y) # or whatever the conversion needs to be ...
end
```

I have to come back and test this in DelaunayTriangulation.jl some more.

DanielVandH:
Note that smoothing is just about shifting the vertices around in this context, so no vertices get added. There probably is a good way to combine the existing mesh refinement methods with this though. But yes this will need to a future effort.

ffreyer:
Does the order of polygons generated match up with the order of points put in? If I put in polygon colors it doesn't match up the way I would expect it to

ffreyer:
E.g.

```julia
ps = [Point2f(x, y) for x in 1:10 for y in 1:10]
cs = [x for x in 1:10 for y in 1:10]
voronoiplot(ps, color = cs)
```

results in (with polygon_color being renamed to just color):

Screenshot from 2023-08-05 17-34-44

DanielVandH:
The polygons are stored as `Dict`s (because not every vertex has to have a polygon when considering constrained triangulations, hence why it's not a vector). in the line

```
gtr = [DelTri.get_generator(vorn, i) for i in DelTri.each_generator(vorn)]
```

it's iterating over the generators, and `each_generator(vorn)` which is just the keys:

```
julia> each_generator(vorn)
KeySet for a Dict{Int64, Tuple{Float64, Float64}} with 50 entries. Keys:
  5
  35
  20
  16
  30
  19
  32
  49
  6
  45
```

So no, order is not preserved. I originally did something like `gtr[i] = ...`, where `i` was from the `i`th polygon, but that assumes that all vertices are inside the triangulation which they don't have to be.

DanielVandH:
For the tessellation as you've created, all vertices will be in the triangulation, so I guess it could be appropriate to want that example you've given to work. Just need to access the `i`th index from the color vector when iterating over the generators.

For the general case where a user provides the tessellation directly, can I provide colors as `Dict`s mapping generators to colors? Or does Makie assume they're vectors?

ffreyer:
I think for PointBased inputs like above it would be the most intuitive to have input colors and points match, and internally permute or discard them. I'm not sure what would would be best if a Tessellation is passed. I don't think Makie cares at all what format you pass in `Voronoiplot`, it should just lower to some color type or vector of color types or vector of numbers before going into poly.

ffreyer:
As a heatmap+ we would want `voronoiplot(xs::Vector, ys::Vector, zs::Matrix)` to convert to a grid with points specified by xs and ys, and cell colors set by `zs`. That's what I want to enable in the end. The points should be working fine now, and colors should get extracted to match the order of points. Now the colors just need to match up with the cells

DanielVandH:
Right. I can set it up so that the polygon colours, if they are a vector,
must have length `num_points(get_triangulation(vorn))`. Internally, I'll
delete any vertices that aren't present and then permute the polygon list
to match accordingly. I should be able to do that on Monday

ffreyer:
Seems like `color = automatic` gets replaced with a flat color before the `plot!` function runs, so it breaks the automatic colors generated there :(

src/basic_recipes/voronoiplot.jl
DanielVandH: I have fixed a bunch of DelaunayTriangulation.jl issues relating to Float32 (by just converting to Float64 internally - Float32 is really not sufficient for what is needed for computing triangulations/tessellations unfortunately, since ExactPredicates.jl is limited to Float64) (https://github.com/DanielVandH/DelaunayTriangulation.jl/pull/74 and https://github.com/DanielVandH/DelaunayTriangulation.jl/pull/75). Hopefully you don't run into any issues anymore.

DanielVandH:
On the latest commit:

```julia
ps = [Point2f(x, y) for x in 1:10 for y in 1:10]
cs = [x for x in 1:10 for y in 1:10]
cs2 = [y for x in 1:10 for y in 1:10]

fig = Figure()
ax = Axis(fig[1, 1])
voronoiplot!(ax, ps, polygon_color = cs)
ax = Axis(fig[1, 2])
voronoiplot!(ax, ps, polygon_color = cs2)
fig
```

dc53eadfb7d466b60e354c1f113885eb

DanielVandH:
I've also added some conversions into `triplot`. Now that [DelaunayTriangulation.jl v0.8.3](https://github.com/DanielVandH/DelaunayTriangulation.jl/releases/tag/v0.8.3) is registered, I think most of the remaining issues are solved.

Is there anything else you think might be needed here still @ffreyer?

ffreyer:
I'll go through and make some more updates to docs and tests and whatever I stumble over. 

Are the default colors in a voronoiplot meaningful or is that just a way to generate distinguishable colors?

DanielVandH:
The default colours don't mean anything, no - just a way to get distinguishable colours as you say. The `(a + b) / 2` is just to convert the coordinates into a single number, so it could be anything.

src/basic_recipes/voronoiplot.jl
```diff
@@ -0,0 +1,191 @@
...
+                # bb_ref = RefValue(Rect3f())
+                # foreach(DelTri.get_generators(vorn)) do (_, p)
+                #     update_boundingbox!(bb_ref, Point2f(p))
+                # end
+
+                # x0, y0, _ = minimum(bb_ref[]) .- extent .* widths(bb_ref[])
+                # x1, y1, _ = maximum(bb_ref[]) .+ extent .* widths(bb_ref[])
+
+                # return (x0, x1, y0, y1)
+                return DelTri.polygon_bounds(vorn, extent)
```
ffreyer, line 156: I noticed that sometimes the poylgon vertices end up pretty far away from the input vertices:

```julia
using Random
Random.seed!(1234)

points = rand(Point2f, 50)
f, ax, p = voronoiplot(points)
f
```

Screenshot from 2023-08-07 16-24-00

I tried switching to a bounding box based on the input positions here to make this less of an issue, but DelaunayTriangulation isn't happy when triangle vertices end up outside the bounding box. Would it be possible to make that work, so we can avoid these far-away vertices here?

DanielVandH:
The polygon vertices come from circumcenters of triangles. When the
triangles are very thin, the circumcenters can be extremely far away as
they have a very high circumradius. For random point sets, thin triangles
are very common, so you will see this problem frequently for random points
as you have shown.

The current code for clipping to a bounding box assumes that only the
boundary generators have their polygons affected, but this problem could
occur for interior generators, which is much harder to account for. I'm not
sure how I would effectively fix this issue honestly. I would have to think
hard about it.

DanielVandH:
Not sure why this isn't replying in the actual comment but anyway.
I thought of some ideas actually, let me try and get to this soon.

ffreyer:
I would guess that for every poylgon generated, you can look for intersection of edges with the bounding box and draw a new edge between the intersection points. As far as I understand, if the bounding box is outside the initial set of points you should always have 0 or 1 vertex outside, meaning 0 or 2 intersections. Though maybe I'm thinking about this wrong

DanielVandH:
Yeah, it will be reasonably straightforward. I was getting confused with a previous implementation of these coordinates. Will try and work on it today.

src/basic_recipes/voronoiplot.jl
```diff
@@ -0,0 +1,246 @@
...
+    empty!(polygons)
+    sizehint!(generators, DelTri.num_generators(vorn))
+    sizehint!(polygons, DelTri.num_polygons(vorn))
+    # TODO: get_polygon_coordinates() errors without a bbox tuple.
+    pseudo_bbox = DelTri.polygon_bounds(vorn, 0.1)
+    for i in DelTri.each_generator(vorn)
+        g = DelTri.get_generator(vorn, i)
+        p = Point2f(DelTri.getxy(g))
+        p in bbox && push!(generators, p)
+        polygon_coords = DelTri.get_polygon_coordinates(vorn, i, pseudo_bbox)
```
ffreyer, line 123: `get_polygon_coordinates()` has some bbox checks in it which make it fail when none is passed, so I'm passing the old auto-generated bbox here. Would be good to avoid this

DanielVandH: The solution I'm working on will hopefully address this. Do you mean a `nothing` is passed? If you paste here something you'd like to work, I will test against that.

FYI I've already got some existing code for the Sutherland-Hodgan algorithm in my branch on DT.jl for this, don't worry about going further with that.

src/basic_recipes/voronoiplot.jl
```diff
@@ -105,21 +104,65 @@ function clip_polygon(poly::Polygon, rect::Rect2)
...
+end
+
 function clip_polygon(poly::Polygon, ::Nothing)
     return poly
 end
 
 function get_voronoi_tiles!(generators, polygons, vorn, bbox)
+    inside(p, bb::Rect2) = p in bb
+    inside(p, c::Circle) = dot(p - c.center, p - c.center) < c.r * c.r
+    inside(p, bb) = true
```
DanielVandH, line 154: Is the aim here to support clipping on arbitrary shapes? Do note that a lot of complications arise even in the rectangle case due to the unbounded polygons. I'm working on generalising the Sutherland-Hodgman algorithm for these cases, but the corner/degenerate cases put you into a rut quickly.

src/basic_recipes/voronoiplot.jl
DanielVandH: Yes, unfortunately it has to be handled by DT. Since the Voronoi polygons are unbounded, special care is needed for the unbounded edges. The bounding box is necessary so that we can plot them. e.g. if you look at some of the polygon indices, you might see something like

```julia
julia> get_polygon(vorn, 1)
6-element Vector{Int64}:
  6
  2
 -1
 -5
  5
  6
```

these negative numbers (ignore their values, just note that they're negative) denote vertices out at infinity.

src/basic_recipes/voronoiplot.jl
DanielVandH: Once I have merged https://github.com/DanielVandH/DelaunayTriangulation.jl/pull/76, we will be able to do some more testing here. Arbitrary (rectangular) bounding boxes will be able to be provided, and a better default for the bounding box will be 

```julia
DelTri.polygon_bounds(vorn, extent, include_polygon_vertices = false)
```

DanielVandH: The default behaviour is much nicer:

```julia
using Random
Random.seed!(1234)

points = rand(Point2f, 50)
f, ax, p = voronoiplot(points)
f
```

1314ff41ec637f1c172055f6c0b378ca

DanielVandH:
Added some tests with custom bounding boxes that don't even have to contain all polygons. The custom bounding boxes will require https://github.com/MakieOrg/Makie.jl/pull/3146 to be merged in cases that the polygons aren't included in the bounding box.

ffreyer:
Did some more cleanup and brought circular clipping back for PolarAxis. That also uses circular clipping by default now, if you use heatmap-style inputs.

> In the image, generators are only included if their polygon appears somewhere in the bounding box. This does make it possible that some generators appear outside of the bounding box as shown, though.

It's easy to exclude these by checking against the bbox, but we actually don't want to hide them, do we?

src/basic_recipes/voronoiplot.jl
```diff
@@ -0,0 +1,228 @@
...
+            end
+        elseif dot(Cp2, Cp2) < r2 # p1 outside, p2 inside
+            p = intersection(p2, p1, circle)
+            push!(output, p, p2)
+        end
+    end
+
+    return Polygon(output)
+end
+_clip_polygon(poly::Polygon, ::Any) = poly
```
DanielVandH, line 90: I'm not sure this can easily work. The clipped polygons against the circle will not match with the actual circle itself, you would need to place arcs between points on the circle. It also doesn't seem to currently work. 

```julia
A = (-3.0, 7.0)
B = (1.0, 6.0)
C = (-1.0, 3.0)
D = (-2.0, 4.0)
E = (3.0, -2.0)
F = (5.0, 5.0)
G = (-4.0, -3.0)
H = (3.0, 8.0)
points = [A, B, C, D, E, F, G, H]
tri = triangulate(points)
vorn = voronoi(tri)
fig = Figure()
ax = PolarAxis(fig[1, 1])
voronoiplot!(ax, vorn, show_generators = true, markersize = 14, color = :white)
ax = Axis(fig[1, 2]) # just to see what it actually looks like 
voronoiplot!(ax, vorn, show_generators = true, markersize = 14, color = :white)
fig
```

32918029b800569033bc49e2171bdad2

The polygons clip together as you can see, and you can see what I mention about needing arcs: The polygon near 45 and 90 has an empty space on the edge connecting two points on the boundary.

DanielVandH:
> It's easy to exclude these by checking against the bbox, but we actually don't want to hide them, do we?

Yeah, I went back and forth on this point. I think it makes sense to keep them.

src/basic_recipes/voronoiplot.jl
DanielVandH: Oh, the issue with the first is because the conversion to a circular bounding box only happens when you pass points, not when you pass a tessellation itself. Is that the behaviour you prefer?

DanielVandH: Passing the circular bounding box itself doesn't seem to be any better actually:

```julia
fig = Figure()
ax = PolarAxis(fig[1, 1])
c = Makie.Circle(Makie.Point2f(0.0, 0.0), 15.0)
voronoiplot!(ax, vorn, show_generators = true, markersize = 14, color = :white, bounding_box = c)
ax = Axis(fig[1, 2]) # just to see what it actually looks like 
voronoiplot!(ax, vorn, show_generators = true, markersize = 14, color = :white)
fig
```

5f637cc088d03f98c9f404629133eb48


But if it's intentional that this should be only when `points` are provided (which I've not tested) then I guess this is fine.

ffreyer: There is a bit of a messy situation here with transform functions. If you do
```julia
ax = PolarAxis(fig[1, 1])
voronoiplot!(ax, vorn)
```
you end up applying the voronoi tessellation in Polar coordinates, which I assume just doesn't make sense. The Polar -> Cartesian transformation will apply later down the line in poly and scatter.

If you pass points directly we can resolve this in the `plot!` method beforehand, i.e. do Polar -> Cartesian with
```julia
transformed = Makie.apply_transform(tf, ps)
```
and then do the voronoi tessellation in Cartesian space. To avoid applying the Polar -> Cartesian transformation twice we remove it with
```julia
transform = Transformation(p.transformation; transform_func=identity)
```

If we wanted to resolve the polar transform in both cases we would have to extract the input points, transform them and then rerun the tessellation. I'm not sure if that's possible (can we recover how the tessellation was set up?) and it also suggests doing things inefficiently (double tessellation if you're not careful). I'd rather just have the `VoronoiTessellation` method be the "I know what I'm doing" version and the `Vector{Point2f}` method be the "Do what I want" version.

> Passing the circular bounding box itself doesn't seem to be any better actually:

Same problem here - when plotting to a PolarAxis and directly passing a tessellation the bounding box applies in polar coordinates.

DanielVandH: That makes sense. Would it be sensible to disallow passing `vorn` like this into a `PolarAxis` then? Or just trust the user to "know what they're doing" as you say? I'm up for either.

ffreyer: > I'm not sure this can easily work. The clipped polygons against the circle will not match with the actual circle itself, you would need to place arcs between points on the circle.

You can make the same argument for the polygons in general. If the boundary is curved, why not all the edges? To me it looks kind of weird if it's just the outer ones: (quick test)
Screenshot 2023-08-12 015801
I think in the end this will just result in dragging out the pr a lot for relatively little gain.

DanielVandH: Fair point. I'm not too fussed either way, just wanted to make sure you are aware. Let us not worry about it :)

src/basic_recipes/voronoiplot.jl
```diff
@@ -177,6 +177,7 @@ function plot!(p::Voronoiplot{<:Tuple{<:DelTri.VoronoiTessellation}})
         if color === automatic
             # generate some consistent distinguishable colors
             cs = [sum(DelTri.get_generator(vorn, i)) for i in DelTri.each_generator(vorn)]
+            reverse!(cs)
```
DanielVandH, line 180: I wanted to make it possible to move a point slightly and not have all the colors change so much, but this makes it hard to also want distinguishable colors. So, just to get better default behaviour, I'll keep `reverse!` here. Don't need to do anything better.

ffreyer:
Are there still things you want to change before merging this pr?

DanielVandH:
@ffreyer No, everything seems to be good now!

DanielVandH:
I've fixed the `global cx` declaration. But I don't understand the StackOverflow errors - they seem unrelated to this since other PolarAxis tests are failing too...

src/makielayout/blocks/polaraxis.jl
```diff
@@ -273,14 +273,9 @@ function draw_axis!(po::PolarAxis, axis_radius)
         color = po.thetaticklabelcolor,
         strokewidth = po.thetaticklabelstrokewidth,
         strokecolor = thetastrokecolor,
-        align = thetatick_align[]
+        align = thetatick_align,
```
DanielVandH, line 276: If I change this to `thetatick_align[]` again, the StackOverflow goes away. Should we go back to that, or is the problem deeper?

ffreyer: Yea this stuff shouldn't get reverted. Git got confused somehow and is doing weird things

ffreyer:
Did a `rebase master`. Hopefully everything is fine now

DanielVandH:
> e: Didn't see that I somehow messed up a bunch of files when going back to this branch after creating the linked PR. Does it matter? Is it just the merged master?

Ah. Guess it wasn't the merged master. Don't know how I managed that. Sorry for mixing up the files like that - thanks for fixing it @ffreyer.

DanielVandH:
I'm not sure how I managed to make the same mistake again? I'm pretty sure all I did was pull within VSCode ...

SimonDanisch:
As long as you don't force push, it should just be a matter of undoing the last commit!
Although, it does seem like it isn't working with `git reset --hard HEAD~1`...
Had to do `git reset --hard 55c2201c6`. 
And yes, pulling from a branch that got forced pushed to isn't straightforward...
To be honest, I always do `git reset --hard HEAD~20` (some big enough number), and then pull again :D
I think the correct way is to do:
```
git reset --hard origin/your-branch
```
I opened a new PR with a corrected version since its easier to work with and actually builds + publishes the docs!
https://github.com/MakieOrg/Makie.jl/pull/3159

Issue #3060, opened 2023-07-11
Cannot plot `contour`s in both CairoMakie and GLMakie


tomchor:
When using `CairoMakie` to animate contours, I get the following error:

```
ERROR: LoadError: AssertionError: length(positions) == length(colors)
```

I posted this on slack a while ago but couldn't solve the issue, so I've decided to post here this time. I also remember someone online somewhere with the same issue as me and no solution, but I can't find that anymore.

Here's the MWE to reproduce this:

```julia
using CairoMakie

fig = Figure()
n = Observable(1)

ax = Axis(fig[1, 1])

N = 10
A = randn(N, N, N)
Aₙ = @lift A[:,:,$n]
contour!(ax, Aₙ)

frames = 1:N

record(fig, "mwe.mp4", frames, framerate=14) do frame
    @info "Plotting frame $frame..."
    n[] = frame
end
```

PS: The error doesn't happen in GLMakie, only usingo CairoMakie. Also for some reason the error seems to happen only in the second frame (i.e. the first frame plots fine), but when I manually try to plot (not animate) the first frame after the fact I get the same error:

```julia
julia> n=1
1

julia> current_figure() # hide
Error showing value of type Figure:
ERROR: AssertionError: length(positions) == length(colors)
...
```

- [x] are you running newest version (version from docs) ?
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x] What platform + GPU are you on?: up-to-date Manjaro with kernel 6.1.31-2, Julia 1.9.1 (from pre-compiled binaries), but I can reproduce this in all machines I've tried

Here's the full error output:

```
ERROR: LoadError: AssertionError: length(positions) == length(colors)
Stacktrace:
  [1] draw_multi(primitive::Lines{Tuple{Vector{Point{2, Float32}}}}, ctx::Cairo.CairoContext, positions::Vector{Vec{2, Float32}}, colors::Vector{ColorTypes.RGBA{Float32}}, linewidths::Vector{Float32}, dash::Nothing)
    @ CairoMakie ~/.julia/packages/CairoMakie/zPic1/src/primitives.jl:165
  [2] draw_multi(primitive::Lines{Tuple{Vector{Point{2, Float32}}}}, ctx::Cairo.CairoContext, positions::Vector{Vec{2, Float32}}, colors::Vector{ColorTypes.RGBA{Float32}}, linewidth::Float32, dash::Nothing)
[...]

PR #3106, opened 2023-07-28
Refactor SurfaceLike trait

ffreyer:
# Description

This pr breaks up the `SurfaceLike` trait and its children `ContinuousSurface` and `DiscreteSurface` into 3 (4) new traits:
- `GridBased` which more or less replaces SurfaceLike. The name change is meant to remove the association with Surface plots which is misleading imo.
    - `CellBasedGrid <: GridBased` which replaces `DiscreteSurface` with no changes
    - `VertexBasedGrid <: GridBased` which replaces `ContinuousSurface` for surface, contour and contourf plots. The only change here is to default to a `1:size(zs, dim)` instead of `0:size(zs, dim)`. This fixes the alignment issue noted in #3060 and a similar issue noted for surface in #2598.
- `ImageLike` which replaces ContinuousSurface specifically for image plots. This no longer produces arrays in the conversion pipeline, but instead just an interval. In backends an image plot is treated as a single quad so I don't think this removes anything

In #2598 there is a picture which shows how vertices are aligned for surface plots:

After this pr they would be aligned to integer x and y values instead, starting at (1, 1).

Example after pr: (should only change placement and scale)

```julia
begin
    f = Figure()
    img = [sin(i + 2 * cos(j)) for i in 1:10, j in 1:10]
    heatmap(f[1, 1], img)
    image(f[1, 2], img)
    a = Axis(f[2, 1])
    surface!(a, img, shading = false)
    xlims!(a, 1, 10); ylims!(a, 1, 10)
    contourf(f[2, 2], img)
    a, p = contour(f[2, 3], img)
    tightlimits!(a)
    f
end
```
Screenshot from 2023-07-28 18-34-27

With this and #3102 adding `voronoiplot`, maybe we can consider #748 done?

TODO:
- [x] Check if CairoMakie surface discoloration comes from this pr

## Type of change

Changes placements of some plots and may break some `convert_arguments` methods.

## Checklist

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -302,7 +302,7 @@ end
 
 
 ################################################################################
-#                                 SurfaceLike                                  #
+#                                  GridBased                                   #
 ################################################################################
 
 function edges(v::AbstractVector)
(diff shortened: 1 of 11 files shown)


ffreyer:
CairoMakie adds some red tones into the surface plot for some reason:

pr

But that also happens on master and has nothing to do with this pr:

master

@SimonDanisch @jkrumbiegel What do you think about these changes? Should we add them to the next breaking release?

jkrumbiegel:
Those changes sound good to me. Can we convert ranges to intervals for `ImageLike`? Should that error when the number of entries is different from the number of pixels? Or should we just be completely breaking and not allow ranges there at all, as they don't really make sense

ffreyer:
Oh right, that was a pretty aggressive change and I didn't actually check if it was ok... 

The behavior for irregular xs/ys and nonlinear transform functions in image is currently:
- GLMakie uses transformed extrema but keeps linear spacing in the quad
- CairoMakie errors
- WGLMakie uses transformed spacing

Heatmap with `interpolate = true` works with GLMakie and WGLMakie, but still errors in CairoMakie.

My assumption was that irregular images didn't work across the board so that an interval didn't remove any functionality. Since that's not true the question is whether we want to actually support that. CairoMakie and GLMakie currently treat an image simply as a single quad/rect, which should be a good bit faster than heatmap. If we want to support irregular images we'd probably repeat most of what heatmap does.

I don't really mind reverting the interval change for images, though I might rethink the naming. With vector xs and ys I would think of image as a top-right aligned cell based grid, which would fit into the GridBased branch of conversions but needs to be contrasted with heatmaps center aligned cell grids.

jkrumbiegel:
> whether we want to actually support that. CairoMakie and GLMakie currently treat an image simply as a single quad/rect

Yeah I think that's actually a very good distinction between heatmap and image that we should emphasize? I'm fine with specifying the boundaries of an image quad via intervals, while it makes sense I think to specify cell centers for the heatmap because you can shift those around. Not sure what we currently do if a range passed to heatmap is n or n+1 long (compared to the number of entries in that corresponding matrix dimension). Do we shift behavior automatically or error for one?

src/conversions.jl
```diff
@@ -320,62 +320,63 @@ function edges(v::AbstractVector)
     end
 end
 
-function adjust_axes(::DiscreteSurface, x::AbstractVector{<:Number}, y::AbstractVector{<:Number}, z::AbstractMatrix)
+function adjust_axes(::CellBasedGrid, x::AbstractVector{<:Number}, y::AbstractVector{<:Number}, z::AbstractMatrix)
     x̂, ŷ = map((x, y), size(z)) do v, sz
         return length(v) == sz ? edges(v) : v
     end
     return x̂, ŷ, z
 end
```
ffreyer, line 328: If n values are passed they are interpreted as centers and expanded to n+1, otherwise we just keep them. CairoMakie, GLMakie and WGLMakie are all ok with getting something other than n+1 values... I guess we should add errors here.

ffreyer:
Simon is also OK with the restrictions to `image`. 

I think the pr was mostly done, but we should probably emphasize the differences between heatmap and image a bit more in their docstrings.

ffreyer:
Should we merge this into the beta branch?

PR #3129, opened 2023-08-05
Tweaks readme

lazarusA:
- adds some additional buttons
- update example plots
- adds `details` so the the readme file is more user friendly (less crowded) and new user can easily start using the package.


jkrumbiegel:
Thank you! Good improvements, I think

jkrumbiegel (approved):

Issue #3137, opened 2023-08-08
zreversed feature in Axis3


Undermine28:
In aerospace, many coordinate systems are defined with a positive Z direction down. Currently, Axis3 does not support flipping the default direction of any of the axis. For example the vehicle-carried normal Earth reference frame:
<img src ="https://github.com/MakieOrg/Makie.jl/assets/59933649/6ff9e382-5ec7-4202-a53c-9fb59c42690f" width="400" class="center" />

### Feature description
A parameter similar to axis xreverse = true and yreverse = true but for axis3.

The critical difference here is that the coordinate system should ideally remain right-handed. Another feature could be to change from right-handed to left-handed. But the default behavior should keep right-handedness for consistency.

### Plot examples of what the final feature should do.
First, we have what is currently the default:
<img src="https://github.com/MakieOrg/Makie.jl/assets/59933649/d989fd10-2d82-4372-95df-751f3a1fcc79" width="800" />

Next, we have the plot with the z+ direction downwards. Notice the y-axis stays in the negative region but also changes direction to maintain right-handedness. 
<img src="https://github.com/MakieOrg/Makie.jl/assets/59933649/6393fde6-d633-4023-845d-bdc243fed9a8" width="800" />
(I made the graphs for illustrative purposes by rotating the data around the x-axis and then editing out and in the - tick markers where appropriate)

PR #3138, opened 2023-08-08
Add `Axis3` axis reversal attributes

jkrumbiegel:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/3137

Implementation could be cleaner I guess but I did the most straightforward thing I could think of and just flipped projection plus xor-ing everything in the decoration code. The Axis3 code is kind of annoying anyway because writing generic code for 3d cases is just confusing with the indexing etc, which is why I held off on doing this for so long in the first place.

Example:

test

## Type of change

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,8 +2,9 @@
 
 ## master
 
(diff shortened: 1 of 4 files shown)


Undermine28:
I compiled this commit locally and integrated it into my plotting. Features work as intended. Thanks a lot!

PR #3141, opened 2023-08-10
Improve CairoMakie lines with array colors

### LINKED ISSUE #3136: Transparent contour lines drawn as overlapping segments in CairoMakie
MWE:
```julia
using CairoMakie
x = range(-π, π; length=50)
z = @. sin(x) * cos(x')
fig, ax = contour(x, x, z, color=RGBAf(1,0,0,0.4), linewidth=6)
save("fig.png", fig)
```
Output:
testfig
Same result when saved to svg or pdf

System information:
```
  │    Machine      LENOVO ThinkPad T14 Gen 3 21CGS0TU00                │
  │    Kernel       6.4.8-arch1-1                                       │
  │    Distro       Arch Linux                                          │
  │    CPU          AMD Ryzen 7 PRO 6850U with Radeon Graphics (16)     │
  │    GPU          AMD ATI Radeon 680M                                 │
```
```
  [13f3f980] CairoMakie v0.10.7
  [e9467ef8] GLMakie v0.8.7
````

### PR DESCRIPTION
**Pull request by jkrumbiegel**
# Description

Fixes #3136 

This now tries to draw as many points of a `lines` plot continuously as possible in CairoMakie, which reduces the number of junction points that are visible when alpha colors are used and improves the look of dashes with solid-color lines as well.

## Before

image

## After

image

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -158,30 +158,22 @@ function draw_multi(primitive, ctx, positions, color, linewidths::AbstractArray,
     draw_multi(primitive, ctx, positions, [color for l in linewidths], linewidths, dash)
 end
 
-function draw_multi(primitive::Union{Lines, LineSegments}, ctx, positions, colors::AbstractArray, linewidths::AbstractArray, dash)
-    if primitive isa LineSegments
-        @assert iseven(length(positions))
-    end
+function draw_multi(primitive::LineSegments, ctx, positions, colors::AbstractArray, linewidths::AbstractArray, dash)
+    @assert iseven(length(positions))
     @assert length(positions) == length(colors)
     @assert length(linewidths) == length(colors)
 
-    iterator = if primitive isa Lines
-        1:length(positions)-1
-    elseif primitive isa LineSegments
-        1:2:length(positions)
-    end
-
-    for i in iterator
+    for i in 1:2:length(positions)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: CairoMakie/src/primitives.jl [hunk 1]
```diff
@@ -199,8 +191,99 @@ function draw_multi(primitive::Union{Lines, LineSegments}, ctx, positions, color
...
+                Cairo.move_to(ctx, this_position...)
+            else
+                # and this is also nan, do nothing
+            end
+        else
+            if this_color == prev_color
+                # this color is like the previous
+                if !this_nan
+                    # and this is not nan, so line_to and set prev_continued
+                    this_linewidth != prev_linewidth && error("Encountered two different linewidth values $prev_linewidth and $this_linewidth in `lines` at index $(i-1). Different linewidths in one line are only permitted in CairoMakie when separated by a NaN point.")
```
ffreyer, line 243: Maybe a bit off-topic, but is this a restriction from Cairo or could we have per-element linewidths here? GLMakie supports those so it would be nice if it worked in CairoMakie too.

SimonDanisch (approved):
Didn't look at the changes in detail, but didn't find any regressions in the refimages :)

jkrumbiegel:
Only thing I found is that the dots in the new testimages look a bit too wide, more like dashes. Not sure what's going on there, in other references the two backends seem to match better.

PR #3146, opened 2023-08-11
Allow for poly to work with empty polygons

DanielVandH:
# Description

This PR makes it possible for `poly` to work with empty polygons. This only required a change for `CairoMakie`'s `polypath` since it works for other backends.

## Type of change

## Checklist


DanielVandH:
As mentioned on Slack, I'm not sure what the best way to add unit tests for this is, since the issue is only found when showing the plot and I can't seem to find how to actually inspect the drawn plot properly. I added a reference image test that checks this for polygons, multipolygons, and also makes sure that the colors are correct when passed as a vector.

ffreyer:
I fixed some more cases that errored and added some conversion tests for those.

I think for the original thing you fixed we'd need to check if `display(poly(Polygon(Point2f[])))` errors on each backend if we wanted to do this outside of refimgs. Though as far as I can tell we don't check the color mapping for multiple polygons anywhere else, so I think it's fine to add the refimg test you made.

Side note: `poly(Polygon[])` gets interpreted as a 3D plot.

ffreyer (approved):
GLMakie tests passed, github actions just failed to finish correctly. Should be good to merge I think

DanielVandH:
> Side note: poly(Polygon[]) gets interpreted as a 3D plot.

I think this is common for empty plots, right? e.g. `scatter(Point2f[])` is a 3D plot, as is `lines(Point2f[])`.

jkrumbiegel:
Currently, yes, for some reason. In the future, probably not because we will go by type not content.

PR #3147, opened 2023-08-12
Add a theme that use LaTeX fonts for everything

Kolaru:
# Description

Most issues are asking to have the sans serif Makie font in LaTeX formulas, this PR only fix the font discrepency issue by putting the LaTeX font everywhere with the new theme.

A prototype fix for the problem in the other direction should come relatively soon.

With the theme I get that kind of results
image

## Type of change

## Checklist


jkrumbiegel:
I like the idea, maybe it's a good opportunity to show in the docs that you can merge themes. So we could call the theme `theme_latexfonts` and merge it once into the default and once into a dark theme or so to show how it works.

jkrumbiegel:
In your docs examples, better use `with_theme` to apply the themes only in a block, or reset to default at the end by calling `set_theme!() # hide`. Otherwise the theme changes might affect following examples

Kolaru:
I simply don't understand the doc build failure for `docs/documentation/theming.md`. It complains that `@L_str` is unavailable, but it should be available from `using CairoMakie` (and it is the same as in `docs/documentation/latex.md` anyway).

(also building the doc locally for either this branch or master error for me -- a different error, maybe because I'm on Windows. That's a separate issue, but it doesn't  help)

jkrumbiegel:
Hm I didn't have these macro issues in the docs in a while, but last time it was some Franklin weirdness that I couldn't really understand.

SimonDanisch:
Ok, I'm giving up... No idea why the `@L_str` works anywhere else, but not here...

SimonDanisch:
I don't really know what to do here... I really don't want to start debugging franklin, and I really don't want this PR to rot, just because of a Franklin bug...
@Kolaru can we somehow change the docs in a way, that it doesn't need the latex label for now? 😓

jkrumbiegel:
Yeah this is really strange.. @tlienart just FYI this is one of those cases where we assume Franklin is doing something which messes with evaluation. But we can't be sure 🤷‍♂️

Kolaru:
It would be a bit unfortunate to not have a LaTeX string as it shows the theming is consistent.

But that's a minor thing, I'll change it before we all succumb to madness.

tlienart:
do you know which example fails? because this seems to work fine too:

````
\begin{examplefigure}{}
```julia
using CairoMakie
CairoMakie.activate!() # hide
using Makie.LaTeXStrings # hide
function example_plot()
    f = Figure()
    for i in 1:2, j in 1:2
        lines(f[i, j], cumsum(randn(50)))
    end
    Label(f[0, :], "A simple example plot")
    Label(f[3, :], L"Random walks $x(t_n)$")
    f
end

example_plot()
```
\end{examplefigure}
````

Screenshot 2023-08-24 at 14 53 01

tlienart:
Ok I won't be able to do much more digging just now but this repo: https://github.com/tlienart/ftest-makie basically is just your `theming.md`; I can build it locally without issue but it fails to be deployed with the same issue you observed (here: https://github.com/tlienart/ftest-makie/actions/runs/5964248687/job/16179044884 ; it looks like it worked because I didn't toggle `fail_on_warning`) 

Failures at deployment stages are harder to fix because we need to figure out exactly what's happening a github machine and why things aren't the same as locally... 

Does Makie.LaTeXStrings require a working latex installation to be available on the machine to fully work for instance? (at this point I'm just guessing, help welcome if you can think of stuff that are different on github action than locally)

tlienart:
Ok I think I found a workaround, can you adapt your utils taking L28 from https://github.com/tlienart/ftest-makie/blob/main/utils.jl

That seems to be the problem, the code in examplefigure is not run in an environment that has the right `using` (somehow). I don't have the full picture in my head but changing that seems to fix the issue:

https://github.com/tlienart/ftest-makie/actions/runs/5968808262/job/16193413162#step:4:1

jkrumbiegel:
I have a theory, it probably really is about the `begin` block. It seems that a `using` inside a top-level `begin` cannot supply macros for resolving the same block. Which makes sense, I guess?

Here's an experiment in a new session with `LaTeXStrings` installed.

image

The first statement fails at load time, so the second one also doesn't work because `using` hasn't run yet. The third one is a single REPL command, but not a `begin` block, so there it works, probably because Julia executes each top-level statement one-by-one and after the first one, the macro in the second can be resolved.

So we might only have seen this error whenever a macro was used in the first cell of a page.

jkrumbiegel:
Although, looking at the file in this PR again, the example added is not the first on the page, and there are even other L strings on that page before it. Including in the first cell...

tlienart:
I'm not sure what your last comment refers to but, your previous comment about `using` in `begin` makes sense; so a straightforward way to get this fixed and merge this PR is to adjust `env_examplefigure` to remove the lines of `code` that start with `using` or `import` and shove those before the `__result` assignment.

SimonDanisch:
I think I tried that in https://github.com/MakieOrg/Makie.jl/pull/3180, but maybe I did something wrong...
Closing this in favor of the new pr.

PR #3154, opened 2023-08-14
PolarAxis improvements (full rlimits, thetalimits, improved controls, argument order)

### LINKED ISSUE #3152: `PolarAxis`: Allow limits on radius and angle
It would be great if a `PolarAxis` could be limited in radius and angle (including with a starting radius that is not 0).

This is very useful to make plots where the radii below a certain point are not of interest; and likewise where only a limited set of angles are needed.  This avoids large empty parts of a circle.

Here I use [GMT](https://www.generic-mapping-tools.org/) to do this:
```sh
awk '
    BEGIN{
        for (r=50; r<=100; r+=10) for (theta=-30; theta<=30; theta+=5) print r, theta
    }' |
    # `-R-30/30/50/100 sets theta limits to -30° to 30°
    # and radius limits to 50 to 100
    # while `-JP5c+a` sets a 5 cm-wide polar plot plotting azimuths clockwise from upwards
    gmt psxy -JP5c+a -R-30/30/50/100 -Sc0.1c -Gblack -Ba30/a50NseW -P > /tmp/p.ps
```

GMT_partial_polar_plot

The cryptic letter options are a bit inscrutable, but the style of simply setting minimum and maximum plot limits on _r_ and _θ_ seems logical to me.  These could be set by a convention on the order of the arguments to the `limits` keyword to `Axis`/`PolarAxis`.  (I don't like the inconsistency between the `-R` order and input data order for _r_ and _θ_!)  Or a new set of properties for a `PolarAxis` like `rlimits` and `thetalimits`.  `rlimits` could default to `(0, nothing)` and `thetalimits` to `(0, 2π)`.

Matplotlib, on the other hand, requires something like:
```python
import matplotlib.pyplot as plt
import numpy as np

fig, ax = plt.subplots(subplot_kw={"projection": "polar"})

ax.set_rlim(50, 100) # By default this makes the origin of the plot start at 50
ax.set_rorigin(0) # This is required to make the inner part of the plot be above 0
ax.set_thetalim(np.radians(-30), np.radians(30))
ax.set_theta_offset(np.radians(90))
ax.set_theta_direction(-1)

n = 50

rs = 50*np.random.rand(n) + 50
thetas = 60*np.random.rand(n) + 30

ax.scatter(thetas, rs)

plt.show()
```
Python_partial_polar_plot

Here I find the requirement to set the origin of the radial axis back to 0 if you adjust the lower _r_ limit unintuitive.  Makie could allow the true centre of the plot to be other than 0 with an option if desired, but that's just a simple transformation of the points and radial axis labels.

## Examples in the wild
From just my domain, these kinds of plots are ubiquitous:

- Mantle tomography on Wikipedia
- Simulation of mantle convection in super-earths
- Core dynamo simulation

### LINKED ISSUE #3158: PolarAxis Grid Z-Order
- [x] are you running newest version (version from docs) ? → Makie v0.19.8
- [x] What platform + GPU are you on? → Ubuntu 22.04.2 LTS + GeForce GTX 1070

In `PolarAxis`, grid and minorgrid are shown in the foreground, while in the normal `Axis`, the grids are in background. I think the z-layering of `PolarAxis` should be changed to that of `Axis`. 

MWE: 
```
using CairoMakie
CairoMakie.activate!()
fig = Figure( resolution = (500,1000) )
layout = fig[1,1] = GridLayout()
pts = Circle( Point((.1,0)), .9 )

ax = PolarAxis(layout[1,1])
poly!( ax, pts, color = :yellow, strokecolor = :black, strokewidth = 1 )

ax = Axis(layout[2,1])
poly!( ax, pts, color = :yellow, strokecolor = :black, strokewidth = 1 )

fig
```

image

### LINKED ISSUE #3178: Add option to switch argument order for plotting in `PolarAxis` (and make the swapped version default)
The new `PolarAxis` currently assumes the arguments of the plot command to be ordered as `r, theta`, identifying the cartesian `x` axis with the radius and the `y` axis with the angle. Event though cartesian coordinates are usually specified in this order, many other plotting libraries take the arguments the other way around:

- https://matplotlib.org/stable/gallery/pie_and_polar_charts/polar_demo.html#polar-plot
- https://pgfplots.net/polar-axes/
- https://ggplot2.tidyverse.org/reference/coord_polar.html

apart from feeling more natural, swapping the order to `theta, r` should make the currently (partly) broken recipe `stairs` work as intended and allow for `hist` to be used without explicitly setting the `direction=:x`:

```
using GLMakie
f = Figure()
ax1 = PolarAxis(f[1, 1])
ax2 = PolarAxis(f[1, 2])

xs = LinRange(0, 2pi, 10)
ys = rand(10)
stairs!(ax1, ys, xs, step=:center)
scatter!(ax1, ys, xs)
hist!(ax2, 2pi .* rand(300), direction = :x, strokewidth = 1, bins = range(0, 2pi, length=37))
f
```

If we (as suggested by @ffreyer on slack) add an option `iso_order=false` which swaps the order, the above code could become:
```
using GLMakie
f = Figure()
ax1 = PolarAxis(f[1, 1], iso_order=false)
ax2 = PolarAxis(f[1, 2], iso_order=false)

xs = LinRange(0, 2pi, 10)
ys = rand(10)
stairs!(ax1, xs, ys, step=:center)
scatter!(ax1, xs, ys)
hist!(ax2, 2pi .* rand(300), strokewidth = 1, bins = range(0, 2pi, length=37))
f
```

in addition, we should make `iso_order=false` the default, to be compliant with the way other large plotting packages are handling things.

(I am not 100% sure on the naming of that option. GIven that `hist` uses `direction=:x` rather than something like `vertical=false`, maybe calling the option `theta=:x` (similar to ggplot) might be more in line with the general conventions?)

ffreyer:
# Changelog

### Breaking changes

- The default argument order changed from `(r, theta)` to `(theta, r)`. This can be changed back with `polaraxis.theta_as_x[] = false`
- Changed `ax.radius` representing rmax to `ax.rlimits` representing (rmin, rmax)`

### Visual/Interactivity changes

- Placement or r and theta ticks changed to work with section views.
- Zooming changed to be compatible with section views. Zooming can be restricted interactively with `ax.rzoomkey = Keyboard.r` and `ax.thetazoomkey = Keyboard.theta` and more permanently with `ax.rzoomlock = false`, `ax.thetazoomlock = true` and `ax.fixrmin = true` (defaults)
- Grids are now placed behind plots. To revert this set `polaraxis.griddepth[] = 8999`.

### New Additions/Features

- Added `delete!(::PolarAxis, ::Plot)`
- Added `rlims!(ax, rmin, rmax)`
- Added `thetalims!(ax, thetamin, thetamax)` and `ax.thetalimits[]`
- Changed `autolimits!(ax[, unlock = true])` to allow the PolarAxis to derive its limits freely and unlock interactions.
- Added `rticklabelrotation = automatic` to orient rtick labels according to the corresponding grid lines.
- Added `zoomspeed = 0.1` to control the speed of zooming
- Added `rzoomkey = Keyboard.r` to restrict zooming to the r direction
- Added `thetazoomkey = Keyboard.t` to restrict zooming to the theta direction
- Added `fixrmin = true` to control whether zooming and translating affects rmin
- Added `rzoomlock = false` and `thetazoomlock = true` to block zooming in the respective directions
- Added `r_translation_button = Mouse.right`  for translating the plot in theta direction
- Added `theta_translation_button = Mouse.right` for translating the axis in theta direction
- Added  `axis_rotation_button = Keyboard.left_control & Mouse.right`to rotate the axis as a whole
- Added `reset_axis_orientation = false` to control whether the reset interaction reorients the axis
- Added `normalize_theta_ticks = true` to control whether theta ticks are normalized to a -2pi..2pi range

### Background Changes

- `sample_density` now applies to clip polygons
- Reordered attributes
- Added `target_thetalims`
- rtick labels now also affect the space reserved for ticks
- Added `AngularTicks` to derive more pleasant theta ticks
- Changed title align to work with numbers

---

# Description

- fixes #3152, adding rlimits and thetalimits to generate circle sectors 
- fixes #3158 by adding `ax.griddepth` and adjusting the default behaviour
- closes #3178 by adding `theta_as_x` and changing the default argument order
- adds translations

## TODO

- [x] reposition view based new limits
- [x] add limit for inner clip and adjust r_0 instead
- [x] try to prevent frequent regeneration of clip poly
- [x] improve automatic rtick placement & alignment 
- [x] ~~improve `MultipleTicks` for angles~~ add `AngularTicks`
- [x] clean up `rlims!`, add `thetalims!`
- [x] fix error on zoom
- [x] fix align not updating with varying theta_0
- [x] add mouse drag to adjust r limits (and maybe theta_0)
- [x] stop drag rotation from affecting the clip polygon ~~detach clip orientation from theta_0 and theta limits. You should be able to adjust~~
- [x] fix r ticks going out of frame when translating in clipped view
- [x] cleanup debug code
- [x] fix title placement in partial PolarAxis

## Review

- [x] default to no distortion, mention effect in docs
- [x] fix clip & spine resolution
- [x] add attribute to control clip and spine resolution
- [x] figure out a better name for `maximum_clip_radius` (is it better?)
- [x] fix rticks pointing inwards with `direction = -1`
- [x] consider bbox of r ticks when reserving space
- [x] add `rtick_rotation = true/false/automatic/angle` attribute 
- ~~reconsider `direction * (theta + theta_0)` vs `direction * theta + theta_0`~~
- [x] reconsider zooming implementation
- [x] reconsider translation implementation
- [x] fix error on zoom
- [x] fix `isfinite(vmax)` error when translating a lot
- [x] improve inner clip resolution (sector)

## Type of change

## Checklist


ffreyer:
For testing:
```julia
begin
    f = Figure()
    theta_0 = Slider(f[1, 1], range = range(0, 2pi, length=101)).value
    thetalims = IntervalSlider(f[2, 1], range = range(0, 2pi, length=101)).interval
    rlims = IntervalSlider(f[3, 1], range = range(0, 10, length=101)).interval
    po = PolarAxis(f[4, 1], thetalimits = thetalims, rlimits=rlims, theta_0 = theta_0)
    bb = map(po.thetalimits, po.target_radius, po.theta_0, po.direction) do _, _, _, _
        Makie.transformed_bbox(po)
    end
    p = scatter!(po, 8 .* rand(1000) .+ 1, 2pi .* rand(1000))
    l = lines!(po, bb, color = :yellow, transformation = Transformation())
    translate!(l, 0, 0, 9000)
    f
end
```

---

I'm currently handling r and theta limits here by adjusting the clip polygon. The Polar transform still works the same as before, i.e. no extra distortion and fully invertable. This however means that if you're interested in a large rmin, rmax with a small difference, like (8, 10), you will only get a slim ring like this:
Screenshot from 2023-08-15 20-27-03

I'm wondering if instead I should add a radial shift to the Polar transform, so I can keep the clip polygon at set radii (e.g. 0.1 and 1.0). With that the shift would make sure that whatever range needs to be visible would fit into the clip poylgon and the ticks would adjust to communicate that.

anowacki:
If it's not too bold to comment, I personally have no problem with a 'slim ring' when I have asked for a small range of _r_, so I would be happy with that.

(I don't have any opinion on the implementation as I don't understand that.)

jkrumbiegel:
I think it would be better to not go the "slim ring" route and instead do it via shift of the transform (a slimmer ring might still be desirable to keep the amount of radial stretching or "aspect ratio" of the data at some desired value.

That would make the transform not fully invertible, but I guess if we throw a clear error if any data clips inwards that's ok?

ffreyer:
> I think it would be better to not go the "slim ring" route and instead do it via shift of the transform (a slimmer ring might still be desirable to keep the amount of radial stretching or "aspect ratio" of the data at some desired value.

Yea I added an `r0` to the transform to translate data in r direction now, with `maximum_clip_radius` to set when that is adjusted instead of the clip area.

> That would make the transform not fully invertible, but I guess if we throw a clear error if any data clips inwards that's ok?

Since we don't use the inverse transform in PolarAxis at all we could, I guess, but I'm not sure if it's good to do that. It's quite easy for stuff to end up with r < 0 through zooming (and now translating) the axis. We probably shouldn't throw an error every time that happens if someone is using the inverse transform in some way. Probably better to just hide/leave stuff at r = 0.

ffreyer:
Some Notes on how I currently have things set up:

- `radius` has been replaced by `rlimits` which include an rmin and rmax. If rmin > 0 a circular clip in the center appears
- `maximum_clip_radius` has been added. If `rmin/rmax` exceed that value radii are shifted such that `(rmin - shift) / (rmax - shift) = maximum_clip_radius`. This keeps the inner clip from getting to close to the outer, i.e. it avoids the slim ring problem
- `thetalimits` have been added so that the PolarAxis can be reduced to for example a 180° view. 
- `theta_0` (now?) acts as an angular shift applying to **all** angles, i.e. it rotates data, grid lines, ticks and the clip polygon. It does however not change angular ticks - these are only affected by `thetalimits` (otherwise 0° in your data would not match 0° in tick labels). `theta_0` can be used to rotate your 180° view to whatever orientation you like.
- zooming (scrolling) can affect the size of the inner clip but never remove or create it
- radial translation (right drag) will generally try to keep data under your cursor which means it currently can remove the inner clip, but not create it. 
- angular translation (disabled by default) will also try to keep data under your cursor. It does not affects `thetalimits` and `theta_0` but does not change the orientation of the clip polygon.

Updated test plot:
```julia
f = Figure()

theta_0 = Slider(f[1, 1], range = range(0, 2pi, length=101)).value
thetalims = IntervalSlider(f[2, 1], range = range(0, 2pi, length=101)).interval
rlims = IntervalSlider(f[3, 1], range = range(0, 11, length=111)).interval

po = PolarAxis(f[4, 1:2], thetalimits = thetalims, rlimits=rlims, theta_0 = theta_0)
po.theta_translation_button[] = Mouse.right
bb = map(Makie.transformed_bbox,
    po.target_radius, po.thetalimits,
    map(x -> x.r0, po.scene.transformation.transform_func),
    po.direction,
    po.theta_0,
)

# these are the real values, not the values set by the slider (axis interaction causes desync)
Label(f[1, 2], map(x -> @sprintf("θ₀ = %0.1f°", x * 180/pi), po.theta_0))
Label(f[2, 2], map(lims -> @sprintf("θ = %0.1f° ..  %0.1f°", (180/pi .* lims)...), po.thetalimits))
Label(f[3, 2], map(lims -> @sprintf("r = %0.2f ..  %0.2f", lims...), po.target_radius))

p = scatter!(po, Point2f.(2, 2pi .* (0:7) ./ 8), color = 1:8, colormap = :RdBu)
lines!(po, range(1, 10, 101), range(0, 2pi, length=101))
p = scatter!(po, Point2f.(10, 2pi .* (0:7) ./ 8), color = 1:8, colormap = :RdBu)
l = lines!(po, bb, color = :yellow, transformation = Transformation())
translate!(l, 0, 0, 9000)
f
```

ffreyer:
New refimg for this pr:

Screenshot from 2023-08-19 21-28-17

Since github actions kept complaining about missing ref images before I added this there must be one missing on master?

ffreyer:
Ok I think this is pretty much done now. I may still try to add some unit tests for this if I can think of something useful... (maybe interactivity?) And fix tests if anything actually wrong pops up.

@jkrumbiegel Is there a way I can have `PolarAxis report to layouting what aspect ratio it wants? Or have suggest a bbox based on one proposed by layouting?

For (user) review, it would be good for me to get some feedback on what feels good or bad. For example:
- Should theta translations be disabled by default? (Use `ax.ax.theta_translation_button[] = Mouse.right` to enable)
- Does zooming and translating feel fine?
- Does zooming with `rmin > 0` feel correct? (i.e. that the inner clip can disappear?)
- Are theta ticks too dense? (In what ranges?)
- Is the r tick placement fine?
- Are docs on limits, theta_0, direction, maximum_clip_radius sufficient? (You can get to the pr docs by clicking "Details" on the documenter/deploy task in CI in this pr)

anowacki:
Other impressions:
- Everything looks nice and the fit of the sections to the available plot width is good.
- Maybe `direction` taking either `1` or `-1` is slightly unexpected compared to other 'either-or'/Boolean settings in Makie.  Perhaps `reverse_theta::Bool` or `thetareversed::Bool` (defaulting to `false`) as options would align more with other options like `xreversed` and `yreversed`.

ffreyer:
> * I find the behaviour around scaling the radial axis unintuitive.  I often want to plot sections of a circle near the outer radius, for example plotting a model of the Earth's interior.  The plot should not by default introduce stretching which distorts the aspect ratio of features in the plot.

Ah that's true, I didn't realize this. I would agree that having distortion without the user explicitly opting in is bad. Changed this to default to 1.0


> * It looks like more points should be used to plot the borders (and maybe set the clip mask?) when lots of scaling isn't applied to the data.  It looks like only about 4 points are used to cover 10° in my example above, which doesn't look so nice when the scaling is taken off.

Also a good point. Fixed

> * `maximum_clip_radius` can be understood once you read the docs, and I don't have a good alternative to suggest, but I couldn't guess what it meant just from the name.  I still don't get what clipping means here.  Is clipping not plotting points outside the theta and r limits?  Some name which implies stretching or scaling might be clearer on first reading.

Hmm, maybe this is leaning too much on the implementation then. To only draw things in the desired limits the PolarAxis is placing two polygons - one big square with a circle sector cutout for theta limits and rmax, and a smaller circle for rmin. The attribute sets what the maximum radius of the inner circle is, in units of rmax.
From your perspective I would say it controls when radial distortion starts to happen. I.e. if it's set to 0.2 you will get radial distortion at `rmin >= 0.2rmax`.

> * When you are only plotting part of a circle, it might look better for the radial tick labels to be outside the plot. Currently they overlap. 

Yes, I thought I had already set it up this way. Fixed now. 

> Note that the labels rotate so they line up with the lines of constant radius.

This seems a bit awkward to me as a hard-coded, because on a full circle it might lead to vertical or upside down labels. I'll add an attribute for this instead 

> * I think the theta ticks are mostly good, but sometimes are too close together and overlap, especially when the default scaling is applied and `(rlimits[end] - rlimits[begin])/rlimits[end]` is small.  (See refimages above.)

I don't think I'll try fixing this. You'd need to figure out space requirements for tick labels and the available space and then adjust things, which would change how much space is required and available. So you'd end up with a cyclic process which is annoying to deal with and a bit of a performance waste, probably. `Axis` also ignores this problem.

> * The docs are great and I understood what things do by reading them.  My only quibble is that I don't have a sense of which way `theta_0` rotates the plot without trying it first!  I think it rotates it a different way depending on whether `direction` is `1` or `-1`, which in retrospect is consistent with the docs to be fair.  Perhaps conceptually easier to handle for users is a `theta_0` which is always the anticlockwise rotation of the 0 point?

I stumbled over this as well. The Polar transform is defined to do `direction * (angle + theta_0)`, which I didn't touch here. To me `direction * angle + theta_0` also seems more intuitive, probably because I'm used to seeing things like `sin(k * t + t_0)`. With tick labels around I'm unsure, because currently `theta_0` always rotates in the direction tick labels show.

> * I personally wouldn't necessarily expect zooming to work on a polar plot because I know it would be hard to satisfy people's expectations.  It also errored when I tried to do it on this branch.

The error was me being too quick to push something. Fixed now.

> * My intuition on zooming is that you are zooming into a region in polar space, so the plot could either become:
>     * a cartesian plot of a section of that space, adjusted to fit the current subfigure size;
>     * or the `rlimits` and `thetalimits` could adjust to be a limited region of that plot so that the available part of the subfigure/layout is filled up.  As you zoom further in, that plot gets closer and closer to a rotated cartesian plot, unless you are near the centre, in which case it becomes a polar plot with a very limited radial range.

It's currently set up to adjust r limits, similar to master. 

Cartesian zoom is rather incompatible with how polar axis is set up. If we do zooming like you've shown we can't represent axis limits with r and theta anymore. I think it would overall just complicate a lot of things.

I'm not sure if I'm following the polar zoom idea with this, but something I could probably set up relatively easily would be a zoom like this:
Screenshot from 2023-08-21 18-07-41
This would follow the general rule that the point under your cursor is preserved when zooming and is compatible with polar limits. What I worry about with this is that you will probably abruptly cut out large chunks of the polar axis when zooming, and it will be hard to do radial zooming. Maybe this is another "Why not both?" case. (You can also do radial zooms with translations when the full circle is visible atm. Maybe that should change too?)

ffreyer:
Oh right, thank you for the review as well. I have not used polar axes a lot so getting input on how things should work is really helpful!

anowacki:
> Hmm, maybe this is leaning too much on the implementation then. To only draw things in the desired limits the PolarAxis is placing two polygons - one big square with a circle sector cutout for theta limits and rmax, and a smaller circle for rmin. The attribute sets what the maximum radius of the inner circle is, in units of rmax.
From your perspective I would say it controls when radial distortion starts to happen. I.e. if it's set to 0.2 you will get radial distortion at rmin >= 0.2rmax.

Thank you for the explanation.  Maybe then a name which includes 'scaling' might be better than one which includes 'clip'.  Perhaps `radius_scaling_fraction` or similar?

> I stumbled over this as well. The Polar transform is defined to do `direction * (angle + theta_0)`, which I didn't touch here. To me `direction * angle + theta_0` also seems more intuitive, probably because I'm used to seeing things like `sin(k * t + t_0)`. With tick labels around I'm unsure, because currently `theta_0` always rotates in the direction tick labels show.

I think so long as the direction it goes is clear in the documentation, it's liveable with.  I realise that changing this would be breaking.

> Cartesian zoom is rather incompatible with how polar axis is set up. If we do zooming like you've shown we can't represent axis limits with r and theta anymore. I think it would overall just complicate a lot of things.

That's absolutely fair enough!

> I'm not sure if I'm following the polar zoom idea with this, but something I could probably set up relatively easily would be a zoom like this:

This looks really cool and I think would work well.

I did get a chance to try the zooming and I have to say that limiting it to the radial direction only felt a little strange.  What you are proposing sounds great.

I would also say that 'theta translation' should be the default.  It was very useful to be able to essentially rotate the view of the data.  In fact I didn't feel I had an intuition for where things were going with radial translation.

ffreyer:
I think I'm about done with round two now. 

> > Hmm, maybe this is leaning too much on the implementation then. To only draw things in the desired limits the PolarAxis is placing two polygons - one big square with a circle sector cutout for theta limits and rmax, and a smaller circle for rmin. The attribute sets what the maximum radius of the inner circle is, in units of rmax.
> > From your perspective I would say it controls when radial distortion starts to happen. I.e. if it's set to 0.2 you will get radial distortion at rmin >= 0.2rmax.
> 
> Thank you for the explanation. Maybe then a name which includes 'scaling' might be better than one which includes 'clip'. Perhaps `radius_scaling_fraction` or similar?

I have it as `radial_distortion_threshold` atm.

> > I'm not sure if I'm following the polar zoom idea with this, but something I could probably set up relatively easily would be a zoom like this:
> 
> This looks really cool and I think would work well.
> 
> I did get a chance to try the zooming and I have to say that limiting it to the radial direction only felt a little strange. What you are proposing sounds great.

I ended up not going with my idea for zooming. I tried and noticed that you also have a degree of freedom in how much you zoom in r and theta direction. Tried a couple of things and ended just directly changing r and theta limits while keeping the ratio between arc length and (rmax - rmin) constant. 

I also adjusted the default to zoom in both directions and added `ax.fixrmin::Bool` so you can stop zooming from changing rmin.

> I would also say that 'theta translation' should be the default. It was very useful to be able to essentially rotate the view of the data. In fact I didn't feel I had an intuition for where things were going with radial translation.

Yea, radial translations were a weird mix of translation and zooming before. Now they just translate and refuse to go to lower r once `rmin = 0`. Theta translations are now on by default, and I also added the option to rotate the whole axis with left_control + Mouse.right.



I also tweaked the default limit behavior so that you can start off with an axis like your showed in the issue (#3152) without having to set limits yourself. I'm not that happy with them yet though. I'm currently padding r and theta limits with 5% padding to each side, which looks weird for wider limits imo.

anowacki:
This is all really amazing!  I think the zooming works better than I could have hoped.  The translation also works really well.  I knew what `fixrmin` would do without reading the docs even.  And the ability to rotate the plot (rather than changing the plot limits) is really useful.

I wonder whether holding down `x` and `y` might fix the translation to be only in the `theta` and `r` directions, just like translating a Cartesian plot?

> I have it as `radial_distortion_threshold` atm.

Just to note that in the code it's actually `radial_distortion_threshhold`, so that's one typo I can help spot.

> I also tweaked the default limit behavior so that you can start off with an axis like your showed in the issue [#3152](https://github.com/MakieOrg/Makie.jl/issues/3152) without having to set limits yourself.

The auto limits seem to work well, but as much as I like plotting only bits of the circle, I think the default should be to always plot all angles and plot the whole circle unless the user has set `rlimits` and `thetalimits`.  Otherwise one will be forever setting `rlimits=(0, nothing), thetalimits=(0, 2pi)`.

Perhaps a middle way could be an option to automatically set the `thetalimits` and `rlimits` from the data, which would be nice sometimes when working out what final limits to set?  Perhaps that would just be `rlimits=(nothing, nothing), thetalimits=(nothing, nothing)`?

I spotted some current issues that can arise when doing lots of zooming and translating:
- It's possible to get the theta labels to end up starting at 180° (and increasing to 540°).  I'm afraid I can't say exactly how I got into this state, but if I can manage it again, perhaps there is something useful I can dump from Makie's state to help work it out?
- Ctrl-clicking doesn't reset the plot back if you have rotated it with Ctrl-right-click.  I don't know whether it should or not, but I did expect it to and it took me a while to work out why it was happening.

ffreyer:
> I wonder whether holding down `x` and `y` might fix the translation to be only in the `theta` and `r` directions, just like translating a Cartesian plot?

I set that up with r and t. Not sure if that's better or worse than x and y, but you can also change it via attributes as a user.

> Just to note that in the code it's actually `radial_distortion_threshhold`, so that's one typo I can help spot.

Ah thanks. For some reason I always add another h there if I don't pay attention.

> > I also tweaked the default limit behavior so that you can start off with an axis like your showed in the issue [#3152](https://github.com/MakieOrg/Makie.jl/issues/3152) without having to set limits yourself.
> 
> The auto limits seem to work well, but as much as I like plotting only bits of the circle, I think the default should be to always plot all angles and plot the whole circle unless the user has set `rlimits` and `thetalimits`. Otherwise one will be forever setting `rlimits=(0, nothing), thetalimits=(0, 2pi)`.
> 
> Perhaps a middle way could be an option to automatically set the `thetalimits` and `rlimits` from the data, which would be nice sometimes when working out what final limits to set? Perhaps that would just be `rlimits=(nothing, nothing), thetalimits=(nothing, nothing)`?

I'll just bring the limits back to `(0.0, nothing)` and `(0.0, 2pi)`. `autolimits!(ax)` already  (almost) does the same as the current defaults

> I spotted some current issues that can arise when doing lots of zooming and translating:
> 
> * It's possible to get the theta labels to end up starting at 180° (and increasing to 540°).  I'm afraid I can't say exactly how I got into this state, but if I can manage it again, perhaps there is something useful I can dump from Makie's state to help work it out?

I thought about whether i should normalize limits like `(2pi, 4pi)` and decided that it's probably better not to. So you can get this right now by just setting your thetalimits to a range like that. For interactions I didn't think about it much anymore, so it became inconsistent. Zoom and translation do some kind of normalization, but not the same. I'll add another attribute to switch between "normalize to -pi .. 3pi" and keeping to the set limit range.

> * Ctrl-clicking doesn't reset the plot back if you have rotated it with Ctrl-right-click.  I don't know whether it should or not, but I did expect it to and it took me a while to work out why it was happening.

Also adding an attribute to change this. The current behavior is a bit weird, it should either hold the orientation of the axis and move move the limits there, or reset the orientation and the limits together.

ffreyer:
Ok, the things you noted should be taken care of now.

Since I switched back to full-circle PolarAxis by default I also adjusted the default interactions:
- `fixrmin = true` by default now and also stop radial translation (avoid circle -> ring)
- `rzoomlock = false` was added, which can restrict zooming in r direction
- `thetazoomlock = true` was added, which stops zooming in theta direction by default (avoid going from circle -> sector)

`autolimits(::PolarAxis)` now unlocks these interactions by default, and also allows `PolarAxis` to choose limits freely.

src/makielayout/blocks/polaraxis.jl
```diff
@@ -362,115 +860,99 @@ end
 ################################################################################
 
 
-function Makie.plot!(
-    po::PolarAxis, P::Makie.PlotFunc,
-    attributes::Makie.Attributes, args...;
+needs_tight_limits(::Surface) = true
```
ffreyer, line 863: One set of test failures (basically everything surface related) comes from this, because this also affects `Axis`. Personally I think surface plots on an Axis defaulting to tight limits is something we want but I can also just revert this.

ffreyer:
The other test failures around PolarAxis have a few sources...

#### PolarAxis Decorations before:
PolarAxis decorations
#### After:
PolarAxis decorations
This comes from making rticks align with the edge when thetalimits are less than 2pi. Should be easy enough to revert, but I personally don't have a strong opinion on which looks better. The outer most grid line being faint comes from moving grid lines into the background. 

#### PolarAxis scatterlines spine before:
PolarAxis scatterlines spine
#### After:
PolarAxis scatterlines spine
This has a different default rtick angle, which is also coming from rtick alignment in polar sectors and is easy to revert if we want, and this has a different zoom level because of
https://github.com/MakieOrg/Makie.jl/blob/f7c3713280e01ed2010bafc2e4bba767934e3a9d/src/makielayout/blocks/polaraxis.jl#L879-L881
I'm not sure how this is set up with Axis but we'll essentially need a `on(_ -> reset_limits!(po), window_open)` to fix it.

#### Voronoiplot with a nonlinear transform, Triplot with nonlinear transformation
Same rtick angle change as above

tl;dr question @jkrumbiegel, @SimonDanisch 
- Do we want to add tight limits by default for surface in Axis here?
- Should I revert the rtick angle and alignment changes for a full PolarAxis?
- How is the "reset limits on display" logic meant to work?

SimonDanisch:
>Do we want to add tight limits by default for surface in Axis here?

I'd say we shouldn't!

>Should I revert the rtick angle and alignment changes for a full PolarAxis?

I like them on a straight line but don't have a strong opinion ;)

jkrumbiegel:
The straight line would be problematic with slightly longer ticks, I'd argue for keeping them horizontal because of that.

ffreyer:
Looking back at this I think there are two specific question about rticks:
1. Should rticks (by default in a full PolarAxis) be placed on a diagonal like before or on a horizontal like now? (I.e. should rtickangle default to 0 or pi/8?
2. Should rticks (by default in a full PolarAxis) always be aligned horizontally like before or align themselves based on the angle they are placed at? (I.e. align themselves to the r or theta direction depending on which one is more horizontal.)


> > Do we want to add tight limits by default for surface in Axis here?
> 
> I'd say we shouldn't!

I think we should, because in a 2d Axis surface is more or less the same as image which has tight limits. But if we don't I'd just surface untightened here too

SimonDanisch:
I meant, we shouldn't do it in this PR, since its unrelated and strictly speaking breaking (by the definition that we need to update reference images).

ffreyer:
Removed/Commented out the surface tightlimits change, so now a surface plot in PolarAxis produces

Screenshot from 2023-09-20 14-48-14

(from test w/o rlims!() call)

---

Default rtickangle is unchanged, so the default rtick placement is still on a straight horizontal line:

Screenshot from 2023-09-20 14-41-41

With explicitly set rtickangle, the individual label rotation is now horizontal if thetalimits are > 1.9pi

Screenshot from 2023-09-20 14-41-59

and ":aligned" otherwise. That means they rotate according to their position, but stay close to upright and horizontal

Screenshot from 2023-09-20 14-42-29

PR #3155, opened 2023-08-15
Add tooling to bump versions

jkrumbiegel:
You should be on current master and then do `include("tooling/bump_versions.jl")` from the REPL. This tries to be a little bit smart in that it checks which `src` directories have changes, and it includes the dependencies to Makie.jl and MakieCore.jl.

@SimonDanisch try it out and let me know what you think


SimonDanisch:
Will try it out for our next release :) I think we can merge it like this and work from there.

PR #3170, opened 2023-08-20
Improve resolution of Mandelbrot example image

### LINKED ISSUE #3049: Thumbnail for the Plot Examples link in the docs could be improved
Currently, scrolling down to the First Steps section in https://docs.makie.org/dev/ shows:

image

I wonder if it would be better to show a better looking example for Plot Examples, which currently comes from https://docs.makie.org/stable/examples/plotting_functions/heatmap/#mandelbrot_heatmap. Rather than using a highly pixelated version (unless this is an intentional choice? in which case feel free to close), which isn't so attractive at least in my opinion, a higher quality Mandelbrot set could be a better initial view. For example, if I change the code that produces this image from

```julia
using CairoMakie


function mandelbrot(x, y)
    z = c = x + y*im
    for i in 1:30.0; abs(z) > 2 && return i; z = z^2 + c; end; 0
end

heatmap(-2:0.1:1, -1.1:0.1:1.1, mandelbrot,
    colormap = Reverse(:deep))
```

to instead use a denser range,

```julia
heatmap(-2:0.001:1, -1.1:0.001:1.1, mandelbrot, colormap = Reverse(:deep))
```

I instead get the image:

image

Which could be more appealing? The time difference is not too significant (obviously takes relatively much longer, but is 0.64s a long time for generating the docs?).

```julia
julia> @time heatmap(-2:0.01:1, -1.1:0.01:1.1, mandelbrot, colormap = Reverse(:deep))
  0.014030 seconds (44.08 k allocations: 4.232 MiB)

julia> @time heatmap(-2:0.001:1, -1.1:0.001:1.1, mandelbrot, colormap = Reverse(:deep))
  0.648334 seconds (44.10 k allocations: 153.999 MiB, 1.20% gc time, 0.38% compilation time)
```

DanielVandH:
# Description

Fixes #3049 

This just makes this figure look nicer so that the landing page has a nicer example image.

## Type of change

## Checklist

### DIFF
--- a/docs/examples/plotting_functions/heatmap.md
+++ b/docs/examples/plotting_functions/heatmap.md
@@ -34,7 +34,7 @@ function mandelbrot(x, y)
     for i in 1:30.0; abs(z) > 2 && return i; z = z^2 + c; end; 0
 end
 
-heatmap(-2:0.1:1, -1.1:0.1:1.1, mandelbrot,
+heatmap(-2:0.001:1, -1.1:0.001:1.1, mandelbrot,
     colormap = Reverse(:deep))
 ```
 \end{examplefigure}


jkrumbiegel:
Yeah I did that on purpose at the time, both because the docs build was already long, and because I thought it looked kind of "pixel-art-y". But I'm also ok with a high-res version :)

DanielVandH:
It's okay to close this then if you prefer @jkrumbiegel. I never thought about the pixel art aspect of it though - since usually the resolution of the Mandelbrot set is especially of interest - but I don't think it matters so much to change it. This image should only take about <1s to produce also

SimonDanisch (approved):

PR #3179, opened 2023-08-24
fix error bar range in recipes

### LINKED ISSUE #3176: Auto limits broken with recipes on 0.19.7 and 0.19.8
- [x] What platform + GPU are you on? Linux x86-64, nVidia, but running CairoMakie

<details><summary>version info</summary>

```julia
julia> versioninfo()
Julia Version 1.9.2
Commit e4ee485e909 (2023-07-05 09:39 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × Intel(R) Xeon(R) E-2288G CPU @ 3.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, skylake)
  Threads: 4 on 16 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4
```

</details>

<details><summary>MWE</summary>

Developed when diagnosing a PR in [MixedModelsMakie](https://github.com/palday/MixedModelsMakie.jl/pull/77)

```julia
using CairoMakie, Random

@recipe(CoefPlot, n) do scene
    return Attributes()
end

function Makie.plot!(ax::Axis, P::Type{<:CoefPlot}, allattrs::Makie.Attributes, n)
    plot = Makie.plot!(ax.scene, P, allattrs, n)

    if haskey(allattrs, :title)
        ax.title = allattrs.title[]
    end
    if haskey(allattrs, :xlabel)
        ax.xlabel = allattrs.xlabel[]
    else
        ax.xlabel = "Estimate and 95% confidence interval"
    end
    if haskey(allattrs, :ylabel)
        ax.ylabel = allattrs.ylabel[]
    end
    reset_limits!(ax)
    ax.yticks = 1:n
    ylims!(ax, 0, n + 1)
    return plot
end

function Makie.plot!(plot::CoefPlot{Tuple{Int}})
    n = plot[1][]
    xs = rand(MersenneTwister(42), n)
    ys = collect(1:n)
    se = 0.1
    scatter!(plot, xs, ys)
    errorbars!(plot, xs, ys, se, se; direction=:x)

    return plot
end

f = Figure()
ax = Axis(f[1,1])
coefplot!(ax, 2)
display(f)

```

</details>

<details><summary>Makie 0.19.8</summary>

makie0 19 8

</details>

<details><summary>Makie 0.19.7</summary>

makie0 19 7

</details>

<details><summary>Makie 0.19.6</summary>

makie0 19 6

</details>

Bisected to [#8ccb6bdd](https://github.com/MakieOrg/Makie.jl/commit/8ccb6bdd)

Possibly related to #3051

I am unable to replicate this when calling `scatter!` and `errorbars!` directly -- it seems to be some interaction with the recipe system.

palday:
# Description

Fixes #3176

Moves to data space instead of pixel space.

## Type of change

## Checklist

- [ ] ~Added or changed relevant sections in the documentation~
- [ ] ~Added unit tests for new algorithms, conversion methods, etc.~
- [ ] ~Added reference image tests for new plotting functions, recipes, visual options, etc.~

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## master
 
+- Fixed regression in determining axis limits [#3179](https://github.com/MakieOrg/Makie.jl/pull/3179)
+
 ## v0.19.8
 
 - Improved CairoMakie rendering of `lines` with repeating colors in an array [#3141](https://github.com/MakieOrg/Makie.jl/pull/3141).
--- a/src/basic_recipes/bracket.jl
+++ b/src/basic_recipes/bracket.jl
@@ -54,7 +54,7 @@ function Makie.plot!(pl::Bracket)
         return to === automatic ? Float32.(0.75 .* fs) : Float32.(to)
     end
 
-    onany(pl, points, scene.camera.projectionview, pl.model, transform_func(pl), 
+    onany(pl, points, scene.camera.projectionview, pl.model, transform_func(pl),
           scene.px_area, pl.offset, pl.width, pl.orientation, realtextoffset,
           pl.style) do points, _, _, _, _, offset, width, orientation, textoff, style
 
@@ -108,7 +108,7 @@ function Makie.plot!(pl::Bracket)
     end
 
     # Avoid scale!() / translate!() / rotate!() to affect these
-    series!(pl, bp; space = :pixel, solid_color = pl.color, linewidth = pl.linewidth, 
+    series!(pl, bp; space = :pixel, solid_color = pl.color, linewidth = pl.linewidth,
         linestyle = pl.linestyle, transformation = Transformation())
     text!(pl, textpoints, text = texts, space = :pixel, align = pl.align, offset = textoffset_vec,
         fontsize = pl.fontsize, font = pl.font, rotation = autorotations, color = pl.textcolor,
@@ -116,8 +116,8 @@ function Makie.plot!(pl::Bracket)
     pl
 end
 
-data_limits(pl::Bracket) = mapreduce(union, pl[1][]) do points
-    Rect3f([points...])
+function point_iterator(pl::Bracket)
+    point_iterator(pl.plots[1])
 end
 
 bracket_bezierpath(style::Symbol, args...) = bracket_bezierpath(Val(style), args...)
--- a/src/basic_recipes/error_and_rangebars.jl
+++ b/src/basic_recipes/error_and_rangebars.jl
@@ -239,7 +239,7 @@ end
 function plot_to_screen(plot, points::AbstractVector)
     cam = parent_scene(plot).camera
     space = to_value(get(plot, :space, :data))
-    spvm = clip_to_space(cam, :data) * space_to_clip(cam, space) * transformationmatrix(plot)[]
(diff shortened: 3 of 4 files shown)


ffreyer:
Rather than reverting this change we should probably have a custom `data_limits` method like bracket for now
https://github.com/MakieOrg/Makie.jl/blob/a4af178f3fac55a2d14d31b0dfbc9724f6d89139/src/basic_recipes/bracket.jl#L119-L121
and figure out a clean way to ignore non data space limits in Axis later, maybe as part of #3040.

palday:
@ffreyer there's already a custom data_limits

https://github.com/MakieOrg/Makie.jl/blob/a4af178f3fac55a2d14d31b0dfbc9724f6d89139/src/basic_recipes/error_and_rangebars.jl#L282-L285

but I don't know enough to know how to update it to fix this problem

ffreyer:
Huh, why doesn't that work then? Is that not getting called when updating limits?

ffreyer:
Oh ok, the problem comes from `errorbars` being wrapped in another plot recipe. In this case `Axis` calls
https://github.com/MakieOrg/Makie.jl/blob/a4af178f3fac55a2d14d31b0dfbc9724f6d89139/src/layouting/data_limits.jl#L170-L172
which goes into
https://github.com/MakieOrg/Makie.jl/blob/a4af178f3fac55a2d14d31b0dfbc9724f6d89139/src/layouting/data_limits.jl#L136-L144
and further into
https://github.com/MakieOrg/Makie.jl/blob/a4af178f3fac55a2d14d31b0dfbc9724f6d89139/src/layouting/data_limits.jl#L76
skipping the custom `data_limits`.

So if we swap out the `data_limits` function for
```julia
function point_iterator(bars::Union{Errorbars, Rangebars})
    point_iterator(bars.plots[1])
end
```
we should be good. If that works could you skim through the other plot recipes in `basic_plots` and `stats` and replace the other overwrites as well?

ffreyer:
`axis3d!` is fine, the other two would be good to change. For hexbin I think you can just call `decompose(Point2f, Rect3f(no, nw))` to keep it simple

jkrumbiegel:
I wonder if this makes sense

```julia
function iterate_transformed(plot) 
     points = point_iterator(plot) 
     t = transformation(plot) 
     model = model_transform(t) 
     # TODO: without this, axes with log scales error.  Why? 
     trans_func = identity # transform_func(t) 
     # trans_func = identity 
     iterate_transformed(points, model, to_value(get(plot, :space, :data)), trans_func) 
 end 
```

Because the transformation / model of a parent plot doesn't affect its children directly, does it? But we seem to want to transform by it (although not really as `trans_func` is set to identity for some reason)

ffreyer:
Yea, but I think it's better to have this be a quick fix and consider changes to `data_limits` more wholistically with #2881 / #3040.

jkrumbiegel:
I'm ok with that

SimonDanisch (approved):

SimonDanisch:
Thank you!

PR #3180, opened 2023-08-25
Tex theme

SimonDanisch:
Continuation of #3147


tlienart:
the change to `utils` is not in the right spot (the spot where you made the change is probably a good idea too).

a more general change is maybe something like this:

```
code_lines = split(code, '\n')
ui_lines = map(l -> startswith(l, r"(?:using|import)"), code_lines)
ui_code = join(code_lines[ui_lines], '\n')
rest_code = join(code_lines[.!ui_lines], '\n')
str = """
```julia:example_figure
$ui_code
__result = begin # hide
$rest_code
end # hide
```

(under the assumption that the `using/import` are at the beginning so that this won't change the ordering of code lines). Anyway what you have (just with `using Makie.LaTeXStrings: @L_str`) is probably enough.

SimonDanisch:
> Anyway what you have (just with using Makie.LaTeXStrings: @L_str) is probably enough.

well it doesn't work :(

tlienart:
did you fix the body of examplefigure?

SimonDanisch:
https://github.com/MakieOrg/Makie.jl/blob/bad94df760587de0bb9533345eab722bfb52f1cd/docs/utils.jl#L46
last commit ;)

tlienart:
That's in showhtml, the change should also be in L84-85

tlienart:
yup now the deployment step is happy

PR #3185, opened 2023-08-27
try pagefind for search

jkrumbiegel:
The `stork` maintainer has archived the project, so longer term we should find something else.

This one already seems to work quite well, it also fixes broken search on preview branches which is nice. But it doesn't yet have the on enter go to search page behavior. (Which is kind of broken right now anyway, because the search results don't load reliably).

### DIFF
--- a/docs/makedocs.jl
+++ b/docs/makedocs.jl
@@ -12,9 +12,23 @@ run(`$(npm_cmd()) install lunr`)
 run(`$(npm_cmd()) install cheerio`)
 
 using Downloads
-stork = Downloads.download("https://files.stork-search.net/releases/v1.4.2/stork-ubuntu-20-04")
-run(`chmod +x $stork`)
-success(`$stork`)
+using Tar
+
+pagefind = let
+    url = if Sys.isapple()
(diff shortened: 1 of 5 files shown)


docs/makedocs.jl
```diff
@@ -12,9 +12,23 @@ run(`$(npm_cmd()) install lunr`)
...
-stork = Downloads.download("https://files.stork-search.net/releases/v1.4.2/stork-ubuntu-20-04")
-run(`chmod +x $stork`)
-success(`$stork`)
+using Tar
+
+pagefind = let
+    url = if Sys.isapple()
+        "https://github.com/CloudCannon/pagefind/releases/download/v0.12.0/pagefind-v0.12.0-aarch64-apple-darwin.tar.gz"
+    elseif Sys.islinux()
+        "https://github.com/CloudCannon/pagefind/releases/download/v0.12.0/pagefind-v0.12.0-x86_64-unknown-linux-musl.tar.gz"
```
SimonDanisch, line 21: I guess it would be nice to have a _jll here ^^

PR #3190, opened 2023-08-29
time doc build

SimonDanisch:


SimonDanisch:
Ok, this seems to work normally. 
Still need to figure out, where the problem lays that suddenly docs take 3 hours to build - sometimes.

PR #3192, opened 2023-08-29
rename CI job

t-bltg:
Small consistency PR (the `Install Julia dependencies` message in CI is misleading since it also runs the tests).


jkrumbiegel:
Thanks!

PR #3135, opened 2023-08-08
Define custom Linestyle type (fix #2851)

Datseris:
This PR defines the new `Linestyle` type for handling custom linestyles. It also fixes #2851 already in its current state. What I don't know:

- Where to put `export Linestyle`.
- Where to put this `Linestyle` definition.
- Where to put the deprecated `convert` method.
- How to test this.

Any help is appreciated thank you all!

PR #3193, opened 2023-08-30
Linestyle type

SimonDanisch:
Continuation of #3135


src/conversions.jl
```diff
@@ -868,38 +868,65 @@ convert_attribute(c::Number, ::key"strokewidth") = Float32(c)
 convert_attribute(c, ::key"glowcolor") = to_color(c)
 convert_attribute(c, ::key"strokecolor") = to_color(c)
 
-convert_attribute(x::Nothing, ::key"linestyle") = x
+####
+## LineStyle conversions
```
jkrumbiegel, line 872: Linestyle

src/conversions.jl
```diff
@@ -868,38 +868,65 @@ convert_attribute(c::Number, ::key"strokewidth") = Float32(c)
...
+
+For example, with `value = [0.0, 4.0, 6.0, 9.5]`
+you start drawing at 0, stop at 4 linewidths, start again at 6, stop at 9.5,
+then repeat with 0 and 9.5 being treated as the same position.
+"""
+struct Linestyle
+    value::Vector{Float32}
+end
+
+to_linestyle(style::Linestyle) = Float32[x - style.value[1] for x in style.value]
```
jkrumbiegel, line 900: this seems weird to me. Why give a first value at all if it's always shifted back to zero?

SimonDanisch: Honestly, I don't know :D This has been there like this, and changing it now would be a breaking change I suppose.

PR #3201, opened 2023-09-01
fix brack data_limits

SimonDanisch:
#3179 introduced a regression in the `data_limits` for brackets, and since our reference images tests still have a way too high threshold, it didn't fail the tests.
@ffreyer, you think there are other places where this happened which are untested?
We should definitely look into the refimages this time before merging!


ffreyer:
https://github.com/MakieOrg/Makie.jl/blob/ac02141c4c87dbf71d06b301f6dc18f5719e6d05/test/makielayout.jl#L381-L405
This one is erroring for me in #3040, possibly for the same reason. It includes most/all of the mixed-space plots and the limits look fine to me on this branch. So that's probably a good sign.
Maybe we can also extend this to check limits to catch these kind of issues in future?

PR #3206, opened 2023-09-02
Refimage algorithm change

jkrumbiegel:
Some refactoring so far, but this PR deliberately errors right now where the metric is applied. That's the point of discussion, what do we want there?


jkrumbiegel:
obsolete due to https://github.com/MakieOrg/Makie.jl/pull/3520

PR #3210, opened 2023-09-04
fix contour boundingbox with `NaN`s

t-bltg:
# Description

Fixes [discord issue](https://discord.com/channels/996787732149981214/1148180700005154876/1148180700005154876) (fix https://github.com/MakieOrg/Makie.jl/issues/3211):

```julia
let
    a = b = 1:3
    c = [0 1 2; 1 2 3; 4 5 6]
    ct1 = contour(a, b, c; levels = [1, 2, 3], labels = true)   # OK
    d = [0 1 2; 1 2 3; 4 5 NaN]
    ct2 = contour(a, b, d; levels = [1, 2, 3], labels = true)   # NOK
end
```

Fix a `CairoMakie` bug when glyphs positions contained a `NaN`, corrupting the Cairo drawing state / context.

## Type of change

## Checklist


Lightup1:
Hi @t-bltg ,
I tried this branch.
There is no error message now, but there are also no contour lines now.

Lightup1:
image

jkrumbiegel:
Which backend was this?

Lightup1:
@jkrumbiegel CairoMakie

t-bltg:
Let me have a look this morning.

jkrumbiegel:
Could be a bug in CairoMakie, too, because there is one little line segment and afterwards nothing, also the axis lines are missing. That usually happens when Cairo silently corrupts the surface state, which can happen with NaNs sometimes.

t-bltg:
Let's reason first with `Nan`-less arrays: `c = [0 1 2; 1 2 3; 4 5 6]`.

What is visually represented is `reverse(c; dims=1)`:
```
3×3 Matrix{Float64}:
 4.0  5.0    6.0
 1.0  2.0    3.0
 0.0  1.0    2.0
```

So this is `contourf(a, b, c; levels)`:
contourf

With `GLMakie` (or `CairoMakie`), `contour(a, b, c; levels, labels=true)`:
glmakie.

Now with `NaN`s, `d = [0 1 2; 1 2 3; 4 5 NaN]`, `contourf(a, b, d; levels)`:
contourf-nan.

So I will expect at least the `1` and `2` levels lines with labels to appear. I'll rework the PR.

Lightup1:
GLMakie works fine now. But CairoMakie produces no contour lines.

t-bltg:
With GLMakie:

glmakie

Lines, but missing 2 contour labels.

t-bltg:
I'm marking as ready for review because I have no ideas left currently.
The contour algorithm output is different whether there is a `NaN` or not.

Lightup1:
Dear @t-bltg ,
Is there any quick fixing way to make Makie ignore NaN?
I tried to replace NaN with 1e6 in some cases it works fine, but in some other cases (when the border is smaller than the specified levels) it will draw multiple lines near the border between normal values and NaN.
Does julia support void element in a matrix which is ignored by Makie?

t-bltg:
> Does julia support void element in a matrix which is ignored by Makie?

I would want `missing` (or `nothing`) to behave how you describe unfortunately it doesn't work either in this case.

jkrumbiegel:
Aside from the NaN-interpolation question discussed above, is this now in mergeable state? With NaN causing the appropriate gaps?

t-bltg:
Current state, output from the code snippet in https://github.com/MakieOrg/Makie.jl/pull/3210#issue-1880608285.

**CairoMakie - no Nans**
cairo_no_nans

**CairoMakie - Nans**
cairo_nans

**GLMakie - no Nans**
gl_no_nans

**GLMakie - Nans**
gl_nans

jkrumbiegel:
Ok so CairoMakie is still buggy

t-bltg:
Is inconsistency between `CairoMakie` and `GLMakie` backends happening at backend level (the way every backend deals with segments) or is this something we can solve in `Makie` ?

jkrumbiegel:
This must be a NaN-handling bug in CairoMakie. It may not actually _draw_ NaNs, they have to be skipped (but one has to be careful to create the right paths in the process). And here it seems it's drawn (or something else that doesn't work) because the axis lines are missing, so Cairo broke in the process of drawing this.

t-bltg:
Oh right I didn't notice the axis lines missing.
Do you have any hint where this might occur ?
I'll investigate before the end of the week ...

t-bltg:
@jkrumbiegel, this is the new output, the bug was when drawing a glyph with `NaN`s.

**GLMakie - no NaNs c**
gl-no-nans-c
**GLMakie - NaNs d**
gl-nans-d

**CairoMakie - no NaNs c**
cairo-no-nans-c

**CairoMakie - NaNs d**
cairo-nans-d

CairoMakie/src/primitives.jl
```diff
@@ -587,6 +565,11 @@ function draw_glyph_collection(
         # offsets and scale apply in markerspace
         gp3 = glyph_pos[Vec(1, 2, 3)] ./ glyph_pos[4] .+ model33 * (glyphoffset .+ p3_offset)
 
+        if any(isnan, gp3)
```
t-bltg, line 568: This is the fix for https://github.com/MakieOrg/Makie.jl/pull/3210#issuecomment-1778933826.

jkrumbiegel:
Could you remove all the non functionality related formatting changes? Makes it harder to see what's up. Also the empty ifs with comments can be left as they are, I know they technically don't do anything but they helped me understand what the code was doing.

t-bltg:
> Could you remove all the non functionality related formatting changes?

Done.

> Makes it harder to see what's up.

I agree for the review, although these changes were make to simplify and make the current code consistent.

> Also the empty ifs with comments can be left as they are

Done.

jkrumbiegel (approved):

jkrumbiegel:
Thank you! Looks good to me

PR #3113, opened 2023-08-01
Rebased version of beta 0.20

SimonDanisch:
Continues #2831 !
Still needs to check, if I rebased correctly and didn't incorrectly apply some of the changes!

## Merged PRs
- https://github.com/MakieOrg/Makie.jl/pull/2598
- https://github.com/MakieOrg/Makie.jl/pull/2746
- https://github.com/MakieOrg/Makie.jl/pull/2346
- https://github.com/MakieOrg/Makie.jl/pull/2544
- https://github.com/MakieOrg/Makie.jl/pull/3082
- #2868
- #3062
- #3106
- #3281
- #3246

## TODOS

- [x] fix flaky test `@test GLMakie.window_size(screen.glscreen) == scaled(screen, (W, H))`
- [x] Merge axis type inferences from #2220 
- [x] Test on different resolution screens, IJulia, Pluto, VSCode, Windowed
- [x] rebase to only have merge commits from the PRs 
- [x] investigate unexpected speed ups
- [x] reset camera settings from tests
- [ ] check doc image generation
- [x] rethink default near/far in Camera3D (compatability with OIT)
- [x] merge #3246
- [x] fix WGLMakie issues/tests:
  - [x] fix line depth issues (see tests: ~~hexbin colorrange~~ (not new), LaTeXStrings in Axis3, Axis3 axis reversal)
  - [x] fix lighting of surface with nan points (fixed in #3246)
  - ~~volume/3D contour artifacts (see 3D Contour with 2D contour slices)~~ not new
  - ~~artifacting in "colorscale (lines)"~~ not new
- [x] GLMakie:
  - [x] slight outline in "scatter image markers" test
  - ~~clipping/z-fighting in "volume translated"~~ not new
- [x] CairoMakie:
  -  ~~Artfiacting in `colorscale (lines)"~~ not new
  - ~~markersize in "scatter rotations" changed?~~ not new
  - ~~color change in "colorscale (poly)"~~ not new
  - ~~transparency/render order of "OldAxis + Surface"~~ not new
  - ~~render order in "Merged color mesh"~~ not new
  - ~~render order of "Surface + wireframe + contour"~~ not new
- [x] Check "SpecApi in convert_arguments" (colors swapped?)

## Fixes the following errors

- fixes #2721 via #2746
- fixes #1600 via #2746
- fixes #1236 via #2746
- fixes https://github.com/MakieOrg/GeoMakie.jl/issues/133 via #2598
[...]

PR #3226, opened 2023-09-11
Unit support for Axes & Recipes, a.k.a axis converts

SimonDanisch:
Continuation of #1347.

#### Axis conversion
* Axis specifies types it can handle and creates objects that track global states for those types (e.g. categorical needs to track all categories of all plots in one axis)
* Axis compatible types => scene compatible types plus more types.
* Axis conversion can return axis parameters, which can be overridden step-by-step, similar to Plots.jl recipes.
* Scene conversions don't handle further axis-related tasks.
* For each plot command, we figure out the axis convert type, and then compare it to the existing axis conversion types in that axis. We then compare them by something like the Base.promotion trait system, and error if e.g. one tries to plot categorical values into a time axis.
* Each axis global (e.g. CategoricalConversion) will have an axis dimension associated and will be assigned to `(dim)ticks` of the axis, whcih then correctly creates axis labels etc.

#### We can implement float32 rescaling as a axis conversion:
* Update data rescaling only when axis limits change.
* The "limits" referred to in rescaling aren't the axis limits but the 2D scene limits.
    * Example: For user-facing limits from 1 to 100 and a log10 scale, scene limits would be 0 to 2. So, a real value of 1000 
       would be transformed to 3 and then rescaled to (3 - 0) / (2 - 0) = 1.5. 
* Example: Some type is converted by the axis to float128, which then gets converted + scaled for the scene to float32.
* The scene then converts float128 to float32.
* Transform functions need to occur before data rescaling
* We can optimize rescaling if inputs are already float32.

This PR right now ports all docs & examples and the actual implementation of unit conversions and bases it against https://github.com/MakieOrg/Makie.jl/pull/3113.
We still need to implement the actual axis conversion more cleanly, which was the Achilles heel for #1347 and the reason it wasn't merged yet. Since most work was getting units, printing and conversions to work, I hope that now with a better plan the axis conversion won't take that long to implement anymore.


ffreyer:
I want to collect some information and thoughts on Float32.

### Conversion Pipeline

Conversion steps (current):
1. `convert_arguments` normalizes type and structure of `plot.input_args`, generating `plot.converted`
2. `transform_func` applies any function to positional data in `plot.converted`
3. `model` moves the result into world space (translation + rotation + scale, technically arbitrary matrix transform)
4. `view` moves data into camera space (translation + basis change)
5. `projection` moves data into clip space, -1..1 range (perspective projection, scaling)
6. viewport/`pixel_space` moves data into pixel space, 0..w range (scaling, translation)

There might be some additional steps to deal with `markerspace`. All of those use matrices from steps 3-6.

Float issues:
1. Truncates to Float32, reducing precision
2. Can do arbitrary things such as moving data to a range with more float precision issues. (E.g. x -> x + 1e6, exp)
3. May introduce precision issues from translation
4. Same
5. Should be save since data gets scaled to -1..1 range before perspective projection?
6. Save

Steps 3-6 are Matrix operations and therefore can and should run on the GPU. We want to use Float32 here. (Note that Axis currently has translation in the projection matrix. We can easily move that to view. TODO: Check PolarAxis, Axis3)

### Fixing things

To fix precision issues while keeping 3-6 on the GPU we need to remove translation partially or fully from `view * model`. Scaling shouldn't matter. For example `1e6 .+ collect(0:10)` has precision issues. Removing translations would move the data into a -5:5 range, fixing the problem. Something like `10 .^ (1:10)` does not have issues. On the full range of 10^10 float precision of e.g. 1 never becomes relevant. Zooming would result in a translation keeping precision high.

In 3D `view = basis_transform * translationmatrix(-eyepositon)` so extracting the translation is easy. (2D can be adjusted to match.) For a "normal" model matrix composed out of a translation, rotation and scale we have

$$\begin{align}
&view \cdot model \cdot p \\
= &basis \cdot translationmatrix(-eyepos) \cdot (scale \cdot rotation \cdot p + translation) \\
= &basis \cdot (scale \cdot rotation \cdot p + translation - eyepostion) \\
= &basis \cdot scale \cdot rotation \cdot (p + rotation^{-1} \cdot scale^{-1} \cdot (translation - eyeposition)) \\
= &basis \cdot scale \cdot rotation \cdot (p - inv(model) \cdot eyeposition)
\end{align}$$

For an arbitrary model matrix we could calculate `M = translationmatrix(-eyepos) \cdot model`, generate a matrix $M'$ with the translating part of M set to 0 and compute a translation matrix $T = inv(M') \cdot M$ such that $M = M' \cdot T$. It is not clear to me whether this fixes float precision issues. However this is also an edge case which we may just want to ignore or disallow for this purpose.

With this we have an extra conversion step `p -> Float32(p + translation(view, model))` applying after step 2 and before step 3. Data in step 1 and 2 would need to be kept at it's original precision to remain accurate.

- Can't we just use eyeposition? - Technically no, as plots could be translated into positions that cause float precision issues
- Can we apply this earlier? - In most cases we can probably just inverse transform the offset, but there might be some transform functions that are not (locally) invertible or cause precision issues themselves. Things like `log(0)` or `InversePolar(-5, 30°)` could also cause issues

### TODO

Following the above I think our Float32 TODO would be:

- remove Float32 conversion from convert_arguments (I guess this means keeping the user type and just adjusting the structure of data (e.g. 3x Vector{<: Real} -> Point3 Vector)?)
- generalize transform_funcs to take arbitrary numeric types
- remove translations from the view matrix (and projection) and keep track of them in `camera.eyeposition`
- add a conversion step between transfunc and model that applies an offset derived from `camera.eyeposition` and `plot.model` to transformed data
- And to allow non-`Real` types I think we'd need to extend `convert_arguments` accordingly

What to do with categorical data is still a question. In this structure they would have to fit in with convert_arguments, but they also need to talk to the axis/scene...

Tangent:
Why is our model matrix defined as `scale_translation * rotation` rather than `rotation * scale_translation`? This means that `scale!(p, Vec3f(1,1,2)); rotate!(p, Vec3f(1,0,0), pi/4)` scales the plot y and z coordinates (world z) rather than just plot z (world y and z). Do we want that?
We're also missing `inv(::Quaternion)`.

SimonDanisch:
Updated TODOs from initial indepth tests:

- [ ] fix convert_arguments which return axis_convert values (units/categorical/dates)
- [ ] set limits with axis_convert values 
- [ ] only update scaling for Dates if value range changes and not finallimits
- [ ] pick up `{n}_dim_convert` from nested recipes, to make recipes using axis_convert values work

ffreyer:
Might be good to merge #3719 and extend it for documentation?

ffreyer (commented):
Some issues I noticed:

The axis only adapts its ticks to a plots conversion if the Axis has been created by the plot. E.g.
```julia
fig = Figure()
ax = Axis(fig[1, 1])
scatter!(ax, u"ns" .* (1:10), u"d" .* (1:10), markersize=20, color=1:10)
```
yields
Screenshot from 2024-04-02 17-13-09

When x values are generated the unit conversion check doesn't apply. E.g. this runs without error:
```julia
f, a, p = scatter(u"ns" .* (1:10), u"d" .* (1:10), markersize=20, color=1:10)
scatter!(a, u"ms" .* (1:10), u"hr" .* (1:10), markersize=20, color=1:10)
scatter!(a, 1:10)
f
```
Screenshot from 2024-04-02 17-24-03

On the other hand this errors:

```julia
scatter(u"m" .* (1:10))
```

ffreyer:
I tried rewriting the docstrings for `should_dim_convert()` and `@convert_targets` to make them clearer but I'm not sure if I actually correctly understood them. So please take another look at them.

Also if I do understand `convert_arguments_typed` correctly it's just there for error checking. Couldn't we simplify this to something like
```julia
conversion_target_type(::PlotTypeOrConversionTrait) = Tuple{type1, type2, ...}

function verify_conversion(PlotType, converted)
    return converted isa conversion_target_type(PlotType)
end
```

test/conversions.jl
```diff
@@ -351,9 +352,10 @@ end
         @test convert_arguments(Heatmap, r1, i2, m3) == (o1, o2, o3)
         @test convert_arguments(Heatmap, v1, r2, m3) == (o1, o2, o3)
         @test convert_arguments(Heatmap, 0:10, v2, m3) == (collect(0f0:10f0), o2, o3)
-        @test_throws ErrorException convert_arguments(Heatmap, m1, m2, m3)
-        @test_throws ErrorException convert_arguments(Heatmap, m1, m2)
-
+        # TODO, this throws ERROR: MethodError: no method matching adjust_axes(::CellGrid, ::Matrix{Int64}, ::Matrix{Int64}, ::Matrix{Float64})
+        # Is this what we want to test for?
+        @test_throws MethodError convert_arguments(Heatmap, m1, m2, m3) === (m1, m2, m3)
+        @test convert_arguments(Heatmap, m1, m2) === (m1, m2)
```
ffreyer, line 358: Something like `[1 2; 1.5 2.5]` doesn't work for heatmaps since the cells are rectangular. So you can't use the extra freedom a Matrix input gives you. I think that's what the missing conversion/test is implying.

ffreyer (commented):
I removed some things that seemed like forgotten comments/prototyping in the last commit. Also tried to rephrase some bits in the documentation.

PR #3227, opened 2023-09-12
Move recursion to data_limits

ffreyer:
# Description

I thought about the changes in #3179 some more and came to the conclusion that moving the recursion to `data_limits` instead of `point_iterator` isn't something we need or should hold back. It's just a simple internal change that is pretty isolated from other bounding box and transformation changes.

This rolls back the data_limits -> point_iterator changes from #3179 and instead updates bbox recursion.

TODO:
- [x] check errorbars, i.e. #3179
- [x] check bracket
- [x] check hexbin

## Type of change

- Internal clean up

## Checklist

### DIFF
--- a/src/basic_recipes/bracket.jl
+++ b/src/basic_recipes/bracket.jl
@@ -116,9 +116,7 @@ function Makie.plot!(pl::Bracket)
     pl
 end
 
-function point_iterator(pl::Bracket)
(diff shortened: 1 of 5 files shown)


SimonDanisch:
This breaks `latex hlines in axis`:

latexhlines

PR #3230, opened 2023-09-14
cycle color for error and range bars

ErickChacon:
# Description

Currently error and range bars do not cycle colors. This PR simply allows to cycle colors when using error and range bars.

## Type of change

## Checklist


jkrumbiegel:
Thanks, I think this makes sense. Probably needs to be merged in the next breaking version, though, just in case people relied on the old behavior.

PR #3231, opened 2023-09-14
Refactor `Slider` and `IntervalSlider` for `dragging` logic

jkrumbiegel:
# Description

Supersedes https://github.com/MakieOrg/Makie.jl/pull/2228

I thought about it some more and now settled on the idea that the user can use the `dragging` status to decide what they want to do with a changed slider value. This is a bit more flexible than having a second output observable, and allows further custom logic depending on the slider drag state change. This approach also showed some smaller problems with the implementation which I therefore refactored a bit.

## Checklist

### DIFF
--- a/GLMakie/src/drawing_primitives.jl
+++ b/GLMakie/src/drawing_primitives.jl
@@ -72,9 +72,9 @@ end
 
 function handle_intensities!(attributes, plot)
     color = plot.calculated_colors
-    if color[] isa Makie.ColorMap
+    if color[] isa Makie.ColorMapping
         attributes[:intensity] = color[].color_scaled
-        interp = color[].categorical[] ? :nearest : :linear
+        interp = color[].color_mapping_type[] === Makie.continuous ? :linear : :nearest
         attributes[:color_map] = Texture(color[].colormap; minfilter=interp)
         attributes[:color_norm] = color[].colorrange_scaled
         attributes[:nan_color] = color[].nan_color
@@ -417,7 +417,7 @@ function draw_atomic(screen::Screen, scene::Scene, heatmap::Heatmap)
         t = Makie.transform_func_obs(heatmap)
         mat = heatmap[3]
         space = heatmap.space # needs to happen before connect_camera! call
-        xypos = map(t, heatmap[1], heatmap[2], space) do t, x, y, space
+        xypos = lift(t, heatmap[1], heatmap[2], space) do t, x, y, space
             x1d = xy_convert(x, size(mat[], 1))
             y1d = xy_convert(y, size(mat[], 2))
             # Only if transform doesn't do anything, we can stay linear in 1/2D
--- a/MakieCore/src/attributes.jl
+++ b/MakieCore/src/attributes.jl
@@ -31,6 +31,7 @@ Attributes(pairs::AbstractVector) = Attributes(Dict{Symbol, Observable}(node_pai
 Attributes(pairs::Iterators.Pairs) = Attributes(collect(pairs))
 Attributes(nt::NamedTuple) = Attributes(; nt...)
 attributes(x::Attributes) = getfield(x, :attributes)
+attributes(x::AbstractPlot) = getfield(x, :attributes)
 Base.keys(x::Attributes) = keys(x.attributes)
 Base.values(x::Attributes) = values(x.attributes)
 function Base.iterate(x::Attributes, state...)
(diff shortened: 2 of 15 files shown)


src/makielayout/blocks/intervalslider.jl
```diff
@@ -178,47 +178,42 @@ function initialize_block!(isl::IntervalSlider)
...
+    end
 
+    onmouseleftclick(mouseevents) do event
+        isl.dragging[] = false
+        notify(isl.selected_indices)
         return Consume(true)
     end
 
     onmouseleftdoubleclick(mouseevents) do event
+        isl.dragging[] = false
```
ffreyer, line 216: This causes an update like this:
```
IntervalSlider drag true
IntervalSlider drag false
IntervalSlider interval (0.0, 10.0)
```
So if you react to `dragging` you won't catch the new interval here. It would be better if you did. Maybe in general the `dragging = false` update should happen after the interval update?

src/makielayout/blocks/intervalslider.jl
```diff
@@ -178,47 +178,42 @@ function initialize_block!(isl::IntervalSlider)
...
+        handle_event(event)
         return Consume(true)
     end
 
     onmouseleftdragstop(mouseevents) do event
-        dragging[] = false
+        isl.dragging[] = false
         # adjust slider to closest legal value
         sliderfractions[] = sliderfractions[]
+        notify(isl.selected_indices)
```
ffreyer, line 192: This just duplicates updates
```
IntervalSlider interval (1.18, 10.0)
IntervalSlider indices (119, 1001)
IntervalSlider drag false
IntervalSlider interval (1.18, 10.0)
IntervalSlider indices (119, 1001)
```
so I think we should just skip the notify here.

src/makielayout/blocks/intervalslider.jl
```diff
@@ -178,47 +178,42 @@ function initialize_block!(isl::IntervalSlider)
...
-        end
-        # linecolors[] = [color_active[], color_inactive[]]
+        return Consume(true)
+    end
 
+    onmouseleftclick(mouseevents) do event
+        isl.dragging[] = false
+        notify(isl.selected_indices)
         return Consume(true)
     end
```
ffreyer, line 213: Clicks also trigger a drag start so you end up with a duplicate update here too:
```
IntervalSlider drag true
IntervalSlider interval (0.43, 10.0)
IntervalSlider indices (44, 1001)
[ Info: click start
IntervalSlider drag false
IntervalSlider interval (0.43, 10.0)
IntervalSlider indices (44, 1001)
[ Info: click end
```

There is an exception to this her though: If you do not move the slider through a click you do not get an interval update through the drag start. So then you only get the click update, which however does not change the values. I'm not sure if that's an update worth pushing out.

src/makielayout/blocks/slider.jl
```diff
@@ -114,23 +114,45 @@ function initialize_block!(sl::Slider)
...
+        sl.dragging[] = false
+        # adjust slider to closest legal value
+        sliderfraction[] = sliderfraction[]
+        fraction = get_fraction(event)
+        newindex = closest_fractionindex(sliderrange[], fraction)
+        selected_index[] = newindex
+
         linecolors[] = [sl.color_active_dimmed[], sl.color_inactive[]]
         return Consume(true)
     end
```
ffreyer, line 138: This has the same update duplication as IntervalSlider

src/makielayout/blocks/slider.jl
```diff
@@ -114,23 +114,45 @@ function initialize_block!(sl::Slider)
...
+            selected_index[] = newindex
+        end
         return Consume(true)
     end
 
     onmouseleftdoubleclick(mouseevents) do event
+        sl.dragging[] = false
         selected_index[] = closest_index(sliderrange[], sl.startvalue[])
         return Consume(true)
     end
```
ffreyer, line 158: Same update order problem as IntervalSlider

Sagnac:
@jkrumbiegel I've tried this specific slider dragging feature in a dev build before and found that it works quite well in its current state. Is there still a plan to continue working on this and eventually merge it or has it been abandoned?

jkrumbiegel:
I never took the time to follow up on @ffreyer's review comments, but in principle I'd still like to get equivalent changes in.

maxfreu:
Hi! Thanks for your efforts already! I just wanted to express interest in this; it would be really cool to have for uses when you don't want to trigger countless calls of a long-running function as you drag. In my use case I'd use the slider to flip through satellite images taken at different times and only want to read the target images from disk.

PR #3235, opened 2023-09-16
MouseStateMachine order fixes

jkrumbiegel:
- move mouseup before click and trigger click before doubleclick, too
- add mouse state machine tests to glmakie tests

I've noticed that the canonical order in browsers is up -> click -> doubleclick so I've changed it to that on our side, too https://www.w3.org/TR/DOM-Level-3-Events/#events-mouseevent-event-order

Actually while doing all of this I've got a stronger feeling that we need to rework our interaction / event model. It's difficult to know what object will trigger first, will consume the event, etc. Maybe we can have a similar "bubbling-up" concept as the web with its DOM, just with our recipes. DataInspector already does something similar.

### DIFF
--- a/GLMakie/test/unit_tests.jl
+++ b/GLMakie/test/unit_tests.jl
@@ -91,6 +91,59 @@ end
         @test Set(last.(scatter_plot_idx)) == Set(99991:99998)
         GLMakie.destroy!(screen)
     end
+
+    @testset "Mouse state machine" begin
+        s = Scene(camera = campixel!, resolution = (400, 400))
+
+        p1 = poly!(s, BBox(100, 200, 100, 200))
+        p2 = poly!(s, BBox(200, 300, 200, 300))
+
+        me = addmouseevents!(s, p1)
+
+        events = []
+
+        on(me.obs) do event
+            push!(events, event)
+        end
+
+        display(s) # don't immediately run next code, needs to wait for render first it seems?
+
+        sleep(0.1)
+
+        s.events.mouseposition[] = (0, 0)
+        s.events.mouseposition[] = (150, 150)
+        s.events.mouseposition[] = (170, 170)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.press)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.release)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.press)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.release)
+        s.events.mouseposition[] = (160, 160)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.press)
+        s.events.mouseposition[] = (150, 150)
+        s.events.mouseposition[] = (140, 140)
+        s.events.mousebutton[] = Makie.MouseButtonEvent(Makie.Mouse.left, Makie.Mouse.release)
+
+        MT = Makie.MouseEventTypes
+
+        @test map(e -> (e.type, (e.px...,)), events) == [
+            (MT.enter, (150, 150)),
+            (MT.over, (170, 170)),
+            (MT.leftdown, (170, 170)),
+            (MT.leftup, (170, 170)),
+            (MT.leftclick, (170, 170)),
+            (MT.leftdown, (170, 170)),
+            (MT.leftup, (170, 170)),
+            (MT.leftclick, (170, 170)),
+            (MT.leftdoubleclick, (170, 170)),
+            (MT.over, (160, 160)),
+            (MT.leftdown, (160, 160)),
+            (MT.leftdragstart, (150, 150)), # not sure about position here? should it be previous where the drag starts?
+            (MT.leftdrag, (150, 150)),
+            (MT.leftdrag, (140, 140)),
+            (MT.leftdragstop, (140, 140)), # should leftup come before this or not?
+            (MT.leftup, (140, 140)),
+        ]
+    end
 end
 
 @testset "emtpy!(fig)" begin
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
huh I never realized we had tests for mousestatemachine in events.jl, maybe these cover what I wanted already

ffreyer:
> huh I never realized we had tests for mousestatemachine in events.jl, maybe these cover what I wanted already

I think I added them with PriorityObservables back in AbstractPlotting. They just test whether things are working as they did before though, not anything more principled. Though I did leave some notes there on things that seemed odd or different from my OS to me.

ffreyer:
> Actually while doing all of this I've got a stronger feeling that we need to rework our interaction / event model. It's difficult to know what object will trigger first, will consume the event, etc. Maybe we can have a similar "bubbling-up" concept as the web with its DOM, just with our recipes. DataInspector already does something similar.

I think using scene/plot/visual order rules on interactions is a bad idea for our low level event interface. Some problems:
- If events can be added from outside, e.g. with `onclick()` you will probably still need priority to sort events independently of when they're added
- When priority and scene/plot/visual order exist separately one must prioritize one over the other. Which one should it be?
    - Something like an escape menu, pause menu, help context would be part of the root scene (low priority in that context) but should  take priority over pretty much everything else. So this wants priority first
    - two overlapping buttons would naturally have the same priority to their interactions. They want visual order first
- When I tried scene/plot order + priority in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/586 I ended up with something relatively slow (800µs per event vs 40µs with normal Observables and 140µs with priority observables if I'm reading my benchmarks correctly), Though maybe that was just a bad implementation...
- scene and plot order is independent of visual order (merged z buffer)
- visual order is dynamic, i.e. can change with plot args and `translate!`
- visual order in 3D is expensive (to calculate or to grab with pick) and even more dynamic due to 3D rotations/camera interactions

I wonder if instead of trying to run scene/plot/visual order alongside or instead of priority we could translate it to a priority value. With a restriction to 2D scenes/plots that would probably fix most of the issues above.

jkrumbiegel:
I thought more along the lines of, the first object to handle a mouse event is the one visually under the cursor, if not that one then the one containing it, and so on, until the scene can get it. Because usually you do want to click the menu that you see is open, not the button below it, right? I'm not sure exactly how javascript handles that, because the article I linked had examples of nested mouseenter events, where you enter three objects at the same time if they are nested divs for example.

The priority thing is difficult because there's no system for saying what priority anything should have. You can always dial it to some value that works in a given situation, but it's not principled as far as I know.

ffreyer:
> I thought more along the lines of, the first object to handle a mouse event is the one visually under the cursor, if not that one then the one containing it, and so on, until the scene can get it. Because usually you do want to click the menu that you see is open, not the button below it, right?

Yea that's true. I think what I want to avoid is creating another recursive structure like I had in the pr. I.e. something like

```julia
function scene_recursion(scene)
    foreach(scene_recursion, scene.children)
    foreach(plot_recursion, scene.plots)
    foreach(evaluate_interaction, scene.interactions)
end

function plot_recursion(plot)
    foreach(plot_recursion, plot.plots)
    foreach(evaluate_interaction, plot.interactions)
end
```

because that just ends up moving through a lot of scenes and plots unnecessarily. 

> What makes the "visually under the cursor" thing a bit more difficult is that if you use not a single pixel but a range, you get multiple objects and then it's not clear which one to choose. But it's important to have a range I think, otherwise it's very difficult to select text, for example. But maybe one should do this one via boundingbox again, I'm not sure. Although whenever we use boundingboxes, we get the layering problem again, where something occluding something else does not keep it from picking up events in its boundingbox. Which you usually don't want, I think.

`pick` would only get you the top most plot, so if we want to be able to go through each plot drawn at a certain position (i.e. the entered A, B, C thing) we need to work with bounding boxes. In 2d it shouldn't be that difficult to figure out the order of objects since they only depends on z translation and scene/plot order (though we may have some differences between backends here). With objects that have simple bounding boxes like UI elements that should also be much faster than fetching pixels from the gpu with `pick`. I would assume the web also works largely with bounding boxes.

For plots with lots of input data calculating bounding boxes may get slow, but I'm not sure if we even want to consider them in a system like this? They wouldn't be UI elements, so they shouldn't really compete with them, should they? If we do want to include them we should probably look into (lazily) caching bounding boxes/data limits and checking against scene bounding boxes before plot bounding boxes to filter irrelevant plots.

Btw, keeping depth buffers between scenes separate would also simplify this. If we have something like
```
parent scene
    axis scene
        axis background decoration sscene
        axis plots scene
        axis foreground decorations scene
    UI scene
        Button scene
        Menu scene
```
and none of these scenes mix depth buffers, then we can check them one by one rather than considering everything at once. (It might make sense to not always do this, so that Button scene and Menu scene share the depth buffer of UI scene. Or to make another object which is more of a container for plots than a new full new scene.) Having those scenes be tightly defined to the space they occupy would also help, since we can start with a super cheap `is_mouse_in_scene_rect` check. 

> The priority thing is difficult because there's no system for saying what priority anything should have. You can always dial it to some value that works in a given situation, but it's not principled as far as I know.

Yea I agree. I think I originally had some constant in there like `DEFAULT_UI_PRIORITY` etc. Not sure why I removed them, but that would probably help a bit. I was thinking that for 2D depth sorting we could have something like
```julia
function register_interaction(user_callback, object, ...[, small_priority = 0])
    is_in_limits(small_priority) || error()
    ...
    callback = on(some_event) do event
        if mouse_in_object_bbox(object)
            return user_callback(...)
        end
        return Cosume(false)
    end
    on(transformation(object).model) do model
        z = get_z_from_model(model) # -10k .. 10k range
        priority = priority_from_z_and_render_order(z, object)
        update_priority(callback, priority + small_priority)
        return Consume(false)
    end
    ...
end
```
for interactions. Updating the callback priority would take care of depth sorting and `mouse_in_object_bbox` would check if the mouse is hovering the object or not. The `small_priority` would allow for limited sorting of interactions within one objects (e.g. 0..99) and the priorities derived from depth and plot order would be limited to some other larger range, e.g. 100 * (1000..9999). That would overall limit interaction priorities to a range of 100_000 .. 999_999 which other event listeners could work around. Maybe I should just try to make a prototype for this?

PR #3242, opened 2023-09-18
Fix struct uniforms

ffreyer:
# Description

I wanted to play around a bit with lighting and noticed struct uniforms are broken. (I think I ran into this before?) There were two issues here:
1. `fieldnames(obj)` needs to be `fieldnames(type)` now
2. Doing a `merge!(data, other)` when iterating `for (k, v) in data` breaks the iteration

The second problem may still cause issues here:

https://github.com/MakieOrg/Makie.jl/blob/d61d05f5cc2a6f5a2678b51c39ff1f2f6b05d537/GLMakie/src/GLAbstraction/GLTypes.jl#L435-L438

I don't know what input hits this branch so I don't know how to test it. 

## Type of change

### DIFF
--- a/GLMakie/src/GLAbstraction/GLTypes.jl
+++ b/GLMakie/src/GLAbstraction/GLTypes.jl
@@ -423,8 +423,10 @@ function RenderObject(
         else
             k in (:indices, :visible, :ssao, :label, :cycle) && continue
             # structs are treated differently, since they have to be composed into their fields
+            # NOTE that merge!(data, ...) will break the iteration, so struct uniforms must
+            # be added after this iteration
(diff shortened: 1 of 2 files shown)


ffreyer:
This doesn't work with structs in Observables

SimonDanisch:
>I don't know what input hits this branch so I don't know how to test it.

None I think 😅 That's why it has been bitrotting!

ffreyer:
I've tried vector uniforms of normal gl types earlier today and they do actually seem to work. (I was tunnel visioning on vector of struct uniforms before.) Could be that those take that path, haven't checked yet.

PR #3244, opened 2023-09-20
Docs wider layout

jkrumbiegel:
This PR widens the desktop-style layout by moving the page contents in a separate section on the right, and the navigation on the left. Combined, these were often way to long. The main page part an also grow a bit larger to make better use of larger screens. The content overview goes into a dropdown on mobile.

grafik

grafik

The search on desktop goes to a button at the top right triggering a modal which has a lot of space. On mobile, the search is still above the navigation in the side bar. 

grafik

grafik

grafik


jkrumbiegel:
Any objections before I merge this?

PR #3246, opened 2023-09-21
Implement multiple lights and more Light types

### LINKED ISSUE #2985: Issue with typing on LScene with lights
The following fails:

```julia
julia> using RPRMakie
julia> let
           fig = Figure();
           radiance = 10000
           lights = [PointLight(Vec3f(0, 0, 20), RGBf(radiance, radiance, radiance))]
           ax = LScene(fig[1, 1]; show_axis = false, scenekw=(lights=lights,))
       end
ERROR: MethodError: no method matching Scene(::Scene, ::Events, ::Observable{GeometryBasics.HyperRectangle{2, Int64}}, ::Observable{Bool}, ::Camera, ::EmptyCamera, ::Transformation, ::Vector{AbstractPlot}, ::Attributes, ::Vector{Scene}, ::Vector{MakieScreen}, ::Observable{ColorTypes.RGBA{Float32}}, ::Observable{Bool}, ::SSAO, ::Vector{PointLight})

Closest candidates are:
  Scene(::Union{Nothing, Scene}, ::Events, ::Observable{GeometryBasics.HyperRectangle{2, Int64}}, ::Observable{Bool}, ::Camera, ::AbstractCamera, ::Transformation, ::Vector{AbstractPlot}, ::Attributes, ::Vector{Scene}, ::Vector{MakieScreen}, ::Observable{ColorTypes.RGBA{Float32}}, ::Observable{Bool}, ::SSAO, ::Vector{Makie.AbstractLight})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:118
  Scene(::Scene, ::Any; kw...)
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:345
  Scene(::Scene; events, px_area, clear, camera, camera_controls, transformation, kw...)
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:303

Stacktrace:
 [1] Scene(; px_area::Observable{GeometryBasics.HyperRectangle{2, Int64}}, events::Events, clear::Observable{Bool}, transform_func::Function, camera::typeof(cam3d!), camera_controls::EmptyCamera, transformation::Transformation, plots::Vector{AbstractPlot}, children::Vector{Scene}, current_screens::Vector{MakieScreen}, parent::Scene, visible::Observable{Bool}, ssao::SSAO, lights::Vector{PointLight}, theme::Attributes, theme_kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:253
 [2] Scene(parent::Scene; events::Events, px_area::Observable{GeometryBasics.HyperRectangle{2, Int64}}, clear::Bool, camera::Function, camera_controls::Makie.PixelCamera, transformation::Transformation, kw::Base.Pairs{Symbol, Vector{PointLight}, Tuple{Symbol}, NamedTuple{(:lights,), Tuple{Vector{PointLight}}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:318
 [3] Scene(parent::Scene, area::Observable{GeometryBasics.HyperRectangle{2, Int64}}; kw::Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol, Symbol}, NamedTuple{(:clear, :camera, :lights), Tuple{Bool, typeof(cam3d!), Vector{PointLight}}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\scenes.jl:346
 [4] initialize_block!(ls::LScene; scenekw::NamedTuple{(:lights,), Tuple{Vector{PointLight}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\makielayout\blocks\scene.jl:23
 [5] _block(::Type{LScene}, ::Figure; bbox::Nothing, kwargs::Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:show_axis, :scenekw), Tuple{Bool, NamedTuple{(:lights,), Tuple{Vector{PointLight}}}}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\makielayout\blocks.jl:374
 [6] _block(::Type{LScene}, ::GridPosition; kwargs::Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:show_axis, :scenekw), Tuple{Bool, NamedTuple{(:lights,), Tuple{Vector{PointLight}}}}}})
   @ Makie C:\Users\User\.julia\packages\Makie\DekzU\src\makielayout\blocks.jl:268
 [7] _block
   @ C:\Users\User\.julia\packages\Makie\DekzU\src\makielayout\blocks.jl:261 [inlined]
 [8] #_#1103
   @ C:\Users\User\.julia\packages\Makie\DekzU\src\makielayout\blocks.jl:253 [inlined]
 [9] top-level scope
   @ REPL[13]:5
```

since `typeof(lights) == Vector{PointLight}`. If instead `typeof(lights) == Vector{AbstractLight}`, it works:

```julia
julia> let
           fig = Figure();
           radiance = 10000
           lights = Vector{Makie.AbstractLight}([PointLight(Vec3f(0, 0, 20), RGBf(radiance, radiance, radiance))])
           ax = LScene(fig[1, 1]; show_axis = false, scenekw=(lights=lights,))
       end
LScene()
```

Code's taken from https://docs.makie.org/stable/documentation/backends/rprmakie/, on:

```julia
julia> versioninfo()
Julia Version 1.9.0
Commit 8e63055292 (2023-05-07 11:25 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 32 × 13th Gen Intel(R) Core(TM) i9-13900K
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, goldmont)
  Threads: 1 on 32 virtual cores

(jl_8jzM0Y) pkg> st
Status `C:\Users\User\AppData\Local\Temp\jl_8jzM0Y\Project.toml`
  [22d9f318] RPRMakie v0.5.5
```

### LINKED ISSUE #3238: Point light color is ignored in GLMakie
```julia
using GLMakie
using FileIO

GLMakie.closeall()

f = Figure()

ls = LScene(f[1, 1]; scenekw = (; lights = Makie.AbstractLight[PointLight(Point3f(100, 100, 100), RGBf(1, 0, 0))]))

mesh!(ls, brain, color = :white)

f
```

Gives this:

grafik

No hint of red to be seen. I've followed the tracks a bit and think the problem is that in `mesh.frag` we have

```glsl
void main(){
    vec4 color;
    // Should this be a mustache replace?
    if (fetch_pixel){
        color = get_pattern_color(image);
    }else{
        color = get_color(image, o_uv, color_norm, color_map, matcap);
    }
    {{light_calc}}
    write2framebuffer(color, o_id);
}
```

and in `surface.jl` we have

```julia
function light_calc(x::Bool)
    if x
        """
        vec3 L      = normalize(o_lightdir);
        vec3 N      = normalize(o_normal);
        vec3 light1 = blinnphong(N, o_camdir, L, color.rgb);
        vec3 light2 = blinnphong(N, o_camdir, -L, color.rgb);
        color       = vec4(ambient * color.rgb + light1 + backlight * light2, color.a);
        """
    else
        ""
    end
end
```

And then we have 

```glsl
vec3 blinnphong(vec3 N, vec3 V, vec3 L, vec3 color){
    float diff_coeff = max(dot(L, N), 0.0);

    // specular coefficient
    vec3 H = normalize(L + V);

    float spec_coeff = pow(max(dot(H, N), 0.0), shininess);
    if (diff_coeff <= 0.0 || isnan(spec_coeff))
        spec_coeff = 0.0;

    // final lighting model
    return vec3(
        diffuse * diff_coeff * color +
        specular * spec_coeff
    );
}
```

So `color` in `blinnphong` which to me should be the light color seems to be the texture color instead unless I'm misunderstanding.

### LINKED ISSUE #3239: Directional light as new default
Right now we don't have the light type directional light at all in GLMakie and WGLMakie. This is a light like the sun, like a point light at infinite distance so that all rays are parallel. This is nice because it's easy to light any object without worrying about anything but the light angle.

Therefore, I think we should add this light type to Makie and actually make it the default. Right now, the default is a point light that is placed at eye position. This is not great I think, because lighting from eye position means there aren't any visible shadow areas, which reduces the three-dimensionality of the lit surfaces. I understand that this choice was made because it was the only simple choice available. For any other position of a point light, one would probably need to take object positions and sizes into account. Not so with a directional light. A good default could be a light coming from top right of eye position.

I already tried tinkering with this a little bit and hit https://github.com/MakieOrg/Makie.jl/issues/3238 in the process. The glmakie light pipeline appeared much more hardcoded to one point and one ambient light to me than I had assumed prior given the `lights` Vector being able to hold an arbitrary number of arbitrary light types. So this might need a larger refactor. Or we add direction-ness as a boolean Switch to PointLight which would change how the Point3f it holds is interpreted, as a position or a direction.

ffreyer:
# Description

## Change Summary

This pr adds/changes:
- (breaking) `backlight` now affects `normal` instead of `lightdir`. This means specular reflections show up with backlight
- (breaking) changed default light direction (position), light intensities, diffuse and specular attributes 
- (minor breaking) change default light to `DirectionalLight` (closes #3239)
- (minor breaking) added smoothing of `diff_coeff` around 0. This softens the edge where normals are perpendicular to the light direction)
- (non-breaking change) `shading` attribute is now `:none/:fast/:verbose` instead of true or false
- (internal/maybe breaking) moved scaling in Axis3 from view to model matrix and stopped connecting scenes/plots with different spaces. The latter is something I want to do for #3013.
- (new) support for multiple light sources (need not be unique, ref #1592)
- (new) `PointLight` now has attenuation parameters (distance based intensity falloff, off by default)
- (new) added `DirectionalLight` (parallel light rays)
- (new) added `SpotLight` (light source illuminating a cone)
- (fix) the light color of `PointLight` (and all others) is now connected (fixes #3238)
- (internal) `colorbuffer()` now polls events to trigger `render_tick` so that it can cause lighting updates
- (internal) light calculation moved to world space which shouldn't change anything visually
- (internal) loosen type restriction on `lights` vector (fix #2985)

## Type of change

- [x] breaking (backlight, default light, smoothing around 0 intensity)

## TODO

- [x] Hook up updates (for individual lights and the lights array)
- [x] implement more simple light types (SpotLight, DirectionalLight, PointLight with attentuation)
- [x] Hook up `shading`
- [x] Update other backends to work with changes made to the lighting interface
- [x] Implement new lights in RPRMakie
- [x] lighting in volume 
- [x] backlight
- [x] record updates
- [x] check/fix missing normals for matcap with NoShading
- [x] fix WGLMakie
- [x] fix laggy directional light in WGLMakie (move camera-space light direction calculation to javascript)
- [x] fix Axis3
- [x] fix images in docs (MultiLightShading and SSAO fail to render correctly)


ffreyer:
I'm following the attenuation from learnopengl/ogre3d for now, which is effectively `1 / (1 + a * distance + b * distance^2)`. Example:
```julia
fig = Figure(backgroundcolor = :black)
sl = Slider(fig[1, 1], range = 0.05:0.05:5.0)

lights = Makie.AbstractLight[
    AmbientLight(RGBf(0.1, 0.1, 0.1)),
    PointLight(RGBf(1,0,0), Point3f(2, 2, 0), 1.0),
    PointLight(RGBf(0,1,0), Point3f(-2, 2, 0), 10.0),
    PointLight(RGBf(0,0,1), Point3f(-2, -2, 0), 100.0),
    PointLight(RGBf(1,1,1), Point3f(2, -2, 0), 10.0),
]
scene = LScene(
    fig[2, 1], show_axis = false,
    scenekw=(camera = cam3d!, lights = lights, backgroundcolor = :black, center = false),
)

p = mesh!(scene, Rect3f(Point3f(-10, -10, 0.01), Vec3f(20, 20, 0.02)), color = :white)
map(z -> translate!(p, 0, 0, -z), sl.value)

update_cam!(scene.scene, Vec3f(0, 0, 10), Vec3f(0, 0, 0), Vec3f(0, 1, 0))

idxs = vcat(1:length(sl.range[]), length(sl.range[])-1:-1:2)
record(fig, "attenuation.webm", idxs, framerate = 30) do idx
    sl.selected_index[] = idx
end
```
[attenuation.webm](https://github.com/MakieOrg/Makie.jl/assets/10947937/ebfff235-1823-4038-91b7-308be110ca09)

I also tried just `a / distance^3` as an attenuation factor beforehand, with a weak directional blue light above:
```julia
lights = [
    AmbientLight(RGBf(0.1, 0.1, 0.1)),
    PointLight(RGBf(0,1,0), Point3f(0, -3, 1)),
    PointLight(RGBf(1,0,0), Point3f(-3, 0, 1)),
    PointLight(RGBf(0,1,0), Point3f(0,  3, 1)),
    PointLight(RGBf(1,0,0), Point3f( 3, 0, 1)),
    Makie.DirectionalLight(RGBf(0, 0, 0.5), Vec3f(0, 0, 1)),
]
scene = Scene(camera = cam3d!, lights = lights)
mesh!(scene, Rect3f(Point3f(-1), Vec3f(2)), color = :white)
r = range(-10, 10, length=101)
surface!(scene, r, r, [0.5 * (sin(x+y) * sin(2x) * sin(3y)) - 1 for x in r, y in r], colormap = [:white, :white])
ps = [light.position[] for light in lights if light isa PointLight]
scatter!(scene, ps, color = :yellow, strokewidth = 1, strokecolor = :black)

screen = display(scene)
```
Screenshot from 2023-09-21 18-46-22

GLMakie/src/drawing_primitives.jl
```diff
@@ -2,6 +2,75 @@ using Makie: transform_func_obs, apply_transform
...
+# TODO: Maybe move this somewhere else?
+function handle_lights(attr::Dict, screen::Screen, lights::Vector{Makie.AbstractLight})
+    # TODO: make this adjustable?
+    max_lights = 64
+    max_params = 5 * 64
+
+    updater = screen.render_tick
+    MAX_PRIORITY = typemax(Int)
+
+    attr[:lights_length] = map(updater, priority = -1000) do _
```
ffreyer, line 14: Updating on `render_tick` doesn't work with `record`.

Updating (many) lights is kind of complicated in general, because we want to react to updates of each field for each light and also when lights are added or deleted to the lights vector.

Listening to each field individually would require attaching and removing listeners when the lights vector changes. Detecting that would either require some kind of Vector wrapper or an update on draw.

Instead of bothering with all of that I thought it would be easier to just check the lights vector once per frame and update it then. I'm using `render_tick` for that atm, but I guess we don't want that to update in `record` to avoid triggering events?

ffreyer:
<details>

<summary>Moving, colored `PointLight`s with different attenuation + white SpotLight.</summary>

```julia
begin
    fig = Figure(backgroundcolor = :black)
    sl = Slider(fig[1, 1], range = 0.05:0.05:5.0)

    # Prepare lights
    angle = Observable(0)
    angle2pos(phi) = Point3f(3cosd(phi), 3sind(phi), 0)
    lights = Makie.AbstractLight[
        AmbientLight(RGBf(0.1, 0.1, 0.1)),
        PointLight(RGBf(1,0,0), map(phi -> angle2pos(phi + 0), angle), 10.0),
        PointLight(RGBf(0,1,0), map(phi -> angle2pos(phi + 120), angle), 100.0),
        PointLight(RGBf(0,0,1), map(phi -> angle2pos(phi + 240), angle), 1000.0),
        SpotLight(RGBf(1,1,1), Point3f(0, 0, 1), Vec3f(0, 0, -1), pi/6),
    ]

    # Set scene
    scene = LScene(
        fig[2, 1], show_axis = false,
        scenekw=(camera = cam3d!, lights = lights, backgroundcolor = :black, center = false),
    )

    # floor
    p = mesh!(scene, Rect3f(Point3f(-10, -10, 0.01), Vec3f(20, 20, 0.02)), color = :white)
    map(z -> translate!(p, 0, 0, -z), sl.value)

    # Random balls
    N_balls = 2_000
    random_colors = !true
    meshscatter!(
        scene, [Point3f(20rand()-10, 20rand()-10, 10rand()-5) for _ in 1:N_balls],
        color = random_colors ? rand(RGBf, N_balls) : (:white)
    )

    # Visualize light source positions
    ps = Observable([light.position[] for light in lights if light isa Union{PointLight, SpotLight}])
    cs = [light.color[]    for light in lights if light isa Union{PointLight, SpotLight}]
    scatter!(scene, ps, color = cs, strokewidth = 1, strokecolor = :white)

    # Place camera above
    # update_cam!(scene.scene, Vec3f(0, 0, 10), Vec3f(0, 0, 0), Vec3f(0, 1, 0))

    # place camera at angle nearby
    update_cam!(scene.scene, Vec3f(8, 1, 4), Vec3f(0, 0, 0), Vec3f(0, 0, 1))

    # record
    idxs = vcat(1:length(sl.range[]), length(sl.range[])-1:-1:2)
    record(fig, "point_lights_and_spot_light.webm", idxs, framerate = 30) do idx
        sl.selected_index[] = idx
        angle[] = angle[] + 4
        ps[] = [light.position[] for light in lights if light isa Union{PointLight, SpotLight}]
        notify(fig.scene.current_screens[1].render_tick)
    end

    # interactive
    # display(fig)
    # @async while isopen(fig.scene)
    #     angle[] = angle[] + 2
    #     ps[] = [light.position[] for light in lights if light isa Union{PointLight, SpotLight}]
    #     sleep(1/60)
    # end
end
```

</details>

[point_lights_and_spot_light.webm](https://github.com/MakieOrg/Makie.jl/assets/10947937/80635fe4-d7ee-4dec-81b6-1cc3545bc80b)

ffreyer:
Added refimgs:

Ambient + SpotLights + PointLights:
Screenshot from 2023-09-26 14-57-39

DirectionalLights:
Screenshot from 2023-09-26 14-57-46

GLMakie/src/screen.jl
```diff
@@ -686,7 +687,7 @@ function Makie.colorbuffer(screen::Screen, format::Makie.ImageStorageFormat = Ma
     ctex = screen.framebuffer.buffers[:color]
     # polling may change window size, when its bigger than monitor!
     # we still need to poll though, to get all the newest events!
-    # GLFW.PollEvents()
+    pollevents(screen)
     # keep current buffer size to allows larger-than-window renders
     render_frame(screen, resize_buffers=false) # let it render
```
ffreyer, line 692: I added `pollevents()` here so lighting updates from `render_tick` can take effect before drawing the frame. Given the note I would assume this breaks output sizes with `save` and `record`, however I have no problems saving/recording at `resolution = (4000, 4000)`.

GLMakie/assets/shader/util.vert
```diff
@@ -239,30 +239,41 @@ out vec3 o_view_pos;
...
     // position in clip space (w/ depth)
     gl_Position = projection * view_pos;
     gl_Position.z += gl_Position.w * depth_shift;
-    // direction to light
-    o_lightdir = normalize(view*vec4(lightposition, 1.0) - view_pos).xyz;
+
+    // direction to light for :fast shading
+    // (normalizing this will result in incorrect lighting)
+    vec4 view_light_pos = view*vec4(lightposition, 1.0);
+    o_lightdir = view_light_pos.xyz / view_light_pos.w - view_pos.xyz;
```
ffreyer, line 267: As noted normalizing here results in incorrect light directions/perceived positions. To illustrate this, consider a light source shining on a rect viewed from the side:

Screenshot from 2023-09-27 16-27-14

If we **don't** normalize `o_lightdir` it will be a vector pointing from the light source to the vertex handled by the vertex shader. Since I placed the light source at (0, 0) here they'll match p1 and p2 in the example. On the way to the fragment shader `o_lightdir` will then get interpolated between different vertices. For example, halfway between the two vertices the vector will point to the  green "Vertex center". Normalizing in the fragment shader will then give correct light-to-vertex-center direction.

If we **do** normalize `o_lightdir` in the vertex shader as before, we will be interpolating the normalized `o_lightdir` vectors. In the sketch above those point to n1 (equal to p1) and n2. Doing the same half-way interpolation for vertices as above, we end up with a `o_lightdir` halfway between n1 and n2, which is marked as "Normalized center". This is not a vector passing through the vertex center position and thus the wrong light direction.

I'm pretty sure the test failure in `Order Independent transparency` is a result of this change. There are likely quite a few more failures that didn't get caught.

ffreyer:
Volume with multiple lights:

[volume.webm](https://github.com/MakieOrg/Makie.jl/assets/10947937/13fd974b-bc7b-4b50-8bfd-69d309baa027)

Volume plots don't differentiate inwards facing vs outwards facing surfaces, so lighting is a bit weird here. I'm not going to try to change this here though.

ffreyer:
Another example: 

<details>

<summary>
Meshscatter with `backlight = 1f0` synchronized with some PointLights moving through a transparent tube
</summary>

```julia
using LinearAlgebra, GeometryBasics

function tube(points, r, segments, closed = false)
    ps = sizehint!(Point3f[], segments * length(points))

    for i in eachindex(points)
        p = points[i]

        forward = Vec3f(0)
        if closed
            im = i > 1 ? i-1 : length(points)
            ip = i < length(points) ? i+1 : 1
            forward += normalize(p - points[im])
            forward += normalize(points[ip] - p)
        else
            if i > 1
                forward += normalize(p - points[i-1])
            end
            if i < length(points)
                forward += normalize(points[i+1] - p)
            end
        end
        forward = normalize(forward)

        # assuming dir != u_z
        right = cross(forward, Vec3f(0,0,1))
        up = cross(right, forward)

        # gen points
        for phi in range(0, 2pi, length = segments+1)[1:end-1]
            push!(ps, p + r * (cos(phi) * right + sin(phi) * up))
        end
    end

    fs = QuadFace[]
    for i in 0 : segments : segments * (length(points) - 2)
        for j in 1:segments
            f = QuadFace(
                i + j,
                i + mod1(j+1, segments),
                i + mod1(j+1, segments) + segments,
                i + j + segments,
            )
            push!(fs, f)
        end
    end
    if closed
        i = segments * (length(points) - 1)
        for j in 1:segments
            f = QuadFace(
                i + j,
                i + mod1(j+1, segments),
                mod1(j+1, segments),
                j,
            )
            push!(fs, f)
        end
    end

    return normal_mesh(ps, fs)
end

begin
    ps = [Point3f(cos(phi), sin(2phi), 0.2sin(phi)) for phi in range(0, 2pi, length = 101)[1:100]]
    m = tube(ps, 0.1, 8, true)

    idx = Observable(1)
    lights = Makie.AbstractLight[
        AmbientLight(RGBf(0.1, 0.1, 0.1)),
        PointLight(RGBf(1, 0, 0),   map(i -> ps[i], idx), 1.0),
        PointLight(RGBf(1, 1, 0),   map(i -> ps[mod1(i+10, length(ps))], idx), 1.0),
        PointLight(RGBf(0, 0.2, 1), map(i -> ps[mod1(i+4,  length(ps))], idx), 1.0),
        PointLight(RGBf(0, 0.8, 0), map(i -> ps[mod1(i+37, length(ps))], idx), 1.0),
        PointLight(RGBf(0, 0.8, 1), map(i -> ps[mod1(i+88, length(ps))], idx), 1.0),
    ]

    fig = Figure(backgroundcolor = :black)
    ax = LScene(fig[1, 1], scenekw = (lights = lights, camera = cam3d!), show_axis = false)
    mesh!(ax, m, color = (:white, 0.4), transparency = true, shading = :verbose)

    pos = map(idx) do i
        Point3f[
            ps[i],
            ps[mod1(i+10, length(ps))],
            ps[mod1(i+4,  length(ps))],
            ps[mod1(i+37, length(ps))],
            ps[mod1(i+88, length(ps))],
        ]
    end
    meshscatter!(ax, pos, markersize = 0.05, shading = :verbose, backlight = 1f0, color = :white) # , marker = Rect3f(Point3f(-1), Vec3f(1)))

    # display(fig)
    # @async while isopen(fig.scene)
    #     idx[] = mod1(idx[] + 1, 100)
    #     sleep(1/30)
    # end

    idxs = 1:100
    record(fig, "race.webm", idxs, framerate = 30) do _
        idx[] = mod1(idx[] + 1, 100)
    end
end
```
</details>

[race.webm](https://github.com/MakieOrg/Makie.jl/assets/10947937/254b62e5-200e-4411-9bf4-11d5d5cdf386)

GLMakie/assets/shader/util.vert
ffreyer: I switched to doing lighting in world space so this is not really relevant to lightdir anymore. Though it probably still applies to camdir. (World vs view/camera space shouldn't matter for lighting. It was using view/camera space because SSAO needed it and lighting could reuse it. Merging lighting between volume and the rest of Makie requires world space though.)

ffreyer:
Did a little bit of performance testing: (first point with verbose lighting has the same setup as fast lighting)
Screenshot from 2023-09-29 14-26-43
That seems like pretty much the expected scaling to me. A bit more up-front cost from having a more complex shader and then roughly linear scaling with the number of lights. The frames are taken with calls to `colorbuffer` on an invisible screen. (Much less overhead than calling record)

Some questions for review:
- I kept `shading = :fast` (old algorithm) as the default. Do we want to switch to `:verbose`?
    - Switching would be a bit slower but maybe not that much?
    - As far as I understand each *element* of the arrays carrying light data counts as one uniform component. I have 16k of these, but I don't know if this is device dependent. If it is that might cause issues.
    - Switching to verbose by default would mean we can do #3239
- Are the `shading` option names good? (none, fast, verbose)

SimonDanisch:
I dont like the names (:verbose, :none, :fast), and I think would prefer an enum, but I guess that's a yak we can shave once this is ready ;)
For defaults, I've come more and more to the conviction that we should go for quality first with optional fastpaths, so I think we should make the good shading the default ;)

SimonDanisch:
Ideally, the user doesn't have to specify the shading mode and things "just work" when adding different light configurations.
I think this could be an option as well, no?

ffreyer:
The "just works" option would be to use the multi-light algorithm by default.

> GL_MAX_FRAGMENT_UNIFORM_COMPONENTS
>    data returns one value, the maximum number of individual floating-point, integer, or boolean values that can be held in uniform variable storage for a fragment shader. The value must be at least 1024. 

Ok, maybe I was too worried about uniform limits. This pr should be adding 576 components with default settings, which is safely below the minimum. 

> I dont like the names (:verbose, :none, :fast), and I think would prefer an enum, but I guess that's a yak we can shave once this is ready ;)

This is at the point where I'm doing this kind of stuff. Maybe
```julia
@enum begin
    NoShading
    FastShading
    MultiLightShading
end
```
then?

SimonDanisch:
>The "just works" option would be to use the multi-light algorithm by default.

No I meant something like:
```julia
lights = [] # NoShading
lights = [one_simple_light] # FastShading
lights = [many_lights...] # MultiLightShading
```
Then you wouldn't need to define the lighting in two places ;)

jkrumbiegel:
cool, I should test the directional light with treemaps for example

ffreyer:
If we do decide to switch to a DirectionalLight as a default we should also adjust `shading = :fast` (as well as WGLMakie and CairoMakie) to use that.

Personally I don't have a strong opinion on DirectionalLight vs PointLight as a default. I could see having a static light be helpful to tell viewing angles and DirectionalLight giving a bit more of a "illuminated by the sun" kind of feel, but it also allows for the area you look at directly to be dark. So I guess it's a bit less helpful for reading off data (which you probably shouldn't try with 3D plots anyway) while giving a bit more clarity on orientation and motion?

ffreyer:
I have no idea how to update the generated javascript files in WGLMakie so that still needs updates. Other than that GLMakie and CairoMakie should be ok, though there are now more test failures because I made the default light static (+ DirectionalLight).

<details>

<summary>
For playing around with defaults
</summary>

```julia
using GLMakie, FileIO, LinearAlgebra, GeometryBasics

begin
    fig = Figure(resolution = (800, 800))

    Label(fig[1, 1:2], "Light Intensities", tellwidth = false)
    ambient = Slider(fig[2, 1], range = 0.0:0.01:1.0, startvalue = 0.55).value
    Label(fig[2, 2], map(x -> "ambient = $x", ambient), width = 100)
    dir_int = Slider(fig[3, 1], range = 0.0:0.01:2.0, startvalue = 1.0).value
    Label(fig[3, 2], map(x -> "directional = $x", dir_int), width = 100)

    Label(fig[1, 3:4], "Light \"Position\"", tellwidth = false)
    phi = Slider(fig[2, 3,], range = range(0, 360, length=361), startvalue = 45).value
    Label(fig[2, 4], map(x -> "phi = $(x)°", phi), width = 100)
    theta = Slider(fig[3, 3], range = range(-90, 90, 181), startvalue = 30).value
    Label(fig[3, 4], map(x -> "theta = $(x)°", theta), width = 100)

    Label(fig[4, 1], "Show Light Ray", tellwidth = false)
    vis = Toggle(fig[4, 2]).active
    Label(fig[4, 3], "Relative to cam", tellwidth = false)
    relative = Toggle(fig[4, 4]).active

    light_dir = Observable(Vec3f(1))

    lights = [
        AmbientLight(map(x -> RGBf(x, x, x), ambient)),
        DirectionalLight(map(x -> RGBf(x, x, x), dir_int), light_dir)
    ]

    earth = FileIO.load(Makie.assetpath("earth.png"))
    sphere = GeometryBasics.mesh(Tesselation(Sphere(Point3f(0), 1f0), 540))

    # Sphere new
    Label(fig[5, 1:2], "New")
    ax = LScene(fig[6, 1:2], show_axis = true, scenekw = (lights = lights, center = false))
    m = mesh!(ax, sphere, color = :white)
    # m = mesh!(ax, sphere, color = :white, specular = Vec3f(0))
    center!(ax.scene)

    onany(phi, theta, relative, ax.scene.camera.view) do phi, theta, rel, _
        cp = cosd(phi); sp = sind(phi)
        ct = cosd(theta); st = sind(theta)
        if rel
            cc = cameracontrols(ax.scene)
            viewdir = normalize(cc.lookat[] - cc.eyeposition[])
            up = normalize(cc.upvector[])
            right = normalize(cross(viewdir, up))
            dir = ct * cp * viewdir - ct * sp * right - st * up
        else
            dir = -Vec3f(ct * cp, ct * sp, st)
        end
        light_dir[] = normalize(dir)
    end

    linesegments!(ax,
        map(dir -> Point3f[Point3f(0), -2dir], light_dir),
        color = :yellow, visible = vis
    )

    # Earth new
    light_dir2 = Observable(Vec3f(1))
    lights2 = [
        AmbientLight(map(x -> RGBf(x, x, x), ambient)),
        DirectionalLight(map(x -> RGBf(x, x, x), dir_int), light_dir2)
    ]
    ax2 = LScene(fig[7, 1:2], show_axis = true, scenekw = (lights = lights2, center = false))
    mesh!(ax2, Sphere(Point3f(0), 1f0), color = earth)
    center!(ax2.scene)
    onany(phi, theta, relative, ax2.scene.camera.view) do phi, theta, rel, _
        cp = cosd(phi); sp = sind(phi)
        ct = cosd(theta); st = sind(theta)
        if rel
            cc = cameracontrols(ax2.scene)
            viewdir = normalize(cc.lookat[] - cc.eyeposition[])
            up = normalize(cc.upvector[])
            right = normalize(cross(viewdir, up))
            dir = ct * cp * viewdir - ct * sp * right - st * up
        else
            dir = -Vec3f(ct * cp, ct * sp, st)
        end
        light_dir2[] = normalize(dir)
    end
    linesegments!(ax2,
        map(dir -> Point3f[Point3f(0), -2dir], light_dir2),
        color = :yellow, visible = vis
    )


    # old sphere
    Label(fig[5, 3:4], "Old")
    pos = Observable(Point3f(0))
    old_lights = [
        AmbientLight(RGBf(0.55, 0.55, 0.55)),
        PointLight(RGBf(1, 1, 1), pos)
    ]
    ax3 = LScene(fig[6, 3:4], show_axis = true, scenekw = (lights = old_lights,))
    m = mesh!(ax3, sphere, color = :white)
    # m = mesh!(ax2, sphere, color = :white, specular = Vec3f(0))
    cc = ax3.scene.camera_controls
    map!(identity, pos, cc.eyeposition)
    update_cam!(ax3.scene, cc.eyeposition[], cc.lookat[])

    # old earth
    pos = Observable(Point3f(0))
    old_lights = [
        AmbientLight(RGBf(0.55, 0.55, 0.55)),
        PointLight(RGBf(1, 1, 1), pos)
    ]
    ax4 = LScene(fig[7, 3:4], show_axis = true, scenekw = (lights = old_lights,))
    mesh!(ax4, Sphere(Point3f(0), 1f0), color = earth)
    cc = ax4.scene.camera_controls
    map!(identity, pos, cc.eyeposition)
    update_cam!(ax4.scene, cc.eyeposition[], cc.lookat[])

    fig
end
```

</details>

jkrumbiegel:
very cool visualizer :) something is not quite correct with the specular highlight:

grafik

see that the specular highlight stays in front even if the directional light moves to the side? this causes a visible cutoff

jkrumbiegel:
Something like this looks much better to me than the current state. And I don't find it disorienting to rotate the brain ;) the sphere looks a bit weird of course

grafik

jkrumbiegel:
@ffreyer I still don't think it's right, shouldn't the specular reflection move with the light? now only the intensity changes. not sure about this, especially with a directional light

ffreyer:
If you lock the light position to the camera it should not move, because the light-to-vertex and camera-to-vertex directions have a constant angle between them. If you keep the light direction constant it should and for me it does.

ffreyer:
I added some smoothing to the `diff_coeff` because the edge that appears where the light direction and normals become perpendicular annoys me.
no_smooth

This comes from `diff_coeff ~ dot(lightdir, normal) = cos(angle)`, which is a linear function near 90° that ends abruptly as it hits 0. The smoothing function I added smooths that part out: (with an exponent as a parameter for how weak the smoothing is)
Screenshot from 2023-10-04 17-17-29

| No smoothing | Exponent 16 | Exponent 8 (current) |
| --- | --- | --- |
| no_smooth | pow16_smooth | pow8_smooth |

jkrumbiegel:
Nice now it looks quite good, I restarted the session, maybe the change with the specular highlight was only picked up after. Now it moves as I expect. This setting looks somewhat nice:

grafik

ffreyer:
Yea changes to shaders require a `GLMakie.closeall()` or session restart to apply.

I'm wondering as well if we should set `diffuse = Vec3f(1)` by default. Currently it's 0.4, which makes the brightness values kind of magical. E.g. 0.33 + 1.15 (for direct light) is somehow not full white, because it's actually 0.33 + 0.4 * 1.15 = 0.79. With `diffuse = 1` and `specular = 0.5` (scaled by the same factor from 0.2) you'd get the same result with `ambient = 0.33` and `directional_intensity = 0.46`, but now you can just add the intensities to get the maximum diffuse brightness.

jkrumbiegel:
Yeah I was wondering about that as well. I think not limiting diffuse by default sounds very reasonable and easier to predict.

ffreyer:
Alright, I made those changes but nudged specular down a little and the light intensities up a little.
Screenshot from 2023-10-04 18-57-35

lazarusA:
I cannot wait for this to be merged 🤯 , what's the plan? The TODO list is almos done 😄 .

ffreyer:
Yea, the only thing that's missing is some WGLMakie updates for the switch from PointLight to DirectionalLight. Since this includes a couple of breaking changes it would go into the 0.20 beta branch. I'm not sure what the exact state of that branch is. I think we all want to get it finished and merged but there are at least the WGLMakie ci failures that need fixing.

ffreyer:
@SimonDanisch I moved the calculation of camera relative light directions to javascript. It seems to work fine to me, but it would be good if you took a closer look. I'm pretty unsure of what I'm doing there and just kind of tried till it worked.

---

In general I'm still a bit unsure about defaults. Quite a few refimgs look darker now, probably mostly because of the angled light direction. In some cases that's helpful, I think, because it helps accentuate curvature

| old | new |
| --- | --- |
| Screenshot from 2023-10-11 18-35-06 | Screenshot from 2023-10-11 18-35-12 |
| Screenshot from 2023-10-11 18-41-57 | Screenshot from 2023-10-11 18-42-00 |
| Screenshot from 2023-10-11 18-51-17 | Screenshot from 2023-10-11 18-51-20 |
| Screenshot from 2023-10-11 18-49-12 | Screenshot from 2023-10-11 18-49-15 |

Some others ended up just looking darker or less detailed. For example:

| old | new |
| --- | --- |
| Screenshot from 2023-10-11 18-38-57 | Screenshot from 2023-10-11 18-38-46 |
| Screenshot from 2023-10-11 18-47-00 | Screenshot from 2023-10-11 18-47-04 |
|Screenshot from 2023-10-11 18-54-18 | Screenshot from 2023-10-11 18-54-22 |

jkrumbiegel:
Maybe somethings not right with Axis3. You know how I'm doing this unusual view and projection matrix, maybe that throws off the math so that the light angle is not right anymore?

ffreyer:
Hmm, moving scale_matrix to `model` means that is also applies with `space != :data` which is kind of annoying

src/interfaces.jl
```diff
@@ -228,11 +228,14 @@ function connect_plot!(scene::SceneLike, plot::Combined{F}) where {F}
             transform!(t, t_user)
             plot.transformation = t
         end
-        connect!(transformation(scene), transformation(plot))
+        if is_space_compatible(plot, parent)
+            connect!(transformation(parent), transformation(plot))
+        end
```
ffreyer, line 233: This only connects transformations of plots and parent plots/scenes if they have compatible spaces. For plots this checks `get(plot, :space, :data)` and for scenes it will always assume `:data` space to be compatible, with for example `:pixel` space in addition with a pixel camera.

jkrumbiegel:
With the reversed axes, the lighting direction obviously changes. I don't think that's good in this camera-relative mode. I would say the light should always come from some angle relative to the main camera axis, and if you pass a direction vector manually it should work in data coordinates, so be affected by scaling. Otherwise it's kind of hard to reason about what it means.

ffreyer:
> With the reversed axes, the lighting direction obviously changes. I don't think that's good in this camera-relative mode. I would say the light should always come from some angle relative to the main camera axis, and if you pass a direction vector manually it should work in data coordinates, so be affected by scaling. Otherwise it's kind of hard to reason about what it means.

With the changes the reversal also moved to the model matrix, so the light direction actually doesn't change. What changes are the normals generated by a surface. I.e. with a z reversal they'll point downwards and not interact with the light coming from above. With `backlight = 1f0` the light will interact with -normal too and you get:

Screenshot from 2023-10-12 20-55-36

We could also make that a default, for surface...

ffreyer:
Fixed this (even though it should be unrelated to this pr)
Screenshot from 2023-10-27 14-45-18

ffreyer:
Seems like this too :shrug: 
Screenshot from 2023-10-27 14-51-51

There also seem to be some issues with line (and maybe scatter) depth order:
Screenshot from 2023-10-27 14-44-42

ffreyer:
As far as I can tell the only issue left here that isn't also on #3113 is with "contour labels 3D". With Axis3 scaling being moved to `model` the text is now being scaled, which makes it tiny. (You can see it with fontsize ~ 300.) The gaps are also too small I think. This requires some changes to `transform_marker`/`scale_primitive` to be fixed, and probably text bounding boxes

PR #3261, opened 2023-09-27
replace bgcolor with backgroundcolor in Legend

### LINKED ISSUE #2101: Inconsistent kwarg names `backgroundcolor` (for `Figure`, `Axis`) and `bgcolor` (for `Legend`)
The keyword to change the background color of a `Legend` is `bgcolor`:

https://makie.juliaplots.org/v0.17.8/api/index.html#Legend

Whereas the keyword to change the background color of a `Axis` or `Figure` is `backgroundcolor`:

https://makie.juliaplots.org/v0.17.8/api/index.html#Axis

It would be nice if these were consistent.

JonasIsensee:
# Description

Fixes #2101 

This PR renames the kwarg `bgcolor` in `Legend` to `backgroundcolor` to be consistent with usage in `Axis` and `Figure`.

## Type of change

- Breaking but with deprecation warning ?

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ with_theme(palette = (; patchcolor = cgrad(cmap, alpha=0.45))) do
     band!(x, sin.(x), approx .+= x .^ 5 / 120; label = L"n = 2")
     band!(x, sin.(x), approx .+= -x .^ 7 / 5040; label = L"n = 3")
     limits!(-3.8, 3.8, -1.5, 1.5)
-    axislegend(; position = :ct, bgcolor = (:white, 0.75), framecolor = :orange)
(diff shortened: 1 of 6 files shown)


jkrumbiegel:
Good idea, can we make this a deprecation instead? For a while we could have both and warn that the other should be used? Or maybe we group this into 0.20 which is already breaking anyway.

JonasIsensee:
Keeping a deprecation for a while shouldn't do any harm.

JonasIsensee:
So, I tried to fix the deprecation again.
I'm absolutely not sure if this is the way you would like it to be...

SimonDanisch (approved):
I like this :) I've stumbling over this myself for quite a few times^^

PR #3280, opened 2023-10-06
Fix WGLMakie's README code snippet for JSServe

DanielVandH:
It didn't work before. Needed to add the missing `do`, and `Pyramid` is not exported so it required `WGLMakie.Pyramid`.


SimonDanisch:
Thanks!

PR #3281, opened 2023-10-06
Blockspec

SimonDanisch:
Similar to [PlotSpec](https://github.com/MakieOrg/Makie.jl/pull/2868), just for having `Observable{Figure}`

* Fixes lots of memory leaks from `delete!(block)` and `delete!(ax/scene, plot)`, especially for WGLMakie
* Also fixes large performance bugs with creating plots (e.g. `bbox(::BezierPath)` is really expensive and was run many times for `markersize=large_array`. See https://github.com/MakieOrg/Makie.jl/issues/3307
* Adds docs & tests for PlotSpec & BlockSpec
* Fixes previous compile time regressions and found a few improvements while searching for regressions
* Refactors plot creation API a bit, to make things like `convert_arguments(...)::FigureSpec` easier to integrate and removes double `convert_arguments` from previous implementation. This should also make https://github.com/MakieOrg/Makie.jl/pull/3226 easier!
* Moves cycler from Axis to Scene, so that Recipes and PlotSpec work with cycling, and cycling is applied more consistently
* Moves update tracking for the GLMakie on_demand_renderloop to conversion pipeline, so that much less observables for tracking are created (more than halfs the size of objects created for each plot)
* Lays the groundwork for a very lightweight and interactive AlgebraOfGraphics

Examples from the docs:
```julia
using GLMakie, DelimitedFiles, FileIO
import Makie.SpecApi as S
GLMakie.activate!() # hide
volcano = readdlm(Makie.assetpath("volcano.csv"), ',', Float64)
r = LinRange(-1, 1, 100)
cube = [(x .^ 2 + y .^ 2 + z .^ 2) for x = r, y = r, z = r]

ax1 = S.Axis(; title="Axis 1", plots=map(x -> S.density(x * randn(200) .+ 3x, color=:y), 1:5))
ax2 = S.Axis(; title="Axis 2", plots=[S.contourf(volcano; colormap=:inferno)])
ax3 = S.Axis3(; title="Axis3", plots=[S.mesh(brain, colormap=:Spectral, color=[tri[1][2] for tri in brain for i in 1:3])])
ax4 = S.Axis3(; plots=[S.contour(cube, alpha=0.5)])


spec_array = S.Figure([ax1, ax2]);
spec_matrix = S.Figure([ax1 ax2; ax3 ax4]);
f = Figure(; resolution=(1000, 500))
plot(f[1, 1], spec_array)
plot(f[1, 2], spec_matrix)
f
```
image

```julia
import Makie.SpecApi as S
struct PlotGrid
    nplots::Tuple{Int,Int}
end

Makie.used_attributes(::Type{<:AbstractPlot}, ::PlotGrid) = (:color,)
function Makie.convert_arguments(::Type{<:AbstractPlot}, obj::PlotGrid; color=:black)
    f = S.Figure(; fontsize=30)
    for i in 1:obj.nplots[1]
        for j in 1:obj.nplots[2]
            ax = S.Axis(f[i, j])
            S.lines!(ax, cumsum(randn(1000)); color=color)
        end
    end
    return f
end

f = Figure()
plot(f[1, 1], PlotGrid((1, 1)); color=Cycled(1))
plot(f[1, 2], PlotGrid((2, 2)); color=Cycled(2))
f
```
image

```julia
struct LineScatter
    show_lines::Bool
    show_scatter::Bool
    kw::Dict{Symbol,Any}
end
LineScatter(lines, scatter; kw...) = LineScatter(lines, scatter, Dict{Symbol,Any}(kw))

function Makie.convert_arguments(::Type{<:AbstractPlot}, obj::LineScatter, data...)
    plots = PlotSpec[]
    if obj.show_lines
        push!(plots, S.lines(data...; obj.kw...))
    end
    if obj.show_scatter
        push!(plots, S.scatter(data...; obj.kw...))
    end
    return plots
end

f = Figure()
ax = Axis(f[1, 1])
# Can be plotted into Axis, since it doesn't create its own axes like FigureSpec
plot!(ax, LineScatter(true, true; markersize=20, color=1:4), 1:4)
plot!(ax, LineScatter(true, false; color=:darkcyan, linewidth=3), 2:4)
f
```
image

```julia
struct MySimulation
    plottype::Symbol
    arguments::AbstractVector
end

function Makie.convert_arguments(::Type{<:AbstractPlot}, sim::MySimulation)
    return map(enumerate(sim.arguments)) do (i, data)
        return PlotSpec(sim.plottype, data)
    end
end
f = Figure()
s = Slider(f[1, 1], range=1:10)
m = Menu(f[1, 2], options=[:scatter, :lines, :barplot])
sim = lift(s.value, m.selection) do n_plots, p
    args = [cumsum(randn(100)) for i in 1:n_plots]
    return MySimulation(p, args)
end

ax, pl = plot(f[2, :], sim)
on(sim; priority=-1) do s
    autolimits!(ax)
end
tight_ticklabel_spacing!(ax)
f
```

blockspec2


GLMakie/src/drawing_primitives.jl
```diff
@@ -196,12 +202,13 @@ function handle_view(array::Observable{T}, attributes) where T <: SubArray
...
+    return lift_convert_inner(value, Key{key}(), Key{Makie.plotkey(plot)}(), plot, screen)
 end
 
-function lift_convert_inner(value, key, plot_key, plot)
+function lift_convert_inner(value, key, plot_key, plot, screen)
     return lift(plot, value) do value
+        screen.requires_update = true
         return convert_attribute(value, key, plot_key)
     end
 end
```
ffreyer, line 214: As far as I can tell all plot update tracking comes from this now? In that case I think it misses updates to camera matrices and lighting, because those get added to a render object from scene data. Though those should probably also be handled in screen.jl?

SimonDanisch: Good point, didn't notice that, since enough is usually going on that it seems to be enough for most plots to rely on those updates...
Yes, I think we should add that here:
https://github.com/MakieOrg/Makie.jl/blob/master/GLMakie/src/screen.jl#L438

src/bezier.jl
```diff
@@ -1,37 +1,39 @@
 using StableHashTraits
 
+const Point2d = Point2{Float64}
```
ffreyer, line 3: This should be in Makie.jl (or maybe types.jl) for better visibility, imo. Maybe also with a 3d and 4d variant, since we'll likely need them if we switch to Float64 for internals in the future.

SimonDanisch: It should actually go into GeometryBasics... Can still do that once we do the refactor, no?

SimonDanisch:
A few Benchmarks:
```julia
using GLMakie, WGLMakie

function test(ax, last_pl, n=500)
    for i in 1:n
        delete!(ax, last_pl)
        last_pl = scatter!(ax, rand(1000); color=rand(1000), markersize=10, colormap=:viridis)
    end
end

f, ax, pl = scatter(rand(1000); color=rand(1000), markersize=10, colormap=:viridis)
fsize = Base.summarysize(f) / 10^6
@time test(ax, pl, 500)

function test2(f, ax, n=100)
    for i in 1:n
        delete!(ax)
        ax, pl = scatter(f[1, 1], rand(1000); color=rand(1000), markersize=10, colormap=:viridis)
        yield()
    end
end
f, ax, pl = scatter(rand(1000); color=rand(1000), markersize=10, colormap=:viridis)
fsize = Base.summarysize(f) / 10^6
@time test2(f, ax)
fsize2 = Base.summarysize(f) / 10^6
```

PR:
test1:
GLMakie:0.6s, 0.475mb, 0.475mb
WGLMakie:3.5s, 7.44mb, 7.45mb

test2:
GLMakie: 2.4s, 8.81mb, 12.1mb
WGLMakie:9.0s, 8.3mb, 9.1mb

Tagged:
test1:
GLMakie:2.5s, 9.5mb, 20.4mb
WGLMakie:3.5s, 7.2mb, 17.7mb

test2:
GLMakie: 3.37s, 19mb, 37mb
WGLMakie:9.3s, 7.5mb, 39mb

PR #3283, opened 2023-10-10
correctly construct Attributes in get_axis

hexaeder:
# Description

Fixes https://github.com/MakieOrg/GraphMakie.jl/issues/142

During `get_axis`, the Attributes for the plot recipe are constructed in a "wrong" way, in which nested attributes do not work as expected. I.e. named tuples in the Attributes become `Observable{Any}` instead of nested `Attributes` objects.

Was introduced in Makie@0.19.11 as part of #3275

MWE:
```julia
using Makie
@recipe(MyPlot, p) do scene
   Attributes(
       nt = (;)
   )
end
function Makie.plot!(mp::MyPlot)
    p = mp[:p]
    @show mp[:not] # see how it looks like in the recipe
    scatter!(mp, p)
end
```
```julia
julia> fig, a, p = myplot(rand(10); nt=(;named=:tuple))
mp[:nt] = Observable{Any}((named = :tuple,)) # "warmup" run
mp[:nt] = Attributes with 1 entry:           # actual run
  named => tuple
```

Not sure if the splat of the dict is the best way to solve this, suggestions welcome. On a different note: it might be nice to pass another attribute, such as `_warmup=true` to the recipe. `GraphMakie` could skip the (possibly expensive) layout algorithm in the warmup run...

## Type of change

## Checklist

### DIFF
--- a/src/figureplotting.jl
+++ b/src/figureplotting.jl
@@ -40,7 +40,7 @@ function get_axis(fig, P, axis_kw::Dict, plot_attr, plot_args)
         delete!(plot_attr, :show_axis)
         delete!(plot_attr, :limits)
         delete!(plot_attr, :color) # Color may contain Cycled(1), which needs the axis to get resolved to a color
-        plot!(proxyscene, P, Attributes(plot_attr), plot_args...)
+        plot!(proxyscene, P, Attributes(;plot_attr...), plot_args...)
         if is2d(proxyscene)
             ax = Axis(fig; axis_kw...)
         else


SimonDanisch:
Oh, that's an annoying corner case... 
Also, those constructors should be equivalent, but guess they aren't -.-

> On a different note: it might be nice to pass another attribute, such as _warmup=true to the recipe

We'll remove the whole proxy_scene business in #3113, so no need for a hacky workaround ;)

hexaeder:
> Also, those constructors should be equivalent, but guess they aren't -.-

Well alternatively, one could define an outer constructor
```julia
Attributes(d::Dict) = Attributes(Dict{Symbol, Observable}(node_pairs(d)))
```
to `MakieCore/src/attributes.jl`, such that the conversion happens explicitly pairwise instead of implicitly in the default inner constructor of `Attributes`.

SimonDanisch:
Fixed in https://github.com/MakieOrg/Makie.jl/pull/3284

PR #3284, opened 2023-10-10
fix constructor for nested attributes

hexaeder:
alternative to #3283


SimonDanisch:
Thanks!

mtfishman:
Thanks for the quick fix! Can you register a new version so we can make use of this, since it is causing issues when using `GraphMakie` (https://github.com/MakieOrg/GraphMakie.jl/issues/142)?

PR #3285, opened 2023-10-10
add timings + used memory to tests

SimonDanisch:


ffreyer:
Would be nice to have some colors in this so slow/expensive tests are easy to spot. Just a some simple thresholds like `if > 10s; color = :red` would be enough for that I think

SimonDanisch:
I think we should just make a nice report at some point, like the compile time benchmark, maybe even with a plot!
If it stays in the CI, almost nobody will look at it anyways.
I dont really have time for that now though, so I think it will stay like this for now :(

PR #3288, opened 2023-10-12
Test against earlier releases

SimonDanisch:
The sudden WGLMakie CI failure is really weird, since nothing important should have changed, its not throwing an error and can't be reproduced locally.
This pr will test some earlier versions of Makie, to make sure its not just some unfortunate change in Makie itself.


SimonDanisch:
That ain't it...

PR #3293, opened 2023-10-13
Improve default near/far

ffreyer:
# Description

In the original camera3d pr I changed near/far to be multiplied by `norm(eyeposition - lookat)` and set the default far to be very large to avoid clipping issues. This is problematic for OIT where depth is used to mix colors because without zooming plots occupy a very small depth range.

To summarize the issues with clipping:
- when zooming in (i.e. looking at details) a static `near` value will eventually become large relative to the limits one looks at. This will cause data to be clipped starting from the camera
- when zooming out a static `far` value will eventually clip the far side of the plot. This usually starts with the axis frame/grid disappearing

This pr makes static near/far values available again and adds a new (default) scaling mode which scales  `far` so that `far = 1` always ends up behind the scene bounding sphere, and calculates `near` to be just in front of it, but no closer than `norm(lookat - eyeposition) * near` (to avoid negative near). This continues to solve the clipping issues as `near` effectively scales with zooming, `far` is always outside/behind the region containing data, and it should improve OIT again by keeping the depth scale close to the minimum necessary.

### Camera setup from the OIT refimg test (scene area not matched exactly)

<details>

<summary> Code </summary>

```julia
begin
    using GLMakie
    using GeometryBasics, LinearAlgebra

    function frustum_snapshot(cam)
        r = Rect3f(Point3f(-1, -1, -1), Vec3f(2, 2, 2))
        rect_ps = coordinates(r) .|> Point3f
        insert!(rect_ps, 13, Point3f(1, -1, 1)) # fix bad line

        inv_pv = inv(cam.projectionview[])
        return map(rect_ps) do p
            p = inv_pv * to_ndim(Point4f, p, 1)
            return p[Vec(1,2,3)] / p[4]
        end
    end


    ex = Point3f(1,0,0)
    ey = Point3f(0,1,0)
    ez = Point3f(0,0,1)
end

begin
    fig = Figure()
    ax = LScene(fig[1, 1])
    cam3d!(ax.scene)
    r = Rect2f(-1, -1, 2, 2)
    for x in (0, 1)
        for (i, a) in enumerate((0.25, 0.5, 0.75, 1.0))
            ps = [Point3f(a, (0.15 + 0.01y)*(2x-1) , 0.2y) for y in 1:8]
            if x == 0
                cs = [RGBAf(1, 0, 0, 0.75), RGBAf(0, 1, 0, 0.5), RGBAf(0, 0, 1, 0.25)]
            elseif x == 1
                cs = [RGBAf(1, x, 0, a), RGBAf(0, 1, x, a), RGBAf(x, 0, 1, a)]
            end
            idxs = [1, 2, 3, 2, 1, 3, 1, 2, 1, 2, 3][i:7+i]
            meshscatter!(
                ax, ps, marker = r,
                color = cs[idxs], transparency = true
            )
        end
    end
    cc = cameracontrols(ax.scene)
    cc.fov[] = 22f0
    update_cam!(ax.scene, cc, Vec3f(0.625, 0, 3.5), Vec3f(0.625, 0, 0), Vec3f(0, 1, 0))

    scene = ax.scene
    cam = scene.camera
    eyeposition = cc.eyeposition
    lookat = cc.lookat
    frustum = map(pv -> frustum_snapshot(cam), cam.projectionview)

    scene2 = LScene(fig[1, 2])
    _cc = Makie.Camera3D(scene2.scene, projectiontype = Makie.Orthographic)
    lines!(scene2, frustum, color = :blue, linestyle = :dot)
    scatter!(scene2, eyeposition, color = :black)
    scatter!(scene2, lookat, color = :black)

    r = Rect2f(-1, -1, 2, 2)
    for x in (0, 1)
        for (i, a) in enumerate((0.25, 0.5, 0.75, 1.0))
            ps = [Point3f(a, (0.15 + 0.01y)*(2x-1) , 0.2y) for y in 1:8]
            if x == 0
                cs = [RGBAf(1, 0, 0, 0.75), RGBAf(0, 1, 0, 0.5), RGBAf(0, 0, 1, 0.25)]
            elseif x == 1
                cs = [RGBAf(1, x, 0, a), RGBAf(0, 1, x, a), RGBAf(x, 0, 1, a)]
            end
            idxs = [1, 2, 3, 2, 1, 3, 1, 2, 1, 2, 3][i:7+i]
            meshscatter!(
                scene2, ps, marker = r,
                color = cs[idxs], transparency = true
            )
        end
    end
    linesegments!(scene2, data_limits(scene), color = :black)

    r = map(frustum) do frustum
        r = Rect3f(frustum)
        Rect2f(origin(r)[Vec(1,2)], widths(r)[Vec(1,2)])
        # Rect3f(frustum)
    end
    m = mesh!(scene2, r, color = :lightblue)
    on(frustum) do frustum
        x, y, z = minimum(Rect3f(frustum))
        translate!(m, Vec3f(0, 0, z))
    end
    notify(frustum)

    fig
end
```

</details>

Screenshot from 2023-10-13 20-30-59

For reference, this is from master:

Screenshot from 2023-10-13 16-28-16

And this from beta-20:

Screenshot from 2023-10-13 16-33-23

## Type of change

### DIFF
--- a/docs/explanations/cameras.md
+++ b/docs/explanations/cameras.md
@@ -62,7 +62,7 @@ ez = Point3f(0,0,1)
 
 fig = Figure()
 scene = LScene(fig[1, 1])
-cc = Makie.Camera3D(scene.scene, projectiontype = Makie.Perspective, far = 3.0)
+cc = Makie.Camera3D(scene.scene, projectiontype = Makie.Perspective)
 
 linesegments!(scene, Rect3f(Point3f(-1), Vec3f(2)), color = :black)
 linesegments!(scene,
@@ -70,7 +70,7 @@ linesegments!(scene,
     color = [:red, :red, :green, :green, :blue, :blue]
 )
 center!(scene.scene)
-    
+
 cam = scene.scene.camera
 eyeposition = cc.eyeposition
 lookat = cc.lookat
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -38,7 +38,7 @@ Settings include anything that isn't a mouse or keyboard button.
 - `fixed_axis = true`: If true panning uses the (world/plot) z-axis instead of the camera up direction.
 - `zoom_shift_lookat = true`: If true keeps the data under the cursor when zooming.
 - `cad = false`: If true rotates the view around `lookat` when zooming off-center.
-- `clipping_mode = :bbox_relative`: Controls how `near` and `far` get processed. With `:static` with get passed as is, with `:view_relative` they get scaled by `norm(eyeposition - lookat)` and with `:bbox_relative` they get scaled such that `far = 1` is at or behind the furthest point of the scene bounding box.
+- `clipping_mode = :bbox_relative`: Controls how `near` and `far` get processed. With `:static` they get passed as is, with `:view_relative` they get scaled by `norm(eyeposition - lookat)` and with `:bbox_relative` they get scaled to be just outside the scene bounding box. (More specifically `far = 1` is scaled to the furthest point of a bounding sphere and `near` is generally overwritten to be the closest point.)
 
 - `keyboard_rotationspeed = 1f0` sets the speed of keyboard based rotations.
 - `keyboard_translationspeed = 0.5f0` sets the speed of keyboard based translations.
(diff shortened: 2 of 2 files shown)


ffreyer:
@SimonDanisch This should be ready to merge. If we want to get OIT closer to what it was we can tweak the weight factor after this.

PR #3296, opened 2023-10-14
Add entry about how to trouble shoot GLMakie crashing on macOS

felixcremer:
This should be fixing #2629 by providing a better documentation about how to circumvent the problem.


SimonDanisch:
Thanks!

Issue #3163, opened 2023-08-17
Handle animations in recipes via `render_tick`


jkrumbiegel:
I've recently tried to implement a way to render mock interaction videos. So the mouse movements and keyboard presses are computed from keyframes and then set before each video frame. This works pretty well, except for things like the delay before axis limit spacing is reset when interacting with `Axis`, or the blinking of the cursor in `Textbox`. That's because these things run on real-world-time timers. And obviously, when running `record` we can easily be faster or slower than real time when rendering. So I thought, it would be ideal if these things could be rendered deterministically. And the candidate solution seems to be that plots can hook into `render_tick` of the `Scene` which should then also not just give a trigger signal but also the time (in what time base I'm not sure yet). But for a `record` loop, this could then be set to exactly the frame time before writing out a frame, therefore giving deterministic animations, no matter at what speed they are rendered. I think it would additionally be great to remove those additional tasks, as I've never really liked them.

@ffreyer any thoughts, as you've implemented `render_tick` for screens I think? I've only seen that there's a mention in the code that `render_tick` should be implemented for `Scene` at some point later, but hasn't been, yet.

ffreyer:
My hold up with moving render_tick to `Events` is that I don't know what to do with WGLMakie

jkrumbiegel:
Ok, if we assume WGLMakie could be handled somewhat similarly to GLMakie, would sending a time make sense? What time should it be in that case, system time at the moment of updating the observable? I guess the best would be "predicted time of showing the currently rendering frame" but that's probably hard to determine..

ffreyer:
By appeal to authority it should not be predictive:

https://docs.unity3d.com/ScriptReference/Time-deltaTime.html
> The interval in seconds from the last frame to the current one (Read Only).

https://www.casualdistractiongames.com/post/2018/12/10/unreal-tick-functions-delta-time-and-the-task-graph
> Delta Time represents the amount of time that has occured since the last tick.

I would say the time difference that game engines use should be in an event variable like this. Alongside that we could add a bunch of other variables, like system time at initialization, system time at last trigger, time since init, maybe number of updates and number of frames. I doubt any of these would slow us down significantly, though maybe some of them are hard to to implement in some backends?

We could also consider adding something like Unity's fixed time. Though I guess that's basically just a `Base.Timer()`...

ffreyer:
For reference, `events(fig).tick` now holds:
```julia
struct Tick
[...]

PR #3299, opened 2023-10-16
Global figure time

Kolaru:
# Description

Implement a global figure time that is incremented when the figure is displayed. This allow to have some animation shown when the GLMakie windows shows up for example and to have everything in sync across the figure.

For example with this PR
```julia
using GLMakie

begin
    fig = Figure()

    axs = [Axis(fig[i, j]) for i in 1:2, j in 1:2]
    xlims = [[-1.2, 0], [0, 1.2]]
    ylims = [[0, 1.2], [-1.2, 0]]
    wavelengths = [
        5 8
        11 14
    ]
    ω = 0.3
    θs = range(0, 2π ; length = 1000)

    for i in 1:2 
        for j in 1:2
            ax = axs[i, j]
            xlims!(ax, xlims[j])
            ylims!(ax, ylims[i])
            w = wavelengths[i, j]

            ρs = @lift [sin(w * (θ + ω * $(fig.time))) for θ in θs]
            xx = @lift $ρs .* cos.(θs)
            yy = @lift $ρs .* sin.(θs)

            lines!(ax, xx, yy)
        end
    end
    
    fig
end

record(fig, "flower.gif", 1:240) do i
    fig.time[] = i/24
end
```
Gives (whenever the figure is displayed)
flower

Documentation will be straightforward once we are set on the semantic, but I have no idea if it would be possible to have the feature tested.

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -931,3 +931,5 @@ function plot2robjs(screen::Screen, plot)
 end
 
 export plot2robjs
+
+render_tick(screen::Screen) = screen.render_tick
--- a/src/display.jl
+++ b/src/display.jl
@@ -161,7 +161,7 @@ function Base.display(figlike::FigureLike; backend=current_backend(),
         screen = getscreen(backend, scene; screen_config...)
 
         if figlike isa Figure
-            sync_global_time(figlike, screen.render_tick)
+            sync_global_time(figlike, render_tick(screen))
         end
 
         display(screen, scene)
--- a/src/display.jl
+++ b/src/display.jl
@@ -159,11 +159,12 @@ function Base.display(figlike::FigureLike; backend=current_backend(),
         scene = get_scene(figlike)
         update && update_state_before_display!(figlike)
         screen = getscreen(backend, scene; screen_config...)
-        display(screen, scene)
-    end
 
-    if figlike isa Figure
-        notify(figlike.displayed)
+        if figlike isa Figure
+            sync_global_time(figlike, screen.render_tick)
+        end
+
+        display(screen, scene)
     end
 
     return screen
--- a/src/display.jl
+++ b/src/display.jl
@@ -140,13 +140,13 @@ function Base.display(figlike::FigureLike; backend=current_backend(),
         """)
     end
 
+
     # We show inline if explicitely requested or if automatic and we can actually show something inline!
     if (inline === true || inline === automatic) && can_show_inline(backend)
         Core.invoke(display, Tuple{Any}, figlike)
         # In WGLMakie, we need to wait for the display being done
         screen = getscreen(get_scene(figlike))
         wait_for_display(screen)
-        return screen
     else
         if inline === true
             @warn """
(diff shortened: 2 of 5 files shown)


ffreyer:
I'd rather have something like this implemented as a (render) tick in Events. In GLMakie that could probably just react to `screen.render_tick` and synchronize with the renderloop, which is preferable to a clock that runs out of sync. For recording that wouldn't work on master since `colorbuffer` doesn't trigger `render_tick`, but would after #3246. With the changes you could have 
```julia
timer = Observable(0)
on(_ -> timer[] = timer[] + time_per_frame, events(fig).tick)
```
to do the same kind of thing as your example. (Or, if we have a frame counter as part of the tick just `map(tick -> f(tick.frames * frame_time), events(fig).tick)`. But frames are conceptually a bit awkward because render_tick triggering doesn't mean a new frame is drawn. It triggers every so often to update events, and then only draws a new frame if that is actually necessary.)

jkrumbiegel:
I outlined this somewhere else, not sure where, but I also wanted to have some sort of time accessible to the rendering logic. However, I would not want that to be `time()` in all cases. The reason is that in `record` you don't care about real time, but I still want to be able to render out videos as they would have looked if played live. So in that case, the timestamps would need to be set to the exact frame times.

Kolaru:
The way I see how it could be done is to have a `frame` observable that trigger at each frame and contain the `dt` to the previous frame, like game engines do.

Something like the following
```julia
struct Frame
    n::Observable  # Frame number because why not
    dt::Observable  # Time from the previous frame
    t::Observable  # Real time from fig beginning
end
```
Then in record or the render loop, you can ignore `dt` and `t` values and have a per-frame callback.

However I don't know how it plays with what @ffreyer said, since AFAIK Makie doesn't have a proper "frame" concept right now.

ffreyer:
No, this is pretty much what I want too. After thinking about it some more I don't think there is a meaningful difference between "I tried to render a new frame and realized it wasn't necessary so I didn't" and "I rendered a new frame" for this. It's just time/work saving, and a bit misleading if someone uses it to judge performance.

My thoughts on the implementation are that we should add a `render_tick` or `delta_time` to Events:
https://github.com/MakieOrg/Makie.jl/blob/0329a70a8832f9b95c28196922c41ff44e767f7f/src/types.jl#L41-L104
This requires changes to the events interface here:
https://github.com/MakieOrg/Makie.jl/blob/master/src/interaction/events.jl
And for GLMakie here, where you can grab `screen._rendertick`:
https://github.com/MakieOrg/Makie.jl/blob/master/GLMakie/src/events.jl

Regarding naming I would prefer calling the struct something else because `Frame` sounds like something visible to me, something related to rendering. Maybe just `DeltaTime` or `RenderTick` instead? You may also want a raw `last_time` in there to calculate time deltas, or maybe change `screen.render_tick` to hold that.

For CairoMakie this should probably only trigger when an image is produced. To be compatible with an interactive Cairo backend it should come from actual drawing, maybe here:
https://github.com/MakieOrg/Makie.jl/blob/0329a70a8832f9b95c28196922c41ff44e767f7f/CairoMakie/src/infrastructure.jl#L10-L57

For WGLMakie we would want the same behavior as GLMakie, but I don't know if we have anything set up for that. Maybe the event pipeline is already synced to rendering and you'd just need to add a notification here:
https://github.com/MakieOrg/Makie.jl/blob/0329a70a8832f9b95c28196922c41ff44e767f7f/WGLMakie/src/events.jl#L51-L58

jkrumbiegel:
If it was added to events we could sync on it inside recipes and blocks right?

ffreyer:
Related issue: #3163

Issue #3301, opened 2023-10-17
Regression: Series ignores color keyword 


felixcremer:
- [x ] are you running newest version (version from docs) ?
   I am running current master.
- [ x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x ] What platform + GPU are you on?
julia> versioninfo()
Julia Version 1.9.2
Commit e4ee485e909 (2023-07-05 09:39 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 PRO 5850U with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver3)
  Threads: 1 on 16 virtual cores
Environment:
  JULIA_IMAGE_THREADS = 1

 
The `series` plotting function ignores the color keyword and thereby does not work for series with more than 7 entries.

```julia
julia> args = ([1, 2, 4], [0.6819493475333241 0.6745600434404886 0.4889977072409113; 0.5746392538285008 0.48370986837296914 0.5911597718895916; 0.5394165394683204 0.7528767672734035 0.15235973948183312; 0.47188215926975174 0.304880337006751 0.42410626104380056; 0.09896354500102733 0.45223723495861734 0.7131082316166479; 0.27589708545765024 0.4734317984443611 0.36709014396134987; 0.003961860036966747 0.122883462295309 0.1857948965006283; 0.6121876274096806 0.3779821159724981 0.20031116825629225; 0.5186604908440713 0.4763178112949824 0.15799975734715332; 0.8099218402069105 0.8248055958641253 0.9740322214419064])
([1, 2, 4], [0.6819493475333241 0.6745600434404886 0.4889977072409113; 0.5746392538285008 0.48370986837296914 0.5911597718895916; … ; 0.5186604908440713 0.4763178112949824 0.15799975734715332; 0.8099218402069105 0.8248055958641253 0.9740322214419064])

julia> colormap = :Set1_5
:Set1_5

julia> n = 10
10

julia> color = resample_cmap(colormap, n)
10-element Array{RGBA{Float32},1} with eltype ColorTypes.RGBA{Float32}:
 RGBA{Float32}(0.894f0,0.102f0,0.11f0,1.0f0)
 RGBA{Float32}(0.5926667f0,0.27622223f0,0.382f0,1.0f0)
[...]

PR #3303, opened 2023-10-17
conditionally remove color arg

SimonDanisch:
Closes https://github.com/MakieOrg/Makie.jl/issues/3301

### DIFF
--- a/src/figureplotting.jl
+++ b/src/figureplotting.jl
@@ -39,7 +39,10 @@ function get_axis(fig, P, axis_kw::Dict, plot_attr, plot_args)
         # Remove arguments may not work with plotting into the scene
         delete!(plot_attr, :show_axis)
         delete!(plot_attr, :limits)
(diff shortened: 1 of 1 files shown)


src/figureplotting.jl
```diff
@@ -39,7 +39,10 @@ function get_axis(fig, P, axis_kw::Dict, plot_attr, plot_args)
         # Remove arguments may not work with plotting into the scene
         delete!(plot_attr, :show_axis)
         delete!(plot_attr, :limits)
-        delete!(plot_attr, :color) # Color may contain Cycled(1), which needs the axis to get resolved to a color
+        if plot_attr.color[] isa Cycled
```
felixcremer, line 42:
```suggestion
        if get(plot_attr, :color, nothing) isa Cycled
```
plot_attr is a Dict therefore it should be done like that.

felixcremer: This would fix the problem but I have not run the tests

PR #3310, opened 2023-10-20
Consider marker bbox in meshscatter data_limits

ffreyer:
# Description

This adjusts `data_limits(::MeshScatter)` to include the size of the scattered meshes with scale and rotations applied.

This sort of fixes one of two issues from #3293 which refimg tests missed:

Screenshot from 2023-10-19 13-28-13

which now looks like this:

Screenshot from 2023-10-20 19-46-32

This is a very different zoom level than the refimg, but that's expected. On master the scattered meshes are considered with 0-width which results in the camera being placed closer. This should mean that meshscatter can clip with default limits given a large enough markerscale or a mesh with vertices far from 0. I would consider that a bug which should be fixed on this branch.

## Type of change

## Checklist


jkrumbiegel:
Agree, as meshscatter markers are quite different from scatter markers, which are usually in screen space, I'd also think they should be included in the limits. Especially if you get clipping otherwise

ffreyer:
I think the other issue is fine as is. All the non-data space objects in the plot are drawn at 50% depth, i.e. halfway between the minimum and maximum z by default. With `near` now being closer to plot area, the 3d plot spans a larger depth range and crosses that 50% depth value earlier, thus clipping through the other plots.

PR #3313, opened 2023-10-24
fix bezier construction for strings starting with m

BeastyBlacksmith:
# Description

Before this `BezierPath("m 0,9 11289,0 0,5129 L 0,5138 0,9 z")` would error on trying to access the empty `command` array.

## Type of change

## Checklist


jkrumbiegel:
Nice, thanks! If you could, it would be great to have a few simple unit tests for svg strings

BeastyBlacksmith:
I added a few tests and detected a few holes. I don't intend to fix those currently though.

jkrumbiegel:
Thanks that's already quite helpful :) I only ever tested a couple of paths that seem not to have used all the operators

PR #3316, opened 2023-10-25
Use MakieReferenceImages repository

jkrumbiegel:
Try to run CI with reference images downloaded from https://github.com/MakieOrg/MakieReferenceImages. If this works, in the next step one could set up automatic PRs against that repository after CI runs.


jkrumbiegel:
We are going to wait with this until 0.20 is out

PR #3325, opened 2023-10-29
Histogram: fix a bug when data changes to allequal values

### LINKED ISSUE #2452: Histogram error if data changes to an allequal vector
```
using GLMakie # Makie v0.18.3, GLMakie v0.7.3
v = Observable([1, 2])
hist(v)
v[] = [1, 1]
```
Produces this error:
```
ERROR: MethodError: Cannot `convert` an object of type Vector{Float64} to an object of type StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
```
Specifically, the histogram must be initialized with an observable vector which is not allequal and then it must change to one that is.

Sagnac:
# Description

Fixes #2452

When plotting Observables if the contained collection changed such that all values were equal the minimum and maximum values would match causing `pick_hist_edges` to return a vector. This in turn would result in a conversion error if the previous value was a range. This is fixed by simply returning a range instead.

## Type of change

## Checklist

### DIFF
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -144,7 +144,7 @@ function pick_hist_edges(vals, bins)
     if bins isa Int
         mi, ma = float.(extrema(vals))
         if mi == ma
-            return [mi - 0.5, ma + 0.5]
+            return (mi - 0.5):(ma + 0.5)
         end
         # hist is right-open, so to include the upper data point, make the last bin a tiny bit bigger
         ma = nextfloat(ma)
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Thank you :) Could you add a simple unit test for this?

Sagnac:
Added

SimonDanisch:
I also prefer the latter :)

Sagnac:
Updated

PR #3343, opened 2023-11-01
Deprecate `resolution` for `size`

jkrumbiegel:
# Description

In 0.20, `resolution` stops being a pixel resolution in GLMakie. It was already not really true for CairoMakie before due to `px_per_unit` scaling. The term `size` is more neutral and also works better with the function `resize!` which we already used to change this value. A warning will now be emitted when a `Scene` is created while `resolution` is in the theme. In that case, the `resolution` value will be preferred over `size` to avoid breaking old code completely.


SimonDanisch:
It's weird that the regressions in the `display` benchmark seem somewhat consistent, but locally I can't reproduce them, and from the code changes I'd be quite surprised as well...
It even seems that most of the time this PR is faster 🤷 
(sd/beta-20 vs this PR)
image

SimonDanisch (approved):

jkrumbiegel:
yeah I don't think that makes so much sense, if so we can follow up later

PR #3346, opened 2023-11-01
Add `strokestyle` attribute to `Box`

nHackel:
Hello! I've recently used the `Box` with cornerradius to create a plot and wanted to adapt the stroke style and found no option to do that.

# Description
This PR adds the `strokestyle` keyword to a `Box`:
box

For the implementation I also had to adapt a check for compatible linestyles in `CairoMakie`, which was incompatible with the return type of `convert_attribute` for `key"linestyle`.

## Type of change

## Checklist


jkrumbiegel:
Hi, thanks for the PR! I think we call that property `linestyle` everywhere, even if other properties have `stroke-` as a prefix. But might as well keep in line with that convention for now

nHackel:
Ah, okay! I'm essentially building my first more involved Makie figure and haven't used many keywords so far. 

In `src/makielayout/types.jl` I saw a few types with `x-` prefix but they also seem to have multiple linestyles per type. I can adapt it to linestyle.

Do you think image reference tests are necessary and if so could you point me in the right direction? I was mainly following your rounded corner PR and haven't strayed too far from files you changed there, so a bit unsure of the testing is setup

jkrumbiegel:
Yeah reference images are good practice, it's hard to unit-test visual attributes. You could add a test for `Box` here https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/figures_and_makielayout.jl, just a `Figure` with a couple boxes, each of which exercises some of the functionality. And it's good to make the differences salient, like using thick strokes, big rounded corners, etc, so that visual changes are picked up well.

SimonDanisch (approved):

jkrumbiegel (approved):
Thanks, looks good!

PR #3360, opened 2023-11-08
change vertical axis with `trimspine == true`

### LINKED ISSUE #2664: y-axis misaligned after axis.ytrimspine
When using `axis.ytrimspine`, the vertical axis becomes misaligned. Its connection to the tick mark at the bottom of the axis is untidy, and the top tick mark doesn't even touch the axis. This becomes more apparent when multiple axes are plotted as the distance of misalignment seems constant, not the angle.

`axis.xtrimspine` has no such problem

I'm using Julia v1.8.5 and CairoMakie v0.10.2. Example below.

```
using CairoMakie
CairoMakie.activate!()

xs = rand(100); ys = rand(100)
fig, axis = scatter(xs, ys)
hidespines!(axis, :t, :r)
axis.xtrimspine = true; axis.ytrimspine = true

save("fig.pdf", fig)
```

markgpritchard:
To fix issue #2664. Top of the y-axis was displaced half a tickwidth to the right when trimspine was true, rather than being made half a tickwidth longer.

# Description

Fixes #2664

When trimspine is true, the bottom of the y-axis was displaced half a tickwidth to the left and the top half a tickwidth to the right. This changes that to make the axis vertical, and extends each end by half a tickwidth so that the axis aligns with the ends of the tickmarks.

## Type of change

- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/src/makielayout/lineaxis.jl
+++ b/src/makielayout/lineaxis.jl
@@ -78,8 +78,8 @@ function create_linepoints(
             return [from, to]
         else
             x = position
-            pstart = Point2f(-0.5f0 * tickwidth, 0)
-            pend = Point2f(0.5f0 * tickwidth, 0)
+            pstart = Point2f(0, -0.5f0 * tickwidth)
+            pend = Point2f(0, 0.5f0 * tickwidth)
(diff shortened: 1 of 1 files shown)


jkrumbiegel (approved):
nice, thanks!

PR #3363, opened 2023-11-10
Docstring improvements for hidedecorations!, etc.

### LINKED ISSUE #2846: Can't turn of the grid of two axis simultaneously
The following code does not give the desired output
```
using CairoMakie

Fig = Figure()

ax1 = Axis(Fig[1, 1], title = "Axis 1")
ax2 = Axis(Fig[1, 2], title = "Axis 2")

hidedecorations!(ax1, grid = false)
hidexdecorations!(ax2, grid = false)

Fig
```
I expect this to give me two axis side by side with no grids, but both grids appear. If by the contrary one of the grids is set to true, then only one grid appears and the other one is hidden.

jwahlstrand:
# Description

Describe what the keyword arguments of `hidedecorations!`, `hidexdecorations!`, etc. do. Also fix a few other docstrings.

Fixes #2846 

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -1167,7 +1167,7 @@ end
 Hide decorations of both x and y-axis: label, ticklabels, ticks and grid.
 Keyword arguments can be used to disable hiding of certain types of decorations.
 
-See also [`hidexdecorations!`](@ref), [`hideydecorations!`](@ref), [`hidezdecorations!`](@ref)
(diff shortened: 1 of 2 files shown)


jkrumbiegel (changes requested):
In principle good, but I think we had problems with `@ref` links in docstrings because we don't use documenter. Maybe just remove the links from those mentions for now?

jwahlstrand:
Sounds good, I removed `@ref`.

jkrumbiegel (approved):

PR #3365, opened 2023-11-11
Histogram: fix a plot updating bug for plot specific attributes

### LINKED ISSUE #3362: Weights attribute in hist functions is not Observable-aware
GLMakie v0.8.12 @  Julia 1.9.3 , reproducible - YES, Windows 11 22H2, Intel on CPU, 

MWE:
```
using GLMakie
vstat = rand(10, 10)

fig = Figure()
ax1 = Makie.Axis(fig[1, 1])
ax2 = Makie.Axis(fig[1, 2])
sl = Slider(fig[2, 1], range = 1:10)    #simple observable

hm = heatmap!(ax1, vstat, colormap=:viridis)

weight = lift(sl.value) do v
    return vstat[v, :]
end
hist!(ax2, 1:10, weights = weight, bins=10) # this should show slider-selected column of the vstat 
fig
```
Problem is that triggering of `weight` observable does not change the hist plot.

Sagnac:
# Description

Fixes #3362

This fixes a bug where dynamic changes to histogram specific Observables would fail to cause the histogram plots to update.

For `hist` this was because the plot updating is handled by watching `widths`, but that didn't receive updates whenever the `normalization`, `scale_to`, and `weights` attributes changed. To fix this we now explicitly observe these attributes and set the points whenever they update.

Note that alternatively watching `points` doesn't work; if that's done instead the updating of the histogram breaks in some cases. Also, the Observables themselves are not passed to `barplot!` in order to avoid redundant updates as `points` and `widths` depend upon some of the same Observables.

For the step histogram, the three aforementioned attributes along with `bins` were not being listened to so as a result the plot would fail to update whenever any of these Observables changed. This is fixed by simply passing the `points` Observable instead of the values to `stairs!`.

`stephist` is different compared to `hist` as there's no `widths` so there are no efficiency gains by managing updates separately. Any updates to the four attributes will update `points` here so the plot updating is handled correctly.

## Type of change

### DIFF
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -78,8 +78,8 @@ function Makie.plot!(plot::StepHist)
     pop!(attr, :normalization)
     pop!(attr, :scale_to)
     pop!(attr, :bins)
-    # plot the values, not the observables, to be in control of updating
-    stairs!(plot, points[]; attr..., color=color)
+    stairs!(plot, points; attr..., color=color)
+    plot
 end
 
 """
@@ -186,6 +186,9 @@ function Makie.plot!(plot::Hist)
     on(plot, widths) do w
         bp[1].val = points[]
         bp.width = w
+    end
+    onany(plot, plot.normalization, plot.scale_to, plot.weights) do _, _, _
+        bp[1][] = points[]
     end
     plot
 end


Sagnac:
Apparently this is an issue with `stephist` as well. None of the plot specific attributes are observed so the plot doesn't update when the Observables change. I've added another commit addressing this.

We could also just pass the `points` Observable to `stairs!` instead of the values, but the comment specifically states passing the values and updating using a different pattern is a design choice.

As for `hist`, an alternative approach is using the same scheme as `stephist` instead of using the `widths` callback:
```diff
diff --git a/src/stats/hist.jl b/src/stats/hist.jl
index 4ea1b50e..4efaf053 100644
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -191,7 +191,7 @@ function Makie.plot!(plot::Hist)
         bp.width = w
     end
     onany(plot.normalization, plot.scale_to, plot.weights) do _, _, _
-        notify(widths)
+        bp[1][] = points[]
     end
     plot
 end

```

src/stats/hist.jl
```diff
@@ -79,7 +79,10 @@ function Makie.plot!(plot::StepHist)
     pop!(attr, :scale_to)
     pop!(attr, :bins)
     # plot the values, not the observables, to be in control of updating
```
Sagnac, line 81:
```suggestion
```

src/stats/hist.jl
```diff
@@ -79,7 +79,10 @@ function Makie.plot!(plot::StepHist)
...
-    stairs!(plot, points[]; attr..., color=color)
+    st = stairs!(plot, points[]; attr..., color=color)
+    onany(plot.normalization, plot.scale_to, plot.weights, plot.bins) do _, _, _, _
+        st[1][] = points[]
+    end
```
Sagnac, line 85:
```suggestion
    stairs!(plot, points; attr..., color=color)
    plot
```
I got the chance to look at this again and think about it some more and the more I consider it the more I realize that we should probably just pass the `points` Observable here. The reason this isn't done for `hist` seems to be because `widths` depends upon some of the same Observables as `points`, so the trick to avoid redundant updates is employed there.

However, `stephist` is different, there's no `widths` so there doesn't appear to be any efficiency gains by managing updates separately. Any updates to `bins`, `normalization`, `scale_to`, and `weights` will update `points` so the plot updating is handled correctly.

src/stats/hist.jl
```diff
@@ -187,5 +190,8 @@ function Makie.plot!(plot::Hist)
         bp[1].val = points[]
         bp.width = w
     end
+    onany(plot.normalization, plot.scale_to, plot.weights) do _, _, _
+        notify(widths)
```
Sagnac, line 194:
```suggestion
        bp[1][] = points[]
```
This is simply more direct, there's no need to go through `widths` here.

Moelf:
Hi, thanks for working on this, I looked at git blame of course it was me who touched this histogram thing last time. I wish there's a way to ping me everything there's histogram related thing happening here...

src/stats/hist.jl
```diff
@@ -187,5 +190,8 @@ function Makie.plot!(plot::Hist)
         bp[1].val = points[]
         bp.width = w
     end
+    onany(plot.normalization, plot.scale_to, plot.weights) do _, _, _
```
Sagnac, line 193:
```suggestion
    onany(plot, plot.normalization, plot.scale_to, plot.weights) do _, _, _
```
This method should be used so that the callback is deregistered on close

Sagnac:
> Hi, thanks for working on this, I looked at git blame of course it was me who touched this histogram thing last time. I wish there's a way to ping me everything there's histogram related thing happening here...

Hi, sorry I should have definitely thought of that; perspective from the last person who worked on it is always helpful

Sagnac:
I've committed the proposed changes and squashed the commits for coherency.

SimonDanisch:
@Moelf, does this look correct now?
@Sagnac, you don't actually need to squash, since we'll squash merge anywasy ;)

Moelf:
from my end I don't think this PR changes how things are visualized, so I don't have problem with it if it "updates" more correctly

SimonDanisch:
Thank you all!

PR #3381, opened 2023-11-17
Replace radial_distortion_threshhold with directly setting radius_at_origin

### LINKED ISSUE #3375: PolarAxis unable to set negative rlimits
```julia
julia> versioninfo()
Julia Version 1.9.4
Commit 8e5136fa29 (2023-11-14 08:46 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 32 × AMD Ryzen 9 3950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver2)
  Threads: 16 on 32 virtual cores
Environment:
  JULIA_NUM_THREADS = 16
```

# Issue

It looks like **PolarAxis** does not support setting negative r limits or plotting lines with negative values. Polar plots are used frequently in electrical engineering to present things like antenna gain patterns, measured on a dB scale ([Example](https://www.mathworks.com/help/antenna/gs/antenna-modeling-and-analysis.html#AntennaModelingAndAnalysisTutorialExample-5)).

I demonstrated this with the MWE below, plotting a fake gain pattern in linear and dB scales on both a regular `Axis` and a `PolarAxis`. The polar plot in dB scale won't display any of the line because it's always <= 0. If I try to set `rlimits` to the appropriate range for this plot, e.g. `rlimits=(-40,0)` the r axis looks like a mess with no ticks remaining.

# Minimum Working Example

```julia
using CairoMakie

pow2db(x) = 20log10(x)

theta_prime = pi
angle(a,b) = min(abs(a-b), 2π-abs(a-b))
gain(theta) = abs(sinc(2.0*angle(theta,theta_prime)))

thetas = range(0, 2pi, length=361)
```

```julia
# Axis plots in linear and dB scale
f_xy = Figure()
ax_xy_lin = Axis(f_xy[1,1], title="Gain", xlabel="Theta [rad]")
lines!(ax_xy_lin, thetas, gain.(thetas))
ax_xy_dB = Axis(f_xy[1,2], title="Gain in dB", xlabel="Theta [rad]")
lines!(ax_xy_dB, thetas, pow2db.(gain.(thetas)))
save("f_xy.png", f_xy)
```

f_xy

```julia
# PolarAxis plots in linear and dB scale
f_polar = Figure()
ax_polar_lin = PolarAxis(f_polar[1,1], title="Gain")
lines!(ax_polar_lin, thetas, gain.(thetas))
ax_polar_dB = PolarAxis(f_polar[1,2], title="Gain in dB")
lines!(ax_polar_dB, thetas, pow2db.(gain.(thetas)))
save("f_polar.png", f_polar)
```

f_polar

ffreyer:
# Description

Fixes #3375 by allowing users to set a radial offset that can be used to map a negative rmin to 0. E.g.:
```julia
pow2db(x) = 20log10(x)    
theta_prime = pi
angle(a,b) = min(abs(a-b), 2π-abs(a-b))
gain(theta) = abs(sinc(2.0*angle(theta,theta_prime)))

thetas = range(0, 2pi, length=361)
rs = pow2db.(gain.(thetas))

f_polar = Figure()
ax_polar_lin = PolarAxis(f_polar[1,1], title="Gain")
lines!(ax_polar_lin, thetas, gain.(thetas))
ax_polar_dB = PolarAxis(f_polar[1,2], title="Gain in dB", radius_at_origin = -80)
lines!(ax_polar_dB, thetas, rs)
f_polar
```

Screenshot from 2023-11-17 14-46-38

Some open questions:
- Should negative rlimits automatically result in radii getting shifted?
- Should zooming be allowed to change `radius_at_origin`?
- Should we adjust `theta_0` to work like r0/radius_at_origin? Currently we have `theta_out = theta_in + theta_0` and `r_out = r_in - r0`

Note that offsetting radii results in distortions:
```julia
phis = range(pi/4, 9pi/4, length=201)
rs = 1.0 ./ sin.(range(pi/4, 3pi/4, length=51)[1:end-1])
rs = vcat(rs, rs, rs, rs, rs[1])

fig = Figure()
ax = PolarAxis(fig[1, 1], radius_at_origin = -2)
lines!(ax, phis, rs)
ax = PolarAxis(fig[1, 2])
lines!(ax, phis, rs)
ax = PolarAxis(fig[1, 3], radius_at_origin = 0.5)
lines!(ax, phis, rs)
fig
```

Screenshot from 2023-11-17 14-57-01

## Changes

- replace `radial_distortion_threshold` with `radius_at_origin` which allows you to set an offset for radii
- add `clip_r::Bool` to PolarAxis and Polar transform, allowing you to set whether to enforce $r \ge 0$ or not
- for `clip_r = true` (old behavior) return `NaN` rather than `r = 0` if $r < 0$.

## Type of change

- [x] (Somewhat minor) Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


ffreyer:
> Some open questions:
> 1. Should negative rlimits automatically result in radii getting shifted?
> 2. Should zooming be allowed to change `radius_at_origin`?
> 3, Should we adjust `theta_0` to work like r0/radius_at_origin? Currently we have `theta_out = theta_in + theta_0` and `r_out = r_in - r0`

Personally I think:
1. is a good idea, since otherwise data is just not shown.
2. is not something we should do. It feels a bit weird for zooming to get "stuck" at `radius_at_origin`, but the other choice would be to allow zooming to distort plots. Probably not good default behavior...
3. is probably good in the long run.

@mikeingold @anowacki @jkrumbiegel @SimonDanisch  Maybe any of you have some more thoughts on this?

mikeingold:
Wow! Thanks for the swift response and PR.

- I would personally prefer PolarAxis automatically adjust `radius_at_origin` to accomodate negative values, when present. If I'm plotting a line with negative values, I'm not sure why I'd intend to generate a plot with them cropped out by default.
- Zooming behavior is an interesting question since it's kind of intrinsically a one-dimensional operation (in vs out) but needs to control two parameters (min, max). I don't have much experience with interactive Makie plots and don't have an interactive PolarAxis in front of me at the moment, but maybe zoom-level could just proportionally scale the min-max range, and a mouse click-drag could be used to slide the `radius_at_origin` to offset the whole range?
- Unless I'm mistaken, the distortion you showed above is an artifact of the transformed coordinate space, and all of the data appears to be plotted at the correct locations, right? Essentially, the offset r min limit causes a spatial transform such that the space isn't visually-equivalent to a Euclidean space, so shapes won't be preserved. Could be worth noting in the docs, but "it's a feature, not a bug".
- Side note: I was also a bit surprised to see that the default behavior wasn't to try to plot negative r values on the opposite theta, i.e. `(r, theta) -> (-r, theta+pi)`. I've seen this occasionally, e.g. [functions like this](https://en.wikipedia.org/wiki/Polar_coordinate_system#Polar_rose).

ffreyer:
We had quite a bit of discussion in #3154 about having a positive r shift as a default and decided not to. You're right that the placement is still correct, but having shapes distort may mislead people when they're not aware of it. Having users explicitly adjust some attribute can be a useful barrier here, I think. 

From that point of view we should have `radius_at_origin` default to 0 to avoid distorting shapes and not have `r_out < 0` draw since $r = \sqrt{x^2 + y^2} \ge 0$. (Currently there is a `max(0.0, r_out)` so some plots will still draw.)

mikeingold:
It sounds like the core principles we'd like to respect are:
- Plot all data that the user asked to be plotted
- Maintain a geometry that is visually consistent with the Euclidean plane

In the case of strictly non-negative data, both principles are satisfied by just pinning `radius_at_origin` to zero, so that seems like a fairly logical default. If a user sees the result and feels like they'd prefer a tighter limit range, then it should probably be configurable to allow this as a non-default option.

In the presence of negative data, though, we can't simultaneously satisfy both principles by default. So then the question in my mind is: which do most users expect should take precedence? In my experience, polar plots with negative values are very common, and the data isn't expected to have a literal physical interpretation.

As a mostly-casual Makie user, my experience with this was roughly: plot with a default `PolarAxis`, get confused by the result, troubleshoot whether my data was bad, see if `lines!` was setup wrong, look for options to manually set `limits`, etc. The change in this PR that allows non-default negative `radius_at_origin` would at least allow users like me to generate these plots, but I'd argue that if I'm asking to plot negative data, then I'm probably already expecting to see negative data. At the very least, I'd appreciate a `@warn`ing that explains why I'm not seeing what I expected.

**tl;dr**, In my opinion:
- This PR improves the Makie experience by enabling polar plots to show negative data, a very common need for (at least) EE's.
- In the case of strictly-positive data, configuring `radius_at_origin = 0` is a good default.
- However, in the presence of negative data, the default should be to show the data offsetting `radius_at_origin`.

src/basic_recipes/triplot.jl
```diff
@@ -239,3 +239,17 @@ function Makie.plot!(p::Triplot{<:Tuple{<:DelTri.Triangulation}})
...
+        # Because the Polar transform is handled explicitly we cannot rely
+        # on the default data_limits. (data limits are pre transform)
+        iter = (to_ndim(Point3f, p, 0f0) for p in p.converted[1][])
+        limits_from_transformed_points(iter)
+    else
+        # First component is either another Voronoiplot or a poly plot. Both
+        # cases span the full limits of the plot
+        data_limits(p.plots[1])
+    end
+end
```
ffreyer, line 255: Test failures come from this adjusting PolarAxis limits a little. It's the same code as we already have for voronoiplots

jkrumbiegel:
> In the presence of negative data, though, we can't simultaneously satisfy both principles by default. So then the question in my mind is: which do most users expect should take precedence? In my experience, polar plots with negative values are very common, and the data isn't expected to have a literal physical interpretation.

I've never used polar plots for real (not common at all in neuroscience / psychology) so I don't have much of an opinion. But this line of reasoning sounds valid to me, and we could adjust r to include all data in case there's negative data.

How are other plotting libraries handling this?

PR #3394, opened 2023-11-21
sanity checks for texture atlas hashing & defaults

### LINKED ISSUE #3264: Bezier hashes might have been removed?
I wanted to check what the hash implementation of bezier commands was again, but all the hash-related lines from that file seem to have been removed. (Even though it says `using StableHashTraits` at the top)

@SimonDanisch this removes a bunch https://github.com/MakieOrg/Makie.jl/pull/3216 but the one that hashed the command based on its vector seems to have been gone for longer even. If this doesn't work right, it will spam the texture atlas.

SimonDanisch:
closes #3264

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1340,42 +1340,263 @@ function convert_attribute(value::Union{Symbol, String}, k::key"algorithm")
     end, k)
 end
 
-const DEFAULT_MARKER_MAP = Dict{Symbol, BezierPath}()
+#=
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Great we tested this, since we do have cache misses because of https://github.com/JuliaArrays/StaticArrays.jl/issues/1222.
Solved for now by writing out the bezierpath values instead of calculating them in each julia version on first call to `default_marker_map`.

PR #3398, opened 2023-11-22
add comment to clarify minus sign

stevengj:
Just adds a comment to clarify that `MINUS = "−"` is the same as U+2212


SimonDanisch:
Thanks, that's helpful!

PR #3413, opened 2023-11-24
LinkAxis defaultfont missing

behinger:
# Description
LinkAxis is requiring a defaultfont for label + xticklabel, which is not given. Maybe these are the right parts to fix it. I put this as a draft because I need to check it locally first.

## Type of change

## Checklist


behinger:
ok, there is much more broken in LinkAxis than I hoped for - I will work locally

PR #3414, opened 2023-11-24
fix bad limits with rotation (no s) attribute

ffreyer:
# Description

Fixes this:

Screenshot from 2023-11-24 16-19-44

until we either deprecate `rotation` or `rotations` as an attribute (for meshscatter). After:

Screenshot from 2023-11-24 16-21-06

## Type of change

## Checklist

### DIFF
--- a/src/layouting/data_limits.jl
+++ b/src/layouting/data_limits.jl
@@ -252,10 +252,11 @@ end
 
 function data_limits(plot::MeshScatter)
     # TODO: avoid mesh generation here if possible
-    @get_attribute plot (marker, markersize, rotations)
+    @get_attribute plot (marker, markersize)
     marker_bb = Rect3f(marker)
     positions = iterate_transformed(plot)
     scales = markersize
+    rotations = to_rotation(to_value(get(plot, :rotation, plot.rotations)))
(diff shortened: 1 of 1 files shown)


SimonDanisch:
We kind of deprecated `rotation` for meshscatter in 0.20!

SimonDanisch:
I'd say we should throw an error when using rotation?

jkrumbiegel:
I also didn't know that. And shouldn't we deprecate rotations instead because we don't usually use plural?

SimonDanisch:
Hm, `rotations` is used in the other primitives and backends, that didn't have the weird GLMakie only `rotation` attribute.
Admittedly, since it was GLMakie only and not widely used, I forgot about adding it to the news.

jkrumbiegel:
Ok but still, wouldn't rotation be more consistent?

SimonDanisch:
Yeah, true!

jkrumbiegel:
Then let's deprecate rotations instead. We could also do a "soft soft deprecation" and just remove mention of rotations everywhere and leave it at that for now. Or warnings, I'm also ok with that

PR #3419, opened 2023-11-26
try upcoming julia-actions/cache version

IanButterworth:
Just seeing what the impact is of https://github.com/julia-actions/cache/pull/71 here.


IanButterworth:
The new version has landed and will be picked up on master given `@v1`

jkrumbiegel:
@IanButterworth haven't really had time to look at the logs carefully, what kind of benefits did you see for Makie?

IanButterworth:
I didn't look carefully here at timing. I just tweaked things to work a bit better with the way concurrency is set up here via different ymls

Issue #3440, opened 2023-12-03
doc: broken `triplot` examples


t-bltg:
No example images in https://docs.makie.org/stable/reference/plots/triplot.

j-fu:
Seems to be a problem with CairoMakie generated svg's. With 0.11.1 they are ok, after upgrade to 0.11.2 I am getting the same thing locally: blurred image in Chrome, empty space in Firefox. Inkscape shows the images properly, though.

jkrumbiegel:
Probably the change in svg salting, maybe it breaks for plot types I didn't test

j-fu:
Something like this. Simple lines are ok, but it fails e.g. with color interpolation on meshes.

PR #3443, opened 2023-12-05
Match url ids but not image data blobs in svg salting

jkrumbiegel:
Fixes https://github.com/MakieOrg/Makie.jl/issues/3440 which was a regression introduced by https://github.com/MakieOrg/Makie.jl/pull/3435 because the regex doing more efficient salting accidentally also matched image base64 blobs, making them invalid:

```html
xlink:href="data:image/png;base64,iVBOR ... SuQmCC-f0f66279"/>
```

### DIFF
--- a/CairoMakie/src/display.jl
+++ b/CairoMakie/src/display.jl
@@ -82,7 +82,11 @@ function Makie.backend_show(screen::Screen{SVG}, io::IO, ::MIME"image/svg+xml", 
     # so the output is deterministic
     salt = repr(CRC32c.crc32c(svg))[end-7:end]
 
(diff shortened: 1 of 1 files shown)


j-fu:
Still broken here: 
[ee78f7c6] Makie v0.20.2 `https://github.com/MakieOrg/Makie.jl#eb3e086`

makieblur

jkrumbiegel:
Can you make an MWE that still fails?

j-fu:
It looks ok with `display`, `inkscape`, but not in the browser, and not with `gimp`.

j-fu:
Ah was confused by the monorepo structure. It was not sufficient to add Makie from the commit, but also had to take CairoMakie from this commit.

Now I can confirm it works - sorry for the confusion & thanks for the quick act!

PR #3469, opened 2023-12-13
Make autolimits! default to current_axis()

KronosTheLate:
This PR makes `Makie.autolimits!` default to using `current_axis()`

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -907,7 +907,11 @@ function autolimits!(ax::Axis)
     ax.limits[] = (nothing, nothing)
     return
 end
-autolimits!() = autolimits!(current_axis())
+function autolimits!()
+    curr_ax = current_axis()
+    isnothing(curr_ax)  &&  throw(ArgumentError("Attempted to call `autolimits!` on `current_axis()`, but `current_axis()` returned nothing."))
+    autolimits!()
+end
 
 function autolimits(ax::Axis, dim::Integer)
     # try getting x limits for the axis and then union them with linked axes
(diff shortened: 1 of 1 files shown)


src/makielayout/blocks/axis.jl
```diff
@@ -897,14 +897,17 @@ function update_linked_limits!(block_limit_linking, xaxislinks, yaxislinks, tlim
...
     autolimits!(la::Axis)
 
 Reset manually specified limits of `la` to an automatically determined rectangle, that depends on the data limits of all plot objects in the axis, as well as the autolimit margins for x and y axis.
+The argument `la` defaults to `current_axis()`.
 """
 function autolimits!(ax::Axis)
     ax.limits[] = (nothing, nothing)
     return
 end
+autolimits!() = autolimits!(current_axis())
```
SimonDanisch, line 910: I think this should check for `current_axis` returning nothing!

KronosTheLate: That sounds like a very good idea. I implemented this check. Below is a demo:
```julia
julia> using GLMakie

julia> Makie.autolimits!
autolimits! (generic function with 4 methods)

julia> function autolimits!()
           curr_ax = current_axis()
           isnothing(curr_ax)  &&  throw(ArgumentError("Attempted to call `autolimits!` on `current_axis()`, but `current_axis()` returned nothing."))
           autolimits!()
       end
autolimits! (generic function with 1 method)

julia> autolimits!()
ERROR: ArgumentError: Attempted to call `autolimits!` on `current_axis()`, but `current_axis()` returned nothing.
Stacktrace:
 [1] autolimits!()
   @ Main ./REPL[5]:3
 [2] top-level scope
   @ REPL[6]:1
```

PR #3490, opened 2023-12-19
Fix & test relocatability

SimonDanisch:


SimonDanisch:
I guess it does pass the test and works on linux + osx... So maybe fine as is?

PR #3491, opened 2023-12-19
add highlevel gui + auto legend/colorbar

SimonDanisch:
This is a proposal for an easy way to add a highlevel GUI and auto legend/colorbar, without any changes to the internal plotting pipeline.
This makes the implementation pretty straightforward and easy to change going forward, at the slight expanse of discoverability & usability.
This works by just wrapping a plot call into `GUI(plot(...))`, which will then create a GUI like this:
GUI
TODOs:
- [ ] tests 
- [ ] more documentation
- [ ] precompiles
- [ ] finalize API
- [ ] what should we do about mutating calls after the call to `GUI`?


src/GUI/gui.jl
```diff
@@ -0,0 +1,98 @@
...
+
+Automatically creates a legend and colorbar for the return value of `plot`.
+Also adds a small UI to save/reset/copy the plot.
+# Example
+
+```julia
+f, ax, pl = GUI(series(rand(7, 20)); legend=(position=:lt, title="legend"))
+```
+"""
+function GUI(faxpl::Makie.FigureAxisPlot; legend=(;), colorbar=(;))
```
asinghvi17, line 67: Could we also add a method which accepts a figure, so people can "opt-in" to displaying their figures (or maybe even Scenes) in this GUI by default, perhaps by setting a switch in activate!?

SimonDanisch:
Alright picked this up again and integrated it with the theming and it can be now turned on by default via theming without wrapping `GUI(fig)`: 
This adds:
* HoverMenu block with Save/Copy/Reset buttons that appears on hover at top of figure
* automatic legend/colorbar creation via Figure(; legend=..., colorbar=..., gui=...)
* Legend(ax; position=:rt) and Colorbar(ax, plot; position=:rt) convenience constructors for overlay positioning
* axiscolorbar function (analogous to axislegend). 
* margin attribute to Colorbar for overlay positioning
* Convert kwargs to Dict{Symbol,Any} early in block constructors to avoid repeated splatting overhead
* added GUIState to Figure for tracking GUI elements

Usage
```julia
# Automatic legend/colorbar from Figure options
scatter(rand(10), label="Data", color=1:10; 
    figure=(; legend=(position=:lt,), colorbar=(position=:rt,)))

# Or via theme
set_theme!(Figure=(; gui=true, legend=(position=:lt,)))
```
TODOS:

* do we actually want to deprecate axislegend/colorbar in favor of `Constructor(ax, ...)`?
* figure_padding should maybe just be `figure=(; padding = 0)` ?

PR #3494, opened 2023-12-20
Fix calculation of focal length for RPRMakie

jkrumbiegel:
With this change, fov looks the same in GLMakie and RPRMakie for me. Before, we just had this hardcoded value for some reason.

grafik
grafik


SimonDanisch:
This might be from the time where the 3d camera didn't use fov for zooming!

Issue #3357, opened 2023-11-07
hexbin fails when one array is all zeros


sethaxen:
`hexbin` errors if one of its arrays is passed all zeros. Passing a non-zero constant array is fine:

```julia
julia> using CairoMakie

julia> hexbin(ones(10), randn(10)) # fine

julia> hexbin(zeros(10), randn(10)) # errors
ERROR: InexactError: Int64(NaN)
Stacktrace:
  [1] Int64
    @ ./float.jl:900 [inlined]
  [2] nearest_center(val::Float32, spacing::Float32, offset::Float32)
    @ Makie ~/.julia/packages/Makie/RgxaV/src/stats/hexbin.jl:232
  [3] (::Makie.var"#calculate_grid#1092"{Observable{Vec{2, Float32}}, Observable{Vector{Float64}}, Observable{Vector{Point{2, Float32}}}})(xy::Vector{Point{2, Float32}}, weights::Nothing, bins::Int64, cellsize::Nothing, threshold::Int64)
    @ Makie ~/.julia/packages/Makie/RgxaV/src/stats/hexbin.jl:123
  [4] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Base ./essentials.jl:819
  [5] invokelatest(::Any, ::Any, ::Vararg{Any})
    @ Base ./essentials.jl:816
  [6] (::Observables.OnAny)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:420
  [7] #invokelatest#2
    @ ./essentials.jl:819 [inlined]
  [8] invokelatest
    @ ./essentials.jl:816 [inlined]
  [9] notify(observable::Observables.AbstractObservable)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206
 [10] plot!(hb::Combined{Makie.hexbin, Tuple{Vector{Point{2, Float32}}}})
    @ Makie ~/.julia/packages/Makie/RgxaV/src/stats/hexbin.jl:179
[...]

PR #3506, opened 2023-12-24
fix `hexbin` corner case

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/3357.

Simplified a bit the implementation of `hexbin` (mostly cleanup, unused leftovers, ...).

## Type of change

## Checklist


t-bltg:
This is the output from the added tests.

Not sure if we want to fix the case with all ones (cause is that the `markersize` is too small, w.r.t. `x` or `y` limits).

```julia
# degenerate case wisth singleton 0
save("x0.png", hexbin([0, 0], [1, 2]))
save("y0.png", hexbin([1, 2], [0, 0]))

# degenerate case with singleton 1
save("x1.png", hexbin([1, 1], [1, 2]))
save("y1.png", hexbin([1, 2], [1, 1]))
```
**x0**
x0
**y0**
y0

**x1**
x1
**y1**
y1

src/stats/hexbin.jl
```diff
@@ -126,11 +120,7 @@ function plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Point2}}})
             d1 = ((_x - nx)^2 + (yweight * (_y - ny))^2)
             d2 = ((_x - nxs)^2 + (yweight * (_y - nys))^2)
 
-            is_grid1 = d1 < d2
```
jkrumbiegel, line 129: This variable is useful I think, without that I wouldn't quite remember what the comparison was for

jkrumbiegel (changes requested):

src/stats/hexbin.jl
t-bltg: Sorry, restored.

t-bltg:
For https://github.com/MakieOrg/Makie.jl/issues/3507, I've sorted the list of tests to be ran in alphabetical order: this is less error prone.

Fixing `statistical_tests.jl` and `zoom_pan.jl` is left for another PR, but at least they are now listed in `runtests.jl` with a `FIXME` note added.

test/runtests.jl
```diff
@@ -10,33 +10,28 @@ using GeometryBasics: Pyramid
 using Makie: volume
 
 @testset "Unit tests" begin
-    @testset "#659 Volume errors if data is not a cube" begin
```
t-bltg, line 13: Note: moved to an added `issues.jl` file.

t-bltg:
I fail to understand why the `Relocability` CI test fails now :/

t-bltg:
> I fail to understand why the Relocability CI test fails all the time now :/

Hum, spurious random CI failure, restored the old `ubuntu-20.4`.

jkrumbiegel:
Hm I find it weird that whether the singleton dimension has only zeros or ones makes a difference whether bins are shown or not. I'm really not sure what a good behavior here is. Also, in the first case, it's kind of weird that the upper bin would include the point in question _right on the edge_ basically. Because the visual impression is that there's some data to the right.

grafik

Actually I've noticed that the way the bins are chosen by default doesn't quite make sense in the y direction:

grafik

The bins extend further than they need to, vertically, to fully envelop the rectangle with the data points. Maybe we should also fix that?

We should check what matplotlib and ggplot are doing with their hexbins for these cases.

t-bltg:
These are examples from matplotlib, adapted from the [hexbin](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hexbin.html) docs:
```python
import numpy as np
import matplotlib.pyplot as plt

np.random.seed(19680801)
n = 300
x = np.random.rand(n)
y = np.random.rand(n)

fig, ax = plt.subplots()
ax.hexbin(x, y, gridsize=(3, 4 // 2))
fig.savefig('py-3x4.png')

fig, ax = plt.subplots()
ax.hexbin([0, 0], [1, 2], gridsize=(2, 2 // 2))
fig.savefig('py-x0.png')

fig, ax = plt.subplots()
ax.hexbin([1, 2], [0, 0], gridsize=(2, 2 // 2))
fig.savefig('py-y0.png')

fig, ax = plt.subplots()
ax.hexbin([1, 1], [1, 2], gridsize=(2, 2 // 2))
fig.savefig('py-x1.png')

fig, ax = plt.subplots()
ax.hexbin([1, 2], [1, 1], gridsize=(2, 2 // 2))
fig.savefig('py-y1.png')
```
**py-3x4**
py-3x4
**py-x0**
py-x0
**py-y0**
py-y0
**py-x1**
py-x1
**py-y1**
py-y1

> The bins extend further than they need to, vertically, to fully envelop the rectangle with the data points. Maybe we should also fix that?

I don't really understand what you mean: this seems to be consistent with what matplotlib does, no ?

Issue #3245, opened 2023-09-21
Observable update bug


jonalm:
```julia
marker=Observable(Union{Char, Symbol}[:cross, :cross])
fig = Figure()
ax = Axis(fig[1, 1])
scatter!(ax, [1,2],[3,4]; marker)
display(fig)

marker[][1] = 'x'
marker[] = marker[]
```
throws an error
```
ERROR: MethodError: Cannot `convert` an object of type Char to an object of type BezierPath
```

while
```julia
marker=Observable(Union{Char, Symbol}[:cross, 'x'])
fig = Figure()
ax = Axis(fig[1, 1])
scatter!(ax, [1,2],[3,4]; marker)
display(fig)

marker[][1] = 'x'
marker[] = marker[]
```
works as expected.

DanielVandH:
What's your `] st`? I don't get any error, and the figure updates correctly:

```julia
julia> marker=Observable(Union{Char, Symbol}[:cross, :cross])
       fig = Figure()
       ax = Axis(fig[1, 1])
       scatter!(ax, [1,2],[3,4]; marker)
       display(fig)

       marker[][1] = 'x'
       marker[] = marker[]
2-element Vector{Union{Char, Symbol}}:
 'x': ASCII/Unicode U+0078 (category Ll: Letter, lowercase)
 :cross
```

SimonDanisch:
Could be a strongly typed conversion somewhere in the internals... @DanielVandH it could be backend specific, which backends are you using?

DanielVandH:
Ah, yes - it's just on GLMakie that it errors like this.

jonalm:
Yeah, sorry. I didn't specify that. It is `GLMakie`.

PR #3508, opened 2023-12-24
fix strong type conversion of markers

t-bltg:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/3245.

Retain the original marker type in the output eltype `Union`, when a marker is converted through the `DEFAULT_MARKER_MAP` dictionary.

Another option is to go for `Any` ?

## Type of change

## Checklist

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1643,7 +1643,10 @@ struct FastPixel end
 Vector of anything that is accepted as a single marker will give each point it's own marker.
 Note that it needs to be a uniform vector with the same element type!
 """
-to_spritemarker(marker::AbstractVector) = map(to_spritemarker, marker)
+function to_spritemarker(marker::AbstractVector{T}) where T
+    spritemarker = map(to_spritemarker, marker)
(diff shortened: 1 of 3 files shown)


src/conversions.jl
```diff
@@ -1643,7 +1643,10 @@ struct FastPixel end
 Vector of anything that is accepted as a single marker will give each point it's own marker.
 Note that it needs to be a uniform vector with the same element type!
 """
-to_spritemarker(marker::AbstractVector) = map(to_spritemarker, marker)
+function to_spritemarker(marker::AbstractVector{T}) where T
+    spritemarker = map(to_spritemarker, marker)
+    convert(Vector{Union{T,eltype(spritemarker)}}, spritemarker)  # retain original type T
```
SimonDanisch, line 1648: I think it would be better to restrict it to all possible converted marker types...
The fact that I don't know it out of the top of my head already shows, that this is under documented, and adding better types here would at least help a little with this :) 
I can take a look!

t-bltg: Yes, as is, this PR breaks a reference test and I must admit I have no idea left on top of my head now. At least, we now know where the issue lies :)

PR #3525, opened 2023-12-30
fix streamplot 3D default arrow size

### LINKED ISSUE #3504: Streamplot 3D seems to be broken
@chillenb has made me aware that Makie.jl is breaking the 3D streamplot:

https://github.com/chakravala/Grassmann.jl/issues/105

It seems to me that this is an issue with `Makie.streamplot` and not an issue on my package.

chillenb:
# Description

Fixes #3504 

The default arrow size in 3D streamplot is set to
`0.2 * minimum(limits.width) / minimum(gridsize)`
per @chakravala's comment in #3504.

The default arrow size in 2D streamplot remains 15 px.

## Type of change


chakravala:
Fixes chakravala/Grassmann.jl#105

src/basic_recipes/streamplot.jl
```diff
@@ -28,7 +28,7 @@ See the function `Makie.streamplot_impl` for implementation details.
         maxsteps = 500,
         color = norm,
 
-        arrow_size = 15,
+        arrow_size = nothing,
```
ffreyer, line 31: This should be `automatic` since we generally use that instead of nothing when things get derived later

chillenb: OK, I've changed `nothing` to `automatic`!

src/basic_recipes/streamplot.jl
```diff
@@ -213,10 +213,18 @@ function plot!(p::StreamPlot)
         rotations = map(x -> x[2], data)
     end
 
+    if p.arrow_size[] === automatic
+        if N == 3
+            arrow_size = 0.2 * minimum(p.limits[].widths) / minimum(p.gridsize[])
+        else
+            arrow_size = 15
+        end
+    end
```
ffreyer, line 222: This is missing an else branch which is why tests are failing. Might also be good to reorganize this so that `p.arrow_size` can change from automatic to some value. I.e. use `arrow_size = map(p.arrow_size, p.limits, p.gridsize) do ...`.

ffreyer (approved):

chillenb:
@ffreyer thanks for cleaning this up. Much appreciated.

PR #3527, opened 2023-12-31
Add voxel plot type

ffreyer:
# Description

This is an experiment with a dedicated voxel rendering function using the meshing ideas presented here: https://www.youtube.com/watch?v=4xs66m1Of4A

The initial commit only implements plain voxel rendering, but can handle 100 Million Voxels without much of an issue for me.

```julia
scene = Scene()
cam3d!(scene)
p = voxel!(scene, rand(UInt8, 1024, 1024, 128)) # 134 217 728 Voxels
l = linesegments!(scene, boundingbox(p))
center!(scene)
display(scene)
```

Screenshot from 2023-12-31 19-34-10

TODO:
- [x] add lighting
- [x] replace `Matrix{UInt8, 3}` restriction with automatic compression
  - [x] add colormaps (compressed id -> color)
  - [x] add texturemaps (id, side -> uv -> texture fetch)
- [x] fast update function (local update of chunk data)
- [x] allow users to control what becomes air/invisible (currently NaN, nothing and missing)
- [x] test actual textures
- [x] improve render order
    - going front to back from eyeposition should improve performance
    - going back to front should enable correct transparency
- [x] implement lowclip and highclip (and colorrange?) by adjusting voxel id generation and colormap
- [x] WGLMakie (requires https://github.com/SimonDanisch/ShaderAbstractions.jl/pull/26)
    - [x] fix rendering issues (moving planes and maybe z-fighting)
- [x] maybe CairoMakie
    - [x] enable tests
- [ ] maybe RPRMakie (TODO: try it)
- [x] allow Colon() in `local_update`
- [x] ~~test alternating x/y/z planes before incrementing the plane index for transparency~~ test some possible improvements to transparency
    - alternating x/y/z direction doesn't work doesn't (easily) work with varied sizes
    - adjusting dimensional order based on view direction results in very noticeable switches
    - shifting planes slightly closer to the viewer along the normal direction doesn't help
- ~~px_per_unit~~ doesn't apply
- [x] test picking
- [x] fix missing z plane when rendering without gap one weaker GPU (and with missing voxels)
- [x] fix render order (consider scaling on voxels)

To discuss:
- Should resizing be possible? (Requires workaround for `resize!(::Array{UInt8, 3})`)
- Should the order of sides in uvmap change?

## Type of change

## Checklist


ffreyer:
Taking an example from the volume docs:

```julia
r = LinRange(-1, 1, 100)
cube = [(x.^2 + y.^2 + z.^2) for x = r, y = r, z = r]
cube_with_holes = cube .* (cube .> 1.4)
voxels = map(x -> ifelse(1.65 <= x <= 1.75, x, NaN), cube_with_holes)
f, a, p = voxel(voxels)
```

image

ffreyer:
Texture mapping that needs more testing:

```julia
voxels = UInt8[
    1 0 2; 0 0 0; 3 0 4;;;
    0 0 0; 0 0 0; 0 0 0;;;
    5 0 6; 0 0 0; 7 0 8;;;
]

uv_map = [ # idx -> LRBT
    Vec4f(0.00, 0.0, 0.5, 0.25),
    Vec4f(0.00, 0.5, 1.0, 0.25),
    Vec4f(0.25, 0.0, 0.5, 0.50),
    Vec4f(0.25, 0.5, 1.0, 0.50),
    Vec4f(0.50, 0.0, 0.5, 0.75),
    Vec4f(0.50, 0.5, 1.0, 0.75),
    Vec4f(0.75, 0.0, 0.5, 1.00),
    Vec4f(0.75, 0.5, 1.0, 1.00),
]

texture = [
    RGBf(0, 0, 0) RGBf(1, 0, 0);
    RGBf(0, 1, 0) RGBf(0, 0, 1);
    RGBf(1, 1, 1) RGBf(1, 1, 0);
    RGBf(0, 1, 1) RGBf(1, 0, 1);
]

f, a, p = voxel(voxels, uvmap = uv_map, color = texture)
```

image

ffreyer:
Using textures from https://www.kenney.nl/assets/voxel-pack for testing.

Texture test without orientations:
```julia
# 9 wide, 10 tall
texture = Matrix(FileIO.load("...")')

uv_map = [
    Vec4f(x, y, x+1/9, y+1/10)
    for x in range(0.0, 1.0, length = 10)[1:end-1]
    for y in range(0.0, 1.0, length = 11)[1:end-1]
]

voxels = UInt8[
    1 0 2; 0 0 0; 7 0 8;;;
    0 0 0; 0 0 0; 0 0 0;;;
    9 0 10; 0 0 0; 11 0 20;;;
]

f, a, p = voxel(voxels, uvmap = uv_map, color = texture)
```

Screenshot from 2024-01-02 17-52-40

With orientations: 
```julia
# 9 wide, 10 tall
texture = Matrix(FileIO.load("...")')

uvs = [
    Vec4f(x, y, x+1/9, y+1/10)
    for x in range(0.0, 1.0, length = 10)[1:end-1]
    for y in range(0.0, 1.0, length = 11)[1:end-1]
]
# -x -y -z x y z
uv_map = Matrix{Vec4f}(undef, 4, 6)
uv_map[1, :] = [uvs[9],  uvs[9],  uvs[8],  uvs[9],  uvs[9],  uvs[8]] # birch
uv_map[2, :] = [uvs[11], uvs[11], uvs[10], uvs[11], uvs[11], uvs[10]] # oak
uv_map[3, :] = [uvs[2],  uvs[2],  uvs[2],  uvs[2],  uvs[2],  uvs[18]] # crafting table
uv_map[4, :] = [uvs[1],  uvs[1],  uvs[1],  uvs[1],  uvs[1],  uvs[1]] # planks

voxels = UInt8[
    1 0 1; 0 0 0; 1 0 1;;;
    0 0 0; 0 0 0; 0 0 0;;;
    2 0 2; 0 0 0; 3 0 4;;;
]

f, a, p = voxel(voxels, uvmap = uv_map, color = texture)
```

Screenshot from 2024-01-02 17-47-33

src/types.jl
```diff
@@ -241,9 +241,9 @@ struct Camera
     resolution::Observable{Vec2f}
 
     """
-    Focal point of the camera, used for e.g. camera synchronized light direction.
+    Direction in which the camera looks.
     """
-    lookat::Observable{Vec3f}
```
ffreyer, line 246: I added this in the lighting pr but didn't end up using it. Swapped it out for view_direction here, which is used to sort planes. Might also be useful in volume rendering?

ffreyer:
The Makie/GLMakie side of this is mostly done, so now would be a good time to try this out. I also wrote a docs page for it: https://docs.makie.org/previews/PR3527/reference/plots/voxel/

@lazarusA @ashwani-rathee Maybe you have some wishes for this?

Some things I've been thinking about but haven't added/changed:
- xs, ys and zs for scaling the chunk. Currently this only works through `translate!` and `scale!`
- allowing chunk size to change (why can't you just resize and `Array{T, 3}`? )
- changing the order of faces for uv lookups

ffreyer:
WGLMakie will continue to fail until https://github.com/SimonDanisch/ShaderAbstractions.jl/pull/26 is merged.

I haven't tested RPRMakie because none of the plugins produce anything but a blank image for me in general. I'm probably doing something wrong but I don't have the will to figure out what. The code is just a stripped down version of meshscatter so it should be fairly easy to debug if something is wrong here.

I renamed this the plot type from `voxel` to `voxels` at some point. I feel like `voxel` is more natural though, so maybe we should change that back?

Otherwise this is pretty much done now.

GLMakie/assets/shader/voxel.frag
```diff
@@ -0,0 +1,118 @@
...
+    voxel_color.rgb = illuminate(o_normal, voxel_color.rgb);
+#endif
+
+    // TODO: index into 3d array
+    ivec3 size = ivec3(textureSize(voxel_id, 0).xyz);
+    ivec3 idx = ivec3(o_uvw * size);
+    int lin = 1 + idx.x + size.x * (idx.y + size.y * idx.z);
+
+    // draw
+    write2framebuffer(voxel_color, uvec2(objectid, lin));
```
ffreyer, line 117: I just realized that with 32 bit integers there is a risk of overflow here. Rendering a 2048x2048x1024 chunk of voxel data eats up the whole range, while taking 4GB of vram and being otherwise no problem for me to render. (At least with random data.) This should be noted in the docs. (And also the todo here needs to be removed)

ffreyer: Added a small section about this to the docs

ffreyer:
WGLMakie still has some z-fighting artifacts

image

GLMakie doesn't have these. I wonder if that's a perhaps a different depth test (e.g. >= instead of >)? 

Fixing this by shifting planes is not possible. To avoid the artifacts above, you'd want to shift planes in +- normal direction (whichever moves them closer to the camera). Doing that would however open up a gap for the front most voxels, so you'd have new artifacts.

ffreyer:
I tried this pr on my Laptop and noticed a plane wasn't rendering on a 500x500x200 chunk. I fairly certain that was a float precision issue that got fixed by switching to integer modulo in the shaders. I wanted to use those in the first place but thought glsl only had float mods for some reason...

While working on that I also noticed the render order was not what I intended it to be so I fixed that too. Good render order should be beneficial for performance but I'm not sure how much it really is. If tests look good this should be ready to be merged again.

PR #3529, opened 2024-01-04
tag v0.20.4

SimonDanisch:
Prepare new release and adds:
```
"""
    available_plotting_methods()

Returns an array of all available plotting functions.
"""
```
and exports available_marker_symbols

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -728,7 +728,7 @@ function update_cam!(scene::Scene, cam::Camera3D)
         far = far_dist * far
     elseif cam.settings.clipping_mode[] === :adaptive
         view_dist = norm(eyeposition - lookat)
-        near = view_dist * near; far = max(radius(bounding_sphere), view_dist) * far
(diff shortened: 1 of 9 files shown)


src/camera/camera3d.jl
```diff
@@ -728,7 +728,7 @@ function update_cam!(scene::Scene, cam::Camera3D)
         far = far_dist * far
     elseif cam.settings.clipping_mode[] === :adaptive
         view_dist = norm(eyeposition - lookat)
-        near = view_dist * near; far = max(1f0, view_dist) * far
```
ffreyer, line 731: The `max(1f0, view_dist)` is supposed to compensate `norm(cam.eyeposition[] - cam.lookat[])` when zooming out and hold far when zooming in. Of course `1f0 / norm(...)` will give different values depending on limits, so sometimes this will work and sometimes not.
The fix now uses the bounding sphere radius instead, which should always be bigger than `norm(...)`, holding far at a safe value.

Issue #3532, opened 2024-01-05
Legend not working when using custom colorrange


howbgl:
- [x] are you running newest version (version from docs) ? Makie `v0.20.4` & CairoMakie `v0.11.5`
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x] What platform + GPU are you on? 
```julia
Julia Version 1.10.0
Commit 3120989f39b (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × AMD Ryzen 5 5625U with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
  Threads: 1 on 12 virtual cores
```

# MWE
```julia
using CairoMakie

f = Figure()
ax = Axis(f[1,1])
for i in 1:8
    lines!(ax,rand(16);colorrange=(1,8),color=i,label="$i")
end
axislegend(ax)
```

Similar results occur when using `scatter!` instead of `lines!` and `f[1, 2] = Legend(f, ax, "Legend")` instead of the call to `axislegend`.

The latest version where the MWE is not broken is Makie `v0.19.5` and CairoMakie `v0.10.6`.

# Stacktrace
```julia
ERROR: Cannot determine a colorrange automatically for single number color value. Pass an explicit colorrange.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] assemble_colors(::Int64, color::Observable{Any}, plot::Lines{Tuple{Vector{Point{2, Float32}}}})
    @ Makie ~/.julia/packages/Makie/fyNiH/src/colorsampler.jl:364
  [3] color_and_colormap!(plot::Lines{Tuple{Vector{Point{2, Float32}}}}, colors::Observable{Any})
    @ Makie ~/.julia/packages/Makie/fyNiH/src/interfaces.jl:14
  [4] color_and_colormap!(plot::Lines{Tuple{Vector{Point{2, Float32}}}})
    @ Makie ~/.julia/packages/Makie/fyNiH/src/interfaces.jl:10
  [5] calculated_attributes!
    @ Makie ~/.julia/packages/Makie/fyNiH/src/interfaces.jl:98 [inlined]
[...]

PR #3536, opened 2024-01-07
fix legend colorrange

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/3532.

## Type of change

## Checklist

### DIFF
--- a/ReferenceTests/src/tests/figures_and_makielayout.jl
+++ b/ReferenceTests/src/tests/figures_and_makielayout.jl
@@ -133,18 +133,34 @@ end
     end
 end
 
-@reference_test "Legend colorrange" begin
+@reference_test "Legend lines colorrange" begin
     f = Figure()
     ax = Axis(f[1, 1])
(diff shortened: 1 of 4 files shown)


t-bltg:
New reference images:

leg1
leg2

jkrumbiegel:
This doesn't copy the colormap I think, so it would break if your test didn't use the default. Could you add the same treatment (and tests) for marker and poly elements as well? Then everything would work the same way.

t-bltg:
> This doesn't copy the colormap I think, so it would break if your test didn't use the default.

Good point, I will check that.

> Could you add the same treatment (and tests) for marker and poly elements as well?

Will do.

t-bltg:
@jkrumbiegel , I've update the examples in https://github.com/MakieOrg/Makie.jl/pull/3536#issuecomment-1880028133, added reference tests, and fixed propagation of the `colormap` attribute.

Does that fix your requests ?

jkrumbiegel:
Thanks again for the PR, I've pushed another PR that does a slightly different approach over here https://github.com/MakieOrg/Makie.jl/pull/3587 but this helped me get that one done

Issue #2974, opened 2023-05-23
Clean up dependencies and maybe use PackageExtension


SimonDanisch:
# Good candidate for PackageExtension, since user needs to load the package to use the feature
* SparseArrays 
* OffsetArrays

# Not sure how to use PackageExtension but packages feel optional
* InteractiveUtils
* Isoband
* KernelDensity
* StatsFuns 
* PolygonOps
* TriplotBase 

# Can be removed completely with little effort
* StructArrays 
* UnicodeFun 
* Match
* Setfield

# Needed only for texture atlas (maybe nice to move out for CairoMakie?)

* SignedDistanceFields 
* StableHashTraits 
* Packing 

We should also check if all `__init__` functions are fast for all dependencies and Makie itself.

bilderbuchi:
Related: #2505. 

> It would make usability a little worse, though, since naively using `density` wouldn't work.

Could you emit an actionable error message for users hitting that situation? `Please install package XYZ to use this feature`?

bilderbuchi:
> Related: #2505 (additional candidate for PackageExtension: `FFMPEG`?)
> 
> > It would make usability a little worse, though, since naively using `density` wouldn't work.
> 
> Could you emit an actionable error message for users hitting that situation? `Please install package XYZ to use this feature`?

Apparently such a technique is now available (from https://github.com/MakieOrg/Makie.jl/issues/3953#issuecomment-2267145860):
> users who try eg `qqplot()` will see an error saying "please execute using Distributions to use qqplot()". It clear and unambiguous what exactly to do.

PR #3546, opened 2024-01-11
Remove deps, e.g. Setfield

SimonDanisch:
Make some progress on https://github.com/MakieOrg/Makie.jl/issues/2974


src/basic_recipes/text.jl
```diff
@@ -359,7 +383,8 @@ function apply_lineheight!(lines, lh)
     for (i, line) in enumerate(lines)
         for j in eachindex(line)
             l = line[j]
-            l = Setfield.@set l.origin[2] -= (i-1) * 20 # TODO: Lineheight
+            ox, oy = l.origin
```
jkrumbiegel, line 386: Should keep the todo comment here as it seems there's a hardcoded line height still?

jkrumbiegel (changes requested):

jkrumbiegel (approved):

PR #3552, opened 2024-01-14
Fix CairoMakie transforms

### LINKED ISSUE #3545: Issues with h/vspan and h/hspan in log space
`vspan`, `vspan!`,  `hspan`, and `hspan!` do not work if the x-axis is log-scaled. 

`vspan!` and `hspan!` have the issue with limit transforms that was discussed in #3271 and partially fixed in #3427. It seems like #3427 fixed the limits in `vspan` and `hspan`, but not their in-place version.

 In addition,  none of the functions produce a visible band when the axes are log-scaled, independent of triggering the limit issue discussed above. The fix of the first issue should be relatively simple, since it's been fixed for `hspan` and `vspan`, but I'm unsure on how to fix the other two.

ffreyer:
# Description

Fixes #3545

Poly was using the scene transform_func rather than the plot transform_func, which can be different. For h/vspan that is the case, causing different results.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/overrides.jl
+++ b/CairoMakie/src/overrides.jl
@@ -45,7 +45,7 @@ end
 function draw_poly(scene::Scene, screen::Screen, poly, points::Vector{<:Point2}, color::Union{Colorant, Cairo.CairoPattern},
         model, strokecolor, strokestyle, strokewidth)
     space = to_value(get(poly, :space, :data))
-    points = project_position.(Ref(scene), space, points, Ref(model))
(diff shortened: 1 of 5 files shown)


ffreyer:
Refimg sets `Axis` transform function to log10 and then plots with `transformation = Transformation()`:

Screenshot from 2024-01-14 15-38-49

If the backend uses the scene transform function this should result in strong differences, e.g. for CairMakie master:

Screenshot from 2024-01-14 15-39-45

PR #3573, opened 2024-01-29
fix redisplaying figures on same screen

SimonDanisch:
The following would retain previously scene objects:
```julia
using GLMakie
GLMakie.activate!(float=true)

screen = GLMakie.Screen(; size=(600, 450))

f1 = Figure()
Axis(f1[1, 1])
display(screen, f1)

sleep(3)

f2 = Figure()
Axis3(f2[1, 1])
display(screen, f2)

display(screen, Figure().scene)
```
This PR fixes it!


jkrumbiegel:
can confirm that this fixed the mwe on my mac

Issue #3519, opened 2023-12-28
CairoMakie `display` mutates package directory


ericphanson:
https://github.com/MakieOrg/Makie.jl/blob/a1d85a1d9ef9ab7397f8197937622634eacb1c8d/CairoMakie/src/display.jl#L34-L56

This causes failures in https://github.com/KristofferC/PkgFsck.jl and doesn't seem like a good idea in general (e.g. I think some folks use read-only shared depots in settings like HPC).

jkrumbiegel:
Agreed, I forgot about that but it's very old behavior and doesn't make any sense :)

SimonDanisch:
Oh yeah,  I guess this should be a tempdir or scratch space...

PR #3588, opened 2024-02-01
Use a temp dir instead of mutating the CairoMakie src dir

jkrumbiegel:
Fixes https://github.com/MakieOrg/Makie.jl/issues/3519

I think a temp dir is the simplest solution. These should theoretically be cleaned up by Julia before exiting, so it shouldn't litter the file system. I also think it's better to make new files rather than reuse one, because a user might look at different versions at the same time.


jkrumbiegel:
@SimonDanisch any idea why this would break relocatability of glmakie?

PR #3612, opened 2024-02-07
Fix CairoMakie erroring when recipes return lists of PlotSpec

asinghvi17:
# Description

CairoMakie suffered from a method ambiguity when a `convert_arguments` returned a list of plotspecs using the new SpecApi.  This PR eliminates that method ambiguity and adds a reference test for it, so we know that all backends support plotlists!

I will probably need some help updating the reference images though!

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed an issue where CairoMakie would unnecessarily rasterize polygons [#3605](https://github.com/MakieOrg/Makie.jl/pull/3605).
+- Fixed a method ambiguity in CairoMakie when plotting lists of PlotSpecs (`plotlist`) [#3612](https://github.com/MakieOrg/Makie.jl/pull/3612).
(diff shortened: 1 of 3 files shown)


asinghvi17:
Hey, just want to bump this - would be nice to get it out there!

SimonDanisch:
On master the test case you added works for me.
Which Julia version did you try?
Also, the test case should be more like:
```julia
struct PlotlistTestType end

function Makie.convert_arguments(::Type{<: Makie.Plot}, obj::PlotlistTestType)
    return [
        Makie.SpecApi.Lines([0, 1], [0, 1]),
        Makie.SpecApi.Scatter([0.5], [0.5]),
    ]
end
plot(PlotlistTestType())
```
Or was there a good reason to define `convert_arguments` for `Lines`?

asinghvi17:
Huh!  Yeah I just tested with latest master and this was not an issue there...maybe my local env was misconfigured?  Will add the test you suggested at a later point.

About the `lines` thing: my usecase was specifically allowing plots to "pass-through", so `lines` seemed like the best candidate just to make sure nothing got in the way there.

Issue #3619, opened 2024-02-09
CairoMakie heatmap pixels overlap


maxfreu:
I want to create a heatmap of a non-square matrix with square pixels. I'm trying to achieve this by setting a colsize of defined aspect. This however leads to somehow staggered pixels and strange effects, as CairoMakie seemingly draws the pixels larger than they actually are. 

Some demo code and the corresponding PDF opened in inkscape:

```julia
fig = Figure(size = mm2pt.((83, 50)))
ax = Axis(fig[1, 1])
r = rand(20,30)
colsize!(fig.layout, 1, Aspect(1, size(r)[1] / size(r)[2]))
heatmap!(ax, r; colorrange = (0, 1))
save("deleteme.pdf", fig)
```

![Screenshot from 2024-02-09 17-10-58](https://github.com/MakieOrg/Makie.jl/assets/67329240/7164a5b3-dc16-4209-94a2-d5fbfcc60b66)

The problem:

![Screenshot from 2024-02-09 17-27-21](https://github.com/MakieOrg/Makie.jl/assets/67329240/9eb24448-8ca6-4023-97d5-61cd50571655)

Any idea how I can make the pixels square without this effect?

Makie 0.20.7, CairoMakie 0.11.8, julia 1.10

jkrumbiegel:
CairoMakie draws heatmap cells larger than they are so that you don't get white antialiasing artifacts at the seams. It should not draw them larger if the adjacent cells are not fully opaque, but maybe in your case the cells are even spanning across more than one neighbor, so they peek through. So we have to check that the overdrawing width is at most some fraction of the neighboring cell so that this cannot happen.

maxfreu:
https://github.com/MakieOrg/Makie.jl/blob/f92b2607de3788a6c2b45eb5fd5546fb634fab3a/CairoMakie/src/primitives.jl#L785-L795

here, right?

PR #3625, opened 2024-02-12
Fix CairoMakie heatmap cell sizes

maxfreu:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/3619

CairoMakie heatmap cells are currently enlarged by a constant factor of 0.5. This PR changes this to a value relative to the pixel size.

Before:

After:
image

There are still some remaining inaccuracies, but its way better than before. This is really important for drawing halfway accurate maps.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -771,7 +771,15 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Unio
     end
 end
 
-function _draw_rect_heatmap(ctx, xys, ni, nj, colors)
+function _draw_rect_heatmap(ctx, xys, ni, nj, colors, margin_factor=0.05f0)
+    # dx and dy are the margin widths that are added to the
+    # heatmap cell sizes further down
+    # adding 5% seems like a good compromise between avoiding artifacts
+    # and not changing pixel sizes too much
+    # how much is added can be controlled via margin_factor
+
+    dx = abs(xys[2,1][1] - xys[1,1][1]) * 0.5f0 * margin_factor
+    dy = abs(xys[1,2][2] - xys[1,1][2]) * 0.5f0 * margin_factor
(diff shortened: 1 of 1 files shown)


maxfreu:
The heatmap_interpolation.png reference image differs slightly, that's probably why the test failed. If you tell me where to find the code generating the reference output, I can try to fix it. Looks like the pixels within this reference image have variable size; I expected all to be the same, but it's easy to adapt the margin on a per-pixel basis.

maxfreu:
Margins are now calculated per individual pixel, which should get rid of the difference in the test image.

maxfreu:
Ok, I think it's not so easy to get this perfectly right, as padding the pixels is already a bit hacky. As the heatmaps only look wrong if NaN is present, I now check for it and:

1) If no NaN is present, everything should stay as it was.
2) If NaN is present, the pixel margins are reduced to 5% the pixel width. This introduces aliasing in some cases, but doesn't oversize the pixels completely.

CairoMakie/src/primitives.jl
```diff
@@ -777,21 +777,20 @@ function _draw_rect_heatmap(ctx, xys, ni, nj, colors, margin_factor=0.05f0)
...
-        dx = abs(p2[1] - p1[1]) * 0.5f0 * margin_factor
-        dy = abs(p4[2] - p1[2]) * 0.5f0 * margin_factor
-
+        
         # Rectangles and polygons that are directly adjacent usually show
         # white lines between them due to anti aliasing. To avoid this we
         # increase their size slightly.
-
+        
         if alpha(colors[i, j]) == 1 && margin_factor > 0
```
asinghvi17, line 785: What happens if using a transparent colormap here?

maxfreu: The problem mostly goes away, but when you set alpha=0.99 for example, there are still some cases where artifacts occur. For example thin lines that shouldn't be there.

PR #3565, opened 2024-01-24
add typed argument conversion

SimonDanisch:
This is step 1 for #3226, to clean up `convert_arguments`, to be able to better know when a conversion was successful or needs axis converts.

This should:
* improve error messages for types that don't convert for the plot type
* allow better recursive application of `convert_arguments` 
* closes #3548

PR #3626, opened 2024-02-13
Documented / validated plot attributes

jkrumbiegel:
## Description

Supersedes https://github.com/MakieOrg/Makie.jl/pull/2966

This PR extends the `@recipe` macro with a second method that takes a `begin` block instead of a `do scene` block. The internal machinery for this recipe sets up similar functions to those already known from `Block`s. With those, we can query whether keywords passed to `Plot` objects are actually valid. The behavior is opt-in by using the new `@recipe` syntax. With the old syntax, we just don't do any validation.

With this new machinery we can also update the docs to be per-attribute like those for `Block`s, with additional examples we can add afterwards. My goal with this PR is not to let it get too huge by adding everything at once. We can add additional features for enhanced usability on top of this later.

## Example

```julia
@recipe MyPlot positions begin
   "This attribute does nothing"
   my_attribute = 123 
end

myplot(; my_attribut = 456)
```
```
ERROR: ArgumentError: Invalid attribute for plot type Plot{myplot}: my_attribut. The available attributes are: my_attribute.
```

# TODOs

- [ ] add failing arrows test back in (I think the new behavior is actually the correct intended one)


ffreyer:
Could we add something like `plot!(..., passthrough = (; internal_attrib = ...))` to skip this check? I think otherwise this breaks some of the hackier stuff where you overwrite defaults generated in backends. That might also help separate hacks from standard usage

jkrumbiegel:
@ffreyer I'd go case by case throught the errors this throws, I'm not sure in what cases what you wrote is applicable. Maybe some can be refactored? Some will be plain bugs (already got some of those). Either we allow-list some specific ones for the backends or we do a passthrough like you suggested. Not sure yet

SimonDanisch:
>Edit: Just found out about interpolate_in_fragment_shader through the CI failure :D

That's an internal attribute like the computed once...
I think they should be completely out of this... We could more clearly prefix all with computed, or add a completely new way to handle those computed attributed ... Which would be nice but I guess will needs more work!

jkrumbiegel:
Computed ones could be added to plot objects after creation, that circumvents the validation mechanism.

ffreyer:
> @ffreyer I'd go case by case through the errors this throws, I'm not sure in what cases what you wrote is applicable. Maybe some can be refactored? Some will be plain bugs (already got some of those). Either we allow-list some specific ones for the backends or we do a passthrough like you suggested. Not sure yet.

To be honest I was just thinking that I added a `debug` uniform in the lines pr and it would be annoying to have to edit MakieCore as well to be able to use it. I guess more generally it's quite useful for prototyping like this.
It might also be useful to some to be able to add extra metadata to plots. E.g. you could add mass and gravity to a scatter plot if you do some kind of physics simulation. Or if a third party makes something that interacts with plots like DataInspector or Legend, it would be useful for them to be able to add some metadata (like label, inspectable, etc) for that.

jkrumbiegel:
> you could add mass and gravity to a scatter plot if you do some kind of physics simulation

I think those usecases, while interesting, distract from the fact that our core library is pretty bug-prone due to not erroring. I'd really want to fix that first before thinking about people with niche use cases. Right now, basically anybody starting to use Makie runs into a lot of frustration first when they mistype anything.

> it would be annoying to have to edit MakieCore as well to be able to use it

To me, MakieCore has served its purpose and should probably be merged back into Makie, now that package extensions exist. That would make developing easier as well, although we can additionally think about developer escape hatches for the validation. If it annoys you during development, you can always comment out the validation function in `recipes.jl` I guess? Or add the attribute in question to the recipe macro and reevaluate it.

ffreyer:
> > you could add mass and gravity to a scatter plot if you do some kind of physics simulation
> 
> I think those usecases, while interesting, distract from the fact that our core library is pretty bug-prone due to not erroring. I'd really want to fix that first before thinking about people with niche use cases. Right now, basically anybody starting to use Makie runs into a lot of frustration first when they mistype anything.

I agree that the main purpose of this pr should be to add that error checking. But I also think we shouldn't just remove the ability to add your own data to a plot. Having `passthrough::NamedTuple` would be pretty simple and it doesn't stop you from doing error checking for everything else. I also makes this less breaking for people that rely on custom attributes and gives us some more time to fix things in case we miss some attributes here.

> > it would be annoying to have to edit MakieCore as well to be able to use it
> 
> To me, MakieCore has served its purpose and should probably be merged back into Makie, now that package extensions exist. That would make developing easier as well, although we can additionally think about developer escape hatches for the validation. If it annoys you during development, you can always comment out the validation function in `recipes.jl` I guess? Or add the attribute in question to the recipe macro and reevaluate it.

Whether I add an attribute in MakieCore, Makie or comment out the validation is all kind of the same. It's also not really that big of a deal I guess.

jkrumbiegel:
> But I also think we shouldn't just remove the ability to add your own data to a plot.

You still can, just not via keyword arg. You can always do `attributes(plt)[:special_key] = "Some stuff"`. Or is that not good enough for your purposes?

ffreyer:
For something like
```julia
on(my_obs) do ...
    val = to_value(get(plot.attributes, :special_key, default))
    ...
end
```
that's fine but for
```julia
onany(my_obs, get(plot.attributes, :special_key, default_obs)) do ...
    ...
end
```
it's not.

jkrumbiegel:
Not sure I quite get it, you mean that the special attribute would not be an observable?

ffreyer:
Oh right it depends on context a bit. GLMakie does the latter, so if you run something like
```
f, a, p = plot(...)
p.attributes[:special_key] = something
```
line by line it will not have the extra attribute when it generates the render object. For other things that's not much of a problem?

jkrumbiegel:
Ah ok I have almost no experience with those render objects so I have not thought about them much. Would it be correct to say though that almost nobody aside from us is going to mess with these, and that you can still pass new properties to the backend by just overwriting the recipe with the additional attribute added?

Then, the remaining use case would be attaching arbitrary data for various reasons, and while that's hacky, it will still be possible as I showed above.

I think you can actually do what you want, too, if you first create the `Plot` (which runs the validator) and then modify that `Plot` before `plot!`ing it. I can trigger a GLMakie error for an unknown data type (String) for gl_convert with this, so I assume it was passed along correctly.

```
p = Plot{lines}((1:500,), Dict{Symbol,Any}())
p.attributes[:special] = "my value"
s = Scene(camera = campixel!)
plot!(s, p)
s
```

(The error also specifying the key is new on this branch)

```
┌ Error: gl_convert for key `special` failed
└ @ GLMakie.GLAbstraction ~/.julia/dev/Makie/GLMakie/src/GLAbstraction/GLTypes.jl:388
ERROR: MethodError: no method matching gl_convert(::String)
```

asinghvi17:
Alternatively you could just not display:
```julia
f, a, p = lines(rand(10));
p.special = :hello
f # display
```
though it is a bit of a pain to remember.

jkrumbiegel:
```
p.special = :hello
```

I wonder if this should actually work on plots, adding random attributes if you misspell one is also not great, no?
Better to force `attributes(plt)[:new_key] = ...`

jkrumbiegel:
This is ready for review now. I would say we keep the second `@recipe` syntax as an internal implementation detail for a bit, this way we also don't have to worry too much about the `@inherit` and `@mixin` syntax, and can modify these going forward however we think it's best.

Merging this PR would then also allow to restructure docs and docstrings for the affected plotting functions, so that we slowly transfer everything to attribute and argument examples like we have started for Blocks.

SimonDanisch:
I haven't had an in-depth look at it yet, but first impression is good! :) 
I don't like the new macro syntax as much, but I'm happy to leave that for a new PR, where we e.g. also add the new macro for typed converts from #3565.

I do think we need to fix the performance regressions as well as we can though. I don't expect checking & documenting the attributes should change compile and runtime as much, so I expect there is something we could do better here (most docstring creation etc should happen at precompile times, and just checking against a dict shouldn't add ~15% overhead for plot creation)!

jkrumbiegel:
Good point about the performance, did not check that yet at all. 

> I don't like the new macro syntax as much, but I'm happy to leave that for a new PR, where we e.g. also add the new macro for typed converts from https://github.com/MakieOrg/Makie.jl/pull/3565.

For now I see this just as a necessary "evil" to support attribute validation, we can iterate on the design of the DSL going forward.

ffreyer:
> ```
> p.special = :hello
> ```
> 
> I wonder if this should actually work on plots, adding random attributes if you misspell one is also not great, no? Better to force `attributes(plt)[:new_key] = ...`

I like that the syntax exists because attributes are more or less the only thing in a plot that users should directly interact with. Adding `attributes(plt)` just seems like an unnecessary inconvenience here. I do think this should have error checking though. Maybe we just add are more explicit "add attribute" method for this, and do checking for normal indexing?

Also would be nice if the error message had the same grouping as the plot docs. That would make it easier to find the attribute you meant to use.

jkrumbiegel:
> do checking for normal indexing

That's what I tried to say, I meant that adding observables via getindex! overload seems dangerous, would be better to make that explicit if wanted. Normal users should basically never want to add observables to plot objects, they'll always want to change existing attributes' values.

> Also would be nice if the error message had the same grouping as the plot docs

You mean the column style I added? The problem is that the docstring is the REPL rendering a markdown object, so I don't control printing in the same way. I would like it if the docstring rendered like that, too.

jkrumbiegel:
Ok performance regression seems to have been changed into slight performance improvement. Any other comments?

ffreyer:
Potential TODO: 

deprecate `meshscatter.rotations` here or in a follow-up pr to close #3414

MakieCore/src/recipes.jl
```diff
@@ -189,6 +189,365 @@ macro recipe(theme_func, Tsym::Symbol, args::Symbol...)
...
+            kwdict = Dict{Symbol, Any}(kw)
+             _create_plot!($funcname, kwdict, args...)
+        end
+        function MakieCore.is_attribute(T::Type{<:$(PlotType)}, sym::Symbol)
+            sym in MakieCore.attribute_names(T)
+        end
+
+        function MakieCore.attribute_names(::Type{<:$(PlotType)})
+            keys(MakieCore.$(attr_placeholder).d)
+        end
```
asinghvi17, line 394: Is it possible to easily override this or add new attributes, when e.g. editing the recipe / adding new rendering methods?

jkrumbiegel: Not sure what you mean, when would you want to do this? You can always reevaluate the edited recipe when you want to make changes

asinghvi17: I was specifically looking at adding another keyword to control `poly` rendering in CairoMakie, so was wondering if that could be picked up on only when CairoMakie is loaded, for example.  Or, if some extension from another package offers a rendering feature, it could hook in to say "I consume this keyword argument so mark it as valid" or something.

jkrumbiegel: No not currently. When discussing backend attributes with Simon we decided it's usually better to go via types for special backend functionality, and that all available attributes should ideally work across all backends. Even if that's not achievable it should always be the goal, so it shouldn't be easy to tack on keywords.

jkrumbiegel:
@SimonDanisch I have converted all of Makie's recipes now, the only thing remaining is that one failing reftest of arrows. But as I write above, I think that was actually a bug before, because the implementation seemed to intend that arrow tips and lines are colored the same automatically.

MakieCore/src/recipes.jl
```diff
@@ -189,6 +189,381 @@ macro recipe(theme_func, Tsym::Symbol, args::Symbol...)
...
+            Attributes(documented_attributes(T).closure(scene))
+        end
+
+        function MakieCore.attribute_default_expressions(T::Type{<:$PlotType})
+            Dict(k => v.default_expr for (k, v) in documented_attributes(T).d)
+        end
+
+        function MakieCore._attribute_docs(T::Type{<:$PlotType})
+            Dict(k => v.docstring for (k, v) in documented_attributes(T).d)
+        end
```
SimonDanisch, line 419:
```julia
        function MakieCore.is_attribute(T::Type{<:$(PlotType)}, sym::Symbol)
            sym in MakieCore.attribute_names(T)
        end

        function MakieCore.attribute_names(T::Type{<:$(PlotType)})
            keys(documented_attributes(T).d)
        end

        function $(MakieCore).default_theme(scene, T::Type{<:$PlotType})
            Attributes(documented_attributes(T).closure(scene))
        end

        function MakieCore.attribute_default_expressions(T::Type{<:$PlotType})
            Dict(k => v.default_expr for (k, v) in documented_attributes(T).d)
        end

        function MakieCore._attribute_docs(T::Type{<:$PlotType})
            Dict(k => v.docstring for (k, v) in documented_attributes(T).d)
        end
```
These functions could be defined outside the macro, can they?

ReferenceTests/src/tests/examples3d.jl
```diff
@@ -199,27 +199,27 @@ end
...
+
+#     pts = vec(Point3f.(x, y, z))
+#     ∇ˢF = vec(∇ˢf.(x, y, z)) .* 0.1f0
+#     surface(x, y, z)
+#     arrows!(
+#         pts, ∇ˢF,
+#         arrowsize=0.03, linecolor=(:white, 0.6), linewidth=0.03
+#     )
+#     current_figure()
+# end
```
SimonDanisch, line 222: What do we need to bring this back?

MakieCore/src/recipes.jl
jkrumbiegel: true, good catch

ReferenceTests/src/tests/examples3d.jl
jkrumbiegel: well the arrows recipe had this part 
```julia
# connect arrow + linecolor by default
get!(attr, :arrowcolor, attr[:linecolor])
```
So I assumed it was wanted that if `linecolor` is set, `arrowcolor` copies this by default. But this wasn't actually the behavior, so that test which sets `linecolor` had black arrow tips (which is the `color` default). That's why the tests failed. So either we fix the behavior to match the old one, or we update the refimages to the new (fixed) behavior.

SimonDanisch: Ok, lets update the reference image then! That's something that should definitely work I think!

jkrumbiegel: yeah not sure, maybe it's simpler that `linecolor` and `arrowcolor` copy from `color` if it's set, not that `arrowcolor` also copies from `linecolor`.

SimonDanisch: Ah, that's a better idea!

jkrumbiegel:
Oh whoops I guess taking out that one method broke validation for plotting functions that don't use the new system

SimonDanisch (approved):
Approving this to be merged into breaking-21.
I think we need to test this a bit (e.g. with some dependencies) and maybe add some more tests for corner cases.
But, we need to do that for the other changes in breaking-21 as well, so I think we should do that one time in the end before tagging 0.21.

PR #3628, opened 2024-02-13
[WIP] Improvements to PlotSpecs and the SpecAPI

asinghvi17:
# Description

This is a WIP set of improvements/changes to the SpecAPI after some usage. 

- Forward PlotList attributes to plots at creation time (this doesn't allow post facto editing, but e.g. `plot(diffeq_solution; linewidth = 0.1)` will at least work).

More changes will come as I keep messing around with this stuff...

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,6 @@ jobs:
           Pkg.test("Makie"; coverage=true)
 
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v3
+      - uses: codecov/codecov-action@v4
         with:
           file: lcov.info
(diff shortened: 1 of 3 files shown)


src/specapi.jl
```diff
@@ -401,6 +401,13 @@ function to_plot_object(ps::PlotSpec)
     return P((ps.args...,), copy(ps.kwargs))
 end
 
+function to_plot_object(ps::PlotSpec, p::PlotList)
+    P = plottype(ps)
```
asinghvi17, line 405: What would people think if, knowing the `default_theme` of `P()::Plot` here, we populate the missing attributes with their default values as Observables in `p::PlotList`, so that users can edit them post-facto and have them pass through like normal Observables?

PR #3642, opened 2024-02-20
Avoid vertical space with `backend_show`

rikhuijzer:
# Description

When using CairoMakie in Pluto.jl, a lot of whitespace is in some cases shown above and below the Makie images. For example, when plotting

```julia
let
  fig = Figure(; size=(1400, 600))
  ax = Axis(fig[1, 1])
  scatter!(ax, Point2f[(100, 300), (200, 400)])
  fig
end
```
with Makie `v0.11.8`, the output looks as follows:

image

The root cause for this seems to be that Makie generates the following HTML:

```html
<img width="1400" height="600" style="object-fit: contain;" src="data:image/png;base64,...">
```

Here, it makes sense that the `width` and `height` are set explicitly because that metadata cannot be set in the base64 encoded image if I'm not mistaken. Next, the `object-fit: contain;` causes the image to always retain the original aspect ratio by rescaling the image width or height when needed. However, `object-fit: contain` will not automatically rescale the HTML element width or height when needed. Instead, those remain fixed and the object will be "letterboxed" (more information at [the Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit)). In the example image, for example, the height remains 600 even though the image height is rescaled to 291. 

This PR suggests to fix that by setting `height: auto` in the style. This tells the browser that the height of the containing element can be rescaled if needed. I've tested this with `Pluto.jl` and `PlutoStaticHTML.jl` and in both cases this will avoid the vertical space.

## Type of change

## Checklist

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -489,6 +489,7 @@ function backend_show(screen::MakieScreen, io::IO, ::Union{WEB_MIMES...}, scene:
     png_io = IOBuffer()
     backend_show(screen, png_io, MIME"image/png"(), scene)
     b64 = Base64.base64encode(String(take!(png_io)))
-    print(io, "<img width=$w height=$h style='object-fit: contain;' src=\"data:image/png;base64, $(b64)\"/>")
+    style = "object-fit: contain; height: auto;"
+    print(io, "<img width=$w height=$h style='$style' src=\"data:image/png;base64, $(b64)\"/>")
     return
 end


jkrumbiegel:
I'm not really knowledgeable enough about CSS and HTML to know if this is the "right" solution, but if it has a better result in more frontends that's fine with me. I had thought that width and height of the img tag communicated the intended size of the image, but that it could still be scaled depending on its placement in a container div. I had chosen these settings at the time because the VSCode plot pane had some max-width setting or so that squished the image under certain circumstances. If height auto fixes that as well, we might be able to remove the `object-fit: contain;` because the image should effectively never have a squished size.

rikhuijzer:
> I had thought that width and height of the img tag communicated the intended size of the image, but that it could still be scaled depending on its placement in a container div.

Yes and I think your observation that `object-fit: contain;` avoids squishing. I have experimented a bit with the settings and it looks like `object-fit: contain` and `height: auto;` are both needed. I think the interpretation of this combination should be something along the lines of "with `object-fit: contain`, the image will be scaled down while retaining the aspect ratio when the width of the containing element becomes smaller than the image width. However, the height will not be scaled, but the image will become letterboxed. With `height: auto`, the element is allowed to downsize too."

Below is an example from PlutoStaticHTML.jl. I've tested the patch for Pluto.jl too, but not with VSCode as I don't have that installed.

## Example

Here is an example from https://huijzer.xyz/posts/latent/. The width of this page is restricted to 705 px on desktop screens via the `franklin-content` class. Further down in the CSS hierarchy, images are restricted to `width: 100%` (so 705 px).

When I change the settings, I see the following (see which settings are enabled/disabled in the bottom of the images).

<br>
<br>

### Figure 1: Squished aspect ratio

image

<br>
<br>

### Figure 2: Too much vertical space

image

<br>
<br>

### Figure 3: Correct

image

rikhuijzer:
`height: auto` is also set for Documenter.jl by default in the CSS (example page: https://turinglang.org/MCMCChains.jl/stable/makie/) and in the Makie docs CSS (example page: https://docs.makie.org/stable/reference/plots/barplot/).

rikhuijzer:
Thanks for updating the CHANGELOG for me and for accepting this patch, @jkrumbiegel 🙂

jkrumbiegel:
thank you!

PR #3646, opened 2024-02-21
Added doc on Makie package extension

kapple19:
# Description

Fixes lack of documentation on adding a package extension for Makie.

## Type of change

Appended package extension implementation guide to `recipes.md`.

## Checklist

### DIFF
--- a/docs/explanations/recipes.md
+++ b/docs/explanations/recipes.md
@@ -341,3 +341,57 @@ GLMakie.activate!() # hide
 ```
 
 \video{stockchart_animation}
+
+## Makie Package Extension
+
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
What I had been thinking about was to add an extension test to our test suite, and this would double as an example to point people to. Being a test, it would not go out of sync and we could be sure that our intended extension mechanism keeps working in future julia versions.

SimonDanisch:
@kapple19 can you add a changelog entry? :) 
https://github.com/MakieOrg/Makie.jl/blob/master/CHANGELOG.md

PR #3650, opened 2024-02-23
Breaking 0.21 release

SimonDanisch:
PR TODO:

- [x] merge https://github.com/MakieOrg/Makie.jl/pull/3558
- [x] merge https://github.com/MakieOrg/Makie.jl/pull/3626
- [x] merge https://github.com/MakieOrg/Makie.jl/pull/3226
- [x] merge #3681
- [x] merge https://github.com/MakieOrg/Makie.jl/pull/3527
- [x] merge #3651
- [x] merge #3716
- [x] merge #3724
- [x] merge #3771

General TODO:
- [x] clean up temporary definitions in src/Makie.jl (merge https://github.com/JuliaGeometry/GeometryBasics.jl/pull/214 and https://github.com/jkrumbiegel/GridLayoutBase.jl/pull/55)
- [x] make a decision about deprecating plot.rotation or plot.rotations (meshscatter, scatter, text)
- [ ] consider how to deal with user attributes (e.g. #3764)


MakieCore/src/basic_plots.jl
```diff
@@ -40,6 +40,30 @@ function generic_plot_attributes(attr)
     )
 end
 
+function mixin_generic_plot_attributes()
+    @DocumentedAttributes begin
+        transformation = automatic
```
asinghvi17, line 45: Should we add `material` here for RPRMakie?

asinghvi17, line 45: Should we add `material` here for RPRMakie?  It's not quite a shading attribute, IMO, since it can also control the height through subsurface materials.

jkrumbiegel: I remember Simon saying he was not super happy with `material` because it has no way of being supported in the other backends. But I'm not sure if it can all be handled by `color` plus types?

asinghvi17: It would be a bit difficult to handle colors passed as values while also setting the material properties (metallicity, etc) if they were all packed in `color`, no?

jkrumbiegel: could always be `color = RPRColor(values, metallicity = ...)` or whatever. @SimonDanisch what's your opinion?

asinghvi17:
This causes GeoMakie to fail when assigning `source` and `dest` projections to separate plots, since it's not in the allow list.  I added it for now but we probably should have an interface for any Makie package to add an extension to the allowlist for this reason.

For example, a ternary plot package may want to assign a `plane`, or some other nonlinear package may have some other global argument which they want to allow at the top level.

My current thoughts are that we could have global list of allowed attributes (we can route consumers through a registration function and don't have to allow them to access this directly) which is picked up by the function `attribute_name_allowlist`.

asinghvi17:
https://github.com/JuliaLang/Pkg.jl/blob/9c6356fa9d838388cc308cfedaec6cd38fdf00ba/src/Types.jl#L1013C20-L1013C85 

We could integrate a fuzzy suggestor into Makie since we know all available attribute names (but this requires a dependency on Pkg).  It could also be implemented as an extension, or we could directly copy the source.

MakieCore/src/basic_plots.jl
```diff
@@ -345,6 +351,8 @@ $(Base.Docs.doc(MakieCore.generic_plot_attributes!))
     linewidth = @inherit linewidth
     "Sets the pattern of the line e.g. `:solid`, `:dot`, `:dashdot`. For custom patterns look at `Linestyle(Number[...])`"
     linestyle = nothing
+    "Sets the type of linecap used, i.e. :butt (flat with no extrusion), :square (flat with 1 linewidth extrusion) or :round."
```
jkrumbiegel, line 354: Nitpicking, but is 1 linewidth extrusion correct? I think it's just half, because the line end is the center of the square, right?

ffreyer: Yes that should only be half

src/interfaces.jl
```diff
@@ -254,7 +254,15 @@ end
 function Plot{Func}(user_args::Tuple, user_attributes::Dict) where {Func}
     # Handle plot!(plot, attributes::Attributes, args...) here
     if !isempty(user_args) && first(user_args) isa Attributes
-        merge!(user_attributes, attributes(first(user_args)))
+        # Only forward attributes that are usable by Plot{Func}
+        # TODO, there are two problems with this:
+        # 1) It's a bit magical which attributes are forwarded, when writing a recipe
+        # 2) We can't warn for wrong attributes, since if we do `plot(p, some_other_plot.attributes, ...)`,
```
jkrumbiegel, line 260: wasn't the warning for wrong attributes the reason why we did this in the first place? I wouldn't like reintroducing that behavior again in a different place..

jkrumbiegel: So what's the scenario where someone does that? Is it likely that there are many other invalid attributes in the collection that are annoying to take out manually? One could also filter them by the keys of the recipe to forward to?

SimonDanisch: >One could also filter them by the keys of the recipe to forward to?

That's what we do here, or do you mean something else?

SimonDanisch:
Seems like we have some regressions in performance, some are expected (e.g. using time), while others are not (since this came with changes that should not have changed performance).
I dont have time to debug this, so I guess we need to merge & tag this with those regressions.

PR #3652, opened 2024-02-23
Reviving CairoMakie/GLMakie integration

asinghvi17:
This is a draft PR from my old GLMakie rasterization pipeline into CairoMakie.

As I recall, the issue was that displaying a child of the main Scene would cause that main Scene to resize to the size of the child's window, creating a loop of sorts.  This was because scenes in a tree share the same events struct.

What I want to do here is figure out how to isolate or unlink the newly constructed screen from all events, so that subscenes can be displayed without resizing the main scene.


SimonDanisch:
This PR seems to be pretty hopelessly out of date... I propose closing it, as we also have a few PRs like https://github.com/MakieOrg/Makie.jl/pull/4150, which will make this a bit easier.
Also, I think I'll put some time into this when improving the GLMakie integration in RPRMakie.

PR #3656, opened 2024-02-26
Update translate! documentation

jamesahoward:
# Description

A basic update to translate! documentation to reflect that Transformable type can also be a plot object, not just a scene.

## Checklist


asinghvi17 (approved):
Thanks!

PR #3657, opened 2024-02-26
Add compat bound on `FFMPEG_jll`

brenhinkeller:
FFMPEG_jll appears to be in deps but not compat; this seems to be causing issues where Pkg chooses to upgrade to the latest FFMPEG_jll v0.6, despite FFMPEG.jl and most other deps (possibly including Makie itself?) not yet being compatible with it


brenhinkeller:
For reference, discussion on Slack: 
https://julialang.slack.com/archives/C67EFTEF3/p1708960255155269
https://julialang.slack.com/archives/CBF3Z1D7V/p1708962327755379

asinghvi17:
Makie is actually compatible with all versions of `FFMPEG_jll` including v0.6.x since it runs fine on CI - should we leave broader compat bounds in that case?

Project.toml
```diff
@@ -75,6 +75,7 @@ Downloads = "1, 1.6"
 FileIO = "1.6"
 FilePaths = "0.8"
 FixedPointNumbers = "0.6, 0.7, 0.8"
+FFMPEG_jll = "0.4"
```
giordano, line 78: This looks wrong?

brenhinkeller: Ah, I am very silly

brenhinkeller:
Ok, here are broader bounds that allow the two extant major versions in that case

PR #3660, opened 2024-02-28
add new Bonito based referenceupdater prototype

SimonDanisch:
- [x] simple prototype that shows all reference images with scores from the unzipped artifact, with simple ui to switch between backends and reference/recording
- [ ] Improve UI to be less confusing and to make it easier to switch between record/reference
- [ ] Simple filtering and grouping for failed tests/below score/new images, only show one backend etc
- [ ] Color code failing reference tests
- [ ] Add updating code, allow to update groups (all failed tests, complete backend, filter?)

### DIFF
--- a/ReferenceUpdater/src/bonito-app.jl
+++ b/ReferenceUpdater/src/bonito-app.jl
@@ -1,0 +1,68 @@
+using Bonito, FileIO, DelimitedFiles
+
+folder = joinpath(pwd(), "ReferenceImages")
+
+App() do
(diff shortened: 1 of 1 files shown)


ffreyer:
Requires https://github.com/SimonDanisch/Bonito.jl/pull/284

ffreyer:
Tests fails for testing purposes. I renamed one refimg and changed one so we can test upload, deletion and changing before merging this

PR #3666, opened 2024-03-01
Unsupported marker warn message hint

J-C-Q:
# Description

When trying to use an unsupported marker, it would be nice to get a hint about (or at least how to find out) which markers are supported.
This just adds this hint to the warn message.

## Type of change


J-C-Q:
Update: Added Changelog

SimonDanisch:
Thank you!

PR #3668, opened 2024-03-02
Move Arc to MakieCore

jw3126:
# Description

This moves an Arc stub to MakieCore, so it can be used without depending on Makie.jl. Personally I think almost all 
recipes defined in Makie.jl should be made available in MakieCore.jl. If there is interest, I will move more recipes over.


jkrumbiegel:
I think the actual way forward is going to be reabsorbing MakieCore into Makie, and having people use pkg extensions to extend Makie's functionality. MakieCore was established before pkg extensions existed, and as this PR shows again, it's very subjective where a line between "core" functionality and the rest would even be.

PR #3671, opened 2024-03-02
Streamline data_limits and boundingbox

ffreyer:
# Description

This adjusts data_limits and boundingbox to be more consistent and well defined:
- `data_limits` now only considers the raw positional data in plots. It never considers transform function, the model matrix, etc.
- `boundingbox` is now strictly in world space, i.e. it applies `transform_func` and `model` before generating a result

Note that this is based on a rather temporary state of #3663 so this pr is likely to fail for reasons not related to this pr.

## Type of change

## Checklist

### DIFF
--- a/src/layouting/data_limits.jl
+++ b/src/layouting/data_limits.jl
@@ -106,10 +106,15 @@ point_iterator(bbox::Rect) = unique(decompose(Point3d, bbox))
 
 
 isfinite_rect(x::Rect) = all(isfinite, x.origin) &&  all(isfinite, x.widths)
+_isfinite(x) = isfinite(x)
+_isfinite(x::VecTypes) = all(isfinite, x)
+scalarmax(x::Union{Tuple, AbstractArray}, y::Union{Tuple, AbstractArray}) = max.(x, y)
+scalarmax(x, y) = max(x, y)
+scalarmin(x::Union{Tuple, AbstractArray}, y::Union{Tuple, AbstractArray}) = min.(x, y)
+scalarmin(x, y) = min(x, y)
 
 extrema_nan(itr::Pair) = (itr[1], itr[2])
 extrema_nan(itr::ClosedInterval) = (minimum(itr), maximum(itr))
-
 function extrema_nan(itr)
     vs = iterate(itr)
     vs === nothing && return (NaN, NaN)
@@ -129,6 +134,12 @@ function extrema_nan(itr)
         vmin = scalarmin(x, vmin)
     end
     return (vmin, vmax)
+end
+
+# used in colorsampler.jl, datashader.jl
+function distinct_extrema_nan(x)
+    lo, hi = extrema_nan(x)
+    lo == hi ? (lo - 0.5f0, hi + 0.5f0) : (lo, hi)
 end
 
 function update_boundingbox!(bb_ref, point)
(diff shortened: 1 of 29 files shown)


src/layouting/text_boundingbox.jl
```diff
@@ -0,0 +1,131 @@
+function boundingbox(plot::Text)
+    @warn """
+    `boundingbox(::Text)` has been deprecated in favor of `Makie.text_boundingbox(::Text)`.
+    In the future `boundingbox(::Text)` will be adjusted to match the other
+    `boundingbox(plot)` functions. The new functionality is currently available
+    as `Makie._boundingbox(plot::Text)`.
+    """
+    return text_boundingbox(plot)
+end
```
ffreyer, line 9: I'm deprecating the old `boundingbox(::Text)` here so we can make it match the other boundingbox functions in the future.

src/layouting/maybe_unused.jl
```diff
@@ -0,0 +1,144 @@
...
+    )
+    rotated = mrot * points
+
+    rmins = minimum(rotated; dims=2)
+    rmaxs = maximum(rotated; dims=2)
+
+    return Rect2(rmins..., (rmaxs .- rmins)...)
+end
+
+=#
```
ffreyer, line 144: I tried cleaning up unused functions in "data_limits.jl" and "boundingbox.jl". All of those are here to simplify review/undoing.

src/layouting/data_limits.jl
```diff
@@ -177,103 +162,15 @@ function update_boundingbox!(bb_ref, bb::Rect)
...
-    positions = iterate_transformed(plot)
-    scales = markersize
-    # fast path for constant markersize
-    if scales isa VecTypes{3} && rotations isa Quaternion
-        bb = limits_from_transformed_points(positions)
-        marker_bb = rotations * (marker_bb * scales)
-        return Rect3d(minimum(bb) + minimum(marker_bb), widths(bb) + widths(marker_bb))
+foreach_plot(f, s::Scene) = foreach_plot(f, s.plots)
+# foreach_plot(f, s::Figure) = foreach_plot(f, s.scene)
+# foreach_plot(f, s::FigureAxisPlot) = foreach_plot(f, s.figure)
```
ffreyer, line 168: It doesn't make sense for data_limits or boundingbox to consider child scenes, as they can have different transforms and cameras. These changes are meant exclude them. Searching the repo doesn't show any other uses for these so I hope this is fine

src/layouting/maybe_unused.jl
```diff
@@ -0,0 +1,144 @@
...
+        cos(angle), -sin(angle),
+        sin(angle), cos(angle)
+    )
+    rotated = mrot * points
+
+    rmins = minimum(rotated; dims=2)
+    rmaxs = maximum(rotated; dims=2)
+
+    return Rect2(rmins..., (rmaxs .- rmins)...)
+end
```
ffreyer, line 142: This triggers a few direct test failures in transformations.jl (i.e. the function is called there)

ffreyer:
There are a couple of GLMakie refimgs failing locally due to changes in ticks, which I think we can ignore in this context, and one Axis3 test failing due to meshscatter markersize not being included in data_limits. I'll think about it a bit more and probably revert that tomorrow.

Issue #3683, opened 2024-03-06
contourf bug


t-bltg:
From https://discourse.julialang.org/t/makie-contourf-giving-incorrect-whitespace-in-contours/111222.

```
using CairoMakie

x = y = LinRange(0, 1, 10)
ymin, ymax = 0.4, 0.6
steepness = 0.1
f(x, y) = (tanh((y - ymin) / steepness) - tanh((y - ymax) / steepness) - 1)
z = [f(_x, _y) for _x in x, _y in y]

fig, ax, cof = contourf(x, y, z, extendhigh=:auto)  # , extendlow=nothing, extendhigh=:magenta
Colorbar(fig[1, 2], cof)
display(fig)

@show cof.plots[1].color[]  # should be finite
```

PR #3684, opened 2024-03-06
fix `contourf` `Inf` value on non-extended case

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/3683.

```julia

x = y = LinRange(0, 1, 10)
ymin, ymax = 0.4, 0.6
steepness = 0.1
f(x, y) = (tanh((y - ymin) / steepness) - tanh((y - ymax) / steepness) - 1)
z = [f(_x, _y) for _x in x, _y in y]

fig, ax, cof = contourf(x, y, z)
Colorbar(fig[1, 2], cof)
save("noext.png", fig)

fig, ax, cof = contourf(x, y, z; extendhigh=:auto)
Colorbar(fig[1, 2], cof)
save("ext.png", fig)
```

**noext**
noext

**ext**
ext

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/contourf.jl
+++ b/src/basic_recipes/contourf.jl
@@ -84,12 +84,12 @@ function Makie.plot!(c::Contourf{<:Tuple{<:AbstractVector{<:Real}, <:AbstractVec
     lowcolor = Observable{RGBAf}()
     map!(compute_lowcolor, c, lowcolor, c.extendlow, c.colormap)
     c.attributes[:_computed_extendlow] = lowcolor
-    is_extended_low = lift(!is_zero_color, c, lowcolor)
+    is_extended_low = lift(!isnothing, c, c.extendlow)
 
     highcolor = Observable{RGBAf}()
     map!(compute_highcolor, c, highcolor, c.extendhigh, c.colormap)
     c.attributes[:_computed_extendhigh] = highcolor
-    is_extended_high = lift(!is_zero_color, c, highcolor)
+    is_extended_high = lift(!isnothing, c, c.extendhigh)
     PolyType = typeof(Polygon(Point2f[], [Point2f[]]))
 
     polys = Observable(PolyType[])
@@ -114,7 +114,6 @@ function Makie.plot!(c::Contourf{<:Tuple{<:AbstractVector{<:Real}, <:AbstractVec
         for (i, (center, group)) in enumerate(zip(levelcenters, isos))
             points = Point2f.(group.x, group.y)
             polygroups = _group_polys(points, group.id)
-            @show center length(polygroups)
             for polygroup in polygroups
                 outline = polygroup[1]
                 holes = polygroup[2:end]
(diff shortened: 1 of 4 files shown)


src/basic_recipes/contourf.jl
```diff
@@ -84,12 +84,12 @@ function Makie.plot!(c::Contourf{<:Tuple{<:AbstractVector{<:Real}, <:AbstractVec
     lowcolor = Observable{RGBAf}()
     map!(compute_lowcolor, c, lowcolor, c.extendlow, c.colormap)
     c.attributes[:_computed_extendlow] = lowcolor
-    is_extended_low = lift(!isnothing, c, lowcolor)
+    is_extended_low = lift(!is_zero_color, c, lowcolor)
```
jkrumbiegel, line 87: I don't remember at all why this was written as it is. But it seems to me this should just replace `lowcolor` with `c.extendlow` because what you check with `is_zero_color` is just set dependent on that:

```julia
function compute_lowcolor(el, cmap)
    if isnothing(el)
        return RGBAf(0, 0, 0, 0)
```

src/basic_recipes/contourf.jl
```diff
@@ -114,6 +114,7 @@ function Makie.plot!(c::Contourf{<:Tuple{<:AbstractVector{<:Real}, <:AbstractVec
         for (i, (center, group)) in enumerate(zip(levelcenters, isos))
             points = Point2f.(group.x, group.y)
             polygroups = _group_polys(points, group.id)
+            @show center length(polygroups)
```
jkrumbiegel, line 117: `@show`

jkrumbiegel (changes requested):

src/basic_recipes/contourf.jl
t-bltg: Let's try that.

jkrumbiegel:
I don't think this fixes the real bug, in my understanding `contourf(x, y, z)` should also have the gap filled, no? For some reason, one of the isobands is treated like it's out of range.

When I do `highs[end] += eps(highs[end])` after the line `highs = levels[2:end]` it actually fills the gap.

t-bltg:
Isn't the root of this bug due to the Float32 -> Float64 conversion in https://github.com/jkrumbiegel/Isoband.jl/blob/28058449fb3c08dc531d851686e920fe6ca3138e/src/Isoband.jl#L27 ?

jkrumbiegel:
I don't think Float32 to Float64 conversion is the problem. Any Float32 is directly representable in Float64 I think, so there should be no mismatch. Somethings definitely wrong here though, would be good to get that logic straight

t-bltg:
Let's try https://github.com/jkrumbiegel/Isoband.jl/pull/2.

Using `float` doesn't hurt (faster).

jkrumbiegel:
superseded by https://github.com/MakieOrg/Makie.jl/pull/3713

PR #3688, opened 2024-03-07
[WIP] Customizable colorrange handling

asinghvi17:
# Description

This PR introduces a function `evaluate_colorrange`, which can be extended by users for their own colorrange types.  It currently has dispatches for `Automatic` and `Function` input, with the default dispatch converting the given `colorrange` to `Vec2{Float64}`.

The objective here is to allow users to pass anything they want to determine the colorrange, so long as it returns appropriate output.  This is a completion of one of the points of discussion at MakieCon.

It should now be possible to implement structs for colorranges which must be centered about zero, quantile-based colorranges, and any other metric.  In addition, functions like `PlotUtils.zscale` can be passed freely, or any user-provided function that returns a 2-tuple or similar.

Unfortunately, _scaling_ colormaps about zero such that the center is at zero but the top and bottom halves have different scales is not possible to implement in the framework of this PR - that would have to be a separate colormap object and would involve a lot more handling.

Still needs tests and examples!

## Type of change

## Checklist

### DIFF
--- a/src/colorsampler.jl
+++ b/src/colorsampler.jl
@@ -244,6 +244,28 @@ colormapping_type(@nospecialize(colormap)) = continuous
 colormapping_type(::PlotUtils.CategoricalColorGradient) = banded
 colormapping_type(::Categorical) = categorical
 
+"""
+    evaluate_colormap(colorrange, color)::Vec2{Float64}
+
+This function must return a `Vec2{Float64}` of the colorrange as `(low, high)`.
+Add additional dispatches to this function for your own custom colorrange objects!
+
+The catch-all definition for this function is to convert any provided `colorrange` into 
+a `Vec2{Float64}`.  By default, two more dispatches are defined for `Makie.automatic` 
+and for `Function`s (but not all callables, since that's not possible to define on).
+
+You can define your method like so:
+
+```julia
+Makie.evaluate_colorrange(colorrange::MyCustomType, color) = ...
+```
+
+and `color` is always an array of numbers.  Be warned that the vector may have only one element.
+"""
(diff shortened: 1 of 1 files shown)


asinghvi17:
Docs failures are unrelated - it seems like `git push` failed.

asinghvi17:
Bump - this would be pretty useful to have!

asinghvi17:
Do people think this is the right approach to take?  I can also change what I'm doing here, but some feedback would be nice.

SimonDanisch:
Seems like an interesting idea... Can we have some tests and docs for this?
@jkrumbiegel do you think the API is ok? 
I guess this will be `heatmap(rand(100, 100), colorrange=(color-> ...))`

asinghvi17:
Pretty much, the idea is to pass e.g `heatmap(sky_data; colorrange = zscale)` for astronomers, or `heatmap(random_unbalanced_data; colorrange = ScaledAbout(0))`.

jkrumbiegel:
What's `colorrange = zscale` supposed to do? I understand wanting a symmetric colorrange about 0, so -abs(extremum), abs(extremum) but a transformation of the limits without affecting the colorscale seems not very powerful? Seems like the two should go together

asinghvi17:
Zscale is basically an algorithm to ignore outliers, pretty useful in imaging where a burnt out pixel always returns the maximum value.

asinghvi17:
This would be one part of the whole - I would imagine some form of colormap wrapper type like `ScaledAbout(colormap, value)` which intercepts the process in `numbers_to_colors` or so.  That would naturally be a lot more powerful, this is more of a convenience function.

felixcremer:
Is it worth it to revive this or did too much change in the last one and a half years?

Issue #1995, opened 2022-05-28
Feature Request: Easy Fan Plots


ParadaCarleton:
A fan-plot is a visually-appealing way to display many confidence (or credible) intervals at once. This makes it easier to look at the uncertainty as a continuous whole, which helps counter bad habits like discounting any values that fall outside a 95% confidence interval. Here's a good example of a fan plot:
![image](https://user-images.githubusercontent.com/71727937/170840968-dc6860bb-d36f-42ea-9c64-41c3c8433d9c.png)

It'd be great if these were easier to make!

asinghvi17:
Could you post some code of how you deal with it right now?  To my perspective this seems appropriate:

```julia
# example code from https://makie.juliaplots.org/stable/examples/plotting_functions/band/#:~:text=color%20%3D%20%3Ared)%0A%0Af-,using%20Statistics,-using%20CairoMakie%0A%0Af
f = Figure()
Axis(f[1, 1])

n, m = 100, 101
t = range(0, 1, length=m)
X = cumsum(randn(n, m), dims = 2)
X = X .- X[:, 1]
μ = vec(mean(X, dims=1)) # mean
lines!(t, μ)              # plot mean line
σ = vec(std(X, dims=1))  # stddev

# end example code
# plot decreasing "confidence intervals" - replace this with real confint calculations
for i in 1:4 # sigma/i
    band!(t, μ + σ ./ i, μ - σ ./ i, color = (:red, 0.3))
end
f
```
![iTerm2 q7Hy1D](https://user-images.githubusercontent.com/32143268/170845418-fc56e6ea-2efc-4b1a-b109-4d78377c253f.png)

Basically, this "stacks" transparency so that each succeeding layer appears more and more opaque.  Thus we go from outside (large width) to inside (small width).

ParadaCarleton:
> Could you post some code of how you deal with it right now? To my perspective this seems appropriate:
> 
> ```julia
> # example code from https://makie.juliaplots.org/stable/examples/plotting_functions/band/#:~:text=color%20%3D%20%3Ared)%0A%0Af-,using%20Statistics,-using%20CairoMakie%0A%0Af
> f = Figure()
> Axis(f[1, 1])
> 
> n, m = 100, 101
> t = range(0, 1, length=m)
> X = cumsum(randn(n, m), dims = 2)
> X = X .- X[:, 1]
> μ = vec(mean(X, dims=1)) # mean
> lines!(t, μ)              # plot mean line
[...]

PR #3695, opened 2024-03-10
add recipes for intervals

aplavin:
# Description

Add a few natural recipes for intervals that make plotting them much more convenient.
Partially closes https://github.com/MakieOrg/Makie.jl/issues/1995.

Only implementation code for now. Should I create complete end-to-end tests with these arguments, or test the `convert_arguments` methods by themselves?

Examples:
image
image

## Type of change

## Checklist


jkrumbiegel:
I'm ok with the band, span and range use, but not with the point conversion where you calculate `mean`.

PR #3697, opened 2024-03-11
support the loop option in html video output

aplavin:
# Description

The `loop` option was only supported for gif output, but is also very convenient in HTML. Adding its support was quite straightforward!

Not sure about testing though: seems like this HTML output is not tested at all for now.

## Type of change

## Checklist


src/ffmpeg-util.jl
```diff
@@ -63,15 +63,12 @@ struct VideoStreamOptions
...
-        if format == "gif"
-            (loop === nothing) && (loop = 0)
-        end
+        (loop === nothing) && (loop = 0)
 
         # items are name, value, allowed_formats
         allowed_kwargs = [("compression", compression, ("mp4", "webm")),
                           ("profile", profile, ("mp4",)),
-                          ("pixel_format", pixel_format, ("mp4",)),
-                          ("loop", loop, ("gif",))]
```
asinghvi17, line 74: I think that `loop` is given to the rendered GIF here, not the HTML.  Can we add this back in?

asinghvi17: Makes sense - thanks!

asinghvi17:
Yeah, I don't think the HTML output is tested either.  That being said, if you put this up in the documentation somehow I can at least move it into a repo branch to make sure it works!

asinghvi17:
`record` just returns a String so I don't think that will work unfortunately

SimonDanisch:
Yeah we should document Record better, it was introduced exactly for this use case.

docs/explanations/animation.md
```diff
@@ -32,11 +32,8 @@ record(fig, "color_animation.mp4", hue_iterator;
         framerate = framerate) do hue
     lineplot.color = HSV(hue, 1, 0.75)
 end
-nothing # hide
```
SimonDanisch, line 35: This should only work if you changed `record` to `Record`, i think?

SimonDanisch:
Good questions, maybe we can add a test?

PR #3700, opened 2024-03-13
Unexport geometry types by default

### LINKED ISSUE #3610: [FR] Do not export geometry basics
Each plotting library defines its own set of geometry types (e.g. Plots.jl GR.jl, Gadfly.jl). In order to avoid namespace pollution, we would benefit from a public API that is entirely defined in terms of native constructs provided by the language such as `Array` and `Tuple` of coordinates. Currently, Makie.jl depends on GeometryBasics.jl and exports these types in the end-user session. My feature request is to not export these by default and ask for an explicit import instead:

```
import Makie.Geometries
```

We addressed this namespace pollution in other packages such as LossFunctions.jl as you can see [here](https://github.com/JuliaML/LossFunctions.jl/blob/master/src/traits.jl). Before that, many traits were exported and conflicted with other projects that are not about loss functions.

Given the widespread usage of geometry across different scientific communities, and the multitude of competing plotting libraries, it would be really great to solve this issue.

juliohm:
# Description

Fixes #3610 

This PR is a work in progress for assessment by the Makie team. The change consists in introducing a local module named `Geom` that users can import explicitly to get access to internal geometry types used by the project. The name `Geom` follows from the Gadfly.jl package, which handles geometries as `Gadfly.Geom.point` for example.

If the PR is welcome, we can continue working on it. All tests already pass because we are still loading GeometryBasics.jl explicitly as a temporary workaround. Hopefully one day, someone with more knowledge of the internals can refactor the codebase to really only depend on the `.Geom` module. This should also help in the future in case the internal geometry types need to be replaced.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/test/rasterization_tests.jl
+++ b/CairoMakie/test/rasterization_tests.jl
@@ -14,8 +14,8 @@ end
     fig = Figure()
     ax = Axis(fig[1,1])
     lp = lines!(ax, vcat(1:10, 10:-1:1))
-    pts = Makie.GeometryBasics.Point2f[(0,0), (1,0), (0,1)]
-    pl = poly!(ax, Makie.GeometryBasics.Polygon(pts))
+    pts = Makie.Geom.Point2f[(0,0), (1,0), (0,1)]
+    pl = poly!(ax, Makie.Geom.Polygon(pts))
 
     @testset "Unrasterized SVG" begin
         @test !svg_has_image(fig)
--- a/CairoMakie/test/runtests.jl
+++ b/CairoMakie/test/runtests.jl
@@ -1,6 +1,7 @@
 using Test
 using CairoMakie
 using Makie.FileIO
+using Makie.Geom
 using ReferenceTests
 
 # Before changing Pkg environment, try the test in #864
--- a/CairoMakie/test/svg_tests.jl
+++ b/CairoMakie/test/svg_tests.jl
@@ -21,8 +21,8 @@ end
     @test svg_isnt_rasterized(begin
         fig = Figure()
         ax = Axis(fig[1,1])
-        poly!(ax, Makie.GeometryBasics.Polygon(Point2.([[0,0],[1,0],[0,1],[0,0]])), color = ("#FF0000", 0.7), label = "foo")
-        poly!(ax, Makie.GeometryBasics.Polygon(Point2.([[0,0],[1,0],[0,1],[0,0]])), color = (:blue, 0.7), label = "bar")
+        poly!(ax, Makie.Geom.Polygon(Point2.([[0,0],[1,0],[0,1],[0,0]])), color = ("#FF0000", 0.7), label = "foo")
+        poly!(ax, Makie.Geom.Polygon(Point2.([[0,0],[1,0],[0,1],[0,0]])), color = (:blue, 0.7), label = "bar")
         fig[1, 2] = Legend(fig, ax, "Bar")
         fig
     end)
@@ -32,10 +32,10 @@ end
     ])))
     @test !svg_isnt_rasterized(poly(rand(Point2f, 10); color = rand(RGBAf, 10)))
 
-    poly1 = Makie.GeometryBasics.Polygon(rand(Point2f, 10))
-    @test svg_isnt_rasterized(poly(Makie.GeometryBasics.MultiPolygon([poly1, poly1])))
-    @test svg_isnt_rasterized(poly(Makie.GeometryBasics.MultiPolygon([poly1, poly1]), color = :red))
-    @test svg_isnt_rasterized(poly(Makie.GeometryBasics.MultiPolygon([poly1, poly1]), color = [:red, :blue]))
+    poly1 = Makie.Geom.Polygon(rand(Point2f, 10))
+    @test svg_isnt_rasterized(poly(Makie.Geom.MultiPolygon([poly1, poly1])))
+    @test svg_isnt_rasterized(poly(Makie.Geom.MultiPolygon([poly1, poly1]), color = :red))
+    @test svg_isnt_rasterized(poly(Makie.Geom.MultiPolygon([poly1, poly1]), color = [:red, :blue]))
 
     @testset "GeoInterface polygons" begin
         using GeoInterface, GeoInterfaceMakie
--- a/GLMakie/experiments/shaderabstr.jl
+++ b/GLMakie/experiments/shaderabstr.jl
@@ -1,5 +1,7 @@
 using ShaderAbstractions: Buffer, Sampler, VertexArray
 
+using Makie.Geom
+
 # Mesh
 mesh(Sphere(Point3f(0), 1f0)) |> display
 mesh(Sphere(Point3f(0), 1f0), color=:red, ambient=Vec3f(0.9))
--- a/GLMakie/test/runtests.jl
+++ b/GLMakie/test/runtests.jl
@@ -1,7 +1,7 @@
 using Makie
 using GLMakie, Test
 using FileIO
-using GeometryBasics
+using Makie.Geom
 using GeometryBasics: origin
 using Random
 using ReferenceTests
--- a/RPRMakie/examples/makie_example.jl
+++ b/RPRMakie/examples/makie_example.jl
@@ -1,4 +1,4 @@
-using GeometryBasics, Makie
+using Makie, Makie.Geom
 using FileIO, Colors
 using RPRMakie
 RPRMakie.activate!(iterations=200)
--- a/RPRMakie/test/lines.jl
+++ b/RPRMakie/test/lines.jl
@@ -1,4 +1,5 @@
-using GLMakie, GeometryBasics, RPRMakie, RadeonProRender
+using GLMakie, RPRMakie, RadeonProRender
+using Makie.Geom
 using Colors, FileIO
 using Colors: N0f8
 RPR = RadeonProRender
--- a/ReferenceTests/src/tests/attributes.jl
+++ b/ReferenceTests/src/tests/attributes.jl
@@ -1,3 +1,5 @@
+using Makie.Geom
+
 @reference_test "glowcolor, glowwidth" begin
     scatter(RNG.randn(10), color=:blue, glowcolor=:orange, glowwidth=10)
 end
--- a/ReferenceTests/src/tests/examples2d.jl
+++ b/ReferenceTests/src/tests/examples2d.jl
@@ -1,3 +1,4 @@
+using Makie.Geom
 
 @reference_test "Test heatmap + image overlap" begin
     heatmap(RNG.rand(32, 32))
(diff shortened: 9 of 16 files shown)


asinghvi17:
What's the motivation for this - conflicts with exports from another package?

juliohm:
Not just that. Makie needs a formal public API as a plotting package. In
terms of UX, a plotting package usually works with the native language
constructs (e.g. tuples, arrays).

In my mind, a plotting package defines a public API with names that include
Figure, Axis, plot, scatter...

Internal geometry types can be specified with kwargs as in marker=:sphere.

The corresponding issue has more details.

Beginners would have a better experience loading any set of geometries for
geometry processing of their choice:

```julia
using MyGeoms
using GLMakie

f = Figure(...)
a = Axis(f[1,1])
plot!(a, Circle(...)) # here Circles comes from MyGeoms
```

Advanced users can get the internal geometries with an explicit import:

```julia
using Makie.Geoms
```

Em qua., 13 de mar. de 2024 14:39, Anshul Singhvi ***@***.***>
escreveu:

juliohm:
What is the decision about this?

Em qua., 13 de mar. de 2024 14:54, Júlio Hoffimann <
***@***.***> escreveu:

juliohm:
cc: @jkrumbiegel

jkrumbiegel:
Not sure about it, I assume our default exports conflict with the ones you'd like to use in Meshes.jl? In principle the very basic types `Point` and `Vec` are not more than tuples, but they do have certain conveniences like `rand(Point2f, ...)` etc that are already in use here and there and which would be annoying to completely rip out. It would be a very breaking change to not offer these types by default anymore, although in principle I agree with the assessment that Makie probably exports too many things by default, and that names like `Point` are just very common.

juliohm:
> Not sure about it, I assume our default exports conflict with the ones you'd like to use in Meshes.jl?

That is one reason, yes. There are many other packages developed for geometric processing that also export the names that Makie.jl exports.

> In principle the very basic types `Point` and `Vec` are not more than tuples, but they do have certain conveniences like `rand(Point2f, ...)` etc that are already in use here and there and which would be annoying to completely rip out.

I believe that this is not an issue. The `rand` function is exported by Julia Base, and will be available as usual. The problem is the export of the geometry types themselves, which conflict with many other alternative approaches to geometry and computation.

> I agree with the assessment that Makie probably exports too many things by default, and that names like `Point` are just very common.

Too many indeed. As I mentioned before, the ideal scenario would consist of a minimum public API for end-users of the plotting library, defined as a MWE:

```julia
using GLMakie

f = Figure()
ax = Axis(f[1,1])
p = plot(obj)
```

Anything beyond this public API calls for an explict import. If you are a Makie dev, then you will likely want a `import Makie.Geom` to bring the internal geometries for additional dispatch. If you are a Meshes.jl user, you will instead rely on the imports provided by the project. Etc.

Unexporting the geometry types is a first step towards a more compact public API in Makie. Besides `Figure`, `Axis` and `plot`, what other high-level concepts do you think are important to export by default for **end-users**?

juliohm:
@jkrumbiegel what is the final decision regarding this change?

juliohm:
Are you planning to discuss it internally in the following days? How decisions like these are handled by the Makie team?

jkrumbiegel:
We discuss things whenever our free time and motivation allows. Things with an uncertain cost benefit ratio tend to take longer. Some things never really reach a final decision, as unsatisfying as that is for whoever opened the issue.

juliohm:
Having a formal review process, or someone responsible for such decisions can really help the project as a whole, as you know. I can't see the cost/benefit ratio in favor of the status quo.

SimonDanisch:
It's unfortunate that it takes this long to reach a decision on this, but there are multiple reasons:

1) It's not a priority, since it doesn't currently affect a wide user base and nobody pays for it
2) It has wide reaching consequences and not an easy deprecation path
3) It's hard to quantify the cost/benefit ratio. Of course for you, who created a package with the same names it's obvious, but I doubt that a wide audience of Makie is currently using Meshes, while tons of user code that is out there which is using Point/Circle/Rect etc. All tutorials, docs etc out there will error going forward, so this is not an easy change, while it currently isn't such a bad situation for a wide user base of Makie, that Meshes.jl has clashing exports.

As an example, the Point2f0 -> Point2f rename is still haunting us to a certain degree, even though we took all extra caution to deprecate it properly, since tutorials and examples using Point2f0 still have a high google ranking and it also turns up in unexpected ways, e.g. co-pilot and chatgpt defaulting to Point2f0 when producing code. And of course, people open issues about it, instead of reading the docs.

So this small change puts a lot of churn on the Makie ecosystem and acting like it's a no brainer which should be decided on in a few weeks isn't helpful.
A formal review process would be amazing, but as long as nobody gets paid for actually sitting down and doing the hours, its not going to happen any time soon.

That said, I do have sympathy to the proposal in general, i just would want to put more time into figuring out the least breaking way.

One proposal @ffreyer and I came up with yesterday would be to have a module that exports a minimal amount of functions instead:

```julia
# only import a minimal set of functions/types:
using GLMakie.Minimal
using CairoMakie.Minimal
using Makie.Minimal
````
This would be not breaking at all, and would give you a way to use Makie along Meshes without clashing exports.
How does that sound?

juliohm:
> It's unfortunate that it takes this long to reach a decision on this

It really is. Hopefully things will change in the future.

> 1. It's not a priority, since it doesn't currently affect a wide user base and nobody pays for it

Perhaps this is not a valid assumption. I'd assume by now that a reasonable chunk of Makie.jl users relies on Meshes.jl for geometry processing work.

> 2\. It has wide reaching consequences and not an easy deprecation path

True. It requires a breaking release and an entry in the NEWS that explains the fix: `using Makie.Geoms`

> 3\. It's hard to quantify the cost/benefit ratio. Of course for you, who created a package with the same names it's obvious, but I doubt that a wide audience of Makie is currently using Meshes, while tons of user code that is out there which is using Point/Circle/Rect etc.

I think you are seeing this with the wrong eyes. These are all open source packages, developed to address practical challenges identified by their authors. This is not a Meshes.jl vs. GeometryBasics.jl issue, it is about a formal specification of public API for Makie.jl. The geometry types are just one piece of the namespace pollution that users of Makie.jl have to deal with today.

> As an example, the Point2f0 -> Point2f rename is still haunting us to a certain degree, even though we took all extra caution to deprecate it properly, since tutorials and examples using Point2f0 still have a high google ranking and it also turns up in unexpected ways, e.g. co-pilot and chatgpt defaulting to Point2f0 when producing code. And of course, people open issues about it, instead of reading the docs.

I guarantee you that users would never face such issues if the Makie.jl public API relied solely on Julia built-in types: tuples for coordinates of points in this case. End-users need to know that a `scatter((0,0))` or a `scatter([(0,0), (1,0)])` works, they **don't** need to know that `scatter(Point(0,0))` exists. This is an internal implementation detail.

> So this small change puts a lot of churn on the Makie ecosystem and acting like it's a no brainer which should be decided on in a few weeks isn't helpful.
> A formal review process would be amazing, but as long as nobody gets paid for actually sitting down and doing the hours, its not going to happen any time soon.

I can share some personal experience here, both as an open-source developer, and as a CEO in the tech industry. People appreciate the **attitude** of the maintainers before they consider putting money on a project. Right now, even if I could reserve some funds to redirect to Makie, I don't feel that issues and PRs are treated with the necessary level of seriousness that would give me the confidence that my money wouldn't be lost. Makes sense? I know that this is a breaking change, but it is breaking in the right direction. I am sure that many of you also see this.

> That said, I do have sympathy for the proposal in general, i just would want to put more time into figuring out the least breaking way.

Thanks!

> One proposal @ffreyer and I came up with yesterday would be to have a module that exports a minimal amount of functions instead:
> 
> ```julia
> # only import a minimal set of functions/types:
> using GLMakie.Minimal
> using CairoMakie.Minimal
> using Makie.Minimal
> ```
> 
> This would be not breaking at all, and would give you a way to use Makie along Meshes without clashing exports. How does that sound?

I don't think this workaround moves the project in the right direction. I see two alternative solutions that are much more beneficial in the long-term:

1. Forget GeometryBasics.jl and use Meshes.jl internally since there is no way the former will catch up the latter in number of features. The gap will continue to increase, day after day, because we are actively working on it, full-time. This solution assumes that Makie.jl aims to become a robust plotting library, which **uses** 3rd-party libraries for geometry processing. If this assumption is invalid, and Makie.jl aims to **become** a geometry processing library, then ignore this solution.

2. Unexport **all** symbols in a breaking release, leaving only the public API, which AFAIK should contain `Figure`, `Axis`, `Plot/plot`, `scatter`, `heatmap`, ... and other plotting aliases. Anything else is an internal implementation detail that can change freely without affecting the user base.

I understand that both these solutions are breaking, but there is no other way around to improve the status quo. You break it once, you announce it, and then the future will be much smoother with less complaints from end users, and less maintenance burden.

SimonDanisch:
>Perhaps this is not a valid assumption. I'd assume by now that a reasonable chunk of Makie.jl users relies on Meshes.jl for geometry processing work.

It doesn't really matter if its a valid assumption or not or if my "perspective is wrong". Fact is, there are not a lot of issues about this, so it is not clear to me either way. 
Not having clear data about user impact makes it a hard decision for me, even if its preferable from a philosophical perspective. From reading Makie user code, I do know though, that Point/Circle/Rect are _heavily_ used across the ecosystem. That's currently my only valid data point for this discussion.

This is a subjective developer opinion, but I'd much rather have less churn for users instead of fixing small design problems. 
My opinion on this has changed over the last 12 years, with a growing Makie user base and having to bear the brunt of "small user facing changes". 
It's easy to say "this is the right thing going forward" without needing to deal with the fall out for the next 3-5 years, which is guaranteed to give me quite a bit of work.

If this PR/issue is all about removing GeometryBasics in favor of Meshes.jl, it's a whole different discussion anyways.

I think `using GLMakie.Minimal` is the only valid way forward for now, and it also gives us some idea of how much people prefer a minimal export version of Makie. If it's used a lot, we can think about introducing it as the default export set.

juliohm:
> It's easy to say "this is the right thing going forward" without needing to deal with the fall out for the next 3-5 years, which is guaranteed to give me quite a bit of work.

You have to remember that part of this work is ours as well. If you opt for solution (1) and rely on Meshes.jl, then we will have tons of new issues to solve in Meshes.jl over the next years coming from Makie.jl users. Is it bad? No. We do work on this because we believe in it. We have more time than you do to work on hairy geometry processing algorithms, and you have more time to evolve the plotting capabilities of Makie.jl.

> If this PR/issue is all about removing GeometryBasics in favor of Meshes.jl, it's a whole different discussion anyways.

It is not. Although this replacement will improve the situation by a lot, it is not enough. Modules like the layout subsystem in Makie export names like `Box` which are clearly not a geometry, and which clash with other libraries out there. I can point many other unsolicited exports as an end-user of Makie.jl, but it is up to you to see it with the same eyes. At the end of the day I am just providing the feedback to improve the Makie.jl experience. You and the other maintainers are the ones who will decide what to do with this feedback.

PR #3708, opened 2024-03-15
If running under VSCode, automatically set `resize_to = :body`

staticfloat:
This gives a better default experience for users plotting inside of VSCode.  It detects the presence of VSCode by looking to see if `VSCodeServer` is in `Base.loaded_modules`, which it should never be unless we're running inside the language server.

## Type of change


SimonDanisch:
Good idea, but this will be sadly a global setting, applying to people not using the plot pane and just developing a Bonito App with inlined plots.
I think we can add a condition in `backend_show` to check for the vscode MIME and then wrap the scene in a `WGLMakie.WithConfig(scene; resize_to=:body)`: https://github.com/MakieOrg/Makie.jl/blob/master/WGLMakie/src/display.jl#L169-L179

PR #3716, opened 2024-03-18
update scatter boundingbox

ffreyer:
Considers marker metrics in `boundingbox(::Scatter)`


ffreyer:
refimg failures are all small changes in the limits of `scatter rotation`. Should be fine to merge

PR #3718, opened 2024-03-18
Use label formatter in barplot.

### LINKED ISSUE #3720: Label formatter not used in barplot
- [x] are you running newest version (version from docs) ? -> `0.20.8`
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) -> yes
- [x] What platform + GPU are you on? -> Linux + Nvidia GPU

```julia
xs = 1:6; ys = rand(6);
fig = barplot(xs, ys; bar_labels=ys, label_formatter=(x->"$(round(Int, x))"))
```
no_bar_plot

RomeoV:
# Description
Before, when the labels were not just the `:x` or `:y` symbols, the label formatter wasn't applied.
Now we apply it, but have to be careful to dispatch the default formatter depending on the input type.
In the PR I define new dispatches for `::String` and `::LaTeXString` types (basically the identity function), which should also fix the failing tests.

## Type of change
Basically just change a typo and add two dispatches.

Fixes #3720.


SimonDanisch:
Thanks! merged in https://github.com/MakieOrg/Makie.jl/pull/3721

PR #3721, opened 2024-03-19
Use label formatter in barplot

SimonDanisch:
Fix for https://github.com/MakieOrg/Makie.jl/pull/3718


RomeoV:
Thanks for merging! I just thought about it once more, perhaps it should be

```julia
bar_label_formatter(label::AbstractString) = label
```

instead of the just merged

```julia
bar_label_formatter(label::String) = label
bar_label_formatter(label::LaTeXString) = label
```
?

PR #3723, opened 2024-03-20
Change how boundingbox(::Text) is deprecated

ffreyer:
Adjusts the `boundingbox(p::Text)` deprecation to `@deprecate boundingbox(p) boundingbox(p, p.markerspace[])`, renames `text_boundingbox()` to `string_boundingbox()` (partially to find calls that need to be updated) and cleans up some of roundabout boundingbox logic used for old deprecation

### DIFF
--- a/src/layouting/boundingbox.jl
+++ b/src/layouting/boundingbox.jl
@@ -13,11 +13,11 @@ the `model` matrix. Plots with `exclude(plot) == true` are excluded.
 
 See also: [`data_limits`](@ref)
 """
-function boundingbox(scenelike, exclude = (p)-> false)
+function boundingbox(scenelike, exclude = (p)-> false, space = :world)
     bb_ref = Base.RefValue(Rect3d())
     foreach_plot(scenelike) do plot
         if !exclude(plot)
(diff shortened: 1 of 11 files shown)


src/layouting/boundingbox.jl
```diff
@@ -13,11 +13,11 @@ the `model` matrix. Plots with `exclude(plot) == true` are excluded.
 
 See also: [`data_limits`](@ref)
 """
-function boundingbox(scenelike, exclude = (p)-> false)
+function boundingbox(scenelike, exclude = (p)-> false, space = :world)
```
asinghvi17, line 16: The order of arguments then has to change so space is first, no?  Either that or it becomes a kwarg...

ffreyer: The pr is only really supposed to change `boundingbox(::Text)` to include space so we don't have to deprecate it again in the future. So if this is problematic we can just remove it for now and deal with it later

ffreyer:
I think this is testing against master rather than breaking, causing refimg failures

SimonDanisch:
Ah, no i updated the refimages for sd/axis-converts

PR #3724, opened 2024-03-20
Deprecate `rotations` for `rotation` for Scatter and MeshScatter

jkrumbiegel:


jkrumbiegel:
@SimonDanisch @ffreyer could one of you have a look at the failure in CairoMakie? A Mat4 ends up in a place where it's not expected there, I assume it has something to do with removing the `key"rotations"` conversion methods but I'm not sure.

ffreyer:
> @SimonDanisch @ffreyer could one of you have a look at the failure in CairoMakie? A Mat4 ends up in a place where it's not expected there, I assume it has something to do with removing the `key"rotations"` conversion methods but I'm not sure.

The value got overwritten for meshscatter. should be fixed now.

jkrumbiegel:
Thanks :) some more errors remain but I'll try to dig into those myself

jkrumbiegel:
@SimonDanisch some WGLMakie shader problem remains, I don't know how to fix that. Not sure if `get_rotations` must be `get_rotation`, that also doesn't quite work on my machine

PR #3733, opened 2024-03-25
Fix texture atlas updates in WGLMakie

### LINKED ISSUE #3232: DataInspector tooltips swallow punctuation on first plot
Using `WGLMakie#sd/beta-20`, on a Macbook Pro M1.  The following long DataInspector tooltip:

```
julia> using WGLMakie
       f = Figure()
       ax = Axis(f[1,1])
       scatter!(ax, [1, 2, 3], [1, 2, 1]; inspector_label = (args...) -> "This is a really long tooltip\nI'm trying to trigger the issue Glen mentions\nMaybe I can do it if I keep on typing\nBut this length seems to work\nwhat's the issue? I can't find it!")
       DataInspector(f)
       f
```

Results in a tooltip without the `?` or `!`:

image

But when I run the exact same code again a second time, the glyphs properly render.

ffreyer:
# Description

Fixes #3232

The centralized texture atlas that WGLMakie uses doesn't forward updates to js. Not on observable updates, not on plot insertion and not on scene insertion.

```julia
using WGLMakie
f, a, p = text("test!") # ! shows up. I assume it's inserted before the js side gets initialized

p[1][] = "test!?°ħ▒-" # ?°ħ▒ do not show up
text!(a, Point2f(0.1), text = "test") # adding plot doesn't update js texture atlas
Label(current_figure()[1, 2], "waa") # scene insertion doesn't either
```

As a test I switched to handling the texture atlas per plot like a normal texture. That fixes the issue. We should fix this by making the centralized texture atlas update though, to avoid creating duplicates for every text and most scatter plots.

I have no knowledge of Bonito but I assume this is wrong:

https://github.com/MakieOrg/Makie.jl/blob/d50dad0a2340c24b84ffce172037bf615506f3d3/WGLMakie/src/three_plot.jl#L41

## Type of change

## Checklist

### DIFF
--- a/WGLMakie/src/WGLMakie.jl
+++ b/WGLMakie/src/WGLMakie.jl
@@ -80,7 +80,7 @@ function __init__()
     atlas = wgl_texture_atlas()
     TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(atlas.data))
     Makie.font_render_callback!(atlas) do sd, uv
-        TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(wgl_texture_atlas().data))
+        TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(sd))
     end
     DISABLE_JS_FINALZING[] = false
     return
--- a/WGLMakie/src/particles.jl
+++ b/WGLMakie/src/particles.jl
@@ -181,7 +181,8 @@ function scatter_shader(scene::Scene, attributes, plot)
 
     if uniform_dict[:shape_type][] == 3
         atlas = wgl_texture_atlas()
-        uniform_dict[:distancefield] = NoDataTextureAtlas(size(atlas.data))
+        # uniform_dict[:distancefield] = NoDataTextureAtlas(size(atlas.data))
+        uniform_dict[:distancefield] = Sampler(map(x -> wgl_texture_atlas().data, TEXTURE_ATLAS))
         uniform_dict[:atlas_texture_size] = Float32(size(atlas.data, 1)) # Texture must be quadratic
     else
         uniform_dict[:atlas_texture_size] = 0f0


WGLMakie/src/WGLMakie.jl
```diff
@@ -80,7 +80,7 @@ function __init__()
     atlas = wgl_texture_atlas()
     TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(atlas.data))
     Makie.font_render_callback!(atlas) do sd, uv
-        TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(wgl_texture_atlas().data))
+        TEXTURE_ATLAS[] = convert(Vector{Float32}, vec(sd))
```
SimonDanisch, line 83: This is wrong, since sd is just the signed distancefield of the newly rendered glyph.
So this _should_ update the texture atlas observable in the browser.
I suspect some race condition, that the plot is rendered first before all glyphs are updated in the texture atlas.

WGLMakie/src/particles.jl
```diff
@@ -181,7 +181,8 @@ function scatter_shader(scene::Scene, attributes, plot)
 
     if uniform_dict[:shape_type][] == 3
         atlas = wgl_texture_atlas()
-        uniform_dict[:distancefield] = NoDataTextureAtlas(size(atlas.data))
+        # uniform_dict[:distancefield] = NoDataTextureAtlas(size(atlas.data))
+        uniform_dict[:distancefield] = Sampler(map(x -> wgl_texture_atlas().data, TEXTURE_ATLAS))
```
SimonDanisch, line 185: This should do pretty much the same as before, just with 10-100x the data transfer, since the texture atlas is huge and lots of plots will then have separate observables which update on write.
The question is, did this actually actually fix it? If yes, I think this must be some kind of update order bug, and this fix changes the order.

PR #3736, opened 2024-03-26
[Docs] add wrap/redirect recipe tutorial

Moelf:
# Description
Adding a new tutorial, as discussed in https://discourse.julialang.org/t/redirect-a-plot-type-in-makie-while-keeping-color-consistent/111975 with @asinghvi17  

## Type of change

- [x] Documentation

## Checklist

### DIFF
--- a/docs/tutorials/wrap-existing-recipe.md
+++ b/docs/tutorials/wrap-existing-recipe.md
@@ -69,8 +69,8 @@ hist(h; color=:red, direction=:x)
 \end{examplefigure}
 
 This almost works, but we see that the keyword arguments are not passed to the `barplot!` function.
-To handle these attributes properly, we need to override/merge the user-passed attributes with the
-default attributes of the underlying plot type (in this case, `BarPlot`):
+To handle these attributes properly, we need to override/merge the
+default attributes of the underlying plot type (in this case, `BarPlot`) with the user-passed attributes:
 
 \begin{examplefigure}{name = "final_result"}
 ```julia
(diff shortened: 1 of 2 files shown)


docs/tutorials/wrap-existing-recipe.md
```diff
@@ -0,0 +1,89 @@
...
+
+\begin{examplefigure}{name = "final_result"}
+```julia
+function Makie.plot!(plot::Hist{<:Tuple{<:MyHist}})
+    scene = Makie.parent_scene(plot)
+    attributes = Makie.default_theme(scene, Makie.BarPlot)
+    for key in keys(attributes)
+        attributes[key] = get(plot.attributes, key, attributes[key])
+    end
+    barplot!(plot, attributes, plot[1])
```
ffreyer, line 83: I think just `barplot!(plot, plot.attributes, plot[1])` is enough. Makie should fill out the remaining attributes when creating the barplot

Moelf: the for loop is so that user-passed options (in this case, `color` and `direction`) gets passed through to the underlying function when applicable.

Moelf: what if `plot.attributes` contain illegal things for the barplot? would Makie know to ignore them?

asinghvi17: That was the main reason we went this route and did not directly pass the attributes through.   At current master this would be fine, but I'm not sure for v0.21.

jkrumbiegel: maybe it would be better to wait with this until 0.21 settles and clarifies how this should work

Moelf: I don't mind waiting, maybe give this PR a label so we remember to revisit when 0.21 releases?

Moelf:
this doesn't actually work. I will open an issue.

PR #3737, opened 2024-03-26
actually update texture atlas on change

SimonDanisch:


ffreyer:
closes #3733
fixes #3232

ffreyer:
Refimg test I added: (WGLMakie)

| Step | master | pr |
| --- | --- | --- |
| 1 | step1 | step1 |
| 2 | step2 | Screenshot from 2024-03-26 15-30-11 |

ffreyer (approved):

SimonDanisch:
The `0.69x slower❌` turned up two times now and seems a bit big to be noise, but, `create` should be a purely Makie.jl based benchmark, and we haven't change a single Makie related function and the Julia changes are minimal as well, so I think we should be able to ignore this

PR #3747, opened 2024-03-29
Prepare 0.20.9

ffreyer:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [0.20.9] - 2024-03-29
```
jkrumbiegel, line 5: this file also needs that [0.20.9] link added at the very bottom, and the unreleased ref changed

PR #3756, opened 2024-04-01
Use geti/j/k instead of indices for DelaunayTriangulation code

DanielVandH:
This isn't strictly needed yet, but I've v1.0 for DelaunayTriangulation in the works that will be removing the indices function (becoming `triangle_vertices`, but that doesn't exist yet) - this would be the only change needed, so it would be helpful to simply replace `indices` early. `geti`, `getj`, and `getk` are public.


ffreyer:
Should we wait with this until you release 1.0 so we can also bump the dependency version? Currently that's still on https://github.com/MakieOrg/Makie.jl/blob/46e14a829340d1cca677991de4a23da36fcbc610/Project.toml#L70

DanielVandH:
The main concern I have with delaying is that, once the version is released, it'd be unusable with Makie for a short period until Makie also gets a patch release, which would be annoying since `triplot` would be commonly used when also using DelaunayTriangulation.jl.

ffreyer:
I don't think you'd be able to upgrade to 1.0 with the "0.8.7" restriction

DanielVandH:
Hm, yes true I overlooked that. I still think this would be useful to have done early, helps me not having to temporarily keep an `indices` function in my development branch also.

asinghvi17:
If the functions already exist now, no reason not to merge I think?

DanielVandH:
I've also now set the version to allow for both v0.8.7 and v1.0 - using `geti`, etc. will make it compatible across both versions.

DanielVandH:
I'm going to close this for now and will come back to it. There's a few other things that need to change that I've somehow never encountered in my devbranch. For now I'm just going to keep versions of these functions in v1.0 that I'll come back to update later.

PR #3758, opened 2024-04-02
Clarify that behavior of markersize defines the size in one dimension.

alecloudenback:
# Description

Fixes # (issue)

The existing docs were not clear to me how `markersize` behaved - i.e. if I doubled the size `x` would I get something akin to `x^2` times larger? Based on experimentation, I think that's true and I tried to make the docs unambiguous and concise (but would be happy to take suggested changes). 

## Type of change

## Checklist

### DIFF
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -383,7 +383,7 @@ Plots a marker for each element in `(x, y, z)`, `(x, y)`, or `positions`.
   Otherwise, one can set one color per point by passing a `Vector{<:Colorant}`, or one colorant for the whole scatterplot. If color is a vector of numbers, the colormap args are used to map the numbers to colors.
 - `cycle::Vector{Symbol} = [:color]` sets which attributes to cycle when creating multiple plots.
 - `marker::Union{Symbol, Char, Matrix{<:Colorant}, BezierPath, Polygon}` sets the scatter marker.
-- `markersize::Union{<:Real, Vec2f} = 9` sets the size of the marker.
+- `markersize::Union{<:Real, Vec2f} = 9` sets the size of the marker (twice the `markersize` is approximately four times as large).
 - `markerspace::Symbol = :pixel` sets the space in which `markersize` is given. See `Makie.spaces()` for possible inputs.
 - `strokewidth::Real = 0` sets the width of the outline around a marker.
 - `strokecolor::Union{Symbol, <:Colorant} = :black` sets the color of the outline around a marker.
--- a/docs/reference/plots/scatter.md
+++ b/docs/reference/plots/scatter.md
@@ -109,8 +109,10 @@ f
 
 #### Markersize
 
-The `markersize` attribute scales the scatter size relative to the scatter marker's base size.
-Therefore, `markersize` cannot be directly understood in terms of a unit like `px`, it depends on _what_ is scaled.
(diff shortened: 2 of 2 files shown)


ffreyer:
I think it would be clearer to talk about scaling the dimensions of a marker rather than how it affects the area. A `Vec2f` markersize scales a marker in x and y direction independently. A `Real` markersize is effectively `Vec2f(markersize)`, scaling the x and y dimension by the same factor.

MakieCore/src/basic_plots.jl
```diff
@@ -383,7 +383,7 @@ Plots a marker for each element in `(x, y, z)`, `(x, y)`, or `positions`.
   Otherwise, one can set one color per point by passing a `Vector{<:Colorant}`, or one colorant for the whole scatterplot. If color is a vector of numbers, the colormap args are used to map the numbers to colors.
 - `cycle::Vector{Symbol} = [:color]` sets which attributes to cycle when creating multiple plots.
 - `marker::Union{Symbol, Char, Matrix{<:Colorant}, BezierPath, Polygon}` sets the scatter marker.
-- `markersize::Union{<:Real, Vec2f} = 9` sets the size of the marker.
+- `markersize::Union{<:Real, Vec2f} = 9` sets the size of the marker (twice the `markersize` is approximately four times as large).
```
asinghvi17, line 386: Perhaps something like:
```julia
- `markersize::Union{<:Real, Vec2f} = 9` sets the size of the marker (think radius, not area).

ffreyer: I think it's better to keep the docstring somewhat vague and simple so people don't get overwhelmed by detail. `markersize` scales markers in x and y direction, either equally by a `Real` or independently by the components of a `Vec2f`. The units depend on `markerspace`.

The details about the final size of the marker can then go into the online documentation for people who actually care about it:
- for characters `markersize` is effectively the same as `fontsize`, i.e. `marker = m, markersize = 10, markerspace = :pixel` results in an `m` with a fontsize of 10px
- for `Circle` and `Rect` are unit-sized, i.e. `marker = Circle, markersize = 10, markerspace = :pixel` draws a circle with diameter = 10px.
- BezierPaths have their own coordinate system which gets scaled by markersize. E.g. if a Bezierpath includes points between 0..10 x 0..10, then `markersize = 10, markerspace = :pixel` will scale it to size of 0..100 by 0..100
- Symbols like `:circle` are BezierPaths with coordinate systems adjusted to work well with characters markers

> it scales width and height of the marker up for values greater than one, or down for values between zero and one
Also this made me think you were talking about scaling markers like how you'd resize an image (png, bmp, etc) at first. So just to clarify, in WGLMakie and GLMakie most markers do have a rasterized base size (32 and 64 respectively) but the resizing is much more accurate than a normal image would be. Circle and Rect markers are the exceptions here - they are effectively vector graphics. But this is all more of an implementation detail and not something we should have in the docs imo.

jkrumbiegel: Hm I was actually thinking that each attribute docstring _should_ contain all the information to really understand it (and the docs would just repeat it). Of course that is somewhat at odds with a manageable-length docstring for the whole plot type. That's why the descriptions of Block attributes have to be requested separately with `Axis.xlabel` etc. which doesn't work for plot types due to dispatch reasons.

alecloudenback: My vote is for concise completeness, with detail put into an extended help section. Is that an agreeable approach?

Issue #3641, opened 2024-02-19
Lines TODO


ffreyer:
This supersedes #2548 and closes it once #3558 gets merged.

---

#### Joinstyle and linecap (from #2548 & #3558)

Rounded joints, caps and extruded caps should be much easier with the added signed distance fields. We just need to adjust their interpretation.

#### Closed Lines / Loops (from #2548)

Add a keyword that makes lines closed loops. This would require adjusting indices + shaders to cleanly connect the line.

#### Bring back per-point linewidth (from #3558)

This is getting removed in #3558 as it complicates things a lot. With a large difference in linewidth the shape of the line segment can become concave, which makes the signed distance fields cut away from sections they should not affect. A potential way to deal with this is to detect concave corners and switch `max(width, length)` to `min(width, length)` for them. 

#### Poly Optimization (from #3558)

`poly` picks the first method definition https://github.com/MakieOrg/Makie.jl/blob/1c6d934dd279466db4f06c897953909b8b10ceee/src/basic_recipes/poly.jl#L11 for `Rect`, which means it relies on `wireframe` which uses `linesegments`. So it will always produce bad corners (fully overlapping, partially overlapping or missing corners). We should make it go into the second method https://github.com/MakieOrg/Makie.jl/blob/1c6d934dd279466db4f06c897953909b8b10ceee/src/basic_recipes/poly.jl#L86 to produce at least 3 clean corners via `lines`

#### Optimize line length calculation/slow path (from #3558)

In order to correctly continue patterns from line segment to segment we need to calculate the cumulative line length on the CPU. Currently we always do a full transformation to pixel space. I think that is unnecessary when if the camera does not have perspective projection or at least if it does not have rotation. 

#### Avoid bending dots across joints (from #3558)

In #3558 this can be done by setting `offset = max(abs(extrusion, halfwidth)` in `process_pattern` however that result in patterns breaking in dense lines (e.g. bracket tests). This might be fixable by limiting halfwidth to the segment length.

#### Sample colormap in fragment shader (from #3558)

Currently colormaps are sampled in the vertex/geometry shaders, which means that we interpolate colors directly rather than interpolating values and resampling the colormap. This leads wrong colors in some cases.

#### WGLMakie artifacts in dense lines (from #3558)

In #3558 one of the tests shows artifacts:

[...]

PR #3771, opened 2024-04-05
Add linecaps and jointstyles

### LINKED ISSUE #871: Allow to set line join and cap in CairoMakie

ffreyer:
# Description

Adds `linecap = :butt / :square / :round` and `jointstyle = :miter / :round` across GLMakie, WGLMakie and CairoMakie for solid lines.

Screenshot from 2024-04-05 16-18-19

closes #871
ref #3641 

## Type of change

Minor but technically breaking change:
- adjusted the default miter_limit to 60° across all backends (previously 60° in CairoMakie and 66.42182152179817° in GLMakie and WGLMakie)

## Checklist


WGLMakie/src/Lines.js
```diff
@@ -450,6 +454,10 @@ function lines_vertex_shader(uniforms, attributes, is_linesegments) {
                 vec2 miter_n1 = is_truncated[0] ? normalize(v0.xy - v1.xy) : normalize(n0 + n1);
                 vec2 miter_n2 = is_truncated[1] ? normalize(v1.xy - v2.xy) : normalize(n1 + n2);
 
+                // Are we truncating based on jointstyle? (bevel)
+                if (int(jointstyle) == 3) // why int
```
ffreyer, line 458: Seems like integer uniforms get converted to floats on their way to the shader?

ffreyer:
Refimages added:

For different jointstyles and liencaps:
Screenshot from 2024-04-07 18-26-59

For miter limit: (notice frequency of truncated miter joints)
Screenshot from 2024-04-07 18-28-32

ffreyer:
Failures in WGLMakie are due to minor changes in aliasing. Should be fine to ignore

ffreyer:
The last couple commits:
- update CairoMakie to use miter_limit
- change miter_limit to be angle based
- change the default miter_limit to 60°, which was the default for CairoMakie and is relatively close to the old defaults for GLMakie and WGLMakie
- add `miter_distance_to_angle` to convert miter distance (either in line direction or euclidean) to angles

ffreyer:
This is good to merge. Just needs a refimg update (maybe also "PolarAxis radial shift and clip" as the miter angle change makes a difference here)

PR #3779, opened 2024-04-09
Add how to for Makie extensions

milankl:
following discussion in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/511

### DIFF
--- a/docs/how-to/makie-extensions.md
+++ b/docs/how-to/makie-extensions.md
@@ -1,0 +1,135 @@
+# Makie package extensions
+
+In the following we described how you can extend Makie in your own package
+without having a hard dependency on Makie.jl.
+[Package extensions](https://pkgdocs.julialang.org/v1.10/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) were introduced
+in Julia v1.9 to allow one package to extend functionality of another
+without the need to add that other package as a dependency.
+These so-called extensions or soft dependencies often define an
+additional method for a function from package A for a type of package B,
+but this method is only loaded when both A and B are imported.
+The documentation of extensions in
+[Pkg.jl](https://pkgdocs.julialang.org/v1.10/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions))
+describes the situation for extending a function your package owns
+for a type another package owns. However, extending Makie externally,
+many cases will work the other way around whereby another
+packages own the type and would like the extend a function they
+do not own, but that lives in Makie.
+
+## Define the Makie extension
+
+First you need to define that Makie is a soft dependency of your
+package `MyPackage.jl`. In its repository, add the folder `ext`
+for extensions and within it a file named `MyPackageMakieExt.jl`
+or similar. This file is then structured as
+
+```julia
+module MyPackageMakieExt
+
+using MyPackage, Makie
+
+"""Define Makie's `heatmap` for `x::,MyType`."""
+function Makie.heatmap(x::MyType)
+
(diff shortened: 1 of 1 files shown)


milankl:
It looks like you don't use the `[Header](@ref)` syntax for links but `\reflink` instead?

milankl:
@jkrumbiegel approve CI?

milankl:
@jkrumbiegel @SimonDanisch approve or request changes?

jkrumbiegel (changes requested):
Thanks for the effort! In general nicely written etc. I just don't really like the example. Of course you can overload `heatmap` like any other function, but these plotting functions are not your usual functions, they are rather entry points into our recipe/theming/plotting system. Which is circumvented by defining methods on them manually. I know you _can_ technically do it, and I know that we don't yet have the canonical solution for that in Makie, but it's still not something the official docs should endorse.

I think I'd rather show how some recipe functions owned by the package can be filled in using an extension. So you define `someplot` and `someplot!` as empty functions in the package and add a recipe on top of that in the extension, like I was doing here https://github.com/jkrumbiegel/MakiePkgExtTest/blob/master/SomePackage/ext/MakieExtension.jl

milankl:
Sorry I don't get this. You'd rather have packages that want a Makie extension not to use the Makie function like `heatmap` but rather come up with their own name, like `visualise_my_data` that then calls Makie functions like `heatmap!`? I'm sure there are good reasons to hook into lower level functions in the Makie logic, but I guess many package developers (like me) aren't necessarily familiar how to do that or why. So they'll presumably think about putting something on top of the top-level functions. [SpeedyWeather.jl](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/ext/SpeedyWeatherMakieExt.jl) does that now, and it's exactly what we need, [Tenet.jl](https://github.com/bsc-quantic/Tenet.jl/blob/master/ext/TenetGraphMakieExt.jl) does it, [Trixi.jl](https://github.com/trixi-framework/Trixi.jl/blob/main/ext/TrixiMakieExt.jl) does it and it seems like there's plenty more if you search for `MakieExt`.

As I said, I thought this would be helpful for other package developers to have some documentation of how to write a Makie extension, but if you'd like to establish another way of writing such extensions then please let me know / change this pull request so that we know how to do it the preferred way + outlining the reasons for it, because I still haven't understood yet what the advantages are to hook into the deeper Makie logic than to extend a top-level function with another method.

jkrumbiegel:
> You'd rather have packages that want a Makie extension not to use the Makie function like heatmap but rather come up with their own name, like visualise_my_data that then calls Makie functions like `heatmap!`

I don't know what the best solution is, the problem is that all the plotting functions defined through `@recipe` follow a special pipeline which takes care that `heatmap(...)` returns `FigureAxisPlot`, `heatmap!(ax, ...)` etc are all defined. Of course one can overload unused methods of these functions and then return whatever, for example a `Figure` with a heatmap plus other stuff. But these then live "outside" of the Makie system and in my opinion it gets harder to understand what's going to happen when you call them.

From my point of view it's better currently to define a simple function if you want to do a full figure with lots of stuff going on. Or a `@recipe` if you want to offer a reusable building block. But if people start overloading the default plotting functions and everybody comes up with their own interface what kind of options they take and how, that's going to be confusing.

I acknowledge that this is happening because we currently don't have a "complex recipe" story in Makie, as it's a complex design issue, and people are going ahead with their custom solutions because they don't have time to wait. That doesn't make the problem I described above go away.

PR #3787, opened 2024-04-16
Upgrade to DelaunayTriangulation v1.0

DanielVandH:
This contains all the changes needed to upgrade to DelaunayTriangulation v1.0.

Once this has been merged, could a new release be made when possible please?


DanielVandH:
Why is the CI not updating the registry to recognise v1 as a possible version?

asinghvi17:
It might be a cache issue, I've purged some so let's see how these new CI runs go

DanielVandH:
Still nothing. Is there something I need to do different?

asinghvi17:
The figure
```
Voronoiplot for a centroidal tessellation with an automatic colormap.png
```
is failing reference image tests.

asinghvi17:
Screenshot 2024-04-17 at 2 34 19 PM
this PR on the left, master on the right

DanielVandH:
Is that image the only issue? Seems from the CI that once it's replaced this should be good to go?

DanielVandH:
Any update on this?

jkrumbiegel:
I think because it did cause different visuals it was considered breaking? So my understanding was it could go into 0.21

DanielVandH:
Is it really different visuals? It's the same output for an identical triangulation, e.g. `triplot(points)` is always the same. The only difference is if someone did `refine!` before trying to plot (which is never done by `triplot` itself), and even then the difference is small.

jkrumbiegel:
Oh ok I didn't know that the changes were dependent on `refine!`. That's ok then, as far as I'm concerned, as people using `refine!` must've loaded DelaunayTriangulation and are therefore subject to its versioning.

asinghvi17:
Merged into https://github.com/MakieOrg/Makie.jl/pull/3814

PR #3794, opened 2024-04-18
Fix rare missing/duplicate pixels in truncated joint

ffreyer:
# Description

#3558 still had an issue where sometimes both line segments would draw a pixel or both discard a pixel in a truncated joint. See for example:

Screenshot from 2024-04-18 16-59-14

The discard was controlled by comparing the distance of a pixel/fragment along line direction from the shared line point. If it is closer to the first line segments it is drawn by it, otherwise by the next one. The problem with this is that the two line segments sometimes disagree on which one is closer, resulting in double or no draw. This happens because the signed distances were computed in the vertex/geom shader and passed as an interpolated vertex attribute. Since the two segments do not share all vertices, the interpolation differs on a float precision level.

The changes in this pr are:
- The calculation of the relevant signed distances now happens in the fragment shader. This avoids issues from interpolation.
- The discard decision is now based on the distance of the `FragCoord` from the shared line point in miter vector direction. All of these values are calculated the same way between the involved segments. (Which should also be true for the old method.)  This should also be a more stable decision than before, since distances along line direction will differ less as the line approaches a 180° turn while the new process doesn't care.

This pr may also be a slight performance improvement since it gets rid of two interpolated vertex attributes.

## Type of change

## Checklist

### DIFF
--- a/GLMakie/assets/shader/line_segment.geom
+++ b/GLMakie/assets/shader/line_segment.geom
@@ -16,9 +16,7 @@ in {{stripped_color_type}} g_color[];
 in uvec2 g_id[];
 in float g_thickness[];
 
-out float f_quad_sdf0;
-out vec3 f_quad_sdf1;
-out float f_quad_sdf2;
+out vec3 f_quad_sdf;
 out vec2 f_truncation;
 out float f_linestart;
 out float f_linelength;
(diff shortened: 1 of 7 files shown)


ffreyer:
Just fyi, refimg tests failed to catch:

| recorded | pr |
| --- | --- |
| Screenshot from 2024-04-19 20-06-39 | Screenshot from 2024-04-19 20-06-35 |

which I fixed now.

Also the issue this pr aims to fix shows up in "GLMakie/colorscale (lines)" but it's very subtle:
Screenshot from 2024-04-19 20-15-16

The issue WGLMakie has with that refimg do not get fixed by this.

PR #3798, opened 2024-04-18
Test #3697

asinghvi17:
Tests #3697 by deploying the docs from a local PR.

### DIFF
--- a/docs/explanations/animation.md
+++ b/docs/explanations/animation.md
@@ -28,14 +28,11 @@ nframes = 30
 framerate = 30
 hue_iterator = range(0, 360, length=nframes)
 
-record(fig, "color_animation.mp4", hue_iterator;
+Record(fig, hue_iterator;
         framerate = framerate) do hue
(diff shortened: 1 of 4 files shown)


asinghvi17:
cc @aplavin

docs/explanations/animation.md
```diff
@@ -28,15 +28,12 @@ nframes = 30
 framerate = 30
 hue_iterator = range(0, 360, length=nframes)
 
-record(fig, "color_animation.mp4", hue_iterator;
+Record(fig, hue_iterator;
```
asinghvi17, line 31: In this case it seems like Franklin isn't displaying the output - that should probably be corrected...

asinghvi17: https://github.com/MakieOrg/Makie.jl/blob/f51eb527806e184690b577e17e630e9c9c521cb9/docs/utils.jl#L93-L99 

needs some kind of dispatch on `VideoStream` objects, to return gifs...

Issue #3741, opened 2024-03-28
Add optional `origin` parameter to `pie`


JohannesNE:
Allow placing a `pie` at a user-specified origin. 

Example of intended effect:
```
fig, ax, pie1 = pie([1,2], color = [:blue, :red])
translate!(pie1, Point2(0.5, 0.5))
fig
```
![download](https://github.com/MakieOrg/Makie.jl/assets/11048760/7dc0e0fb-b881-49fd-ab8e-84cddf90072a)

ffreyer:
This just requires adding an attribute in the `@recipe` block and translating the `points` generated here:
https://github.com/MakieOrg/Makie.jl/blob/master/src/basic_recipes/pie.jl

SimonDanisch:
Why not use `translate!(plot, x, y, z)`?

SimonDanisch:
Maybe we should add a recipe then, to do this kind of plot, more with a scatter like syntax?

JohannesNE:
It's a weird type of plot, so maybe a separate recipe is  overkill(?)

My own usecase is actually just to make a [circle sector](https://en.wikipedia.org/wiki/Circular_sector). @jkrumbiegel suggest using `pie` and `translate!` (or drawing it with `BezierPath`).
Maybe I should open an other issue for creating a circle sector instead? (like `arc` but with fill).

ffreyer:
Implemented by #4027 via merge of #4122

PR #3806, opened 2024-04-25
Add circle sector recipe

JohannesNE:
# Description

Adds recipe for creating a circle sector. 
`arc` is used as template for this, but instead of many lines imitating an arc, it creates a filled path using `BezierPath`. (Using code modified from an example from @jkrumbiegel ).

Related to #3741. And solves the problem I initially wanted to use the offset `pie` for.

Example:
```julia
using CairoMakie
f = Figure() 
Axis(f[1, 1], aspect = DataAspect())

for i in 1:10
    sector!(Point2f(i, 0), 0.4, 0, 2π*i/10, inner_radius = 0.2)
end

f
```
download

(btw, is there a smart way to trim this figure, which is also in the docs?)

I'm quite new to Julia/Dev, so I have a few questions:
- Should I make a test for this, or are the examples in the docs enough?
- An If-else statement is used to draw a sector if `inner_radius` is 0 or a "fat" arc if `inner_radius` is > 0. Is there a smarter way to do this? 

## Type of change

## Checklist


jkrumbiegel:
Thanks for the PR! I feel like this is so similar to `pie` that it's a bit confusing to have both. It has a scatter-like signature in that the position is the first positional arg(s) but other than that, I think an array-enabled `offset`  or `position` keyword for `pie` would do a similar thing, no?

jkrumbiegel:
Hm although pie has this cumulative effect where one passes the fractions in order, maybe one would have to remodel that behavior a little then. So that one can pass array-valued offsets as well so it's start-stop for each?

jkrumbiegel:
Ok I guess reusing it for pie internals makes more sense. Pie predates the bezier stuff, hence the vertices.

JohannesNE:
> Ok I guess reusing it for pie internals makes more sense. Pie predates the bezier stuff, hence the vertices.

But that is a separate PR right?, or should I try to include it in this?

jkrumbiegel:
If it's a scatter-like thing then I'd probably change the signature to `PointLike` style, so only the positions as positional args (with the default conversions from the PointLike trait). And the other settings as attributes, accepting either scalar or vector

JohannesNE:
Well, I did not consider whether it should be scatter-like or something else. 
Really, I just wanted something similar to `arc` but drawing sectors, so it seemed simplest to just make it behave similar to `arc`. 
Maybe it is more appropriate to add a `Sector` feature to GeometryBasics.jl and use `poly`, like for drawing a `Circle` or a `Rect` (thought I have no idea how to do that).

ffreyer:
Is this redundant with #4027? (pie has a position argument and `inner_radius` now, so you can build a sector with it)

PR #3816, opened 2024-04-29
Add wrap recipe tutorial

SimonDanisch:
Fixes for #3736


asinghvi17:
Should this be merged into the v0.21 branch instead, since it references Makie v0.21?  Or we could put that in a warning block...

Issue #3164, opened 2023-08-17
WilkinsonTicks is undocumented


rkube:
I'm trying to set yticks at 0.1 spacing:
```
    ax_s = Axis(fig_s[1, 1], 
        limits = (-0.05, 0.05 + maximum(values_x), 0.0, 1.1 * maximum(values_y)),
        xticks=WilkinsonTicks(0.1))
```
error is:

```
ERROR: MethodError: no method matching WilkinsonTicks(::Float64)

Closest candidates are:
  WilkinsonTicks(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ Makie ~/.julia/packages/Makie/uAmck/src/makielayout/types.jl:59
  WilkinsonTicks(::Int64; k_min, k_max, Q, granularity_weight, simplicity_weight, coverage_weight, niceness_weight, min_px_dist)
   @ Makie ~/.julia/packages/Makie/uAmck/src/makielayout/ticklocators/wilkinson.jl:1
  WilkinsonTicks(::Int64, ::Int64, ::Int64, ::Vector{Tuple{Float64, Float64}}, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64)
   @ Makie ~/.julia/packages/Makie/uAmck/src/makielayout/types.jl:59

Stacktrace:
```

The documentation suggests that `WilkinsonTicks` is the default tick finder but does not explain how to use it. The documentation itself is not helpful either:

```
help?> WilkinsonTicks
search: WilkinsonTicks

  No documentation found.

  Summary
  ≡≡≡≡≡≡≡≡≡

  struct WilkinsonTicks

  Fields
  ≡≡≡≡≡≡≡≡

  k_ideal            :: Int64
  k_min              :: Int64
  k_max              :: Int64
  Q                  :: Vector{Tuple{Float64, Float64}}
  granularity_weight :: Float64
  simplicity_weight  :: Float64
  coverage_weight    :: Float64
  niceness_weight    :: Float64
  min_px_dist        :: Float64

```

jkrumbiegel:
`WilkinsonTicks` uses PlotUtils.optimize_ticks pretty much directly so if you're motivated, you could transfer the information from that docstring over to Makie in a PR: https://github.com/JuliaPlots/PlotUtils.jl/blob/master/src/ticks.jl#L51

PR #3717, opened 2024-03-18
Document WilkinsonTicks

laikq:
# Description

Fixes #3164 

Adds documentation for WilkinsonTicks, mostly copying the description from PlotUtils. Does this cause a license issue?

## Type of change

Adds documentation

## Checklist

I have not run tests yet because my laptop is really slow. But I guess this change shouldn't cause anything to break anyways.

PR #3819, opened 2024-04-30
document `WilkinsonTicks`

t-bltg:
# Description

Fix https://github.com/MakieOrg/Makie.jl/issues/3164.

- supersedes https://github.com/MakieOrg/Makie.jl/pull/3717 by instead dynamically interpolate the docstring from `PlotUtils.optimize_ticks` (default values are identical);
- rename `tickvalues -> ticklocations` for consistency with the aforementioned docstring;
- remove unused `min_px_dist`;
- update `CHANGELOG`.

Resulting docstring:
```julia
help?> Makie.WilkinsonTicks
  Wilkinson's algorithm for positioning ticks.

  optimizeticks(xmin, xmax; extendticks::Bool = false, Q = [(1.0,1.0), (5.0,
  0.9), (2.0, 0.7), (2.5, 0.5), (3.0, 0.2)], kmin = 2, kmax = 10, kideal = 5,
  granularityweight = 1/4, simplicityweight = 1/6, coverageweight = 1/3,
  nicenessweight = 1/4, strictspan = true, span_buffer = nothing)

  Find some reasonable values for tick marks.

  This is basically Wilkinson's ad-hoc scoring method that tries to balance
  tight fit around the data, optimal number of ticks, and simple numbers.

  Arguments:
  ==========

    •  xmax:
       The maximum value occurring in the data.

    •  xmin:
       The minimum value occurring in the data.

    •  extend_ticks:
       Determines whether to extend tick computation. Defaults to false.

    •  strict_span:
       True if no ticks should be outside [xmin, xmax]. Defaults to true.

    •  Q:
       A distribution of nice numbers from which labellings are sampled.
       Stored in the form (number, score).

    •  k_min:
       The minimum number of ticks.

    •  k_max:
       The maximum number of ticks.

    •  k_ideal:
       The ideal number of ticks.

    •  granularity_weight:
       Encourages returning roughly the number of labels requested.

    •  simplicity_weight:
       Encourages nicer labeling sequences by preferring step sizes that
       appear earlier in Q. Also rewards labelings that include 0 as a
       way to ground the sequence.

    •  coverage_weight:
       Encourages labelings that do not extend far beyond the range of
       the data, penalizing unnecessary whitespace.

    •  niceness_weight:
       Encourages labellings to produce nice ranges.

  Returns:
  ========

  (ticklocations::Vector{Float64}, x_min, x_max)

  Mathematical details
  ====================

  Wilkinson’s optimization function is defined as the sum of three components.
  If the user requests m labels and a possible labeling has k labels, then the
  components are simplicity, coverage and granularity.

  These components are defined as follows:

:$

  \begin{aligned} &\text{simplicity} = 1 - \frac{i}{|Q|} + \frac{v}{|Q|}\
  &\text{coverage} = \frac{x{max} - x{min}}{\mathrm{label}{max} -
  \mathrm{label}{min}}\ &\text{granularity}= 1 - \frac{\left|k - m\right|}{m}
  \end{aligned} :$

  and the variables here are:

    •  q: element of Q.

    •  i: index of q ∈ Q.

    •  v: 1 if label range includes 0, 0 otherwise.
```

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Document `WilkinsonTicks` [#3819](https://github.com/MakieOrg/Makie.jl/pull/3819).
 - Improved thread safety of rendering with CairoMakie (independent `Scene`s only) by locking FreeType handles [#3777](https://github.com/MakieOrg/Makie.jl/pull/3777).
 
 ## [0.20.9] - 2024-03-29
--- a/src/makielayout/ticklocators/wilkinson.jl
+++ b/src/makielayout/ticklocators/wilkinson.jl
@@ -1,25 +1,28 @@
-function WilkinsonTicks(k_ideal::Int; k_min = 2, k_max = 10,
-        Q = [(1.0,1.0), (5.0, 0.9), (2.0, 0.7), (2.5, 0.5), (3.0, 0.2)],
-        granularity_weight = 1/4,
-        simplicity_weight = 1/6,
-        coverage_weight = 1/3,
-        niceness_weight = 1/4,
-        min_px_dist = 50.0)
-
+function WilkinsonTicks(
+    k_ideal::Int;
+    k_min = 2, k_max = 10,
+    Q = [(1.0, 1.0), (5.0, 0.9), (2.0, 0.7), (2.5, 0.5), (3.0, 0.2)],
+    granularity_weight = 1/4,
+    simplicity_weight = 1/6,
+    coverage_weight = 1/3,
+    niceness_weight = 1/4
+)
     if !(0 < k_min <= k_ideal <= k_max)
         error("Invalid tick number specifications k_ideal $k_ideal, k_min $k_min, k_max $k_max")
     end
 
-    WilkinsonTicks(k_ideal, k_min, k_max, Q, granularity_weight, simplicity_weight,
-        coverage_weight, niceness_weight, min_px_dist)
+    WilkinsonTicks(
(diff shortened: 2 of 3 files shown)


jkrumbiegel:
Good idea, I think it would be best to add the default arguments manually in the docstring and after that add a line like 

```
`WilkinsonTicks` is a thin wrapper over `PlotUtils.optimize_ticks`, the docstring of which is reproduced below:
```

This way it's clearer that we didn't write that part with Makie in mind.

Also, could you revert the unrelated formatting change of the commented out code, please.

t-bltg:
> Good idea, I think it would be best to add the default arguments manually in the docstring and after that add a line like

Thanks, I've added the suggestion, with the default values to the wrapper.

> Also, could you revert the unrelated formatting change of the commented out code, please.

Could you clarify which changes you are referring to ?
There is no consensus about the code style to use in `Makie` :confused: .
https://github.com/MakieOrg/Makie.jl/pull/2520 was made for that purpose (avoiding style change discussions), but it doesn't seem to move forward.

jkrumbiegel:
grafik
 that one

I'd be all for deciding for a formatter style and enforcing that for consistency. It's just a bit disruptive whenever it happens. Maybe after 0.21 is a good point as I want to focus on docs and other usability stuff anyway.

jkrumbiegel (approved):

t-bltg:
To whoever will be merging this PR, please retain credits when squashing for @laikq 's contribution in https://github.com/MakieOrg/Makie.jl/pull/3717.

jkrumbiegel:
you can squash your other commits into one and force push back

PR #3824, opened 2024-05-05
Add docstrings to bezier path building blocks

jkrumbiegel:


src/bezier.jl
```diff
@@ -1,17 +1,37 @@
...
+A path command for use within a `BezierPath` which starts a new subpath at the given point.
+"""
 struct MoveTo
     p::Point2d
 end
 
 MoveTo(x, y) = MoveTo(Point2d(x, y))
 
+"""
+    LineTo(p::Point2)
```
asinghvi17, line 16: It looks like these methods will only take `Point2d` by the default constructor, though?  May want to add an extra `LineTo(p::VecTypes) = Point2d(p)` or something.

jkrumbiegel: all the default conversions would work

jkrumbiegel: I can change the signature so it doesn't actually look constrained in the constructor

PR #3845, opened 2024-05-09
CairoMakie: Allow restricting PDF version

barucden:
# Description

This PR allows users to restrict the version of an output PDF like this:
```julia
fig = Figure()
ax = Axis(fig[1, 1])
x = range(0, 10, length=100)
y = sin.(x)
save("version_1-4.pdf", fig, pdf_version="1.4")
save("version_1-7.pdf", fig, pdf_version="1.7")

shell> file version_1-4.pdf
version_1-4.pdf: PDF document, version 1.4, 1 page(s)

shell> file version_1-7.pdf
version_1-7.pdf: PDF document, version 1.7
```

This feature was discussed on [Discourse](https://discourse.julialang.org/t/cairomakie-restrict-pdf-version/113939) (and also [earlier](https://discourse.julialang.org/t/pdf-version-used-when-saving-a-figure-in-cairomakie/102282))

## Type of change

## Checklist

I need guidance regarding documentation (where to best put this) ~and testing (I am not sure how to check a PDF's version in Julia)~. 

Also I was thinking that the `CAIRO_PDF_VERSION_*` constants together with `restrict_pdf_version!` should perhaps go to Cairo.jl. What do you think?

### DIFF
--- a/CairoMakie/src/cairo-extension.jl
+++ b/CairoMakie/src/cairo-extension.jl
@@ -74,3 +74,10 @@ function get_render_type(surface::Cairo.CairoSurface)
     typ == Cairo.CAIRO_SURFACE_TYPE_IMAGE && return IMAGE
     return IMAGE # By default assume that the render type is IMAGE
 end
+
+function restrict_pdf_version!(surface::Cairo.CairoSurface, v::Integer)
+    @assert surface.ptr != C_NULL
+    0 ≤ v ≤ 3 || throw(ArgumentError("version must be 0, 1, 2, or 3 (received $v)"))
+    ccall((:cairo_pdf_surface_restrict_to_version, Cairo.libcairo), Nothing,
+          (Ptr{UInt8}, Int32), surface.ptr, v)
+end
--- a/CairoMakie/src/screen.jl
+++ b/CairoMakie/src/screen.jl
@@ -68,6 +68,15 @@ function surface_from_output_type(type::RenderType, io, w, h)
     end
 end
 
+@enum PDFVersion PDFv14 PDFv15 PDFv16 PDFv17
+function pdfversion(version::AbstractString)
+    version == "1.4" && return PDFv14
+    version == "1.5" && return PDFv15
+    version == "1.6" && return PDFv16
+    version == "1.7" && return PDFv17
+    throw(ArgumentError("PDF version must be one of '1.4', '1.5', '1.6', '1.7' (received '$version')"))
+end
+
 """
 Supported options: `[:best => Cairo.ANTIALIAS_BEST, :good => Cairo.ANTIALIAS_GOOD, :subpixel => Cairo.ANTIALIAS_SUBPIXEL, :none => Cairo.ANTIALIAS_NONE]`
 """
@@ -85,6 +94,7 @@ to_cairo_antialias(aa::Int) = aa
 * `pt_per_unit = 0.75`
 * `antialias::Union{Symbol, Int} = :best`: antialias modus Cairo uses to draw. Applicable options: `[:best => Cairo.ANTIALIAS_BEST, :good => Cairo.ANTIALIAS_GOOD, :subpixel => Cairo.ANTIALIAS_SUBPIXEL, :none => Cairo.ANTIALIAS_NONE]`.
 * `visible::Bool`: if true, a browser/image viewer will open to display rendered output.
+* `pdf_version::String = nothing`: the version of output PDFs. Applicable options are `"1.4"`, `"1.5"`, `"1.6"`, `"1.7"`, or `nothing`, which leaves the PDF version unrestricted.
 """
 struct ScreenConfig
     px_per_unit::Float64
@@ -92,6 +102,14 @@ struct ScreenConfig
     antialias::Symbol
     visible::Bool
     start_renderloop::Bool # Only used to satisfy the interface for record using `Screen(...; start_renderloop=false)` for GLMakie
+    pdf_version::Union{Nothing, PDFVersion}
+
+    function ScreenConfig(px_per_unit::Real, pt_per_unit::Real,
+            antialias::Symbol, visible::Bool, start_renderloop::Bool,
+            pdf_version::Union{Nothing, AbstractString})
+        v = isnothing(pdf_version) ? nothing : pdfversion(pdf_version)
+        new(px_per_unit, pt_per_unit, antialias, visible, start_renderloop, v)
+    end
 end
 
 function device_scaling_factor(rendertype, sc::ScreenConfig)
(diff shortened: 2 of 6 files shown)


CairoMakie/src/screen.jl
```diff
@@ -213,6 +214,11 @@ function apply_config!(screen::Screen, config::ScreenConfig)
     aa = to_cairo_antialias(config.antialias)
     Cairo.set_antialias(context, aa)
     set_miter_limit(context, 2.0)
+
+    if get_render_type(surface) === PDF
+        restrict_pdf_version!(surface, config.pdf_version)
```
asinghvi17, line 219: This should be optional, so that output remains unchanged.

CairoMakie/src/cairo-extension.jl
```diff
@@ -74,3 +74,23 @@ function get_render_type(surface::Cairo.CairoSurface)
     typ == Cairo.CAIRO_SURFACE_TYPE_IMAGE && return IMAGE
     return IMAGE # By default assume that the render type is IMAGE
 end
+
+const CAIRO_PDF_VERSION_1_4 = 0
+const CAIRO_PDF_VERSION_1_5 = 1
+const CAIRO_PDF_VERSION_1_6 = 2
+const CAIRO_PDF_VERSION_1_7 = 3
```
asinghvi17, line 81: Should these be an `@enum`?

```suggestion
@enum CAIRO_PDF_VERSION begin
	CAIRO_PDF_VERSION_1_4
	CAIRO_PDF_VERSION_1_5
	CAIRO_PDF_VERSION_1_6
	CAIRO_PDF_VERSION_1_7
end

File: src/theming.jl [hunk 3]
```diff
@@ -89,7 +89,8 @@ const MAKIE_DEFAULT_THEME = Attributes(
         pt_per_unit = 0.75,
         antialias = :best,
         visible = true,
-        start_renderloop = false
+        start_renderloop = false,
+        pdf_version = "1.7"
```
asinghvi17, line 93: Would this cause a change compared to current output?

barucden: It is my understanding that cairo sets the PDF version to 1.7 by default (see here https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/src/cairo-pdf-surface.c?ref_type=heads#L518), so I believe that "restricting" the version to 1.7 is the same as not restricting at all.

CairoMakie/src/cairo-extension.jl
barucden: I followed the style of Cairo.jl (see https://github.com/JuliaGraphics/Cairo.jl/blob/master/src/constants.jl#L70) with the intention of potentially moving that code over to Cairo.jl.

CairoMakie/src/screen.jl
barucden: You mean to set `pdf_version` to `nothing` (or similar) by default and skip restricting the PDF version in that case?

asinghvi17: Yep, you elaborated more on this in the last comment but the Cairo version may change, or the user may use their own local Cairo which may not support that specific PDF version, so best to be safe

barucden: The version can now be an empty string, in which case no restriction is applied.

barucden:
I have added an entry to changelog and documented the new config option.

asinghvi17 (approved):
Looks good to me!

barucden:
Rebased to resolve new conflicts

barucden:
The error is not related imho:
```
GLMakie/Normals of a Cat.png: Test Failed at /home/runner/work/Makie.jl/Makie.jl/ReferenceTests/src/runtests.jl:117
  Expression: score <= threshold
   Evaluated: 0.5715945363044739 <= 0.05
```

barucden:
CI's green ✅ Thank you, Simon.

barucden:
Rebased to fix conflicts.

CairoMakie/src/screen.jl
```diff
@@ -85,13 +85,15 @@ to_cairo_antialias(aa::Int) = aa
 * `pt_per_unit = 0.75`
 * `antialias::Union{Symbol, Int} = :best`: antialias modus Cairo uses to draw. Applicable options: `[:best => Cairo.ANTIALIAS_BEST, :good => Cairo.ANTIALIAS_GOOD, :subpixel => Cairo.ANTIALIAS_SUBPIXEL, :none => Cairo.ANTIALIAS_NONE]`.
 * `visible::Bool`: if true, a browser/image viewer will open to display rendered output.
+* `pdf_version::String = ""`: the version of output PDFs. Applicable options are `"1.4"`, `"1.5"`, `"1.6"`, `"1.7"`, or the default, `""`, which leaves the PDF version unrestricted.
```
jkrumbiegel, line 88: I'd use `nothing` instead of empty string for the default

CairoMakie/src/screen.jl
```diff
@@ -85,13 +85,15 @@ to_cairo_antialias(aa::Int) = aa
...
 """
 struct ScreenConfig
     px_per_unit::Float64
     pt_per_unit::Float64
     antialias::Symbol
     visible::Bool
     start_renderloop::Bool # Only used to satisfy the interface for record using `Screen(...; start_renderloop=false)` for GLMakie
+    pdf_version::String
```
jkrumbiegel, line 96: Instead of storing any string here, and then validating later, I would create a struct `PdfVersion` or so which allows conversion from `String` so that you validate at creation of the screen config. Internally, the `PdfVersion` would store the integer that you defined elsewhere.

barucden: Okay, I think I implemented the changes as requested.

CairoMakie/src/screen.jl
```diff
@@ -68,6 +68,23 @@ function surface_from_output_type(type::RenderType, io, w, h)
...
+function pdfversion(version::AbstractString)
+    version == "1.4" && return PDFv14
+    version == "1.5" && return PDFv15
+    version == "1.6" && return PDFv16
+    version == "1.7" && return PDFv17
+    throw(ArgumentError("PDF version must be one of '1.4', '1.5', '1.6', '1.7' (received '$version')"))
+end
+
+function cairo_pdf_version(version::PDFVersion)
+    version === PDFv14 && return 0
```
jkrumbiegel, line 81: this is not necessary, enums already go from 0 to N:

```julia
julia> @enum ENUM A B C

julia> A
A::ENUM = 0

julia> Int(A)
0
```

barucden: done

CairoMakie/src/cairo-extension.jl
```diff
@@ -74,3 +74,10 @@ function get_render_type(surface::Cairo.CairoSurface)
     typ == Cairo.CAIRO_SURFACE_TYPE_IMAGE && return IMAGE
     return IMAGE # By default assume that the render type is IMAGE
 end
+
+function restrict_pdf_version!(surface::Cairo.CairoSurface, v::Integer)
```
jkrumbiegel, line 78: would be simpler to type the `v` with the enum here, then you don't need yet another check. I'd guess you don't even have to manually convert it to Int32 in the ccall

barucden: I would like to move `restrict_pdf_version!` to Cairo.jl eventually, where they stick to numerical constants instead of enums, so I'd prefer to keep the argument as `Integer`.

You are right that the manual conversion in the ccall is not needed. I removed it.

barucden:
This PR was taken over by #3885.

PR #3856, opened 2024-05-12
Tweak boundingbox handling for better compatability with nonlinear transforms

ffreyer:
# Description

This is in reference to https://github.com/MakieOrg/GeoMakie.jl/pull/207#issuecomment-2105952989

The pr mostly moves some code around, so that it's now `apply_transform(transform, bbox::Rect3d)`s job to transform the `data_limits` of a plot. That way a nonlinear transform can be aware that it's transforming a bounding box and implement specific code for it. I'm hoping that is a good compromise between performance, simplicity and correctness.

The more correct alternative would be to generate all the points relevant to a plot, transform each individually and generate a bounding box afterwards. This would be easier for people who implement their own transformations, as it should just work. But on the other hand it would be slower, because of all the transformations that are often unnecessary.

@asinghvi17 Do you think this is a reasonable solution, or do you think we should use the transform-first-bbox-later approach?

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- `boundingbox` now relies on `apply_transform(transform, data_limits(plot))` rather than transforming the corner points of the bounding box [#3856](https://github.com/MakieOrg/Makie.jl/pull/3856)
+
 ## [0.21.0] - 2024-05-08
 
 - Add `voxels` plot [#3527](https://github.com/MakieOrg/Makie.jl/pull/3527).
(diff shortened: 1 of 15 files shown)


asinghvi17:
Thanks @ffreyer!  This looks like a good compromise to me - and of course specific transforms can implement that sampling on their own.

One such case would be a cartesian to spherical transform, where the boundingbox would just be a line on the sphere if the cartesian domain fills the sphere exactly, but that transformation would be aware of this and could account for it in its `apply_transform(transf, ::Rect)` method.

asinghvi17:
Just checked, this provides correct bounds in GeoMakie!
download-17

asinghvi17:
The CairoMakie CI failure is pretty odd - the meshes are somehow no longer intersecting in the way they used to.

New
Merged color Mesh

Old
Merged color Mesh

ffreyer:
Yea I don't know what's happening there. Maybe it's some float precision thing causing a different render order. It's not a change that should delay this though

SimonDanisch:
Maybe this even recorvers the old behaviour. For 0.21, I updated this reference image with a similarly small change!

SimonDanisch (approved):

Issue #3847, opened 2024-05-09
conversion traits not taken into account in 0.21


aplavin:
In 0.21, `convert_argument(trait, args...)` methods are not really called with user-specified arguments.
That's due to the new `expand_dimensions` function called before `convert_arguments` with conversion trait: https://github.com/MakieOrg/Makie.jl/blob/d157f9cf907971b703f2a2de2dbcb3cd91475c64/src/conversions.jl#L17-L21.

Noticed when trying to update KeyedArray conversions (in AxisKeysExtra.jl) to the new Makie version, specifically conversions like https://github.com/JuliaAPlavin/AxisKeysExtra.jl/blob/0000000065d69673f7f40c0f93f01ba0fbbbda34/ext/MakieExt.jl#L10-L13. MWE:

```julia
julia> using AxisKeys, Makie

julia> Makie.convert_arguments(ct::PointBased, x::KeyedArray{<:Any,1}) = convert_arguments(ct, only(axiskeys(x)), AxisKeys.keyless_unname(x))

julia> convert_arguments(Scatter, KeyedArray([1,2,3], x=10:10:30))
# 0.20, expected:
(Point{2, Float32}[[10.0, 1.0], [20.0, 2.0], [30.0, 3.0]],)
# 0.21, unexpected:
(10:10:30, Point{2, Float64}[[1.0, 1.0], [2.0, 2.0], [3.0, 3.0]])
```

- [x] are you running newest version (version from docs) ? bug only appears in 0.21, while 0.20 works fine
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x] What platform + GPU are you on? any platform

asinghvi17:
Defining `Makie.expand_dimensions(::PointBased, k::KeyedArray) = (only(axiskeys(x)), AxisKeys.keyless_unname(x))` seems to fix it, though I imagine that's not the ideal way forward...

SimonDanisch:
It's a separate step to divide a user defined type into it's x and y components

PR #3859, opened 2024-05-12
propagate args as-is to convert_arguments(trait, args...)

aplavin:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/3847
The Makie 0.21 version doesn't propagate `convert_arguments` arguments to `convert_arguments(trait)` directly, making some common conversions impossible (#3847 for example). Turns out, the fix (this PR) is pretty simple.

## Type of change

## Checklist

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -16,9 +16,6 @@ function convert_arguments(T::Type{<:AbstractPlot}, args...; kw...)
     # Meaning, it needs to be a conversion trait, or it needs single_convert_arguments or expand_dimensions
     CT = conversion_trait(T, args...)
 
-    # Try to expand dimensions first, as this is the most basic step!
-    expanded = expand_dimensions(CT, args...)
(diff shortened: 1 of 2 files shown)


SimonDanisch:
The new `expand_dimensions` was introduced for a reason, so this should not be the correct fix...
I'm surprised that this passes tests though, so I guess I need to investigate this a bit.
In 0.21 you will need to start overloading `expand_dimensions` for your example. This should get documented a bit better!

SimonDanisch:
Ok thank you! I still think this was covering some edge case and that for your overload to work properly you will need to switch to overloading `expand_dimensions`. But I may be mistaken, those last changes where done a bit in a rush.
I will need to take a look at it tomorrow.

SimonDanisch:
We will need to think more about how to streamline x,y,z arguments in Makie and how to decouple it from `convert_arguments`.
Since i couldn't find any corner case that stopped working with this, we might as well merge it for now!

SimonDanisch:
Thank you!

PR #3861, opened 2024-05-13
Move docs to vitepress

jkrumbiegel:


docs/makedocs.jl
```diff
@@ -1,93 +1,106 @@
...
-# cd(@__DIR__); serve(single=false, cleanup=true, clear=true, fail_on_warning = false)
+include("figure_block.jl")
 
-cd("__site") do
-    run(`$pagefind --site . --root-selector .franklin-content`)
-end
+makedocs(;
+    # modules=[Makie],
+    sitename="Makie.jl",
+    format=DocumenterVitepress.MarkdownVitepress(;
```
asinghvi17, line 36: should add a `description` kwarg to MarkdownVitepress for SEO

docs/src/explanations/colors.md
```diff
@@ -117,8 +109,67 @@ text!(ax, 4, 0, text = ":viridis",
...
-{{colorschemes}}
+### misc
+
+These colorschemes are not defined or provide different colors in ColorSchemes.jl
+They are kept for compatibility with the old behaviour of Makie, before v0.10.
+
+```@example colors
+using Markdown # hide
+using Makie.PlotUtils # hide
+Core.include(@__MODULE__, "colormap_generation.jl") # hide
```
asinghvi17, line 123: Couldn't this be regular `include`?

jkrumbiegel: threw a UndefVarError, if Documenter uses some kind of bare modules, they wouldn't have an include defined

asinghvi17:
Screenshot 2024-05-15 at 7 15 54 AM
I feel like we should decrease the size of the sidebar, it's taking up too much space and I did like the extra horizontal room in the Franklin docs.

jkrumbiegel:
@asinghvi17 the problem with the images is rather that I don't know a way yet to show them at their intended size (the figure size, not the pixel size). In Franklin, I had some code that read those dimensions out so that `width` and `height` in the html could be set properly. Not sure how that could be done with Documenter as the images go through the MIME display mechanism, I don't save them myself. Do you have an idea?

That problem happens only with pngs of course, but I have not been using svgs because they seem to be rolled up into big chunks of text and on larger pages the results were sometimes faulty (empty figures etc, something seems to be going on there in the processing). Also something to look into.

asinghvi17:
Documenter always executes example blocks in order (so a race condition is impossible), so we can have the example set the value of some `Ref` in the MakieDocsHelper module, and then modify the output dict (I have code for this somewhere) to pass some struct `(backend, size, figure)` directly to DocumenterVitepress's rendering method.

lazarusA:
>  that width and height in the html could be set properly.

I think we could always define some css style for it directly [here](https://github.com/MakieOrg/Makie.jl/blob/jk/vitepress/docs/src/.vitepress/theme/style.css), something like 

```
img {
width: 60%;
height: auto;
}
```

asinghvi17:
The issue is more that we don't know the `px_per_unit` setting, so we have to have some way to propagate the figure size through.

PR #3864, opened 2024-05-13
Improve model and transform_func interaction with Axis limits

### LINKED ISSUE #3562: Fit objects into scene after Makie transformations
From a Discourse [thread](https://discourse.julialang.org/t/how-to-translate-and-scale-makie-image/108741), here is a MWE by @eliascarv where the image is partially displayed:

```julia
using Makie
using GLMakie

@recipe(ScaledImage, img, x, y) do scene
  Attributes()
end

function Makie.plot!(plot::ScaledImage)
  image!(plot, plot.img, interpolate=false)
  scale!(plot, plot.x[], plot.y[])
  # how to adjust the axis after scaling the plot?
end

img = rand(1:9, 10, 10)
scaledimage(img, 2.5, 1.5)
```
image

ffreyer:
# Description

This is an attempt at resolving issues with model transformations (`scale!`, `translate!`) in Axis with or without tranform_func by relying on `boundingbox` instead of `data_limits` and undoing the transform_func afterwards. My gut feeling is that this will result in domain errors for some transformations in some situations... 

This should fix #3562 and similar issues.

Based on #3856 since there are some overlapping issues (hexbin typo, model application with NaN)

## Type of change

## Checklist

### DIFF
--- a/ReferenceTests/src/tests/examples2d.jl
+++ b/ReferenceTests/src/tests/examples2d.jl
@@ -1151,10 +1151,10 @@ end
 
     fig = Figure(size = (600, 600))
     # Create a recipe plot
-    ax, plot_top = heatmap(fig[1, 1], randn(10, 10))
+    ax, plot_top = heatmap(fig[1, 1], randn(10, 10), colormap = [:transparent])
     # Plot some recipes at the level below the contour
-    scatterlineplot_1 = scatterlines!(plot_top, 1:10, 1:10; linewidth = 20, markersize = 20, color = :red)
-    scatterlineplot_2 = scatterlines!(plot_top, 1:10, 1:10; linewidth = 20, markersize = 30, color = :blue)
+    scatterlineplot_1 = scatterlines!(ax, 1:10, 1:10; linewidth = 20, markersize = 20, color = :red)
+    scatterlineplot_2 = scatterlines!(ax, 1:10, 1:10; linewidth = 20, markersize = 30, color = :blue)
     # Translate the lowest level plots (scatters)
     translate!(scatterlineplot_1.plots[2], 0, 0, 1)
     translate!(scatterlineplot_2.plots[2], 0, 0, -1)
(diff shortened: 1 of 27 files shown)


ReferenceTests/src/tests/examples2d.jl
```diff
@@ -1151,10 +1151,10 @@ end
 
     fig = Figure(size = (600, 600))
     # Create a recipe plot
-    ax, plot_top = heatmap(fig[1, 1], randn(10, 10))
+    ax, plot_top = heatmap(fig[1, 1], randn(10, 10), colormap = [:transparent])
     # Plot some recipes at the level below the contour
-    scatterlineplot_1 = scatterlines!(plot_top, 1:10, 1:10; linewidth = 20, markersize = 20, color = :red)
-    scatterlineplot_2 = scatterlines!(plot_top, 1:10, 1:10; linewidth = 20, markersize = 30, color = :blue)
+    scatterlineplot_1 = scatterlines!(ax, 1:10, 1:10; linewidth = 20, markersize = 20, color = :red)
+    scatterlineplot_2 = scatterlines!(ax, 1:10, 1:10; linewidth = 20, markersize = 30, color = :blue)
```
ffreyer, line 1157: This test failed because of what I'd say is undefined behavior - plotting to a primitive plot. 
Specifically because `data_limits` has a specific overload for Heatmap, thus using heatmap limits, while `boundingbox` attempts to recognize primitives by their lack of child plots. Since heatmap has children here, boundingbox uses scatter + lines for limits. 
Plotting to a primitive also results in other weirdness - the heatmap doesn't show and the axis uses tight limits.

ffreyer:
Added refimg is only meant to test whether the Axis correctly bring splots in view that have scale + translation + transform_func:
image

ReferenceTests/src/tests/examples2d.jl
SimonDanisch: Oh, plotting to a primitive outside a recipe?!  That's something we shouldn't do in any case ;) Didn't realize we actually do this in the tests!

SimonDanisch (approved):

ffreyer:
We should merge #3856 first, rebase and let tests run here before merging

PR #3872, opened 2024-05-17
fix RPRMakie with Material

SimonDanisch:


SimonDanisch:
Huh, somehow the GLMakie cat image test became instable and seems to reset to different limits randomly.
Maybe some synchronization issue.
This is unrelated to this PR, will need to investigate this later. @ffreyer maybe you have an idea? I suspect it's related to one of the bb PRs

PR #3881, opened 2024-05-22
`empty!` the GLMakie screen for reuse instead of `close`ing and reopening

### LINKED ISSUE #2453: GLMakie window is closed and re-opened upon rendering of a new plot
As of 9f767a36640369950d662377579dca5d0d0c52ba (according to git bisect), the GLMakie rendering window is closed and re-opened between rendering of plots. On i3, this has the side-effect that the window pops up in a new container and "forgets" it's previous size.

I usually mark the GL window as floating and place it over my REPL but now that it reopens every time, that no longer works. I am not sure whether this change of behavior is also visible with other window managers.

jmert:
# Description

Fixes #2453

This fixes a problem observed on Linux (across at least a couple of different desktop environments) with windows flashing away and coming back in a different place (can be a different monitor!) when reusing the shared singleton screen.

## Type of change


SimonDanisch:
This is likely not correct and can't be merged like this, but good to know that this fixes it...

jmert:
The second commit eliminates (locally for me) the test errors that occurred on the last CI run. I tracked down the issue to being caused by the process of saving to a file stopping the render loop, but the reused singleton window would auto-close when its render loop was stopped (after switching to the `empty!` process instead of `close`) during the saving process.

Just encouraging the save-to-file code path to use screens from the pool (which still get `close`d before being re-added to the pool for reuse, IIUC) rather than the singleton solved the particular error cases on my machine, but like you said, this may be a bit fragile and not the ideal solution.

jmert:
The second commit caused an accumulation of screens and therefore a ballooning memory use. The latest commit simplifies by just copying what seems to be the only line in the `close` implementation which doesn't relate to controlling the visibility/closed state of the window, and this also seems to resolve the test errors locally for me.

jmert:
The debug output to compare to https://github.com/MakieOrg/Makie.jl/issues/2453#issuecomment-1532594689 is

```julia
julia> using GLMakie

julia> ENV["JULIA_DEBUG"] = "GLMakie"
"GLMakie"

julia> scatter(rand(10))
┌ Debug: new singleton screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:328
┌ Debug: reopening screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:295
┌ Debug: Applying screen config! to existing screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:341
┌ Debug: display scene on screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:407

julia> scatter(rand(10))
┌ Debug: reusing singleton screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:323
┌ Debug: Leaving renderloop, cause: stopped renderloop
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:953
┌ Debug: empty screen!
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:570
┌ Debug: reopening screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:295
┌ Debug: Applying screen config! to existing screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:341
┌ Debug: display scene on screen
└ @ GLMakie ~/.julia/dev/Makie/GLMakie/src/screen.jl:407
```

SimonDanisch (approved):
Thanks, this seems to be working! :)

PR #3883, opened 2024-05-23
don't remove underlying VideoStream file when doing save()

aplavin:
# Description

For some reason, the file was removed when calling `save(path, ::VideoStream)`. This breaks all following `show` calls: they, clearly, rely on the video file still being present.

## Type of change


SimonDanisch:
Hm, that was a lazy clean up, assuming that one records a video and then when saving it, it isn't needed anymore.
I think we'll need to move this into a finalizer for VideoStream to not have this accumulate.

SimonDanisch:
Yes, I guess we'll need to do that then ;)

SimonDanisch:
Thank you!

PR #3907, opened 2024-05-30
Add line loop rendering

ffreyer:
# Description

Changes GLMakie and WGLMakie to detect and render loops passed to `lines` as such. I.e. if someone passes a sequence of points `[A, B, ..., C, A]` bounded by the size of the array or by NaNs to lines, then the generated line will be rendered as a closed loop.

For GLMakie this is done by adjusted the sequence of vertices to `[C, A, B, ..., C, A, B]` with the generated indices and marking C and B as `valid = 3`. The shader then detects whether C-A and A-B appear at the start/end of line, where they don't draw to avoid repeated drawing, or in the middle of a line where they do. This detection requires a nan/invalid point between a loop and the next line/loop.

For WGLMakie we don't need to mark and detect C and B at the start/end of the loop, as having a NaN before/after them automatically causes the line segment to be discarded. So here we just adjust the duplicated vertices around a line segment for loops and tweak the linelength calculation to not count the discarded segments.

## Type of change

## Checklist


ReferenceTests/src/tests/primitives.jl
```diff
@@ -84,8 +84,28 @@ end
...
+        nan, [Point2f(1, 1)], nan
+    )
+
+    f, a, p = lines(loop((-1, -1)), linewidth = 20, linecap = :round, alpha = 0.5)
+    lines!(ps, linewidth = 20, linecap = :round, alpha = 0.5)
+    lines!(vcat(nan, nan, line((1, 1)), nan), linewidth = 20, linecap = :round, alpha = 0.5)
+    f
+end
+
+@reference_test "Miter Limit" begin
```
ffreyer, line 108: I think we forgot to enable this in/before the 0.21 pr?

ffreyer:
The test failures are from loops now closing e.g.:
Screenshot from 2024-05-30 17-13-48

jkrumbiegel:
Yeah I wouldn't consider that breaking

SimonDanisch (approved):

PR #3911, opened 2024-05-30
Fix Block macro when Makie is only imported

asinghvi17:
# Description

Fixes the Block macro trying to access Makie by name when it is called.  Instead, all uses of Makie-specific things are either qualified by interpolated module (`$(Makie).method`) or types are directly interpolated (`::$(Scene)`).

This test and criterion is a little extreme, but it's better this way since we can be 100% sure that the macro only refers to types and not symbols which may or may not be resolved correctly, depending on how the user has imported Makie.

I initially ran into this issue when updating GeoMakie - there were a few undefined variable errors, which were pretty simple to trace back to the block definition.

## Type of change

## Checklist


jkrumbiegel:
instead of interpolating all names into the ast, the `esc` call at the end should be removed, and only the bare necessities should be escaped

asinghvi17:
How would you refer to the type if it's in a different scope, though?  Wouldn't that be unavailable to the non-`esc`'ed code?

jkrumbiegel:
@asinghvi17 maybe look at https://github.com/MakieOrg/Makie.jl/blob/master/MakieCore/src/recipes.jl#L399-L469, I think I did it better there without escaping. I probably got confused with the function definition semantics in macros before.

asinghvi17:
Gotcha, looks like most things are interpolated there and only a few use `esc`.  Will give it a shot, thanks!

src/makielayout/blocks.jl
```diff
@@ -10,13 +10,13 @@ macro Block(_name::Union{Expr, Symbol}, body::Expr = Expr(:block))
...
+    type_expr = _name isa Expr ? _name : :($_name <: $(Makie).Block)
     name = _name isa Symbol ? _name : _name.args[1]
     structdef = quote
         mutable struct $(type_expr)
-            parent::Union{Figure, Scene, Nothing}
-            layoutobservables::Makie.LayoutObservables{GridLayout}
-            blockscene::Scene
+            parent::Union{$(Figure), $(Scene), Nothing}
+            layoutobservables::$(Makie).LayoutObservables{$(Makie).GridLayout}
+            blockscene::$(Makie).Scene
```
SimonDanisch, line 19: We could also consistently interpolate the hole type?

Issue #1662, opened 2022-02-10
Lift from single Observable


fbanning:
Sometimes when I play around with lifting and I refactor some of my code, I end up trying to lift a single Observable but doing `x = @lift($var)` throws

https://github.com/JuliaPlots/Makie.jl/blob/bac1f40315d5b2ee170d467842910dead853a28a/src/interaction/liftmacro.jl#L29

A user can circumvent this by doing either `x = lift(var) do v; v; end` (arguably crappy to read) or just `x = var` which in the end do *almost* the same thing. The big difference is that the former also registers `x` as a listener of `var`.

In the end that's also what I would intuitively expect the lift macro to do as well when only a single Observable is passed. So maybe these two options could be added to the error message to make it clearer to the user why lifting from a single Observable does not work in the macro.

Or (and I would quite frankly prefer this), we could even make the hacky lift-do solution from above the default behaviour of `@lift` instead of it throwing an error? Probably there are good reasons against it, so I'm happy to learn something here. :)

jkrumbiegel:
I mean I guess we could remove the error, I don't remember why I disallowed it at the time. We can just use `identity`, that's the same as your `do`

fbanning:
> We can just use `identity`, that's the same as your `do`

I'm not sure if it's really the same.

```julia
julia> a = Observable(1)
Observable{Int64} with 0 listeners. Value:
1

julia> b = identity(a)
Observable{Int64} with 0 listeners. Value:
1

julia> a
Observable{Int64} with 0 listeners. Value:
1

julia> a[] = 2
2

julia> b
Observable{Int64} with 0 listeners. Value:
2
```

This updates `b` as expected but doesn't add `b` as a listener to `a`. Using `identity` is therefore the same as doing `b = a` but not the same as the `do`-block in my example above.

I think it would be nice if user can rely on the `@lift` macro to add new variables as listeners to the existing Observable. Therefore I think the `do` solution would be better than just using `identity`. Would you agree?

jkrumbiegel:
Sorry I meant `lift(identity, obs)`

fbanning:
Ah yes, that works as expected. I'll try to prepare a PR for this, if you don't mind.

PR #3915, opened 2024-05-30
generalize lift() and @lift()

### LINKED ISSUE #3801: `lift` should support non-observables
Currently, `lift(f, x)` only works if `x` is an observable, but errors _(or does a completely different thing, because `lift` happens to be `=== map`)_ for regular `x` values.

But it would be natural and convenient to make `lift(f, x::Observable) = current behavior`, and `lift(f, x::Any) = f(x)`. This would enable preparing plotting arguments without manually handling observable and non-observable cases, like
```julia
# works no matter if x is a regular value or observable
y = lift(myfunc, x)
plot(y)
```

---

_**UPD**: use `lift` and `@lift` from MakieExtra.jl (no piracy), they support non-observables properly._

aplavin:
`@lift`: allow single observable, or no observables at all
`lift`: allow operating on non-observables

Motivation:
this enables preparing plotting arguments without manually handling observable and non-observable cases, like
```julia
# works no matter if x is a regular value or observable
y = lift(myfunc, x)
# or
y = @lift myfunc($x)

plot(y)
```
Fixes https://github.com/MakieOrg/Makie.jl/issues/3801 and #1662.

---

**UPD:** this PR appears to be effectively rejected here, but this exact `lift`/`@lift` functionality is made available in MakieExtra.jl.

### DIFF
--- a/src/interaction/observables.jl
+++ b/src/interaction/observables.jl
@@ -1,5 +1,15 @@
-# lift makes it easier to search + replace observable code, while `map` is really hard to differentiate from `map(f, array)`
-const lift = map
+# mimick Observables.jl map() signature to forward directly:
+lift(f, arg::AbstractObservable, args...; kwargs...) = map(f, arg, args...; kwargs...)
+# handle the general case:
+function lift(f, args...; kwargs...)
+    if !any(a -> isa(a, AbstractObservable), args)
+        # there are no observables
+        f(args...)
+    else
+        # there are observables, but not in the first position
+        lift((_, as...) -> f(as...), Observable(nothing), args...; kwargs...)
+    end
+end
 
 """
 Observables.off but without throwing an error
--- a/src/interaction/liftmacro.jl
+++ b/src/interaction/liftmacro.jl
@@ -24,22 +24,15 @@ end
 """
 Replaces every subexpression that looks like a observable expression with a substitute symbol stored in `exprdict`.
 """
-function replace_observable_expressions!(exp::Expr, exprdict)
+function replace_observable_expressions(exp::Expr, exprdict)
     if is_interpolated_observable(exp)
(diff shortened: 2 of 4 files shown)


src/interaction/observables.jl
```diff
@@ -1,5 +1,15 @@
...
+lift(f, arg::AbstractObservable, args...; kwargs...) = map(f, arg, args...; kwargs...)
+# handle the general case:
+function lift(f, args...; kwargs...)
+    if !any(a -> isa(a, AbstractObservable), args)
+        # there are no observables
+        f(args...)
+    else
+        # there are observables, but not in the first position
+        lift((_, as...) -> f(as...), Observable(nothing), args...; kwargs...)
+    end
```
asinghvi17, line 11: If we're doing this, then `lift` should always return an Observable, no?

jkrumbiegel:
Not sure if I like enabling non-observable arguments. In what use cases do you get either observable args or not? The attributes in recipes are all converted for example. I find an explicit conversion clearer, but I also understand the convenience aspect.

EdsterG:
> Not sure if I like enabling non-observable arguments.

IMO the behavior of `lift` needs to be changed to:
1) The proposed behavior in this PR
or
2) Error on non-observable x

Both options will reduce the likelihood of error, which is the biggest problem this PR is attempting to solve.

ryofurue:
I was caught by this issue and spent sometime to figure out why `@lift` did the wrong thing on a non-Observable object.  I did know what `map` is because it's a standard part of the Julia language but I didn't know anything about `lift`.

So, what about making it an error to use `lift` on non-Observables?
```
lift(f, x::Observable) = map(f,x)
lift(f, x) = error("lift is defined only for an Observable x")
```
?

PR #3916, opened 2024-05-30
support alpha argument in band()

aplavin:
# Description

The `alpha` argument was just silently ignored before this PR. This makes it actually work.

## Type of change

## Checklist


jkrumbiegel:
@SimonDanisch is mesh expected to support `alpha`?

SimonDanisch:
This is a CairoMakie issue, so I don't think this is the correct fix.
GLMakie works fine:
```julia
using GLMakie
xs = 1:0.2:10
ys_low = -0.2 .* sin.(xs) .- 0.25
ys_high = 0.2 .* sin.(xs) .+ 0.25

band(xs, ys_low, ys_high, alpha=0.5)
```
image
I'm guessing the problem is in https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/overrides.jl?

jkrumbiegel:
A right of course, it's just the override that ignores it.

SimonDanisch:
Sorry for the long wait, these overrides are a bit hacky and not that great to review for me, since it's not quite clear to me what the "correct" way of doing things is here.
I think we should just merge this, and clean up the overrides once we support having backends decide which plots are "atomic".
@jkrumbiegel does that sound good?

SimonDanisch:
Thank you!

PR #3922, opened 2024-06-02
Add code to fake interaction videos with blocks

jkrumbiegel:


jkrumbiegel:
This is mostly ready now, examples look pretty good I think. I don't really see a way around the increased rendering time, as we can't go parallel and CI machines are just not that quick. But it seems worth it to me to actually show the UI in action.

SimonDanisch:
I think the increase is fine, considering that it's still not the slowest CI ;)

PR #3940, opened 2024-06-06
Force Observable{Any} in `apply_expand_dimensions`

asinghvi17:
# Description

Fixes #3938 

This is a pretty small change, just specifies that we're creating `Observable{Any}` out of the values returned by `expand_dimensions`.

## Type of change

Delete options that do not apply:

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] Added unit tests for new algorithms, conversion methods, etc.

PR #3941, opened 2024-06-06
As/expand dims types

SimonDanisch:
#3940 to run benchmark!


SimonDanisch:
This should mainly show up in the `create` compile time benchmark, which seems to be fine :)

PR #3949, opened 2024-06-08
Ensure that the axis `type` kwarg is converted to a value when constructing a new axis

asinghvi17:
# Description

If plotting, e.g. 
```julia
heatmap(rand(10, 10); axis = (; type = LScene))
```
I ran into the error that the parameter `AxType` sent to `set_axis_attributes!` and `_block` was an Observable.  This PR fixes that by simply ensuring that whatever is passed to the functions is always a value, not an Observable.

## Type of change

## Checklist


src/figureplotting.jl
```diff
@@ -103,8 +103,8 @@ function create_axis_for_plot(figure::Figure, plot::AbstractPlot, attributes::Di
         return nothing
     end
     bbox = pop!(axis_kw, :bbox, nothing)
-    set_axis_attributes!(AxType, axis_kw, plot)
-    return _block(AxType, figure, [], axis_kw, bbox)
+    set_axis_attributes!(to_value(AxType), axis_kw, plot)
+    return _block(to_value(AxType), figure, [], axis_kw, bbox)
```
SimonDanisch, line 107: I think we only need this in `pop!(axis_kw)`

test/figures.jl
```diff
@@ -163,11 +163,17 @@ end
...
 
     f = Figure()
     @test_throws ArgumentError lines(f[1, 1], 1:10, axis = (aspect = DataAspect()))
     @test_throws ArgumentError lines(f[1, 1][2, 2], 1:10, axis = (aspect = DataAspect()))
+
+    # The axis type needs special treatment, since it's not meant to be changed later on
+    # and must be interpreted as a value on plot creation.
+    @test_nowarn lines(1:10; axis = (; type = LScene))
+    @test_nowarn lines(1:10; axis = (; type = Axis3))
+    @test_nowarn lines(1:10; axis = (; type = PolarAxis))
```
SimonDanisch, line 178: These tests don't test the type being an observable, do they?

asinghvi17: True, I'll make them all `axis = Attributes(...)`.

asinghvi17:
```suggestion
    # and must be interpreted as a value on plot creation.
    @test_nowarn lines(1:10; axis = Attributes(; type = LScene))
    @test_nowarn lines(1:10; axis = Attributes(; type = Axis3))
    @test_nowarn lines(1:10; axis = Attributes(; type = PolarAxis))

**Comment by asinghvi17** [comment]
I just got bit by this again with `heatmap(-180:10:180, -90:10:90, rand(18, 9); axis = (; type = GeoAxis))` in GeoMakie.

**Comment by SimonDanisch** [comment]
I cant reproduce neither ...
Are you sure that you dont have `GeoAxis = Observable(GeoAxis)` or something?
I also really dont know how this should happen unless you do `GeoAxis = Observable(GeoAxis)`

**Comment by asinghvi17** [comment]
Ah, maybe it's because some of the functions I'm using wrap all kwargs (including `axis`) in Attributes before passing to `plot`.  If you try to plot a DimensionalData DimArray 
```julia
using DimensionalData, Makie
heatmap(rand(X(1:10), Y(1:10); axis = (; type = LScene))
```

that should reproduce it.  Admittedly this is because DimensionalData recipes are a mess, but IMO we should still make this more robust to people passing pre-constructed Attributes in!  (Or just splatting attributes, which still leads to observable entries)

SimonDanisch:
I think Makie should throw an error if the Axis type is an Observable and this should be fixed in DimensionalData...
Why would we accept the axis type as an observable, if you can't change it?

SimonDanisch:
I'm closing this for the above reasons.
For a better error, I think we may as well create a new PR.

asinghvi17:
That sounds good, this is fixed in the latest version of DimensionalData btw

PR #3951, opened 2024-06-11
Add overview pages

jkrumbiegel:


docs/figure_block.jl
```diff
@@ -5,18 +5,129 @@ Documenter.Selectors.order(::Type{FigureBlocks})  = 8.0 # like @example
...
+        """)
+    end
 end
 
 
 
 function Documenter.Selectors.runner(::Type{FigureBlocks}, node, page, doc)
+    title = first(Iterators.filter(page.elements) do el
+        el isa Markdown.Header{1}
+    end).text[]
```
asinghvi17, line 129: This will only work for the first text element - I suggest mdflatten on the node.

docs/figure_block.jl
```diff
@@ -5,18 +5,129 @@ Documenter.Selectors.order(::Type{FigureBlocks})  = 8.0 # like @example
...
+        for page in pages
+            fileinfos = o.d[page]
+            pagename, _ = splitext(basename(page.path))
+            println(io, "### [$(page.title)](@id overview_$(pagename))")
+            println(io)
+            println(io, """<div :class="\$style.container">""")
+            for fileinfo in fileinfos
+                println(io, """
+                <a href="./$pagename.html#example-$(fileinfo.id)">
+                    <img src=\"./$(fileinfo.filename)\" />
```
asinghvi17, line 91: I'm not sure this will work with Vitepress because of rollup - may want to switch this to including the PNG base64 directly.

asinghvi17 (commented):
This looks pretty good to me!  Couple of comments though.

docs/figure_block.jl
jkrumbiegel: what do you mean by first text element, could there be more? I'm not sure why it's even a vector

asinghvi17: If there's any code that's one element, then text, then code again, etc.  For example,
```
# `qqnorm` and `qqplot`
```
expands to
```julia
julia> convert(MarkdownAST.Node, Markdown.parse("# `qqnorm` and `qqplot`"))
@ast MarkdownAST.Document() do
  MarkdownAST.Heading(1) do
    MarkdownAST.Code("qqnorm")
    MarkdownAST.Text(" and ")
    MarkdownAST.Code("qqplot")
  end
end
```

asinghvi17: and 
```julia-repl
julia> mda = convert(MarkdownAST.Node, Markdown.parse("# `qqnorm` and `qqplot`"))
@ast MarkdownAST.Document() do
  MarkdownAST.Heading(1) do
    MarkdownAST.Code("qqnorm")
    MarkdownAST.Text(" and ")
    MarkdownAST.Code("qqplot")
  end
end


julia> Documenter.MDFlatten.mdflatten(mda)
"qqnorm and qqplot\n\n"
```

jkrumbiegel:
@asinghvi17 one problem is that the link id change doesn't work correctly, the syntax leaks into the page 
image

asinghvi17:
I'll check out what's going on in DocumenterVitepress...

asinghvi17:
I think this is actually a Documenter issue - headers are processed before the expander pipeline is run, so these generated headers are not processed into Documenter.AnchoredHeader.

jkrumbiegel:
Ah probably it's because the heading stuff happens before I splice in the new markdown, but not after that. So the headers are not resolved correctly. But that's no problem then, as I was only adding the ids to avoid the headings conflicting with the canonical ones, but they are currently not considered at all I think.

asinghvi17:
Yep that seems fine.  As long as you aren't linking to them via `@ref` from within Documenter.

PR #3954, opened 2024-06-11
Improve accuracy of set framerate in GLMakie

ffreyer:
# Description

When targeting a framerate GLMakie uses `sleep()` which is pretty bad at its job:
Screenshot 2024-06-08 154030

And thus the frametimes are pretty off. For the default 30fps GLMakie targets we should have 1/30 = 0.033s delay between frames but I get around 0.045:
Screenshot 2024-06-10 220324

This pr adds some bookkeeping for the used time per frame. If one frame sleeps for too long the next frame will know about it and sleep for less time. This averages out the frametimes to match the requested framerate.

Screenshot 2024-06-11 214118

An alternative option would be have a `while (current_time < frametime) yield()` which would be better for stability but I assume would be worse for power usage.

## Type of change

## Checklist


jkrumbiegel:
I saw something on discourse where they were using a different sleep function for higher accuracy, is that an option?

ffreyer:
I came across https://discourse.julialang.org/t/accuracy-of-sleep/5546 / https://github.com/JuliaLang/julia/issues/12770 but on windows `Base.Libc.systemsleep()` behaves the same as `sleep()` for me.

src/utilities/timing.jl
```diff
@@ -0,0 +1,154 @@
...
+    min_sleep::Float64
+    budget::Float64
+    last_time::UInt64
+    
+    running::Bool
+    task::Union{Nothing, Task}
+
+    function BudgetedTimer(callback, delta_time::Float64, running::Bool, task::Union{Nothing, Task}, min_sleep = 0.015)
+        return new(callback, delta_time, min_sleep, 0.0, time_ns(), running, task)
+    end
```
ffreyer, line 16: I tried timing `sleep(0.0001)` here but it ended up being inconsistent. Most are 0.015s for me, but some are <0.005s and I assume async tasks and GC can also prolong them. So I just switched to a default value instead.

ffreyer: Since I wasn't sure anymore I did some testing on what effect a wrong `min_sleep` has. Doesn't seem like there are any significant changes:
Screenshot 2024-07-25 134212
Screenshot 2024-07-25 134657

Issue #1667, opened 2022-02-12
Is interactively zooming no longer available for Axis3?


Enterprixe:
Recently, I have been trying some Makie plots. In some videos from the recent past (summer 2021), I saw that there was the possibility to interactively zoom using GLMakie. However, as of right now, that does not seem to be a possibility anymore. Is this intended behaviour?

SimonDanisch:
Axis3 never had zoom, you need to use LScene for that...

Issue #1655, opened 2022-02-09
Add a :free viewmode for Axis3


cadoubs:
Axis3 `viewmode` options `:fitzoom`, `:fit` and `:stretch` block zooming and panning. 

It would be nice to have a `:free` or `:nofit` mode for inspecting 3d details and have access to new view angles/position like it is possible for 2D plots

j-fu:
May be this can be also combined with interactive zoom and pan functionality like in LScene.

bjarthur:
i believe `:free` has been implemented for Axis3 now and so this issue can be closed:

```
help?> Transcriptome4D.Axis3.viewmode
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future
  │  versions:
  │
  │    •  Transcriptome4D.Axis3

  Default value: :fitzoom

  The view mode affects the final projection of the axis by fitting the axis cuboid into the
  available space in different ways.

    •  :fit uses a fixed scaling such that a tight sphere around the cuboid touches the
       frame edge. This means that the scaling doesn't change when rotating the axis (the
       apparent size of the axis stays the same), but not all available space is used.
       The chosen aspect is maintained using this setting.
    •  :fitzoom uses a variable scaling such that the closest cuboid corner touches the
       frame edge. When rotating the axis, the apparent size of the axis changes which
       can result in a "pumping" visual effect. The chosen aspect is also maintained
       using this setting.
    •  :stretch pulls the cuboid corners to the frame edges such that the available space
       is filled completely. The chosen aspect is not maintained using this setting, so
       :stretch should not be used if a particular aspect is needed.
    •  :free behaves like :fit but changes some interactions. Zooming affects the whole
       axis rather than just the content. This allows you to zoom in on content without
       it getting clipped by the 3D bounding box of the Axis3. zoommode = :cursor is
       disabled. Translations can no also affect the axis as a whole with control + right
       drag.

  Example 1

  fig = Figure()
  
  for (i, viewmode) in enumerate([:fit, :fitzoom, :stretch])
      for (j, elevation) in enumerate([0.1, 0.2, 0.3] .* pi)
  
          Label(fig[i, 1:3, Top()], "viewmode = $(repr(viewmode))", font = :bold)
  
          # show the extent of each cell using a box
          Box(fig[i, j], strokewidth = 0, color = :gray95)
  
          ax = Axis3(fig[i, j]; viewmode, elevation, protrusions = 0, aspect = :equal)
[...]

PR #3958, opened 2024-06-13
Add clipping planes

### LINKED ISSUE #3908: Cutting planes (volumetric visualization)
### Feature description

For visualization of volumes that are complex or not very transparent, it would be nice to combine the advantages of `volume` and `volumeslices` through implementation of a "cutting plane": on one side of the plane, the volume is rendered, but on the other side it is as if that portion of the object had been cut away. One could then animate the position of the cutting plane to scan through the volume, always seeing the exposed surface. Think, for example, of progressively shaving away the surface of a [geode](https://assets3.fossilera.com/sp/615738/geodes/quartz-var-amethyst-calcite.jpg) and seeing each newly-exposed face.

Here's one example:
https://youtu.be/IEP0mL8P7GA?feature=shared&t=60

ffreyer:
# Description

Redo/Continuation of #2783.

Adds a `clip_planes` attribute to plots, allowing for up to 8 clipping planes to be defined (hard limit set in OpenGL). This will allow for `Axis3` to clip a world space rectangle, which is the main blocker for zooming (and camera translation?) in Axis3.

Example Rect-based clip:

```julia
f = Figure()
a = LScene(f[1, 1])

# inherit from scene by default for Axis3
a.scene.theme[:clip_planes][] = Makie.planes(Rect3f(Point3f(-0.75), Vec3f(1.5)))

# allow overwrite to e.g. visualize planes
linesegments!(
    a, Rect3f(Point3f(-0.75), Vec3f(1.5)), clip_planes = Plane3f[],
    fxaa = true, transparency = false, linewidth = 3)

p = mesh!(Sphere(Point3f(0,0,1), 1f0), transparency = false, color = :orange, backlight = 1.0)
r = range(-pi, pi, length = 101)
surface!(-pi..pi, -pi..pi, [sin(x + y) for x in r, y in r], transparency = false)
scatter!(r, r)
f
```

Screenshot 2024-06-13 150140

TODO:
- [x] basic GLMakie implementation
- [x] basic WGLMakie implementation
- [x] basic CairoMakie implementation
- [x] use parent plot instead of parent scene for automatic replacement for simpler attribute handling (i.e. avoid every plot needing to inherit explicitly)
- [x] include clipping planes in boundingbox calculation
- [x] fix volume
- [x] fix line clipping (inconsistent length)
- [x] fix line flickering in WGLMakie
- [x] maybe adjust voxels to clip full voxels
- [x] care for `space` (ignore anything but :data?)
- [x] CairoMakie overrides
- [x] fix boundingboxes clipping (see voxel example)
- [x] fix line color interpolation

Related #1655, #1667
Closes #3908

## Type of change

## Checklist


asinghvi17:
A CairoMakie implementation would have multiple approaches:
- For lines and linesegments, cut the line at the clipping plane.  This will require a 3D geometry framework.
- For scatter and meshscatter, don't render markers at the clipping plane.  Some markers will not be partially clipped but I think that's fine.
- For mesh, actually cut the mesh segments that intersect the plane.  Again requires a 3D geometry framework.  For now we can also simply not render those mesh elements which are completely outside the plane.
- The overridden recipes will also need handling, that would be a bit simpler though.

ffreyer:
> A CairoMakie implementation would have multiple approaches:
> 
>   * For lines and linesegments, cut the line at the clipping plane.  This will require a 3D geometry framework.
>   * For scatter and meshscatter, don't render markers at the clipping plane.  Some markers will not be partially clipped but I think that's fine.
>   * For mesh, actually cut the mesh segments that intersect the plane.  Again requires a 3D geometry framework.  For now we can also simply not render those mesh elements which are completely outside the plane.
>   * The overridden recipes will also need handling, that would be a bit simpler though.

I think of these changes as mostly for 3D, so I'm looking to keep CairoMakie rather simple. Lines probably need the same manual clipping code across CairoMakie, GLMakie and WGLMakie so I'll implement that in CairoMakie too. Maybe voxels too. But everything else will probably just be a form of "if vertex is clipped don't draw". E.g. all the mesh-like plots just drop a face when one vertex is clipped atm.

ffreyer:
I'm wondering if I should rename `clip_planes` to `clipping_planes`

jkrumbiegel:
sounds good, otherwise it's a bit like you want to clip planes

ffreyer:
Refimg tests I've added:
Screenshot 2024-06-18 195401
Screenshot 2024-06-18 195410
Screenshot 2024-06-18 195426
Screenshot 2024-06-18 195435

ffreyer:
[d1ca4bf](https://github.com/MakieOrg/Makie.jl/pull/3958/commits/d1ca4bf173951c967c42b369080001e289eab025) fixes an issue where lines just disappear in CairoMakie when one of their coordinates is far enough outside the screen limits. From Discord:
```julia
data = 1e7*[0, 0.5, 10];
fig = Figure(); 
ax = Axis(fig[1,1]); 
ylims!(ax, 0, 1000); 
lines!(ax, 0.0:0.5:1.0, 1e3*[0, 0.5, 10]); 
lines!(ax, 0.0:0.5:1.0, data); 
fig
```
| Before | After |
| --- | --- |
| Screenshot 2024-06-27 163708 | grafik |

The fix I'm doing here is just adding clip planes for the scene boundary (i.e. -1..1 in clip space). This results in a point at the scene edge being added and points outside being discard. Thus the orange line draws:


Adding that caused "Lines form outside" to fail in CairoMakie due to different colors. The reason for that was that color interpolation did not consider clipping. So even if, for example, the second point was moved due to clipping, it would still be assigned the same color, thus squishing the interpolation on a shorter segment. 
Clip planes make this easier to see. If we draw a line with color interpolation, then adding a clip plane should not affect the colors of the visible part. This is now the case:
| Before | After |
| --- | --- |
| Screenshot 2024-06-27 165229 | Screenshot 2024-06-27 165151 |

"Lines from outside" should be expected to fail as the refimg doesn't adjust color interpolation to clipping.

PR #3974, opened 2024-06-19
validate plot attributes later, for axis specific plot attributes

SimonDanisch:
Fix for https://github.com/MakieOrg/GeoMakie.jl/pull/245
Exakt place of validation is up for discussion, but I think it shouldn't hurt to not validate in the constructor?
The problem is, that axis creation needs a finished plot object, but that means we can't create the axis and dispatch on it before handling axis specific attributes:
```julia
using GeoMakie, GLMakie # you could use GLMakie too

lons = -180:180
lats = -90:90
field = [exp(cosd(l)) + 3(y / 90) for l in lons, y in lats]
sp = surface(lons, lats, field; axis=(; type=GeoAxis), source="+proj=longlat +datum=WGS84")
```
The source attribute is plot specific, and then creates a plot AND axis specific transform.
Open for discussion, but I'd not like to have this be something like `plot(...; (; user_attributes=(; source="...")))`, which is the only alternative I see right now.


asinghvi17:
Would this mean that the attribute validation only works on top-level plots, and not subplots made in recipes?

src/scenes.jl
```diff
@@ -454,7 +454,8 @@ function Base.push!(plot::Plot, subplot)
     push!(plot.plots, subplot)
```
asinghvi17, line 454: We could add the check here as well, to ensure full coverage and so that e.g. attributes passed through recipes are caught.

```suggestion
    MakieCore.validate_attribute_keys(Plot{Typ}, plot.kw)
    push!(plot.plots, subplot)
```

asinghvi17:
Thanks!  Could we get a version out with this?  I can do the changelog/bump PR when I get to my hotel but it would be great to release a GeoMakie that can be precompiled before JuliaCon :D

PR #3924, opened 2024-06-03
fix keyboard entry for WGLMakie Textbox

bjarthur:
# Description

fixes https://github.com/SimonDanisch/Bonito.jl/issues/154

users can now enter text into a Makie.Textbox from a web app created with Bonito.jl

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)

PR #3980, opened 2024-06-20
move throttle_function to Bonito

bjarthur:
Bonito's StylableSlider needs to be throttled.  see https://github.com/SimonDanisch/Bonito.jl/pull/233 (which needs to be merged before this one).  instead of duplicating `throttle_function` in Bonito, move it there, and have WGLMakie rely on that.

this is neither a bug fix nor new feature nor breaking change to Makie, but does improve user experience in Bonito while minimizing code duplication.

that being said and done, since Makie text boxes now work in Bonito (https://github.com/MakieOrg/Makie.jl/pull/3924, yay!!!), i have no need for Bonito sliders as my entire GUI can be in Makie now.  but other folks still might find this PR useful.


SimonDanisch:
thanks!

PR #3985, opened 2024-06-24
Update style.css

lazarusA:
this should give us extra space for the content section. As in https://luxdl.github.io/DocumenterVitepress.jl/dev/style_css


jkrumbiegel:
@lazarusA I thought this behavior was hardcoded in the theme?

lazarusA:
well, the default yes. Which uses the `page =[]` syntax. But, since we already have a `config.mts` file is really straight forward to change the layout.

ffreyer:
What's the state with this pr? (And #3987) Did we not want this? Did we forget about it?

lazarusA:
We don't wanna it anymore, is outdated, let's close this.

PR #3986, opened 2024-06-24
Don't assume poly always has two children

jkrumbiegel:
Otherwise the fallback logic doesn't work when `poly` gets an external override which plots a `poly` internally again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


asinghvi17:
Good point - should we also alter how we do the polygon drawing then, so that we only make the decision to go the vector or mesh route once we hit a `poly` plot with mesh and lines child plots?

asinghvi17:
It looks like this also forwards to the vector `draw_poly` when it is applicable for the child plot?  Just to confirm.

jkrumbiegel:
@asinghvi17 i didn't check but I think it just continues the normal way, so it should work

PR #3987, opened 2024-06-24
Update style.css

jkrumbiegel:
Copy of https://github.com/MakieOrg/Makie.jl/pull/3985 so we can see a build

### DIFF
--- a/docs/src/.vitepress/theme/style.css
+++ b/docs/src/.vitepress/theme/style.css
@@ -172,3 +172,37 @@ mjx-container > svg {
   --vp-c-sponsor: #91dd33;
   --vitest-c-sponsor-hover: #e51370;
 }
+
+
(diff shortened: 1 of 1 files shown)


docs/src/.vitepress/theme/style.css
```diff
@@ -171,4 +171,38 @@ mjx-container > svg {
...
+}
+
+
+.VPDoc.has-aside .content-container {
+  max-width: 100% !important;
+}
+.aside {
+  max-width: 200px !important;
+  padding-left: 0 !important;
+}
```
lazarusA, line 183:
```suggestion
```
this removes the hard coded size for the right panel, however, it goes back to the default size, which is quite big for my taste. On the other hand we might recover the proper alignment.

PR #4008, opened 2024-07-06
Remove randomisation of insertion order in calls to `triangulate` in `triplot`/`voronoiplot`/`triplot`

DanielVandH:
Hopefully fixes the issue observed in #4003 as I've discussed with @SimonDanisch. Locally, running the test over and over gives the same centroidal tessellation. 

The test was actually broken before DelaunayTriangulation@1.0.4 (i.e. it always gave a different image), not sure why it suddenly started breaking though. 

Do I need to update the changelog for this?


DanielVandH:
As discussed, the issues with the failing tests seems to be an issue of the results changing depending on the iteration order (which seems to have changed between release and 1.6) of the `Dict`s used within DelaunayTriangulation.jl. 

A better solution to these issues would be for me to redesign the tests that involve either `refine!` or `centroidal_smooth` (since these algorithms can fall into local minima that might depend on the iteration order) so that iteration order does not matter. I'll try and find the time to do that. I do think it would still be good to also disable randomisation as this PR implements

PR #4025, opened 2024-07-13
[WIP] Shadertoy and shader plot types

SimonDanisch:
adds:
```julia
shadertoy(rect2d, frag_shader)
```
To run shadertoy shaders like this: 
image
image

TODOs:
* map arbitrary uniforms to shader (using ShaderAbstractions)
* allow geometry and vertex shader (using a new plot type)
* clean up and docs


SimonDanisch:
I tried cleaning it up to merge it, but somehow run again into a very frustrating OpenGL bug, where the screen just stays blank without any error and I have no idea where to look for the problem...
I fixed that before, as can be seen by the screenshots, so I have no idea how it came back...Was frustrating enough to find it in the first place, and I dont really remember what it was in the end :( Something small and stupid like a uniform type mismatch or so...
Maybe it's platform dependent, I can try to see what my nvidia gpu says about this.

ffreyer:
Fixed the blank screen thing (caused by transparency = true which changes what the shader outputs are).

Btw, my thoughts with adding a "glplot" type were more manual than this. I wanted something that would allow you to set all shader types and have arbitrary vertex inputs and uniforms. Conceptually something like
```julia
glplot!(scene, 
    vertex_inputs...; # 1+ arrays mapping to buffers
    shader = [vertex_shader, fragment_shader], # 1+ vertex shader, 1+ fragment shader (0+ other types?)
    uniforms... # any number of uniforms as kwargs
)
```
that just does some boilerplate and integration stuff in the background. E.g. include "fragment_output" and handle outputs, conversions to gl types (maybe difficult for 1d textures?), mesh conversion paths, add and handle a few core attributes (visible, ssao, fxaa, transparency). (That doesn't mean I want this pr to change. Just wanted to say I had something else in mind)

GLMakie/src/drawing_primitives.jl
```diff
@@ -1087,7 +1087,7 @@ function draw_atomic(screen::Screen, scene::Scene, plot::Makie.ShaderToy)
                 view=templates)
         end
         get!(uniforms, :ssao, Observable(false))
-        get!(uniforms, :transparency, Observable(true))
+        get!(uniforms, :transparency, Observable(false))
```
SimonDanisch, line 1090: Lol

docs/src/reference/plots/shadertoy.md
```diff
@@ -0,0 +1,68 @@
+# spy
```
asinghvi17, line 1:
```suggestion
# shadertoy
```

PR #4026, opened 2024-07-13
consider model transformations in float32 conversions

### LINKED ISSUE #4022: Translated plots are bouncing around on Macs

```julia
using GLMakie

f, a, p = linesegments(rand(2));

translate!(p, 661000, 661000, 0)
# scale!(p, 2, 2, 1)
f # this is a problem

# investigation code
Makie.update_state_before_display!(f)
f
notify(p[1])
Makie.patch_model(a.scene.float32convert.scaling[], p.model[])


f2, a2, p2 = linesegments(rand(Point2{Float64}, 2) .+ Point2{Float64}(661000, 661000)) # works fine
```

ffreyer:
# Description

Fixes #4022

This pr changes how model matrices are handled with Float32Convert. It mainly boils down to changes in `patch_model()` and `apply_transform_and_f32_conversion()`. They both now have a couple of branches handling a couple of different situations:
1. `if is_float_safe(scale, trans) && is_identity_transform(float32convert)` is the case without Float32 precision issues. This leaves `model` as is and effectively only applies transform_func and casts to Float32 in `apply_transform_and_f32_conversion()`. 
(The logic of the branchings is that an inactive float32 conversion implies Float32 safe world space, and a Float32 safe model matrix then implies Float32 safe transformed space)
2. `elseif is_float_safe(scale, trans) && is_rot_free` is the case where transformed data is not safe, but model is and we can permute it with the Float32Convert due to it not having a rotation component. In this case the Float32Convert gets adjusted, model does not, and the adjusted Float32Convert gets applied in `apply_transform_and_f32_conversion()`.
3. `elseif is_rot_free` is the case where the model matrix is not safe and can be absorbed into the Float32Convert. That happens, changing the Float32Convert and setting the model matrix to identity. `apply_transform_and_f32_conversion()` then effectively applies the model matrix via Float32Convert without literally doing the matrix multiplication. (This is merged with the previous branch there)
4. `else` is the case where the model matrix contains rotation and the Float32Convert is active. As far as I can tell we cannot permute Float32Convert with model here, and thus `apply_transform_and_f32_conversion()` applies the model matrix on the CPU. `patch_model()` outputs an identity matrix in this case. (Operation order is `f32convert(apply_model(apply_transform(input)))`)

TODO:
- [x] check up on `text_quads()`
- [x] update WGLMakie
- [x] keep model intact for `transform_marker = true` if it's a save scale+translation matrix
- ~~figure out what to do with z translations for dim conversions (heatmap, image, surface?)~~ skipping this
- ~~fix GLMakie heatmaps with CPU rotation-full model application~~ skipping this
- ~~consider updating `mouseposition()`~~ should be fine?

## Type of change

## Checklist


ffreyer:
The approach I tried for the first commit is:
- no float32 convert -> leave things as they are
- float32 convert + model matrix without rotation -> move scaling and translation into float32 convert, replace model with identity
- float32 convert + model matrix with rotation -> apply model matrix on CPU and replace it with identity for the GPU side

I'm realizing now that this doesn't work with `transform_marker = true` in scatter (and text) since the model matrix may not be the real one anymore. So I guess we'll either need to adjust those shaders as well or come up with a different solution

src/camera/projection_math.jl
```diff
@@ -223,8 +223,88 @@ function transformationmatrix(translation, scale, rotation::Quaternion{T}) where
...
+        end
+
+        q1 = 0.5 * sx / q4
+        q2 = 0.5 * sy / q4
+        q3 = 0.5 * sz / q4
+        rot = Quaternion{T}(q1, q2, q3, q4)
+
+        return trans, scale, rot
+    end
+end
```
ffreyer, line 287: This isn't being used but I think it would be better to have it than not. It might be useful if we ever need to decompose the model matrix into its components. (Though it's somewhat dangerous because this only works with a specific order of matrices. Other orders need minor adjustments to work.)

src/float32-scaling.jl
```diff
@@ -68,11 +70,80 @@ function f32_convert_matrix(ls::LinearScaling, space::Symbol)
...
+
+        # works with rotation component as well
+        trans, scale = decompose_translation_scale_matrix(model)
+        is_rot_free = is_translation_scale_matrix(model)
+
+        if is_float_safe(scale, trans) && is_identity_transform(f32c)
+            # model should not have Float32 Problems and f32c can be skipped
+            # (model can have rotation here)
+            f32c_obs[] = f32c
+            model_obs[] = Mat4f(model)
```
ffreyer, line 120: Pulling model translation and scaling into the float32convert (and probably also just applying model on CPU) messes with render order if plots are at equal z. This should fix that and maybe also improve performance a bit

ffreyer:
@asinghvi17 The GLMakie side of this should be ready if you want to try it with Tyler. 

If no other issues come up with this I'll probably just update WGLMakie and add tests etc.

asinghvi17:
Tyler now works great, but there are issues when scrolling to the corners of the map, it almost feels like the scroll point is not respecting the mouse position.

ffreyer:
> Tyler now works pretty well, but there are issues when scrolling to the corners of the map, it almost feels like the scroll point is not respecting the mouse position.

`mouseposition()` is converting to world space before float32 conversions apply, i.e. it should match Axis ticks. `mouseposition_px` doesn't care. I think that should all be fine. If you have an example of it not working as you expect I can look into it.

SimonDanisch:
Could this be a regression on master? I also noticed, that Axis doesn't zoom onto the mouse position accurately anymore (did it ever?), without being on this branch...

ffreyer:
TODO:
Check if WGLMakie is doing conversion twice things relying on mesh rendering

PR #4034, opened 2024-07-15
Docs version picker

jkrumbiegel:


asinghvi17:
Documenter emits this file to the root of gh-pages but I haven't figured out how to reference it in the docs yet.

jkrumbiegel:
@asinghvi17 I've added those scripts here https://github.com/MakieOrg/Makie.jl/blob/b9bfa873c2ad540838e791429da7a86f67475291/docs/src/.vitepress/config.mts#L20-L21

The question was just how to turn the information from them into a dropdown menu

asinghvi17:
Thanks, we should probably upstream this to DocumenterVitepress as well once it's updated.

jkrumbiegel:
Yeah sure that would be great!

PR #4037, opened 2024-07-15
add setindex() for Attributes

aplavin:
# Description

It's often useful to create a copy of Attributes with some value changed. `Base.setindex` is the function that does exactly this.

Afaict (correct me if I'm wrong), the implementation is correct for this purpose: if the new value is observable it just gets assigned, otherwise the value is converted to an observable and assigned. I tried assigning non-observables directly, but this affects the original Attributes as well, not just the copy.

## Type of change

## Checklist

### DIFF
--- a/MakieCore/src/attributes.jl
+++ b/MakieCore/src/attributes.jl
@@ -94,6 +94,13 @@ function Base.setindex!(x::Attributes, value::NamedTuple, key::Symbol)
     return x[key] = Attributes(value)
 end
 
+function Base.setindex(x::Attributes, value::Observable, key::Symbol)
+    y = copy(x)
+    y[key] = value
+    return y
+end
+Base.setindex(x::Attributes, value, key::Symbol) = setindex(x, Observable(value), key)
+
 function Base.setproperty!(x::Union{Attributes, AbstractPlot}, key::Symbol, value)
     if hasfield(typeof(x), key)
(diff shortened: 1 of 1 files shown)


ffreyer:
Can you give some examples on what this would be useful for? I can only think of attribute handling within a recipe, but there I would rather have something more specialized.

SimonDanisch:
Why not `attributes = copy(attributes); attributes[:attribute] = ...` and friends?
I'm not sure `setindex` is such a great fit for something as large and heavy and mutable as `Attributes`.

ffreyer:
My idea of a function for this would be something like `applicable_attributes(PlotType, parent_attributes; overwrites_and_additions...)` which would
- create a new `Attributes` object (or maybe `default_theme(PlotType)`)
- copy attributes from parent_attributes if they are applicable to the given PlotType
- replace/add attributes passed as kwargs, probably also with validation

We could also have this check if all applicable attributes have been passed. That might be helpful in upping the quality of recipes by making them consciously decide between making child plot attributes reachable or not.

asinghvi17:
Don't we already have `Makie.shared_attributes`?

asinghvi17:
I had the wrong order, sorry about that:
```julia
julia> b = merge(Attributes(x=2), a)
Attributes with 1 entry:
  x => 2
```

PR #4040, opened 2024-07-17
make the backend choice more clear

SimonDanisch:


docs/src/index.md
```diff
@@ -91,8 +91,8 @@ set_theme!() # hide
 
 First, choose one (or more) of the four backend packages (more info under [What is a backend](@ref)):
 
-- [GLMakie](@ref) (OpenGL based, interactive)
-- [CairoMakie](@ref) (Cairo based, static vector graphics)
+- ** [GLMakie](@ref) (OpenGL based, interactive, most feature complete and prefered backend) **
```
asinghvi17, line 94:
```suggestion
- ** [GLMakie](@ref) (OpenGL based, interactive, most feature complete and preferred backend) **

**Comment by asinghvi17** [comment]
Looks good to me, aside from the typo :D

**Comment by jkrumbiegel** [comment]
I don't know, maybe it's pedantic but "most feature complete" doesn't make sense to me. Especially because it's totally unclear what "complete" means. And I'm also not sure if I would call GLMakie "preferred", so many people use cairomakie because they have no strong need for interactivity or 3D. Maybe it could be the "suggested" backend for people who want interactivity, and CairoMakie for those who want vector graphics.

**Comment by asinghvi17** [comment]
I agree with your point, that being said the distinction between "good at interactivity" and "good at vector graphics" may not be totally clear to a newcomer.  

Maybe we could instead recommend GLMakie for `True 3D graphics and interactivity` and CairoMakie for `2D static plots with vector graphics`?  Something like that might help people understand when to and when not to use the backends more immediately.

**Comment by ffreyer** [comment]
Plots.jl has a overview table and some short Pros & Cons lists to explain this kind of stuff:
https://docs.juliaplots.org/stable/backends/#At-a-glance

Maybe that would be better for differentiating backends? Or maybe a high level feature table (e.g. rating 2D, 3D, Interactivity, performance and maybe noting specialties like vector graphics, ray tracing, etc?)

**Comment by SimonDanisch** [comment]
Maybe we can have that too, but I think that quickly becomes overwhelming, so we should have a simple, quick to choose version in getting started.

File: docs/src/index.md [hunk 2]
```diff
@@ -1,12 +1,12 @@
-````@raw html
+![image](https://github.com/user-attachments/assets/2a2ffe40-ecc1-4531-94b3-12d66503586e)````@raw html
```
jkrumbiegel, line 1: what's this?

PR #4043, opened 2024-07-18
fix spy and FastPixel marker

SimonDanisch:
FastPixel didn't allow to change markersize and also ignored markerspace=:pixel/:data.
This fixes FastPixel, making makie much more usable for large point clouds.
Also, this makes `spy` usable, which hasn't gotten any love since it was written and was barely usable.


GLMakie/src/drawing_primitives.jl
```diff
@@ -423,8 +423,21 @@ function draw_atomic(screen::Screen, scene::Scene, @nospecialize(plot::Union{Sca
...
+            gl_attributes[:markerspace] = lift(plot.markerspace) do space
+                space == :pixel && return Int32(0)
+                space == :data && return Int32(1)
+                return error("Unsupported markerspace for FastPixel marker: $space")
+            end
+            camc = scene.camera_controls
+            if camc isa Camera3D
+                gl_attributes[:upvector] = lift(x-> Vec3f(normalize(x)), camc.upvector)
+            else
+                gl_attributes[:upvector] = Vec3f(0, 1, 0) # 2d camera has always y up
```
ffreyer, line 440: I think instead of this we should add upvector to `scene.camera`. That way custom camera controllers and Axis3 can still work with this without further backend changes.

SimonDanisch: Yeah I was also thinking about that.... Maybe it's time for that!

src/basic_recipes/spy.jl
```diff
@@ -12,24 +15,37 @@ spy(0..1, 0..1, x)
...
+    x = to_interval(xs, "x")
+    y = to_interval(ys, "y")
+    return (x, y, convert(SparseArrays.SparseMatrixCSC, matrix))
 end
 
+needs_tight_limits(::Spy) = true
+
 function plot!(p::Spy)
     rect = lift(p, p.x, p.y) do x, y
         xe = extrema(x)
```
ffreyer, line 51: Maybe better to use `minmax(endpoints(x)...)` to allow reversed Intervals since `extrema(1..0)` errors?

SimonDanisch: Hm yeah, but reversing still won't work since we won't actually  look up the points in reverse.

ffreyer: Oh right, this should just be left/rightendpoint. I'm not sure if rect allows negative widths, so maybe we should replace it with a (x, y, w, h) tuple?

Issue #607, opened 2020-05-20
Embedding Makie in a GTK app


LeMinaw:
Is it possible to embed Makie in a GTK application, for exemple using a `GtkGLArea` widget as OpenGL context for GLMakie ?

https://github.com/JuliaPlots/Makie.jl/issues/568 discuss this for the Cairo backend.

(btw, coming back to the Makie ecosystem after a year or so, and I'm amazed by its progress! Maybe you remember me, Simon, I'm the one who made the logo ;-) )

SimonDanisch:
Hey, welcome back! :)
Btw, I think I don't mention you for the logo, and when I noticed, I already forgot the exact name so I couldn't find you :D Feel free to create a PR to correct this!

Gtk integration is possible, but Gtk segfaulted on my, and there is no OpenGL support on OSX.
This may have changed, but that made it a nogo for me to invest time into this ;)

LeMinaw:
> Btw, I think I don't mention you for the logo, and when I noticed, I already forgot the exact name so I couldn't find you :D Feel free to create a PR to correct this!
No problem, no need to clutter docs or readme files with this imo 😉)

> Gtk integration is possible, but Gtk segfaulted on my, and there is no OpenGL support on OSX.
> This may have changed, but that made it a nogo for me to invest time into this ;)
The lack of OpenGL support in OSX is a shame, but I can personnaly can live without it.

Wich UI toolkit do your recommand for use with Makie ? Maybe Qt5 through QML.jl ?

SimonDanisch:
MakieLayout, I guess :D

LeMinaw:
MakieLayout is great, but I find it somewhat limited when apps start to be complex. For exemple, it lacks features like tabs or multiple dialog boxes.

Also, I don't really like Makie's font rendering, it looks pretty blurry:
![image](https://user-images.githubusercontent.com/3968360/82469574-7b3bb700-9ac4-11ea-85b7-4b0b270d01b6.png)
Left, random text in MS Word for comparsion.
Right, some text rendered in Makie through MakieLayout.

I guess this is because it uses OpenGL to draw characters ?

SimonDanisch:
Yeah, and there is still a bug in the opengl text rendering... In theory, it should look much nicer :D

SimonDanisch:
There is also: https://github.com/Gnimuc/CImGui.jl which works much better with opengl!

LeMinaw:
[...]

PR #4073, opened 2024-07-30
Improve support for embedding GLMakie

JamesWrigley:
# Description

I don't think this fixes any specific issue, but it's related to #607.

TL;DR it refactors GLMakie to be aware of whether or not it owns `Screen.gl_screen`, which makes embedding GLMakie a lot easier since it won't attempt to do GLFW-specific things anymore.

The GLMakie API can now basically be used as-is without having to copy-and-paste parts to avoid doing incompatible things. I tested it with CImGui, but it should also be useful for other packages like QML.jl and GtkMakie.jl (cc @barche and @jwahlstrand). I don't think this is breaking in general since `Screen` isn't public, but it will be breaking for packages like GtkMakie.jl and CImGui.jl because the `Screen` type changed.

Here's how it simplified the CImGui integration: https://github.com/Gnimuc/CImGui.jl/commit/53e6117

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -158,6 +158,7 @@ $(Base.doc(MakieScreen))
 """
 mutable struct Screen{GLWindow} <: MakieScreen
     glscreen::GLWindow
+    owns_glscreen::Bool
     shader_cache::GLAbstraction.ShaderCache
     framebuffer::GLFramebuffer
     config::Union{Nothing, ScreenConfig}
@@ -185,6 +186,7 @@ mutable struct Screen{GLWindow} <: MakieScreen
 
     function Screen(
             glscreen::GLWindow,
+            owns_glscreen::Bool,
             shader_cache::GLAbstraction.ShaderCache,
             framebuffer::GLFramebuffer,
             config::Union{Nothing, ScreenConfig},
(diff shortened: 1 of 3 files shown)


jwahlstrand:
I would happily adapt Gtk4Makie.jl to use this if it's merged. Thanks!

GLMakie/src/screen.jl
```diff
@@ -388,19 +405,28 @@ end
...
+    screen = screen_from_pool(config.debugging; window)
     apply_config!(screen, config; start_renderloop=start_renderloop)
     if !isnothing(resolution)
         resize!(screen, resolution...)
     end
     return screen
 end
 
-set_screen_visibility!(screen::Screen, visible::Bool) = set_screen_visibility!(screen.glscreen, visible)
+function set_screen_visibility!(screen::Screen, visible::Bool)
```
SimonDanisch, line 421: Should these error instead of returning?
I Think for Makie to work correctly, these will need to be overloaded for `Screen{T}`, so they should not end up here for screens not owned by makie, no?
Same for the other functions that noop, when not owning!

JamesWrigley: Hmm, I originally thought that it flat-out shouldn't be allowed for Makie to touch the window but I guess that should be left up to the integration. Changed it to an error in e2edca596.

SimonDanisch:
Thank you!

PR #4087, opened 2024-08-05
Data inspector fixes (mostly for bar plots)

EdsterG:
# Description

- Fixes tooltip location for dodged, stacked, and horizontal barplots. Before this PR it would snap to the data position, after this PR it snaps to the top center of the visible bar.
- Fixes wireframe rendering. Before this PR wireframe would overlap the tooltip, after this PR wireframe is below the tooltip.
- ~Fixes tooltip flickering. Before this PR tooltip would flicker on mouseover, after this PR tooltip doesn't disappear on mouseover.~

## Type of change

## Checklist


src/interaction/inspector.jl
```diff
@@ -841,6 +851,10 @@ function show_data(inspector::DataInspector, plot::BarPlot, idx)
         a.indicator_visible[] = true
     end
 
+    pos = plot[1][][idx]
+    if plot.direction[] === :x
+        pos = reverse(pos)
+    end
```
EdsterG, line 857: I'm not sure if the data should be flipped for user defined functions when the direction is horizontal. I assume direction is expected to be a visual change only. In that case, maybe the code should be:
```
    pos = plot[1][][idx]
    if to_value(get(plot, :inspector_label, automatic)) == automatic
        if plot.direction[] === :x
            pos = reverse(pos)
        end
        tt.text[] = position2string(pos)
    else
        tt.text[] = plot[:inspector_label][](plot, idx, pos)
    end
```

One can also make a case for never reversing the data, even in the automatic case. Curious to hear what the devs think.

src/interaction/inspector.jl
```diff
@@ -830,7 +839,8 @@ function show_data(inspector::DataInspector, plot::BarPlot, idx)
                 linewidth = a.indicator_linewidth, linestyle = a.indicator_linestyle,
                 visible = a.indicator_visible, inspectable = false
             )
-            translate!(p, Vec3f(0, 0, a.depth[]))
+            wf_depth = a.depth[] * 0.999 # put the wireframe below the tooltip
+            translate!(p, Vec3f(0, 0, wf_depth))
```
EdsterG, line 843: This feels like a bit of hack, but it works. Is there a different standard for ensuring z ordering?

src/interaction/inspector.jl
```diff
@@ -310,6 +310,10 @@ function on_hover(inspector)
                     should_clear = false
                     break
                 end
+            elseif (plotfunc(plt) === text && plt.parent.parent == inspector.plot) || (plotfunc(plt) === mesh && plt.parent == inspector.plot)
+                # prevent tooltip flickering when mouse hovers over the current inspectors tooltip
+                should_clear = false
+                break
```
EdsterG, line 316: Would like some feedback from the devs about this case. It seems to work in practice but blindly calling `plt.parent.parent` seems a bit fragile. Is there a better way to determine if the mouse is hovering over a tooltip bubble?

src/interaction/inspector.jl
EdsterG: Even with this depth fix, the wire mesh sometimes bleeds into the triangle part of the tooltip. Only when it's drawn to the right, but not when it's drawn to the left, see:
Screenshot 2024-08-05 at 11 51 52 AM
Screenshot 2024-08-05 at 11 52 05 AM

ffreyer: Actually I think this should be removed and replaced by a depth_shift like a couple other indicators have. That way it's independent on camera details.

src/interaction/inspector.jl
ffreyer: Yea it would be better to check this we other way around. i.e. compare with the child plots in `inspector.plot.plots`. That way you have a consistent object at least.

Also can you describe a bit more what kind of flickering this is supposed to prevent? Maybe with some example code/video?

src/interaction/inspector.jl
ffreyer: I generally thought of those positions as positions in the actual Axis/world. So if the bars end up at `reverse(pos)` with `direction[] === :x` I'd swap it for both.

src/interaction/inspector.jl
EdsterG: Hmm, I'm not sure what you have in mind. I've tried a few variations with surprising results.

This snippet solves the problem, but completely ignores the specified `a.depth` attribute:
```
            p = wireframe!(
                scene, bbox, model = model, color = a.indicator_color,
                linewidth = a.indicator_linewidth, linestyle = a.indicator_linestyle,
                visible = a.indicator_visible, inspectable = false
            )
            push!(inspector.temp_plots, p)
```

This snippet uses the depth attribute and a depth shift value but also doesn't solve the problem. Probably because the translate call overwrites the depth shift value?
```
            p = wireframe!(
                scene, bbox, model = model, color = a.indicator_color,
                linewidth = a.indicator_linewidth, linestyle = a.indicator_linestyle,
                visible = a.indicator_visible, inspectable = false,
                depth_shift = -1f-3,
            )
            translate!(p, Vec3f(0, 0, a.depth[]))
            push!(inspector.temp_plots, p)
```

src/interaction/inspector.jl
EdsterG: Ok, the PR currently swaps it for both so I'll leave it as is.
```
    pos = plot[1][][idx]
    if plot.direction[] === :x
        pos = reverse(pos)
    end
    if to_value(get(plot, :inspector_label, automatic)) == automatic
        tt.text[] = position2string(pos)
    else
        tt.text[] = plot[:inspector_label][](plot, idx, pos)
    end
```

EdsterG: > I generally thought of those positions as positions in the actual Axis/world.

Actually, does that mean that `pos` should be the tooltip/bar position instead of the data point? Right now `pos` is the defined as the data point. The data point is different from the bar position in the stacked/dodged case.

src/interaction/inspector.jl
EdsterG: Here's flickering behavior I'm referring to:

https://github.com/user-attachments/assets/9184a2f9-e408-4d97-abfa-de16ab96a6a8

EdsterG: Thanks for the suggestion, I flipped it to compare against child plots.

src/interaction/inspector.jl
ffreyer: Just doing
```
p = wireframe!(scene, bbox, ..., depth_shift = -1f-3)
```
with no translate should place the wireframe in front of bbox. In fact you shouldn't even need to set depth_shift since wireframe has it at <0 by default. 
depth_shift acts after model (i.e. also translate!()) and camera transformation when everything is normalized to a -1..1 depth range. It's meant to resolve issues where two things are plotted at the same depth, but one should be in front.

The reason why triangle doesn't behave right is probably the missing `o[3]` here: 
https://github.com/MakieOrg/Makie.jl/blob/3977fac9184237aa21bfba953ca2a51094d4dbec/src/basic_recipes/tooltip.jl#L179
I thought I fixed that already but I guess not.

src/interaction/inspector.jl
ffreyer: I need to correct myself a bit. At some point we changed this to be in the plots input space rather than world space (after model, transform_func) because that seems to be more useful with things like log plots.
I'm not sure what exactly the position should be. For most plots it's the position the tooltip points at and that's the only thing that makes sense. Here it's less clear. I'd lean towards it being that but I think updating the default label to work with stacked barplots is more important here.

src/interaction/inspector.jl
EdsterG: Thank you for the suggestions. I just pushed both changes.

src/interaction/inspector.jl
EdsterG: I think I understand. To confirm, the code that fixes label **location** is actually the code above:
```
    bbox = plot.plots[1][1][][idx]
    tt_frac = if plot.direction[] === :y
        Point2f(0.5, 1.0)
    else
        Point2f(1.0, 0.5)
    end
    tt_pos = fractionpoint(bbox, tt_frac)
    proj_pos = shift_project(scene, apply_transform_and_model(plot, to_ndim(Point3f, tt_pos, 0)))
```

The code here just specifies the label **data**:
```
    pos = plot[1][][idx]
    if plot.direction[] === :x
        pos = reverse(pos)
    end
    if to_value(get(plot, :inspector_label, automatic)) == automatic
        tt.text[] = position2string(pos)
    else
        tt.text[] = plot[:inspector_label][](plot, idx, pos)
    end
```

> At some point we changed this to be in the plots input space rather than world space (after model, transform_func) because that seems to be more useful with things like log plots.

Good point about log plots. For label data it's more useful to have access to the input space data.

EdsterG:
Since there are quite a few discussions happening in this PR, I think it might be a good idea to break out some of the obvious fixes into separate PR's to prevent blocking on unrelated change.
1) tooltip rendering fix
2) tooltip location fix
3) tooltip flickering fix

If you think this is a good idea, I can create PR's for the first 2.

src/interaction/inspector.jl
EdsterG: It looks like `pick_sorted` doesn't return anything that's occluded, which is desirable in general. However, in this case it would be helpful if `pick_sorted` was able to return plots that are occluded by the tooltip. Then it would be easy to prevent flickering and still update the tooltip based on what's occluded by it.

ffreyer: Yea I think I went through these considerations/problems before and settled on giving picking some range and hoping for the best.

Picking occluded things is in some sense impossible with the current implementation. Picking works by "drawing" a plot id + vertex index to a buffer while rendering and reading that when a picking function is called. That means you only have one value per pixel, one thing you can pick.

To address the flickering issue on a more fundamental level I see two options - write a ray based picking system or allow plots to opt out of picking during rendering.

For ray based picking you'd essentially want
```julia
for scene in all_scenes
    if !is_mouseinside(scene)
        continue
    ray = view_ray(scene) # depends on camera matrices
    for plot in scene.plots
        if collides_with_bbox(plot, ray) # because full intersection/collision checks can be very expensive
            idx, pos = intersection(plot, ray) # ~ position_on_plot()
            if # output marks valid intersection
                # return/collect collision data
```
A good chunk of that exists in ray_casting.jl already. However bounding boxes would probably become a bit of a pain point. You'd want to lazily cache them so that they one get recalculated when they could have changed and are needed.

The other option should be much less work on a surface level - add a generic attribute and effectively add `if (pickable)` to a few shaders in GLMakie and WGLMakie. On a detail level it gets more complicated though, because drawing order matters. If the occluded plot is drawn first this should work, otherwise not. Messing with the drawing order in GLMakie is fairly likely to produce some artifacts. Though it might also be fine if we just set `overdraw = true` for tooltips. 

Either way, these things shouldn't happen here. Maybe instead of locking the tooltip when its hovered it's better to increase the default picking `range` so that things outside the tooltip can be detected. That is more or less the original bandaid fix for this anyway.

EdsterG: Those are all good points. I'll remove the tooltip flicking change from this PR and defer the proper solution to a different time.

ffreyer:
Ok I've tried using this pr now and it's definitely an improvement. I'd be happy to merge this as is, but I wonder now if we should change the placement style of the tooltip. Some other plots like mesh have the tooltip follow the mouse position, which would effectively fix the flickering issue and maybe also look better. Do you think that would be a good change?

src/interaction/inspector.jl
```diff
@@ -815,13 +815,18 @@ function show_data(inspector::DataInspector, plot::BarPlot, idx)
...
-    pos = plot[1][][idx]
-    proj_pos = shift_project(scene, apply_transform_and_model(plot, to_ndim(Point3f, pos, 0)))
+    bbox = plot.plots[1][1][][idx]
+    tt_frac = if plot.direction[] === :y
+        Point2f(0.5, 1.0)
+    else
+        Point2f(1.0, 0.5)
+    end
+    tt_pos = fractionpoint(bbox, tt_frac)
+    proj_pos = shift_project(scene, apply_transform_and_model(plot, to_ndim(Point3f, tt_pos, 0)))
```
ffreyer, line 825:
```suggestion
    proj_pos = Point2f(mouseposition_px(inspector.root))
```
to switch to a tooltip anchored to the mouse position

EdsterG:
I think in general it would work well for bar plots, however there seems to be something wrong with the projection. Where the selected bar is not always the bar right under the mouse, which can make the tooltip a little confusing.

If you have a lead on the issue where the selected bar is not the one immediately under the mouse, I think your suggestion would work very well when it's fixed.

https://github.com/user-attachments/assets/6235e9df-b365-46f6-b636-2d9628f404c5

ffreyer:
That looks like the picking index is off by one for mesh in WGLMakie. It's happening regardless of the change I proposed at least. If you're otherwise happy with mouse-anchored tooltips I would say we add that change here and merge this.

EdsterG:
Sounds good, pushed the mouse-anchored tooltip. Thanks for all of your help.

ffreyer (approved):

Issue #3977, opened 2024-06-19
`select_point` is broken 


Datseris:
The interactive function `select_point` is broken and I don't know how to fix it... :( 

It doesn't really select a point, but it makes the rectangle for zooming to be visible and sometimes it also zooms.

MWE:

```julia
using GLMakie
fig, ax = scatter(rand(100, 2))
spoint = select_point(ax.scene; marker = :circle)
on(spoint) do pos
    print(pos)
end
display(fig)
```

```
  [ee78f7c6] Makie v0.21.3
```

asinghvi17:
It's working fine for me on mac, but when I drag the mouse even a little then it does activate the rectangle zoom.  Could that be the issue you're having?


https://github.com/MakieOrg/Makie.jl/assets/32143268/454754e6-2847-4568-950b-8cb87d902ff4

The appropriate points are printed in the REPL as well.

Datseris:
yeah this is the issue I was having. But for me it _also_ zooms in the rectangle if I drag the mouse while clicking. For you it doesnt?

JamesWrigley:
On latest Makie it seems to be completely broken:
```julia
julia> let
           f, ax, _ = scatter(rand(100))
           display(f)
           
           select_point(ax.scene)
       end
ERROR: Circle{Float32}(Float32[0.0, 0.0], 1.0f0) is not a valid scatter marker shape.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] marker_to_sdf_shape(x::Circle{Float32})
    @ Makie ~/.julia/packages/Makie/rEu75/src/utilities/texture_atlas.jl:419
  [3] (::GLMakie.var"#222#229")(m::Circle{Float32})
    @ GLMakie ~/.julia/packages/GLMakie/z5T26/src/drawing_primitives.jl:401
  [4] #map#218
    @ ~/.julia/packages/Makie/rEu75/src/scenes.jl:167 [inlined]
  [5] map(::GLMakie.var"#222#229", ::Union{Scene, Plot}, ::Observable{Circle{Float32}})
    @ Makie ~/.julia/packages/Makie/rEu75/src/scenes.jl:164
  [6] (::GLMakie.var"#219#226"{GLMakie.Screen{GLFW.Window}, Scene, Scatter{Tuple{Vector{Point{2, Float32}}}}})(gl_attributes::Dict{Symbol, Any})
[...]

PR #4090, opened 2024-08-06
Implement mouseposition(::Axis) to account for axis scales

JamesWrigley:
# Description

Previously `mouseposition()` didn't take the axis scaling into account and would return positions in scaled coordinates (e.g. with `log10` scaling it would return `log10.((x, y))`). Now it inverts the scaling to give the true data coordinates.

Also removed the mention of `hovered_scene()` since that's not implemented yet. I tried to write a test for this but couldn't figure out how, I expected this would show `(50, 0.5)`:
```julia
@testset "Mouse position" begin
    f = Figure(size=(100, 100))
    ax = Axis(f[1, 1])
    lines!(ax, randn(100))
    e = events(f)

    e.mouseposition[] = (50, 50)
    @show mouseposition(ax)
end
```

But instead it showed `mouseposition(ax) = [2.093023224625478, 2.444444408019385]`, so I must be doing something wrong. I did it test it with CImGui.jl though and it does work.

## Type of change

## Checklist


ffreyer:
`events(anything).mouseposition` is in screen/pixel units relative to the root window/figure.

To me the old behaviour is what I would expect, so I wouldn't consider this a bug fix. With `mouseposition(x) = mouseposition(get_scene(x))` already working before this, I would consider this pr breaking and I wouldn't be surprised if it literally breaks something in Makie. We would have to check if there are any calls to `mouseposition(::Axis)` before merging this and most likely replace them with `mouseposition(get_scene(...))`.

JamesWrigley:
Isn't `events(anything).mouseposition` different? According to the [docs](https://docs.makie.org/stable/explanations/events#Point-Picking):

> To correctly place a new marker we will also need to get the mouseposition in axis units. Makie provides a function that does just that: mouseposition([scene = hovered_scene()]). There is also a convenience function for the pixel space mouseposition relative to a specific scene mouseposition_px([scene = hovered_scene()]). Both of these will usually be different from events.mouseposition which is always in pixel units and always based on the full window.


The `mouseposition(scene)`docstring says it will return a position in data coordinates, but if an axis scaling is applied it definitely doesn't do that:
log-tooltip

Code:
```julia-repl
julia> let
           f, ax, _ = lines(rand(100))
           ax.xscale = log
           text_pos = Observable(Point2f(1))
           x = tooltip!(text_pos, "Foo")
           on(events(ax).mouseposition) do pos
               text_pos[] = mouseposition(ax)
               x.text = "events(ax).mouseposition: $(pos)\nmouseposition(ax): $(text_pos[])"
           end
           
           f
       end
```

How else would I get the mouse position in data coordinates?

> To me the old behaviour is what I would expect, so I wouldn't consider this a bug fix. With `mouseposition(x) = mouseposition(get_scene(x))` already working before this, I would consider this pr breaking and I wouldn't be surprised if it literally breaks something in Makie. We would have to check if there are any calls to `mouseposition(::Axis)` before merging this and most likely replace them with `mouseposition(get_scene(...))`.

I guess in a sense it is breaking, but to me this is a correctness bug so it was doing the wrong thing before anyway.

JamesWrigley:
I had a look at the other potential uses of `mouseposition(::Axis)`, and AFAICT they're also broken. The methods in Makie.jl that would be affected are:
- `select_rectangle()`/`select_line()`/`select_point()`; e.g. this is how `select_line()` currently behaves:
  select-line-log

  `select_rectangle()` has the same problem though it's currently slightly broken (#4099), and `select_point()` is quite broken (#3977) so I couldn't test it.
- `_addmousevents!()`, this ones quite internal so I'm not sure how to test it properly.

asinghvi17:
There are also situations where people encode noninvertible transforms (or at least transforms for which `Makie.inverse_transform` is not defined), so we should make sure that all the axis interactions at least work with this...

src/interaction/interactive_api.jl
```diff
@@ -177,15 +177,22 @@ function screen_relative(scene::Scene, mpos)
...
 """
 mouseposition(x) = mouseposition(get_scene(x))
 
+function mouseposition(ax::Axis)
+    pos = mouseposition(get_scene(ax))
+
+    # `pos` has the axis scaling already applied to it, so to get the true data
+    # coordinates we have to invert the scaling.
+    return Vec2{Float64}(inverse_transform(ax.xscale[])(pos[1]),
+                         inverse_transform(ax.yscale[])(pos[2]))
```
asinghvi17, line 193: This should probably be:

```julia
return Vec2{Float64}(Makie.apply_transform(Makie.inverse_transform(ax.scene.transformation.transform_func[]), pos))
```

but you'd have to test that.

src/interaction/interactive_api.jl
```diff
@@ -177,15 +177,22 @@ function screen_relative(scene::Scene, mpos)
 end
 
 """
-    mouseposition(scene = hovered_scene())
+    mouseposition(ax/scene)
```
asinghvi17, line 180:
```suggestion
    mouseposition(scene::Scene)
    mouseposition(ax::Axis)
```
maybe?  Not necessary but perhaps clearer.

Scenes can also be arbitrarily transformed so we should probably implement these changes there as well!

src/interaction/interactive_api.jl
ffreyer: Yea there is also the question of "Why just Axis?". Transform functions can be part of any Scene and technically also be different per plot. And if it is per plot it becomes unclear to me what that space even refers to on a Scene global thing like mouseposition()

JamesWrigley:
> There are also situations where people encode noninvertible transforms (or at least transforms for which `Makie.inverse_transform` is not defined), so we should make sure that all the axis interactions at least work with this...

According to the docs that isn't allowed? https://docs.makie.org/stable/reference/blocks/axis#xscale
> To use a custom function, you have to define appropriate methods for Makie.inverse_transform, Makie.defaultlimits and Makie.defined_interval.

I don't know who's right here :sweat_smile:

ffreyer:
Axis relies on `inverse_transform` as well so if that's not defined Axis breaks.

But more generally we probably don't want `mouseposition(::Axis)` to be different from `mouseposition(::SomeOtherAxisLikeBlock)` or `mouseposition(::Scene)` and those may use non-invertible transforms. Maybe GeoMakie has some of those?

src/interaction/interactive_api.jl
JamesWrigley: Yep that implementation works, changed in 4abb55536.

> Yea there is also the question of "Why just Axis?". Transform functions can be part of any Scene and technically also be different per plot. And if it is per plot it becomes unclear to me what that space even refers to on a Scene global thing like mouseposition()

I don't think it has to be just Axis, but that's the one that I as a user care most about it.

src/interaction/interactive_api.jl
JamesWrigley: Fixed the docstring in 4abb55536, I'll work on adding the transforms for scenes. Do you have any pointers on how I could test that? I haven't had much luck with that so far.

JamesWrigley:
> I also wanted to mention that "data" space is not very clearly defined atm. Sometimes it's the plot input space, sometimes it's after transform func + model. That's something we have to separate at some point.

Hmm, do you mean in Makie specifically or in general? My impression is that in general it's fairly well defined, for example in matplotlib: https://matplotlib.org/stable/users/explain/artists/transforms_tutorial.html#data-coordinates

src/interaction/interactive_api.jl
JamesWrigley: So if we were to apply this to all scenes then we wouldn't need to specialize on `Axis` and this would work:
```julia
mouseposition(x) = mouseposition(get_scene(x))

function mouseposition(scene::Scene = hovered_scene())
    pos = to_world(scene, mouseposition_px(scene))

    inv_transform = Makie.inverse_transform(scene.transformation.transform_func[])
    return Vec2{Float64}(Makie.apply_transform(inv_transform, pos))
end
```

That code gives the correctly transformed numbers, but I get some odd results:
makie-scene-transform

I'm guessing there's code elsewhere that's assuming `mouseposition(ax)` won't apply the scene transform.

asinghvi17: Yep that's probably in the `select_[point/rectangle/line]` routines.  They might have worked in "camera" space (transformed space, so after `apply_transform` and `transformationmatrix`).

JamesWrigley: I think those should actually be doing the right thing now because the line is being drawn correctly, but luckily the only other place in Makie.jl where `mouseposition()` is used is in the mouse state engine which seems to expect the untransformed position. Fixed that in 6622ec4cc but I don't know if the API is very good. Maybe it'd be better to have a separate function?

ffreyer:
In Makie

JamesWrigley:
After thinking about it a bit more, what if we still require custom transforms to always implement `inverse_transform()` but suggest that it can be set to `identity` if the transform is non-invertible? I think that could be breaking, but it seems safer than having a default `inverse_transform()` and less error-prone than special-casing all the code to handle cases where `inverse_transform()` isn't defined.

JamesWrigley:
I changed the name of the argument to `invert_transform` in ca919dae0 and documented that `mouseposition()` can return the position in either data coordinates or scene coordinates.

JamesWrigley:
(gentle bump)

JamesWrigley:
(bump)

ffreyer:
I generally like this change now, but I would still consider it breaking if it changes the default behavior because that has been around for years. I do think that it should change the default behavior though. I don't know when the next breaking version comes around, but I'd imagine it's not that far in the future with the GeometryBasics changes we have been working on.

In the future I'd say we want `mouseposition(scenelike, space)` but that requires us to split up the current ":data" space into data/input space, transformed space and world space. I'm not sure if it's a good idea to future proof this with `space = :data` and `space = :world` as kwargs. We should probably split up `space` first...

Something I do think we should change, is that this should consider the scenes model transformation, i.e. `scene.transformation.model[]` as well. The transformation pipeline goes `input coordinate > apply transform_func > apply model > apply projectionview > apply viewport -> pixel coordinate` and currently this is just skipping the model step. Usually that doesn't change anything because scenes usually have `model = I`, but for correctness we should consider it. The transformation here would be
```julia
p = ( inv(scene.transformation.model[]) * to_ndim(Point4d, to_ndim(Point3d, p, 0), 1) )[Vec(1,2)]
```

We also need to still check if this causes any issues elsewhere. For example, it would change `MouseEvent.data` in 
https://github.com/MakieOrg/Makie.jl/blob/65a718195a30fc4f3775f5457357e1ba35248745/src/makielayout/mousestatemachine.jl#L212
which is used for most interactive things in makielayout. E.g. 

> makie-scene-transform

is a related to that. It assumes transforms to not be included yet and includes them here: 
https://github.com/MakieOrg/Makie.jl/blob/65a718195a30fc4f3775f5457357e1ba35248745/src/makielayout/interactions.jl#L161
There might be more cases like that in Block, in the camera folder and in the interaction folder. 

@jkrumbiegel What are your thoughts on having the "data" mouse position in `MouseEvent` include inverse transformations?

PR #4092, opened 2024-08-06
Treat pseudolog10 and Symlog10 like log10

JamesWrigley:
# Description

The effect is that the tick labels get pretty-printed like with log10. This required a new type union for the regular log functions because they have different limits from pseudolog10/Symlog10.

## Type of change

## Checklist


JamesWrigley:
Looking at the CI history, I think the WGLMakie tests were broken from https://github.com/MakieOrg/Makie.jl/pull/3980.

SimonDanisch:
>Looking at the CI history, I think the WGLMakie tests were broken from https://github.com/MakieOrg/Makie.jl/pull/3980.

I updated test images for #4043, you were unlucky to run the ci just after the images where updated and just before rebasing against master ;)

JamesWrigley:
Oh dear, I'll rebase again :sweat_smile:

JamesWrigley:
(from my side this is ready to go BTW)

SimonDanisch:
@jkrumbiegel, does that look correct?

jkrumbiegel:
This needs tests / reference tests, ideally exercising these changes for multiple different scenarios where you can use these nonstandard log transforms. Not sure what happens when the functions expecting log ticks encounter 0 or negative values, for example.

JamesWrigley:
I added a reference test in 774187496, this is what GLMakie gives me when I run it:
Non-standard log tick formatting

I also added another fix in a2c03f772 to support log formatting ticks are explicitly given, with a reference test:
Log scale with manual ticks

JamesWrigley:
With luck the unit tests should be fixed in 9976f6574 :crossed_fingers: They were actually testing that the log ticks *weren't* formatted with a log formatting.

JamesWrigley:
(bump)

JamesWrigley:
(bump)

ReferenceTests/src/tests/figures_and_makielayout.jl
```diff
@@ -133,6 +133,37 @@ end
...
+    a8, _ = lines(f[4, 2], xs2, xs2; axis=(; title="Symlog(1)", xscale=Makie.Symlog10(1)))
+
+    st = Stepper(f)
+    Makie.step!(st) # generated ticks
+
+    a1.xticks[] = [10, 100]
+    a2.xticks[] = 2 .^ [4, 6]
+    a3.xticks[] = exp.([3, 4])
+    a4.xticks[] = [0, 60, 80, 90, 100]
+    a5.xticks[] = [-10, 0, 10]
```
ffreyer, line 159: User supplied ticks with `pseudolog10` are broken here. This range should result in `-10^1, 10^0, 10^1` but generates `10^-1.0413927, 10^0, 10^1.0413927`.

JamesWrigley: Thanks for catching that, I'll try to fix it next week.

jkrumbiegel:
Is the refimage from above still current where xs vs xs is plotted? On the left it says 10 ^ -1 while it should be -10 ^ 1

PR #4096, opened 2024-08-07
Update datashader docs

asinghvi17:
We don't necessarily need to specify the axis anymore, since the default axis for all plots is `Axis`.

Should we also consider hiding the `async=false` from the docs?  If so then users can copy paste immediately and get full performance.  Otherwise, using this interactively can segfault because of the number of events...


SimonDanisch:
Actually I'd rather default to `async=false` since I noticed that it segfaults from time to time 😅

asinghvi17:
I got segfaults with async=false...

asinghvi17:
I guess these changes can be merged now though, and I'll try to reproduce the segfault at some point...

docs/src/reference/plots/datashader.md
```diff
@@ -15,13 +15,10 @@ using DelimitedFiles
 airports = Point2f.(eachrow(readdlm(assetpath("airportlocations.csv"))))
 fig, ax, ds = datashader(airports,
     colormap=[:white, :black],
-    # use type=Axis, so that Makie doesn't need to infer
-    # the axis type, which can be expensive for a large amount of points
-    axis = (; type=Axis),
     # for documentation output we shouldn't calculate the image async,
```
SimonDanisch, line 18: Why would we remove this?

SimonDanisch:
Thanks!

PR #4102, opened 2024-08-08
support plot(interval, func) for rangebars and band

aplavin:


asinghvi17 (approved):
This looks good to me!

PR #4106, opened 2024-08-08
require version in bug report

EdsterG:
# Description

Update bug report template to require package version. I think this will be helpful for reproducibility and tracking when bugs are fixed. Could also leave the original question and add this new one? Maybe even request Julia version?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report-.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-.md
@@ -7,6 +7,6 @@ assignees: ''
 
 ---
 
-- [ ] what version of (W)(GL)Makie are you running?
+- [ ] what version of Makie are you running? (`]st -m Makie`)
 - [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
 - [ ] What platform + GPU are you on?
--- a/.github/ISSUE_TEMPLATE/bug-report-.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-.md
@@ -7,6 +7,6 @@ assignees: ''
 
 ---
 
-- [ ] are you running newest version (version from docs) ?
+- [ ] what version of (W)(GL)Makie are you running?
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report-.md
```diff
@@ -7,6 +7,6 @@ assignees: ''
 
 ---
 
-- [ ] are you running newest version (version from docs) ?
+- [ ] what version of (W)(GL)Makie are you running?
```
asinghvi17, line 10:
```suggestion
- [ ] what version of (W)(GL)(Cairo)Makie are you running?
```
but this could get confusing.  Maybe:
```suggestion
- [ ] what version of `Makie` and its backend (`(W/GL/Cairo)Makie` are you running?  (e.g. `]st -m Makie GLMakie`)
```

ffreyer: Honestly just the Makie version is enough. The backends are released in synch with Makie and we use the Makie version for pretty much everything.

asinghvi17: Usually yes but not always - the backends can sometimes desync back to the AbstractPlotting days where there was no Makie compat...@kellertuer recently had a similar issue and posted on Slack about it, for example.

ffreyer: Wouldn't that still give us either an ancient version or none?

asinghvi17: That's true...I guess we can always ask for the backend version in the issue!

asinghvi17:
```suggestion
- [ ] what version of Makie are you running? (`]st -m Makie`)

asinghvi17 (approved):
Looks good to me!

PR #4108, opened 2024-08-08
A few more notes of what we expect from a PR

SimonDanisch:
Wording up for debate, and glad to add some more infos ;)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,11 @@
+# Info
+
+* Please don't force push to PRs, it removes the history, creates bad notifications, and we will squash and merge in the end.
+* Feel free to ping for a review after a few days (@simondanisch, @ffreyer). We can't guarantee a review in a certain time frame, but we can guarantee to forget about PRs after a while.
+* The easier a PR is to review, the easier we can merge it:
+    * The more tests you add, the easier it is to see that your change works
+    * The clearer the problem being solved the easier.
+
 # Description
 
 Fixes # (issue)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,11 @@
+# Info
+
+* Please don't force push to PRs, it removes the history, creates bad notifications, and we will squash and merge in the end.
+* Feel free to ping for a review after a few days (@simondanisch, @ffreyer). We can't guarantee a review in a certain time frame, but we can guarantee to forget about PRs after a while.
+* The easier a PR is to review, the easier we can merge it:
+    * The more tests you add, the easier it is to see that your change works
+    * The clearer the problem being solved the easier.
+
```
EdsterG, line 8: Maybe this could be hidden behind an HTML comment? Similar to the JupyterLab pull request template: https://raw.githubusercontent.com/jupyterlab/jupyterlab/main/.github/pull_request_template.md

EdsterG: It'll be visible to all, and won't be rendered if the user doesn't delete those lines.

asinghvi17:
```suggestion
<!--
* Please don't force push to PRs, it removes the history, creates bad notifications, and we will squash and merge in the end.
* Feel free to ping for a review after a few days (@simondanisch, @ffreyer). We can't guarantee a review in a certain time frame, but we can guarantee to forget about PRs after a while.
* The easier a PR is to review, the easier we can merge it:
    * The more tests you add, the easier it is to see that your change works
    * The clearer the problem being solved the easier.
-->

**Comment by JamesWrigley** [comment]
Apologies if I was the motivation for the force-pushing advice :face_with_peeking_eye: I thought merging would be held up by the merge conflicts on the changelog but I guess you'll resolve those before merging.

**Comment by ffreyer** [comment]
In related news I updated https://github.com/orgs/MakieOrg/projects/12 a bit today to help us get more organized with pull requests. This should now automatically:
- add new pull requests as "Work in progress"
- move approved pull requests to "Ready to merge"
- move merged pull requests to "Merged"
- move closed pull requests to "Closed"

I would also like to have new pull requests start as drafts and stay that way until the author deems them ready to review/merge. I always find it hard to recognize whether is at that stage so I think that would be quite helpful.

**Comment by ffreyer** [comment]
Should we also suggest/request access rights to prs?

File: .github/PULL_REQUEST_TEMPLATE.md [hunk 2]
```diff
@@ -1,3 +1,12 @@
+# Info
+
```
EdsterG, line 2: Is this expected to be different from `Description` below? Since the content is no longer rendered, might be confusing to have two related headers.
```suggestion
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,12 @@
...
+<!--
+* Please don't force push to PRs, it removes the history, creates bad notifications, and we will squash and merge in the end.
+* Feel free to ping for a review after a few days (@simondanisch, @ffreyer). We can't guarantee a review in a certain time frame, but we can guarantee to forget about PRs after a while.
+* The easier a PR is to review, the easier we can merge it:
+    * The more tests you add, the easier it is to see that your change works
+    * The clearer the problem being solved the easier.
+-->
```
asinghvi17, line 9:
```suggestion
* Allowing write access on the PR lets us merge more easily, since we can make edits to the PR directly and update it if it gets out of sync with `master`.
-->
```

SimonDanisch: I think we do that already automatically? Or is there a way to deny that?
I at least seem to have write access to most PRs.

asinghvi17: I believe it's a checkbox that has to be actively selected.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,13 @@
+# Info
+
+<!--
```
asinghvi17, line 3:
```suggestion
<!--
# Info
```

asinghvi17 (approved):
This looks good to me!  Should we get this in?  Can always iterate on it later.

ffreyer:
Maybe we should also add some Notes on how refimages work. I.e. explain the structure
```julia
@reference_test "name of test" begin
    # code of test
    ...
    # make sure the last line is a Figure, FigureAxisPlot or Scene
end
```
and note that we will add the reference images when the pr gets merged. Maybe also instruct people to post an image of the refimg so we can verify it's correct before merging

PR #4116, opened 2024-08-11
WIP: reduce overlap among labels

### LINKED ISSUE #757: Prevent overlapping text

timholy:
# Description

Might close #757 (not all the examples, but it was pointed out that this is probably infeasible anyway)

This provides a mechanism to reduce the visual overlap among text labels in a plot. Sadly I can't demo this because Makie doesn't build for me locally right now (`ERROR: LoadError: KeyError: key :stroke_depth_shift not found`). So you'll have to imagine it: if you create a scatter plot and add text labels to each point, this provides an algorithmic approach to shifting the labels to reduce the overlaps. The key idea is to set this up as a minimization problem, where the objective function to be minimized is schematically

    sum(overlap_area(i, j) for j > i) + sum(abs2, x - x0)

where `overlap_area(i, j)` is the number of pixels of overlap between the extent-boxes of labels `i` and `j`, and `x-x0` is the concatenated vector of displacements of all the labels from their original positions.

This currently is a "trial balloon" to start a conversation about the general approach. As there have been many [algorithms proposed to solve this problem](https://en.wikipedia.org/wiki/Automatic_label_placement), we may want to discuss whether we like this approach before I put any more work into it. 

TODO:
- [ ] get Makie building locally so I can test this (a nearly identical version of this code works on a released version of Makie)
- [ ] get tests working and expand them
- [ ] add documentation
- [ ] write the gradient by hand and eliminate the new dependency on ForwardDiff

## Type of change

## Checklist


JamesWrigley:
> Sadly I can't demo this because Makie doesn't build for me locally right now (`ERROR: LoadError: KeyError: key :stroke_depth_shift not found`).

Just wanted to mention that I ran into this too, and for me the problem was not `dev`'ing `MakieCore`.

timholy:
Seems like there must be a `[compat]` bump that should have happened but didn't. This should be a build-time error, not a runtime error.

timholy:
From slack, for when I have time to get back to this:

> Text has offset which is in screen space (like the text itself) so it lends itself to being the thing that your repel algorithm modifies. The original data space positions would be unchanged by that (the positional args)

timholy:
Update on why this isn't moving forward: I've now seen other cases where overlap with non-text objects is also an issue. I'm uncertain that a text-only solution will be satisfying, but going beyond rectangular bounding boxes to objects that may be outlines rather than solid fills seems like a hard problem.

PR #4117, opened 2024-08-11
Fix GLMakie shader compatibility problems with arrays of arrays

ffreyer:
# Description

#3958 introduced `g_clip_planes[][8]` in sprites.geom which works on my PC and apparently also CI, but doesn't on my Laptop. This pr moves the clipping code to the geometry shader so that the result doesn't need to be passed around.

## Type of change


ffreyer:
>  ERROR: LoadError: ArgumentError: Package StaticArrayInterface does not have Requires in its dependencies:

> [0d7ed370] StaticArrayInterface v1.7.0

But that does have Requires as a dependency?

https://github.com/JuliaArrays/StaticArrayInterface.jl/blob/v1.7.0/Project.toml#L11

PR #4120, opened 2024-08-12
Bump version numbers for compat reqs

timholy:
A fresh checkout of Makie doesn't precompile unless you `dev` to the
versions in the repo. This indicates that there are changes not yet
captured by the compatibility requirements. Bump the version numbers so
that error message happens much earlier and the path towards resolution
is clearer.

If the changes that triggered these compilation faillures (awesome PRs like #3958, #4058, #3954) are instead breaking, then the version numbers I've picked here are incorrect.


SimonDanisch:
>A fresh checkout of Makie doesn't precompile unless you dev to the
versions in the repo.

I'm not sure what you're referring to and how this PR changes this? You mean when checking out Makie#master, the backends don't get added with the correct version?

A new release is immanent and we usually don't bump the versions before that.
We have a script for that https://github.com/MakieOrg/Makie.jl/blob/master/tooling/bump_versions.jl and usually it requires a few more steps before we can release.

timholy:
Sounds reasonable. Maybe the CONTRIBUTING.md could have a little something about this? Issues I discovered:
- presumably, then, it's quite routine that contributors need to `dev` some of the within-repo packages. Is there (or should there be) a script for that?
- success or failure in building the packages can depend on whether you have a `LocalPreferences.toml` that disables precompilation workloads

timholy:
To address your question:

> I'm not sure what you're referring to and how this PR changes this?

What it does is change a mysterious precompilation error into a package compatibility error. For people who notice that this is a monorepo, that should make it pretty obvious what to change.

SimonDanisch:
Yeah, I understand the issue better after seeing the exact issue you run into when looking at your PR ;) 
Maybe we can also put it here: https://github.com/MakieOrg/Makie.jl/pull/4108
We have a note in the readme: https://github.com/MakieOrg/Makie.jl?tab=readme-ov-file#developing-makie
But yeah I get that the situation isn't ideal :( 
But bumping versions before we tag seems difficult, at least because we usually only know if we need to bump MakieCore after all PRs have been merged.

timholy:
Ah, I hadn't noticed that (it's folded by default, which is probably good). I should have read more carefully.

On balance I don't think anything more needs to be done, you've already done what you plausibly can. Thanks, and sorry for the noise!

PR #4125, opened 2024-08-12
Fast path for heatmap

SimonDanisch:
This brings back the fast path for heatmaps, if there are no transformations and the heatmap is on a regular grid.

```julia
using WGLMakie, BenchmarkTools
@btime(
    begin
        heatmap!(s, (-1, 1), (-1, 1), data)
        display(s)
    end,
    setup = (data=rand(4000, 4000); s = Scene())
)
# PR, fix
212.649 ms (4410 allocations: 38.34 MiB)
# Master
295.968 ms (4341 allocations: 541.18 MiB)
```
Before, WGLMakie would create an `NxM` mesh, which is quite big (as can be seen in the allocations).
This was slower and lots of data to transfer.

This PR needed to adjust the conversion pipeline to not convert any `1..x` to `[1...x]`, which brings it's own set of problems that still needs cleaning up for this PR to pass tests.


SimonDanisch:
The WGLMakie failure seems to be an harmless artifact of slightly different interpolations from mesh -> image rendering for a regular heatmap.
@ffreyer, what do you think, I'd just update the reference Image and have this merged.

ffreyer:
Yea, they look harmless and sometimes even a bit better

GLMakie/src/glshaders/mesh.jl
```diff
@@ -50,7 +50,7 @@ function draw_mesh(screen, data::Dict)
         color_norm = nothing
         fetch_pixel = false
         texturecoordinates = Vec2f(0) => GLBuffer
-        uv_transform = Vec4f(1, 1, 0, 0)
+        uv_transform = Mat{2,3,Float32}(1, 0, 0, -1, 0, 1)
```
ffreyer, line 53: Maybe we should fix this in surface as well?

WGLMakie/src/imagelike.jl
```diff
@@ -163,26 +163,22 @@ function limits_to_uvmesh(plot, f32c)
...
 
-    # TODO, this branch is only hit by Image, but not for Heatmap with stepranges
-    # because convert_arguments converts x/y to Vector{Float32}
-    if px[] isa Makie.Interval && py[] isa Makie.Interval && Makie.is_identity_transform(t)
+    if px[] isa Makie.EndPoints && py[] isa Makie.EndPoints && Makie.is_identity_transform(t)
         rect = lift(plot, px, py) do x, y
-            xmin, xmax = Makie.endpoints(x)
-            ymin, ymax = Makie.endpoints(y)
+            xmin, xmax = extrema(x)
+            ymin, ymax = extrema(y)
```
ffreyer, line 169: Not sure if this is something to do here or in #3989, but a lot of these extrema probably need to be replaced with splatting or `x[1], x[2]` if we want to support reversed images, heatmaps etc

SimonDanisch: Yeah, I wouldn't want to do it here, and sticking with extrema makes it easier to find once we get to it.

Issue #3302, opened 2023-10-17
Alignment of tick marks and grid lines in Axis3


ldonetti:
In the MWE below, the tick marks of the x and y axes are slightly misaligned with the grid lines. I noticed that *all* images in the docs regarding [Axis3](https://docs.makie.org/stable/reference/blocks/axis3/) show the same issue, so I wonder whether this is a bug or a "feature". In the latter case, I wonder if there is any workaround to obtain grid lines correctly aligned with tick marks. 

MWE (with CairoMakie, but with GLMakie the issue is the same):
```
using CairoMakie
surface(ones(4,4), axis=(;type=Axis3))
```
![mwe](https://github.com/MakieOrg/Makie.jl/assets/80170096/075b94d3-cf27-4769-96c9-2de54d4f8427)

I am running the newest version and I don't think that the rest of the questions in the issue template are relevant since the problem can also be observed in several images of the package documentation.

**Edit**: lowering the resolution and exporting to a vector format (pdf or svg) increase the relative distance between the tick and the grid. 
![image](https://github.com/MakieOrg/Makie.jl/assets/80170096/98a1d5e0-1497-4094-bf3d-46d9222b6ff7)

Issue #3792, opened 2024-04-17
Avoid clipping on 3D zoom to be able to inspect meshes up close


Kevin-Mattheus-Moerman:
- [ ] are you running newest version (version from docs) ?
- [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [ ] What platform + GPU are you on?

I am hoping to create a visualisation for meshes whereby I can zoom in as much as I want (and for any orientation) without clipping occurring. So far I've been unsuccessful with setting `far` and `near`. The below gif shows the issue as well. Any help would be appreciated. 

Here is an example code: 
```julia

using GeometryBasics
using GLMakie

## BUILD CUBE 

# Create vertices       
s = 1/sqrt(3.0)

V = Vector{GeometryBasics.Point{3, Float64}}(undef,8)
V[1 ] = GeometryBasics.Point{3, Float64}( -s, -s, -s)
V[2 ] = GeometryBasics.Point{3, Float64}( -s,  s, -s)
V[3 ] = GeometryBasics.Point{3, Float64}(  s,  s, -s)
V[4 ] = GeometryBasics.Point{3, Float64}(  s, -s, -s)
V[5 ] = GeometryBasics.Point{3, Float64}( -s, -s,  s)
V[6 ] = GeometryBasics.Point{3, Float64}( -s,  s,  s)
V[7 ] = GeometryBasics.Point{3, Float64}(  s,  s,  s)
V[8 ] = GeometryBasics.Point{3, Float64}(  s, -s,  s)
    
# Create faces
F = Vector{QuadFace{Int64}}(undef,6)
F[1 ] = QuadFace{Int64}(1,2,3,4)
F[2 ] = QuadFace{Int64}(8,7,6,5)
F[3 ] = QuadFace{Int64}(5,6,2,1)
F[4 ] = QuadFace{Int64}(6,7,3,2)    
F[5 ] = QuadFace{Int64}(7,8,4,3)    
F[6 ] = QuadFace{Int64}(8,5,1,4)    

M = GeometryBasics.Mesh(V,F)

## VISUALIZATION

fig = Figure(size=(800,800))

# ax1 = Axis3(fig[1, 1], aspect = :data, xlabel = "X", ylabel = "Y", zlabel = "Z", title = "Cube mesh")

ax1 = LScene(fig[1,1],show_axis=true, scenekw = (backgroundcolor=:white,)) 
[...]

PR #4131, opened 2024-08-14
Add zooming & translation controls to Axis3

ffreyer:
# Description

Now that #3958 has been merged we can implement zooming and translation for Axis3 without data spilling out of he Axis limits. 

<details>
<summary> Old Description </summary
The first commit shows of a prototype for zooming.

I don't aim to work on this immediately. Instead I'm hoping that this pr can spark some discussions on how zooming and translations should work with Axis3, so that the end result is more polished.

Some questions:
- What point should zooming relate to? 
    - The center? 
    - A point offset from the center based on the mouse position? 
    - A data point under the mouse?
- Should up/down mouse movements strictly map to z-axis translations, and left/right mouse movements strictly map to xy-axis translations? Or should they freely map to the corresponding (x, y, z) vectors?
- How should translation speed scale with zoom? Should we set a maximum based on data limits to avoid losing sight of data?
- What things should we port from `Camera3D`/LScene? E.g.:
    - Directional restrictions for translations when pressing x/y/z and translating
    - Focus on clicked object/point on object (left alt + left click)
    - Recenter hotkey (left ctrl + left click)
    - Keyboard controls
</details>

I extended the following interactions to work with Axis3:
- `ScrollZoom` which comes with `x/y/zzoomlock` and `x/y/zzoomkey` Axis3 attributes like Axis does, and additionally a `zoommode` which controls whether zoom is centered on the Axis3 area (default) or on the cursor (1)
- `DragPan` which comes with `x/y/ztranslationlock` and `x/y/ztranslationkey` rather than `...pan...` as pan technically refers to a rotation of the camera.
- `LimitReset`

(1) "zoom centered on the cursor" really means that we want to keep a point under the cursor at the same location as we zoom in or out. With perspective projection in 3D this depends on the depth value of the point, which changes as we zoom. Without knowing exactly where that point is, we cannot fulfill that condition. And even if we ray-cast to find the coordinate, another point at a different depth that initially overlapped will move away as we zoom. Because of that I think it's better to just zoom towards the well-defined center.

TODO:
- [x] update documentation
- ~~maybe add interaction for "center on clicked object"~~
- [x] Reuse naming scheme and interactions from Axis
- [x] Add tests, check failures
- [x] fix axis reversal
- [x] fix CairoMakie text clipping
- ~~Discuss whether clipping related changes count as breaking or bug fixes~~ I think it doesn't actually change anything...?
- [x] test every possible viewmode with different perperspectiveness, axis reversals and aspect settings

Related issues: #3792, #1655, #1667

## Type of change

## Checklist


asinghvi17:
Zooming for Axis3 is basically just shrinking `finallimits`, right?

ffreyer:
I don't know yet. In the prototype code it is, but I don't know if that's a good way to do it. that code was really just to test if `clip_planes` allow me to cut off plots outside the Axis bbox while keeping the Axis decorations.

GLMakie/src/GLAbstraction/GLRender.jl
```diff
@@ -10,10 +10,10 @@ function setup_clip_planes(robj)
...
         glEnable(GL_CLIP_DISTANCE0 + UInt32(i))
     end
-    # TODO: This crashes?
-    # for i in N:7
-    #     glDisable(GL_CLIP_DISTANCE0 + UInt32(i))
-    # end
+    # TODO: This crashes? But is also necessary...
+    for i in max(0, N):7
+        glDisable(GL_CLIP_DISTANCE0 + UInt32(i))
+    end
```
ffreyer, line 16: Not sure in what context this was crashing but it's not crashing anymore on my laptop or PC. 
Without this turning clip_planes on once will turn them on forever, with random data. This causes plots to disappear randomly. Maybe this should be a hotfix instead of being part of this pr...

SimonDanisch:
I think we will want to make Axis3 the default axis with this, right?
Problems:
### Zooming into volumes
```julia
contour(rand(50, 50, 50); axis=(; type=Axis3)) # volume(...)
```
Zooming into this gives the error:
```julia
AssertionError: modelinv[4, 4] == 1
```
###  aspect=... different from axis2
* single numbers dont work (needs to_ndims)
* DataAspect is :data    
(@jkrumbiegel, I guess this is simply a mistake?)

Kevin-Mattheus-Moerman:
I tried this with: 

```
using GLMakie
using FileIO

fig = Figure(size=(800,800))
fig
```
However we get this unwanted clipping: 
Peek 2024-08-29 17-35

If we use the other environment e.g. like this: 
```
using GLMakie
using FileIO

mesh(brain,color = :white,  shading = FastShading, transparency=false)
```
We get the desired behaviour: 
Peek 2024-08-29 17-38

Can the `Axis3` behaviour be made fully equivalent?

ffreyer:
There is a `zoommode = :cursor` that keeps what's under the cursor in place (or tries to at least) (this is with the cursor at the center of the screen):

| zoommode = :center | zoommode = :cursor |
| --- | --- |
| clip | clip2 |

```julia
using GLMakie
using FileIO

begin
    fig = Figure(size=(800,800))
    ax1.zoommode[] = :cursor
    record(fig, "clip2.mp4", 1:60, fps = 30) do _
        e = events(fig)
        e.mouseposition[] = (400, 400)
        e.scroll[] = (0.0, 0.5)
    end
end
```

And for reference, this is what no clipping would look like:

no_clip

Kevin-Mattheus-Moerman:
@ffreyer I should add that in MATLAB you can also do the following when you have multiple axes in the same figure window. I imagine that it looks nightmare-ish but I actually like this :smile:
 

https://github.com/user-attachments/assets/5d5e79b6-c7fa-4e9a-9049-70d43a7cd748

SimonDanisch:
Looks like what matlab does is LScene but with better ticks (and maybe better camera controls) ;)

ffreyer:
On an implementation detail level it seems like it doesn't do a camera at all in the sense Makie uses it and instead just does interaction through `Transformation`. But yea, looks LScene but better

Kevin-Mattheus-Moerman:
@SimonDanisch @ffreyer just to remind you that actually classic default MATLAB behaviour is nearly exactly your current `Axis3` implementation. Here is classic MATLAB with code for completeness (and you can see it also suffers from snappy rotation): 
```matlab
figure;
title('test','fontSize',25);
patch('Faces',F,'Vertices',V,'FaceColor','w','EdgeColor','k');
view(3); axis tight; axis equal; axis vis3d; camlight('headlight');
drawnow;
```
Peek 2024-08-30 16-30

If I turn clipping off you get this: 
```matlab
figure;
title('test','fontSize',25);
patch('Faces',F,'Vertices',V,'FaceColor','w','EdgeColor','k');
h=gca;
h.Clipping='off';
view(3); axis tight; axis equal; axis vis3d; camlight('headlight');
drawnow;
```
Peek 2024-08-30 16-32

But in my opinion both of these are useless because: 
* With clipping zooming in becomes useless (and I'd argue the axis limits/labels are pretty much not useful either). 
* Without clipping the axis labels are ridiculous (not understandable) as they appear sometimes in the middle of the thing visualised or seem to bear no relationship to what you are looking at. 

This is why I implemented the trickery above, so as @ffreyer said, it is mostly a type of transformation that is applied to the whole axis. In think this is the only way to get natural feeling "CAD design" like explorability of what one is developing. Would it be hard to turn off clipping and to stop dynamically updating the axis limits? With those two changes perhaps everything could work? I cannot stress enough how important it is to be able to zoom in and check the meshes when developing and debugging mesh related and geometry processing codes. Thanks for continuing efforts on this. 

Should we have an online meeting about these features? I'd be happy to join the discussion if you are looking for help.

ffreyer:
> # MATLAB
> 
> The video below is for MATLAB visualization (I turned on edges so we can see the mesh better). By default MATLAB uses clipping and the behaviour for rotation etc by default is very similar to what you have here now. But here I turned clipping off and used my own axis implementation.
>
> MATLAB_axis_low.mp4

Did some prototyping for this in #4294.

Kevin-Mattheus-Moerman:
@ffreyer oh that looks very promising!

jkrumbiegel:
I'd be ok in principle to add an "unhinged" mode to Axis3 where it doesn't try to place itself in the layout rectangle anymore. Currently, all its modes are geared towards that, so that the box aligns nicely with other 2d axes sharing the same layout. That would let you "zoom" into it by just increasing the area assigned to it, which wouldn't change clipping etc. I think that's what Frederic is trying in that linked PR.

Maybe we could also reuse some of that axis code for LScene, that axis works a bit differently in the way that text is placed, for example.

ffreyer:
This is done from my side. Tests fail due to scatter markers not getting cut off by scene boundaries anymore and some negligible sub-pixel movements of axis decorations.

Something that may still be worth discussing is how Axis3 content and decorations interact with the region layouting assigns to the Axis3. Currently the content is locked to that region, i.e. if you do `mesh!(ax, ...)` that mesh is clipped at the boundary assigned by the layout. Decorations are not constrained to this region, because they weren't before and I tried to make this less breaking. (This is why sometimes a x/y/z labels will end up drawing over another layout element)

The new refimg marks the area the content is restrained to (i.e. computedbbox) with pale orange. So here the cat would be cut off at the red border. (The blue region is with protrusions subtracted and the different cells are changes in aspect and xreversed)
GLMakie_Axis3 viewmodes, xreversed, aspect, perspectiveness_step-1

ffreyer:
> Something that may still be worth discussing is how Axis3 content and decorations interact with the region layouting assigns to the Axis3. Currently the content is locked to that region, i.e. if you do `mesh!(ax, ...)` that mesh is clipped at the boundary assigned by the layout. Decorations are not constrained to this region, because they weren't before and I tried to make this less breaking. (This is why sometimes a x/y/z labels will end up drawing over another layout element)

Example for this:

```julia
using FileIO, GLMakie
cat = load(Makie.assetpath("cat.obj"))

fig = Figure()
Box(fig[1, 1:3], color = :lightblue)
Box(fig[3, 1:3], color = :lightblue)
Box(fig[2, 1], color = :lightblue)
Box(fig[2, 3], color = :lightblue)
ax = Axis3(fig[2, 2], viewmode = :free)
mesh!(ax, cat, color = :orange)
fig # and some zooming/interacting
```

image

jkrumbiegel:
Looks pretty good @ffreyer! The only thing I didn't like while trying it out was that if you're zoomed and panned to some position off-center, you lose your position quickly when further scaling or rotating because the reference point is still the center of the axis. I think it would be better to see if any plot objects in the axis can be picked, and if so, use the picked point as the pivot.

ffreyer:
I think I avoided doing these kinds of things because a lot of the times I tried to be smart and fancy with Camera3D it backfired. 

Some options for rotating are: (for viewmode = :free)
- rotate around axis center (current) - this is fairly easy to do, you just need to shift your coordinate system to the right origin
- rotate around scene center - this is also easy, and would cause the axis to rotate like in LScene. I.e. if you translate it away from the center, it rotates on a sphere surface around the center.
- rotate around cursor is tricky because of perspective projection...
    - rotate around ray-cast intersection between cursor and content - both the content and cursor move so what's under your cursor will change.
    - rotate around settable point - Camera3D can do this with `left_alt + left_click`. I think it's a decent solution but I'm not sure if it's good for general use. If you're looking at surface details of a mesh, rotating around a point on the mesh sounds useful. When you have multiple meshes or you care about the mesh as 3D object, you would probably prefer rotating around the center of the mesh.

Zooming is somewhat similar. You can:
- zoom relative to the axis center (current) - easy to do, same as rotation, but not what you want
- zoom towards the scene center - also easy like the respective rotation. Camera3D can do this with `cam3d!(a, zoom_shift_lookat = false)`. Also not what you want
- zoom toward cursor is tricky...
    - zoom towards cursor based on offset from center - is what `zoom_shift_lookat = true` does. I find this annoying when zooming out. I often end up pushing stuff I want to look at off-screen or into a small space when zooming out. Also still allows zooming past the object
    - zoom towards point under cursor - doesn't have issues with cursor/content movement and would allow for zooming to never enter/pass the content. But now zoom speed can fluctuate depending on what your cursor hovers. Also has the zoom-out problem

I would be fine adding a switch for scene vs axis centered rotation and zooming. Maybe also separating the axis center from the rotation + zoom center and adding extra interactions to control it (like Camera3D's alt-click, a secondary translation, a secondary reset). Those sound like relatively "dumb" interactions to me and therefore things you can relatively easily reason with.

ffreyer:
I realized that I could also just reuse the `viewmode != :free` translation to allow you to center the plot. Now it's setup like this (with viewmode = :free):
- right drag translates the content of the axis, allowing you to move the thing you want to rotate around and zoom towards into the center
- ctrl + right draw translates the entire axis (in the window)
- shift + left click now exists independently of ctrl + left click and re-centers content in the scene. ctrl - click reverts zooming, axis translations and rotations (maybe these tasks should be distributed differently?)
- alt + click has been added to center what the cursor hovers

Also did a bit of cleanup and fixed a rendering bug caused by `near` being too small:
Screenshot 2024-11-23 172520
I also noticed that #3584 is reappearing here, because frame lines are projected manually

src/camera/projection_math.jl
```diff
@@ -390,7 +393,7 @@ function project(proj_view::Mat4{T1}, resolution::Vec2, point::Point{N, T2}) whe
     # at this point the visible range is strictly -1..1 so FLoat64 doesn't matter
     p = (clip ./ clip[4])[Vec(1, 2)]
     p = Vec2{T}(p[1], p[2])
-    return (0.5 .* (p .+ 1) .* (resolution .- 1)) .+ 1
+    return 0.5 .* (p .+ 1) .* resolution
```
ffreyer, line 396: Tick alignment with frame lines got worse when I moved the frame to data space. So I spent too long trying to figure out why. 

I thought it might be a Float32 vs Float64 thing, so cleaned up the remaining Float32 types in Axis3. Didn't help. I tried forcing lines to convert on the CPU to make sure conversions don't happen till the end. Didn't help.

Then I got frustrated and moved the CairoMakie line point projection to Makie so I could revert moving the frame to data space without having clipping issues. And the alignment issues were still there. Then I tried using that code for ticks as well and it finally went away.

So i tried to figure out why and started comparing. Didn't notice the `resolution - 1`. Noticed that clip space points were very different, due to the CairoMakie code clipping to a -1..1 box. So I changed things up to avoid it and there was still a 0.5-1px difference. And then I finally noticed this line, changed it and I the issue seems to be fixed now.

Anyway, fixes #3302.

jkrumbiegel:
I'm not sure if I like the translation feature because you cannot really control which axes get translated how much when dragging over a 2D plane. I guess I would put that one behind the ctrl key if at all

ffreyer:
The WGLMakie failure is just a refimg I forgot to update

PR #4136, opened 2024-08-15
fix selected list in `WGLMakie.pick_sorted`

### LINKED ISSUE #4115: [WGLMakie] selected bar isn't the one directly under the mouse
- What version of Makie are you running? `0.21.6`, `0.21.5`
- Can you reproduce the bug with a fresh environment ? `Yes`
- What platform + GPU are you on? `MacOS`

The selected bar is not always the bar right under the mouse, which can make the tooltip a little confusing. It doesn't seem to be an off-by-one bug ether. ~More like a off-by-one-half because the next bar is outlined as soon as the mouse is halfway past the first bar.~ It seems to be off by some constant offset in pixel space.

Any thoughts on where to look to find the root cause?

```
using WGLMakie
fig = Figure()
ax = Axis(fig[1,1], xreversed = true, yaxisposition = :right)
DataInspector(ax)
barplot!(ax, 1:40, 1:40, direction=:x)
fig
```

https://github.com/user-attachments/assets/6235e9df-b365-46f6-b636-2d9628f404c5

EdsterG:
# Description

Fixes #4115.

The problem was with `picked_plots`. It was a map from plot id to mesh index, where the mesh index would also be the last one read. So `pick_sorted` would only compute the distance to one of the meshes, specifically the one that was last written to `picked_plots`. This PR stores a list of mesh indices for each plot id, allowing `pick_sorted` to compute the distance to all of them.

Before and after videos:

https://github.com/user-attachments/assets/5c7c7cd8-3435-43c0-a549-f75c62a0e8f7

https://github.com/user-attachments/assets/d8ce1a92-737e-4f03-b61f-805351877eff

## Type of change

## Checklist


EdsterG:
This PR fixes the MWE in #4115 but reveals an off-by-one bug when plotted with `direction=:x`. I don't think the off-by-one bug is in this PR, but maybe the devs have some thoughts.

EdsterG:
I found the bug but am not sure how to best handle it. `plot_matrix` matrix is scaled by `px_per_unit` so in my case it's 4x the size of `dx*dy` which means the distance check is only performed on the first 1/4th of the data.

https://github.com/MakieOrg/Makie.jl/blob/9e156624877e6d5e51648c47d735dac5fd540823/WGLMakie/src/wglmakie.js#L634-L653

Based on the GLMakie code I'm assuming `plot_matrix` needs to be down sampled to be `dx` by `dy`. Any recommendations on how best to do this @SimonDanisch @ffreyer? Or should I multiple `dx`/`dy` by `px_per_unit` in the `pick_sorted` loop?

EdsterG:
Regarding your suggestion, doesn't the same apply to `pick_closest` and `pick`?

EdsterG:
> I don't see any problems with :x and :y after that.

By the way, the problem only happens with PPU > 1. So technically this PR is ready to merge because it fixes the original issue.

ffreyer:
> > I don't see any problems with :x and :y after that.
> 
> By the way, the problem only happens with PPU > 1. So technically this PR is ready to merge because it fixes the original issue.

Ah ok that makes sense then

> Regarding your suggestion, doesn't the same apply to `pick_closest` and `pick`?

In GLMakie just pick_closest and pick_sorted. Not sure about WGLMakie.

EdsterG:
Any thoughts @SimonDanisch?

EdsterG:
Merged changes from #4137. This is ready for review.

ffreyer:
Ok, the only problem left that I can see when testing is with
```julia
begin
    scene = Scene(size = (400, 400))
    campixel!(scene)
    scatter!(scene, [100, 105, 105, 110, 110], [100, 105, 110, 105, 110])
    on(events(scene).mouseposition) do mp
        @info mp
        # closest
        plt, idx = Makie.pick(scene, mp, 75)
        @info plt !== nothing
    end
    display(scene, px_per_unit = 2.71)
end
```
With 50 radius I thought it might just be swallowing the `false` cases, but if you increase the radius you're actually missing picks. It stops printing at ~25 in the example above. Maybe WGLMakie is swallowing errors here?

EdsterG:
Thank you for the detailed investigation. I can't seem to recreate what you're describing on my machine. I get printouts regardless of where the mouse is located. Even in the opposite corner:
```
[ Info: (368.0, 316.0)
[ Info: false
[ Info: (369.0, 316.0)
[ Info: false
[ Info: (370.0, 314.0)
[ Info: false
[ Info: (370.0, 313.0)
[ Info: false
[ Info: (371.0, 313.0)
[ Info: false
[ Info: (371.0, 313.0)
[ Info: false
```

EdsterG:
Or did you mean it stops printing within 25px of the plot?

EdsterG:
I periodically get time out errors:
```
┌ Warning: Error in window event callback
│   exception =
│    Timed out
│    Stacktrace:
│      [1] error(s::String)
│        @ Base ./error.jl:35
│      [2] evaljs_value(session::Bonito.Session{Bonito.SubConnection}, js::Bonito.JSCode; error_on_closed::Bool, timeout::Float64)
│        @ Bonito ~/.julia/packages/Bonito/1lY3Q/src/session.jl:285
│      [3] evaljs_value(session::Bonito.Session{Bonito.SubConnection}, js::Bonito.JSCode)
│        @ Bonito ~/.julia/packages/Bonito/1lY3Q/src/session.jl:244
│      [4] pick_closest(scene::Scene, screen::WGLMakie.Screen, xy::Tuple{Float64, Float64}, range::Int64)
│        @ WGLMakie ~/Repositories/labs-monorepo/sandbox/Eddie/dev/Makie/WGLMakie/src/picking.jl:38
│      [5] pick(scene::Scene, xy::Tuple{Float64, Float64}, range::Int64)
│        @ Makie ~/.julia/packages/Makie/GWb7z/src/interaction/interactive_api.jl:86
│      [6] (::var"#11#12")(mp::Tuple{Float64, Float64})
│        @ Main ./REPL[13]:8
│      [7] #invokelatest#2
│        @ ./essentials.jl:819 [inlined]
│      [8] invokelatest
│        @ ./essentials.jl:816 [inlined]
│      [9] notify
│        @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
│     [10] setindex!(observable::Observable, val::Any)
│        @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
│     [11] macro expansion
│        @ ~/Repositories/labs-monorepo/sandbox/Eddie/dev/Makie/WGLMakie/src/events.jl:61 [inlined]
│     [12] (::WGLMakie.var"#80#83"{Dict{Any, Any}, Scene, Events})()
│        @ WGLMakie ./task.jl:514
└ @ WGLMakie ~/Repositories/labs-monorepo/sandbox/Eddie/dev/Makie/WGLMakie/src/events.jl:116
```

For me these timeout errors are the reason it stops printing the boolean value (the mouse position is always printed).

However, I also get these errors on `v0.21.6` and `v0.21.7` so it seems unrelated to this PR.

EdsterG:
I periodically get time out errors:
```
┌ Warning: Error in window event callback
│   exception =
│    Timed out
│    Stacktrace:
│      [1] error(s::String)
│        @ Base ./error.jl:35
│      [2] evaljs_value(session::Bonito.Session{Bonito.SubConnection}, js::Bonito.JSCode; error_on_closed::Bool, timeout::Float64)
│        @ Bonito ~/.julia/packages/Bonito/1lY3Q/src/session.jl:285
│      [3] evaljs_value(session::Bonito.Session{Bonito.SubConnection}, js::Bonito.JSCode)
│        @ Bonito ~/.julia/packages/Bonito/1lY3Q/src/session.jl:244
│      [4] pick_closest(scene::Scene, screen::WGLMakie.Screen, xy::Tuple{Float64, Float64}, range::Int64)
...
```

For me these timeout errors are the reason it stops printing the boolean value (the mouse position is always printed).

However, I also get these errors on `v0.21.6` and `v0.21.7` so it seems unrelated to this PR.

EdsterG:
After thinking about it some more, this PR might trigger timeout error more often. The min dist loops scale quadratically with PPU. In your example, `px_per_unit = 2.71` which means the min dist loops will have `~7.3x` more iterations than the broken version. Naturally, fixing PPU increases the time spent in the min dist loops, which can trigger timeouts more often.

ffreyer:
Probably should've mentioned that I don't get this with Bonito in a Browser. Happens in vscode. And it only happens if the mouse is below the point for me. In a browser I always get false which is also not good but I don't know if that has anything to do with the pr. I can test a bit more tomorrow.

I get the time out too, but I don't think this pr is to blame.

ffreyer:
As far as I can tell this has also fixed the time-out warnings

EdsterG:
> Oh it was doing two loops with dx, instead of one with dx and one with dy

Wow, good catch! Looks like `pick_closest` wasn't crashing because it had a workaround to check the length of `plot_matrix`. Now that it's fixed that check is no longer necessary because `plot_matrix` is guaranteed to be of length `dx*dy`.

EdsterG:
The timeout warning are fixed for me too.

SimonDanisch:
Thank you all :)

PR #4137, opened 2024-08-16
Apply px per unit in `pick_closest`/`pick_sorted`

EdsterG:
# Description

Apply px per unit in `pick_closest`/`pick_sorted`. ~I haven't applied this change to GLMakie because I had some more questions about the suggestion in #4136.~ Instead of blocking #4136, I decided to open a separate PR for this fix.

Before this PR, PPU wasn't accounted for in `pick_closest`/`pick_sorted`. A `PPU < 1` would crash GLMakie and a `PPU > 1` would cause erroneous tooltip behavior in GLMakie/WGLMakie.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Apply px per unit in `pick_closest`/`pick_sorted` [#4137]https://github.com/MakieOrg/Makie.jl/pull/4137
+
 ## [0.21.6] - 2024-08-14
 
 - Fix RectangleZoom in WGLMakie [#4127](https://github.com/MakieOrg/Makie.jl/pull/4127)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Apply px per unit in `pick_closest`/`pick_sorted` [#4137]https://github.com/MakieOrg/Makie.jl/pull/4137
+- Apply px per unit in `pick_closest`/`pick_sorted` [#4137](https://github.com/MakieOrg/Makie.jl/pull/4137).
 - Reduce updates for image/heatmap, improving performance [#4130](https://github.com/MakieOrg/Makie.jl/pull/4130).
 - Fix rendering of `band` with NaN values [#4178](https://github.com/MakieOrg/Makie.jl/pull/4178).
 
--- a/GLMakie/src/picking.jl
+++ b/GLMakie/src/picking.jl
@@ -67,17 +67,26 @@ end
 # Skips one set of allocations
 function Makie.pick_closest(scene::Scene, screen::Screen, xy, range)
     isopen(screen) || return (nothing, 0)
-    w, h = size(scene)
+    w, h = size(scene) # unitless dimensions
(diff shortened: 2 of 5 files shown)


EdsterG:
Referencing the suggestion from https://github.com/MakieOrg/Makie.jl/pull/4136#issuecomment-2293418017

```
# check xy in scene size...
x0, y0 = max.(1, floor.(Int, px_per_unit .* (xy .- range)))
x1, y1 = min.([w, h], ceil.(Int, px_per_unit .* (xy .+ range)))
dx = x1 - x0; dy = y1 - y0

# inline rect pick w/o applying px_per_unit again

x, y =  px_per_unit .* xy .+ 1 .- Vec2f(x0, y0)
# remaining distance calc + sorting as is
```

In pick sorted, `w, h = size(scene)` what are the units on these? If it's not pixels, `min.([w, h], ceil.(Int, px_per_unit .* (xy .+ range)))` won't work.

Inside pick native, `w, h = size(screen.root_scene)` are these units different from the ones in my previous question?

EdsterG:
I pushed a commit based on some assumptions that I am not able to verify.

ffreyer:
You can set px_per_unit with `display(fig, px_per_unit = ...)`

EdsterG:
Pushed a few more fixes. Now the only open question I have is `scene` vs `screen.root_scene` in `pick_sorted`? Should I assert they're the same? If they are expected to differ sometimes, which one should I use? If they are expected to be the same, why the redundancy?

EdsterG:
I've added a test that will fail without this fix. This PR is ready for review @ffreyer.

ffreyer:
I think WGLMakie still has some issues with sorting. I've been testing with
```julia
begin
    scene = Scene(size = (400, 400))
    campixel!(scene)
    scatter!(scene, [100, 102, 102, 104, 104], [100, 102, 104, 102, 104])
    on(events(scene).mouseposition) do mp
        # sorted
        picks = Makie.pick_sorted(scene, mp, 50)
        N = length(picks)
        i = N > 0 ? picks[1][2] : nothing
        @info N, i

        # closest
        # plt, idx = Makie.pick(scene, mp, 50)
        # @info plt !== nothing
    end
    display(scene, px_per_unit = 2.71)
end
```

For `pick_sorted` you should see the index (i) changing as you move the cursor around the blob of markers, but I only see 5. I thought this might be because multiple plots end up with the same (default) distance but that doesn't seem to be the case.

pick_closest also doesn't print anything if I'm inside the WGLMakie window but outside the pick range. No idea why. Also wouldn't hold up the pr for this

EdsterG:
Yeah, #4136 is important to fix the remaining WGLMakie issues. If you include the changes from the other PR, your example will work correctly.

ffreyer:
If both prs are done I'd say we should recombine them so it's easier to review

EdsterG:
> If both prs are done I'd say we should recombine them so it's easier to review

They're done from my perspective. I just haven't received much feedback on the other PR and though this fix is less "controversial", therefore I split to them to reduce friction. I'm also happy to re-combine if that'll help get them in faster.

EdsterG:
Merged into #4136.

PR #4138, opened 2024-08-16
Make the "visualize the camera's view box" example actually runnable

asinghvi17:
# Description

Simple docs PR to wrap the function in a figure object and use GLMakie.


ffreyer:
I think so? 
This was meant more as an example to run yourself and play around with, so you can get an intuition of what the camera shows and how the different controls and settings affect the visible area. It's probably not that useful as a static image in the docs and would at least need to be rotated better

Issue #4140, opened 2024-08-16
Loop export throw deprecation warnings and crashes CI


ChrisRackauckas:
https://github.com/MakieOrg/Makie.jl/blob/master/CairoMakie/src/CairoMakie.jl#L18-L23

```julia
Makie.Combined is deprecated, use Plot instead.
ERROR: LoadError: use of deprecated variable: Makie.Combined
Stacktrace:
 [1] eval(m::Module, e::Any)
   @ Core ./boot.jl:385
 [2] top-level scope
   @ ~/.julia/packages/CairoMakie/1O1dN/src/CairoMakie.jl:20
 [3] include
   @ ./Base.jl:495 [inlined]
 [4] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base ./loading.jl:2222
 [5] top-level scope
   @ stdin:3
in expression starting at /home/runner/.julia/packages/CairoMakie/1O1dN/src/CairoMakie.jl:1
in expression starting at stdin:3
Lattice Simulation Plotting: Error During Test at /home/runner/.julia/packages/SafeTestsets/raUNr/src/SafeTestsets.jl:30
```

asinghvi17:
Looks like a version mismatch, do you have the raw logs available?

asinghvi17:
Is this running in `julia --depwarn=error` by any chance?  We could not export deprecated symbols from the backends by checking `Base.isdeprecated(Makie, sym)` but that sounds breaking to me...

ChrisRackauckas:
> Is this running in julia --depwarn=error by any chance?

Yes, but Makie probably shouldn't be throwing depwarns every single using anyways so it's worth solving.

> We could stop exporting deprecated symbols from the backends by checking Base.isdeprecated(Makie, sym) but that sounds breaking to me...

You can also just fix the deprecation. Instead of deprecating the binding, deprecate the call. So make it into a standard function but when called with any arguments throw a depwarn and pass it to the right function. It's subtly different in the sense that importing but not using the function is not deprecated, which it shouldn't be since you are doing exactly that so you're currently relying on that behavior.

ChrisRackauckas:
```julia
function Combined end
[...]

PR #4143, opened 2024-08-17
Avoid deprecated bindings on re-exported items

ChrisRackauckas:
Fixes https://github.com/MakieOrg/Makie.jl/issues/4140


SimonDanisch:
Thanks, that seems like a better way :)

SimonDanisch (approved):

PR #4151, opened 2024-08-20
Improve Issues

ffreyer:
# Description

Similar to #4108, this pr aims to make some improvements to our Issues (and Discussions) infrastructure. Feel free to give your opinions and suggestions.

Some general thoughts:
- Issues should be actionable, i.e.
  - not be question, loose feature requests, planning or other open ended posts. Those should be Discussions
  - they should deal with one (preferably small) problem. So no roadmaps or other kinds of issue collections (Again, Discussions)
  - not be so old that their validity is unclear (but this is for issue cleanup)
- duplicates should be closed in favor of the first such issue
- We should take labels more seriously to make it easier to find specific issues. E.g.
  - tag what is affected, i.e. Makie/GLMakie/WGLMakie/CairoMakie/RPRMakie (maybe github docs, CI?)
  - tag the type of issue, i.e. bug/regression/documentation/enhancement
  - categorize issues by problem groups, e.g. plot attributes, rendering, DataInspector, transformations, 2D/3D
  - remove labels that are not applicable or have few related issues, e.g. dependencies, naming, planning, question, vizcon, GeometryBasics

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report-.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-.md
@@ -4,7 +4,6 @@ about: Report an error or incorrect behavior
 title: ''
 labels: bug
 assignees: ''
-
 ---
 
 - [ ] what version of Makie are you running? (`]st -m Makie`)
--- a/.github/ISSUE_TEMPLATE/bug-report-.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-.md
@@ -1,6 +1,6 @@
 ---
-name: 'Bug report '
-about: Create a report to help us improve
+name: Bug report
+about: Report an error or incorrect behavior
 title: ''
 labels: bug
 assignees: ''
@@ -10,3 +10,5 @@ assignees: ''
 - [ ] what version of Makie are you running? (`]st -m Makie`)
 - [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
 - [ ] What platform + GPU are you on?
+- [ ] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)
+- [ ] Please provide an minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.
--- a/.github/ISSUE_TEMPLATE/regression.md
+++ b/.github/ISSUE_TEMPLATE/regression.md
@@ -1,0 +1,14 @@
+---
+name: Regression
+about: Something that used to work but doesn't anymore
+title: ''
+labels: regression
+assignees: ''
+---
+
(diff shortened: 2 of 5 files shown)


ffreyer:
I'm planning to go through all the open issues at some point and close the ones that are too old to act on and add tags as I go. I think after that it'll be clearer what's actually useful to have there.

.github/ISSUE_TEMPLATE/regression.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Regression
+about: Something that used to work but doesn't anymore
```
asinghvi17, line 3: What's the difference between this and a bug, from an end user perspective?  Should this be something we ask them to mention in the bug report category?

ffreyer: I think having this as an option would help reduce some work for us. If the user knows something worked beforehand they can categorize it correctly with this and we don't have to. If they don't know they can just call it a bug and we can change the label.
And I think separating bugs from regressions is good. If everything is just a bug it becomes kind of meaningless to even have that label.

asinghvi17: Sounds good to me!  We should probably clarify that in the bug report template as well, to avoid confusion.

.github/ISSUE_TEMPLATE/bug-report-.md
```diff
@@ -1,12 +1,13 @@
 ---
-name: 'Bug report '
-about: Create a report to help us improve
+name: Bug report
+about: Report an error or incorrect behavior
 title: ''
 labels: bug
 assignees: ''
-
 ---
```
asinghvi17, line 8:
```suggestion
<!-- 
If this is about something that worked in an older version of Makie, 
please file a regression report by going back to the issue page 
and selecting the "Regression" template.
-->

.github/ISSUE_TEMPLATE/regression.md
ffreyer: The

> about: Something that used to work but doesn't anymore

is visible when you select a template for your issue. I think that should be enough?

ffreyer:
Might also be a good idea to have 
\`\`\`julia
\# Your code here
\`\`\`
as part of the templates for people that are new to the syntax

.github/ISSUE_TEMPLATE/regression.md
asinghvi17: I guess?  Personally my eyes might skip over that but if it's sufficiently obvious then it should be fine.

SimonDanisch (approved):

asinghvi17:
Should we get this in?

SimonDanisch (approved):

PR #4157, opened 2024-08-21
Fix clip state corruption

ffreyer:
# Description

Re-enables the `glDisable(GL_CLIP_DISTANCE<N>)` lines without which OpenGL can get into a state where clipping is enabled but the values are not set. In that state plots will randomly get clipped. 

I got to this state in #4131 with some tests that include multiple Axis3, each setting 6 clip planes via ax.scene.theme.clip_planes. I don't think I managed to reproduce it by just setting one clip plane in a plot.

In #3958 I had some trouble with this code and had it disabled because of that. I'm not sure what exactly the problem was, but now I have no issues with it on my Laptop or PC.

## Type of change

## Checklist

### DIFF
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -77,7 +77,11 @@ function mixin_generic_plot_attributes()
         inspector_clear = automatic
         "Sets a callback function `(inspector, plot, index) -> ...` which replaces the default `show_data` methods."
         inspector_hover = automatic
-        "TODO: docs"
+        """
+        Clip planes offer a way to do clipping in 3D space. You can set a Vector of up to 8 `Plane3f` planes here, 
+        behind which plots will be clipped (i.e. become invisible). By default clip planes are inherited from the 
+        parent plot or scene. You can remove parent clip_planes by passing `Plane3f[]`. 
+        """
         clip_planes = automatic
     end
 end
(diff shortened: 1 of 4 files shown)


SimonDanisch (approved):

MakieCore/src/basic_plots.jl
```diff
@@ -77,7 +77,11 @@ function mixin_generic_plot_attributes()
         inspector_clear = automatic
         "Sets a callback function `(inspector, plot, index) -> ...` which replaces the default `show_data` methods."
         inspector_hover = automatic
-        "TODO: docs"
+        """
+        Clip planes offer a way to do clipping in 3D space. You can set a Vector of up to 8 `Plane3f` planes here, 
+        behind which plots will be clipped (i.e. become invisible). By default clip planes are inherited from the 
+        parent plot or scene. You can remove parent clip_planes by passing `Plane3f[]`. 
+        """
```
asinghvi17, line 84:
```suggestion
        """
        Clip planes offer a way to do clipping in 3D space. You can set a Vector of up to 8 `Plane3f` planes here, 
        behind which plots will be clipped (i.e. become invisible). By default clip planes are inherited from the 
        parent plot or scene. You can remove parent `clip_planes` by passing `Plane3f[]`. 
        """
```
Minor formatting change

asinghvi17 (commented):
I can't speak to the OpenGL stuff, this looks good to me after trying it locally though.

PR #4242, opened 2024-08-26
Fix lines with OffsetArrays in CairoMakie

asinghvi17:
# Description

Allows `lines(::OffsetArray{T, 1})` to work again.  CairoMakie had an issue with the new line clipping code where it assumed that the indexing of the point array was one based, which has now been fixed.  The final array returned from clipping is guaranteed to be a regular `Vector`, though.

Also adds a new reference test.

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/utils.jl
+++ b/CairoMakie/src/utils.jl
@@ -203,8 +203,8 @@ end
         f32convert = Makie.f32_convert_matrix(scene.float32convert, space)
         transform = Makie.space_to_clip(scene.camera, space) * model * f32convert
         clip_points = Vector{Vec4f}(undef, length(points))
-        @inbounds for i in eachindex(points)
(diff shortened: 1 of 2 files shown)


asinghvi17:
I guess there's also a question of whether we want to patch this or simply ensure that we don't propagate offsetarrays into the deeper code.  Personally I'd prefer to have the code robust to exotic array types, but it's up for discussion.

asinghvi17:
CI failures are because I added the new reference image.

SimonDanisch:
>CI failures are because I added the new reference image.

Well, yes, but because the example is broken:
image

jkrumbiegel:
Note that the failure you get when there are missing images is different from the ones you get when there's an error. In the former case, an additional status is posted like "3 missing reference images" or so. The reftests themselves won't fail for missing images.

asinghvi17:
Gah, this is what I get for coding late at night...:D

ffreyer (approved):

PR #4270, opened 2024-08-27
Avoid gl_ClipDistance in geometry shader

### LINKED ISSUE #4148: Precompilation segfault on WSL2 with GLMakie 0.10.6
- [X] what version of Makie are you running? (`]st -m Makie`)
```
  [e9467ef8] GLMakie v0.10.6
  [ee78f7c6] Makie v0.21.6
``` 
- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
Yes

- [X] What platform + GPU are you on?
Windows running WSL2 (on the Linux side) with Ubuntu 20.04, Intel Iris Xe Graphics.

```julia
pkg> activate --temp
(jl_knUBOc) pkg> add Makie GLMakie
 ⋮
Precompiling project...
  ✗ GLMakie
  2 dependencies successfully precompiled in 105 seconds. 232 already precompiled.
  1 dependency errored.
  For a report of the errors see `julia> err`. To retry use `pkg> precompile`

julia> using GLMakie
Precompiling GLMakie
        Info Given GLMakie was explicitly requested, output will be shown live

[2007] signal (11.1): Segmentation fault
in expression starting at /home/tim/.julia/packages/Makie/GWb7z/precompile/shared-precompile.jl:4
unknown function (ip: 0x7f6cb5bfb2ec)
unknown function (ip: (nil))
Allocations: 39480059 (Pool: 39455302; Big: 24757); GC: 59
```

However, GLMakie v0.10.5 works fine.

ffreyer:
# Description

Fixes #4148

Based on some testing with Simon the cause of the issue seems to be `gl_ClipDistance` being used in a geometry shader. This pr replaces that with just not generating scatter geometry. That is equivalent to what the previous implementation did, in effect, as clipping was/is only based on the scatter position.

This is a bit different from the version Simon tested so it needs to be checked again. (Who knows, maybe WSL also doesn't like geometry shaders not generating geometry)

## Type of change

## Checklist

### DIFF
--- a/GLMakie/assets/shader/sprites.geom
+++ b/GLMakie/assets/shader/sprites.geom
@@ -65,18 +65,25 @@ flat out uvec2 f_id;
 out vec2 f_uv;
 flat out vec4 f_uv_texture_bbox;
 flat out vec2 f_sprite_scale;
-out float gl_ClipDistance[8];
 
-void process_clip_planes(vec3 world_pos)
+
+bool is_clipped(vec3 world_pos)
 {
+    // We clip scatter points based on the user position rather than the 
+    // sprite vertex positions.
     // distance = dot(world_pos - plane.point, plane.normal)
     // precalculated: dot(plane.point, plane.normal) -> plane.w
-    for (int i = 0; i < num_clip_planes; i++)
-        gl_ClipDistance[i] = dot(world_pos, clip_planes[i].xyz) - clip_planes[i].w;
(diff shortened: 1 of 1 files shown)


hexaeder:
Works for me! I was able to precompile GLMakie and plot something after `]add GLMakie#ff/fix_wsl`

PR #4275, opened 2024-08-28
support !-functions in plotfunc() and func2type()

aplavin:
# Description

These are internal functions, but still are useful for generic code: stuff like "get the recipe type from a plotting function".
This PR makes them support !-functions as well. Eg, `plotfunc(scatter!) === scatter`, `func2type(scatter!) === Scatter`.

## Type of change

## Checklist


SimonDanisch:
```
 Expression: plotfunc(hist!) === hist
  UndefVarError: chopsuffix not defined
  ```
Ah yeah, already suspected this isn't 1.6 :( 
At some point we may want to drop 1.6, but maybe for now we can just do something that works on 1.6?

SimonDanisch:
thanks!

Issue #3672, opened 2024-03-03
Strange behavior when inspecting lines in WGLMakie


kbarros:
This is a follow-on to https://github.com/MakieOrg/Makie.jl/issues/3503. Not an immediate blocker for us; we've moved away from WGLMakie due to multiple issues (https://github.com/SunnySuite/Sunny.jl/issues/211). GLMakie gives an overall more robust experience, and we're happy with it.

The following snippet leads to strange behavior, as shown in the movie below.

```julia
using WGLMakie
begin
    segments = Point2f[(0, 0), (0, 1), (1, 0), (1, 2)]
    color = [:red, :black, :blue, :black]
    inspector_label(_, i, _) = (@show Int(i); ["hello", "world", "nice", "day"][mod1(i, 4)])
    fig, ax, pl = linesegments(segments; color, inspector_label, linewidth=10)
    DataInspector(ax)
    display(fig; backend=WGLMakie)
end
```

https://github.com/MakieOrg/Makie.jl/assets/290205/6f84876d-7e97-4636-90e4-c6d2a9a3f9b2

kbarros:
If anyone else encounters this, a workaround for us was to use `arrows` instead of `linesegments`. Setting `arrowsize = 0` draws just the line part of an arrow, as a cylinder. With `arrows`, the inspector pop-up works well in both GLMakie and WGLMakie.

EdsterG:
Seems like the index returned by `pick_closest` differs between GLMake and WGLMakie and the ray tracing code has an offset that only works with GLMakie indexing.

What WGLMakie version did you record your video with?

kbarros:
I tried again just now with the master versions of WGLMakie, Bonito, etc. There is still some strange behavior, but not as strange as before. Mousing over the red segment is now fixed: It shows a hover window with the text "hello" in the right position. Mousing over the blue segment now correctly shows a hover window with the correct text "nice". However, in the second case, the "nice" hover is in the wrong the position (it shifts leftward as the mouse position is raised vertically).

EdsterG:
The position is due to the “wrong” index returned by `pick_closest`.

SimonDanisch:
Yeah the shaders in WGLMakie and GLMakie work a bit different and its hard to do exactly the same thing in them for the index.
And it was also not 100% clear to me which is the better behaviour.
Maybe we should first figure out the correct behaviour for the returned index and then figure out how to change (W)GLMakie.

[...]

PR #4282, opened 2024-08-28
fix nan handling in WGLMakie

### LINKED ISSUE #4109: [WGLMakie] Constantly getting `Invalid text boundingbox` errors.
- [x] What version of Makie are you running? `0.21.8`
- [x] Can you reproduce the bug with a fresh environment ? `Yes`
- [x] What platform + GPU are you on? `MacOS`

~Unfortunately, I don't have a MWE right now. It's much easier to repro with a complex figure. I've been trying to debug this locally and would like some pointers from the devs.~ See https://github.com/MakieOrg/Makie.jl/issues/4109#issuecomment-2278753158 for MWE.

I've narrowed it down to this function:
```
function string_boundingbox(x::Text{<:Tuple{<:AbstractArray{<:GlyphCollection}}})
    if x.space[] == x.markerspace[]
        pos = to_ndim.(Point3d, x.position[], 0)
    else
        cam = (parent_scene(x).camera,)
        transformed = apply_transform(x.transformation.transform_func[], x.position[])
        pos = Makie.project.(cam, x.space[], x.markerspace[], transformed) # TODO: vectorized project
    end
    return string_boundingbox(x[1][], pos, to_rotation(x.rotation[]))
end
```

Specifically the first case, `pos` is invalid because `x.position[]` is `[NaN, NaN, NaN]` when `x.space[] == x.markerspace[]`. What are some reasons that make `x.position[]` invalid and where can I poke around next?

EdsterG:
# Description

Fixes #4109 #3672

This PR is a proof of concept, mainly to solicit feedback from the developers. Instead of reading the instance id from the shader, use the index that's already computed when handling nans and creating line loops. It's not ideal that the index needs to be serialized and passed to the front end as it adds additional overhead, but I don't think the shaders current have enough information to infer the index otherwise.

## Type of change

## Checklist


EdsterG:
Would be nice to get some feedback/direction. @ffreyer

ffreyer:
I don't have time to think about this more deeply atm because of the GeometryBasics refactor. 

I remember that there were some issues with the number of (varying) buffer variables when we reworked the WGLMakie line shaders. I think that was with vertex -> fragment shader buffers, but I'm not sure anymore. That would be something to check.

If there are no issues with those limits then I think this approach is fine for now. The code in Shaders.js looks like there is some encoding or decoding missing for ints, I'd guess either with serialize_tree on the Julia side or that attribute_type function?

EdsterG:
Thank you for the context switch and reply to this PR.

I haven't run into any issues with the number of buffer variables in the shaders, and I've been plotting complex figures with this change for over a week now. Was there a more direct check you had in mind?

I took another look at `Shaders.js` and was able to find where the missing encoding was. Cleaned up the code and tested. Everything works well.

ffreyer:
Did some googling again to figure out what I'm actually talking about. There is a GPU-dependent maximum number of vec4's that can go into a shader stage. I thought it was (also) WebGL Version dependent, maybe not? All GPUs allow 8+, most 12+, see https://opengles.gpuinfo.org/displaycapability.php?name=GL_MAX_VARYING_VECTORS&esversion=2

We have:
- 4x vec3 pos (or was it vec4?)
- 2x vec4 color
- 2x float lastlen
- 2x linewidth, counting as 0.5x vec4

where vec4 and vec3 count for 1 slot, vec2 for 0.5 slots and float for 0.25 slots, so 4+2+0.5+0.5 = 7 total. (See https://stackoverflow.com/questions/26682631/webgl-shaders-maximum-number-of-varying-variables) You're adding another pair of uints, which should count as half a vec4 slot if I understand things correctly. So I think we're fine here.

Is this producing the same indices as GLMakie?

EdsterG:
> Is this producing the same indices as GLMakie?

This is producing the same indices, though maybe in a round-about way. The code returns `lineindex_start` and then 1 is added in `pick_native` which makes the index match GLMakie.
https://github.com/MakieOrg/Makie.jl/blob/86177428baa835d88ee733d4be57b5340de70053/WGLMakie/src/picking.jl#L22

I chose to leave the `+1` to minimize the overall diff but can remove it and have the code return `lineindex_end` instead. Let me know which you prefer, if any.

Also, it looks like the tests are failing because offset arrays can have negative indices. I can revert `lineindex` to be an int but that'll immediately cut the max number of line points in half. It looks like the shader only has 16 bits of precision for the index, so going from uint to int is quite significant in my opinion, 65535 points down to 32767 points.
https://github.com/MakieOrg/Makie.jl/blob/86177428baa835d88ee733d4be57b5340de70053/WGLMakie/src/wglmakie.js#L539
I'll need to think about how to do this in a way that'll preserve the full uint range. Please let me know if you have any suggestions.

ffreyer:
Now that I think about it I wonder why we let OffsetArrays get this far. @asinghvi17 Is there any good reason to not convert them in convert_arguments?

ffreyer:
Ok, I think we should handle OffsetArray in convert_arguments then and not complicate picking by considering OffsetArrays there. I guess we can have something like `enumerate(eachindex(ps))` as a bandaid fix here until we move the conversion? Or maybe work with values(ps)?

EdsterG:
Ok, I added a `convert_argument` definition for point based offset vectors. Tests pass now.

ffreyer:
I tested this locally, seems to be working fine.

@SimonDanisch Is there a way we can test picking in WGLMakie?

@asinghvi17 Do you think we need any other OffsetArray conversions for lines/PointBased?

SimonDanisch:
I think `pick(scene, mousepos)`  should just work fine? Unless it requires advanced opengl which electron on CI doesn't offer (unlikely I think, since it renders the Volumes just fine).

EdsterG:
@SimonDanisch would you point me to a working example? I need a screen object to test pick and I can't find a single WGLMakie test that generates a screen object in CI. The following code just hangs in a headless machine:
```
    fig, ax, plt = lines(1:5, [1:2; NaN; 4:5]);
    screen = display(fig); # hangs
    screen = display(WGLMakie.Screen(fig.scene), fig); # hangs
    screen = display(WGLMakie.Screen(fig.scene, visible=false), fig); # also hangs
    screen = Makie.getscreen(fig.scene) # return nothing
```

SimonDanisch:
On headless you need:
```
using Bonito
import Electron
edisplay = Bonito.use_electron_display()
```
Which should already be part of the WGLMakie CI!

EdsterG:
Hmm, I can't seem to get this working locally. I just get the following error:
```
julia> plot,idx = pick(ax.scene, point_px)
ERROR: BoundsError: attempt to access 0×0 Matrix{Tuple{Union{Nothing, AbstractPlot}, Int64}} at index [1, 1]
Stacktrace:
 [1] throw_boundserror(A::Matrix{Tuple{Union{Nothing, AbstractPlot}, Int64}}, I::Tuple{Int64, Int64})
   @ Base ./essentials.jl:14
 [2] checkbounds
   @ ./abstractarray.jl:699 [inlined]
 [3] getindex(::Matrix{Tuple{Union{Nothing, AbstractPlot}, Int64}}, ::Int64, ::Int64)
   @ Base ./array.jl:918
 [4] pick(::Scene, screen::WGLMakie.Screen, xy::Vec{2, Float64})
   @ WGLMakie ~/.julia/packages/WGLMakie/80NMe/src/picking.jl:67
 [5] pick(scene::Scene, xy::Vec{2, Float64})
   @ Makie ~/.julia/packages/Makie/aQT24/src/interaction/interactive_api.jl:73
 [6] top-level scope
   @ REPL[29]:1
```

Would you please post a MWE for pick without nan and I can then modify it to test the NaN case.

ffreyer:
Maybe we should just handle that in a separate pr then. We'll need to make one for picking other plots anyway

EdsterG:
What do you think @SimonDanisch? Shall we merge as is?

SimonDanisch:
I'm fine with that :)

Issue #4268, opened 2024-08-27
scroll text if width of Textbox is exceeded


bjarthur:
`Textbox`'s for which the `width` is specified currently don't display any further characters after the box is filled.  the user can keep typing, but the letters don't appear.  you can move the cursor back into the box and delete characters, and what you typed to the right edge of the box will appear.

a more useful behavior to me and perhaps others because this is how bokeh works, is to scroll everything to the left as the user keeps typing after the box is full.

PR #4293, opened 2024-08-30
scroll text if width of Textbox is exceeded

bjarthur:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/4268

## Type of change

## Checklist


bjarthur:
not sure how to add a test.  happy to do so if someone suggests a way

bjarthur:
again, could i please get a review?  here's demo code and a movie of the amazing new functionality this PR offers to Makie :)

```
julia> using GLMakie

julia> f = Figure()

julia> Textbox(f[1,1], width=100)
```

https://github.com/user-attachments/assets/ff19fe50-c6dd-499c-8fbe-4b9f34b8c8ce

jkrumbiegel (commented):
This should have a visual reference test I think. Textbox doesn't have its own one so a new one can be added in figures_and_makielayout.jl. All the branches that currently codecov complains about should be handled there, like one normal textbox, one with text scrolled.

With regards to the implementation, I guess this all could be handled a bit more cleanly by a `translate!` on the text and the cursor plot objects instead of modifying their input args. That would correspond more to the idea of "scrolling", too. I don't really like `textplot.converted[1][] = [Point3f(oldv...` as at some point it gets hard to reason about what influences what. Textbox is already not so easy to understand.

bjarthur:
thanks for the suggestion regarding the implementation!  i have refactored to use `translate` and indeed it is much more understandable.  working on tests...

bjarthur:
i have added visual reference tests, but am having trouble uploading them.  when i click on "Update reference images with selection", a popup appears with a "v0.21.0" tag.  irrespective of whether i stick with that or choose another (e.g. "v0.22.0", or "master") i get an error on the julia REPL.  is this because i don't have write permissions to the required repo?

jkrumbiegel:
Yes you need commit rights to use the reference updater. Contributors can only add new test code which notifies committers in the PR status that new images are needed

bjarthur:
thanks @jkrumbiegel for the info.  let me finish the tests and i'll ping you for another review.

bjarthur:
ready for review again.

codecov must not consider the reference tests as those three lines are definitely used there

bjarthur:
the ReferenceTests report that "Violin plots differently scaled.png" is different.  how could that possibly be?

SimonDanisch:
That's not a problem of the pr ;) we've been updating the reference images and there's likely a Missmatch

jkrumbiegel (approved):

SimonDanisch:
Thanks :)

PR #3550, opened 2024-01-13
Standardize view and projection matrices

ffreyer:
# Description

This prepares some things for #3226:
- move translations from projection to view matrices
- track eyeposition in Axis and PolarAxis
- rework Axis3 to use `perspectiveprojection` and `lookat` without modification (though I split up `lookat` into two steps)
- add inv(::Quaternion)

And some random minor cleanup of Axis3 docs (indention, wrong render order in W/GLMakie, double-plotting of Label) + camera controller (name change).

<details>
<summary>Axis3 comparsion</summary>

Axis3 has more extensive changes so things could be different. Here are some new vs old comparisons with docs:

 #### Aspect

Screenshot from 2024-01-13 14-39-03

Screenshot from 2024-01-13 14-39-03

#### Viewing angles

Screenshot from 2024-01-13 14-41-01

Tick placement seems to be slightly different (e.g. bottom left z is cut off more, top left z's overlap more(?))

Screenshot from 2024-01-13 14-41-35

Slightly less space to the title here as well

#### Perspectiveness

Screenshot from 2024-01-13 14-42-26

#### viewmodes

Screenshot from 2024-01-13 14-50-56

Also checked this with `perspectiveness = 0.6`, no visible changes.

#### Tight frustum testing/debugging

This will show you the generated frustum (i.e. what area of the scene will be rendered) with th view space bounding box (+ bounding sphere for :fit) of the Axis3.

```julia
using GeometryBasics

function camera_debug(gridpos, ax::Axis3)
    function frustum_snapshot(cam)
        r = Rect3f(Point3f(-1, -1, -1), Vec3f(2, 2, 2))
        rect_ps = coordinates(r) .|> Point3f
        insert!(rect_ps, 13, Point3f(1, -1, 1)) # fix bad line

        inv_pv = inv(cam.projectionview[])
        return map(rect_ps) do p
            p = inv_pv * to_ndim(Point4f, p, 1)
            return p[Vec(1,2,3)] / p[4]
        end
    end

    scene = ax.scene
    cam = scene.camera
    frustum = map(pv -> frustum_snapshot(cam), cam.projectionview)

    scene2 = LScene(gridpos, show_axis = false)

    _cc = Makie.Camera3D(
        scene2.scene, projectiontype = Makie.Orthographic,
        near = 0.001f0, far = 1000.0f0, clipping_mode = :static
    )
    lines!(scene2, frustum, color = :blue, linestyle = :dot)

    # bbox is Axis3 specific
[...]

PR #4294, opened 2024-08-30
Prototype Suggestions from #4131

ffreyer:
# Description

This is a quick-ish prototype for the suggestion made in #4131, i.e. apply zoom and translation not just to the axis content, but also decorations.

Changes boil down to:
- add ax.zoom_mult and ax.lookat and have interactions affect camera matrices through them
- move ticks, tick labels, axis labels and axis frame to from the blockscene to the 3D scene to avoid different Scene/viewport base clipping
- merge #3550 to not completely break `viewmode` (Though I noticed that this is not allocating enough space for ticks yet. Also not sure if the rotation based resizing/zooming that viewmode does makes sense with these changes...)

Example:

https://github.com/user-attachments/assets/ba2725e7-b713-4885-a2b7-c29aadbd42ce

```julia
using GLMakie
using FileIO

begin
    fig = Figure(size=(600,600))
    ax1.zoommode[] = :cursor
    fig
    record(fig, "hack2.mp4", 1:150, fps = 30) do idx
        e = events(fig)
        if idx < 30
            e.mouseposition[] = (300, 300)
            e.scroll[] = (0.0, 0.5)
        elseif idx < 60
            if e.mousebutton[] != Makie.MouseButtonEvent(Mouse.right, Mouse.press)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.right, Mouse.press)
            end
            e.mouseposition[] = e.mouseposition[] .+ 3
        elseif idx < 120
            if e.mousebutton[] != Makie.MouseButtonEvent(Mouse.left, Mouse.press)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.right, Mouse.release)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
            end
            e.mouseposition[] = e.mouseposition[] .- (4, 0)
        elseif idx < 150
            e.scroll[] = (0.0, -0.5)
        end 
    end
end

```

### DIFF
--- a/GLMakie/test/unit_tests.jl
+++ b/GLMakie/test/unit_tests.jl
@@ -465,5 +465,5 @@ end
     cam = ax.scene.camera
 
     @test robj.uniforms[:resolution][]     == screen.px_per_unit[] * cam.resolution[]
-    @test robj.uniforms[:projectionview][] == cam.projectionview[]
+    @test robj.uniforms[:projectionview][] == Mat4f(cam.projectionview[])
 end
--- a/src/basic_recipes/tooltip.jl
+++ b/src/basic_recipes/tooltip.jl
@@ -173,20 +173,20 @@ function plot!(p::Tooltip{<:Tuple{<:VecTypes}})
         scale!(mp, s, s, s)
 
         if placement === :left
-            translate!(mp, Vec3f(o[1] + w[1], o[2] + align * w[2], o[3]))
+            translate!(mp, Vec3f(o[1] + w[1], o[2] + align * w[2], o[3] - 1f0))
             rotate!(mp, qrotation(Vec3f(0,0,1), 0.5pi))
         elseif placement === :right
-            translate!(mp, Vec3f(o[1], o[2] + align * w[2], o[3]))
+            translate!(mp, Vec3f(o[1], o[2] + align * w[2], o[3] - 1f0))
             rotate!(mp, qrotation(Vec3f(0,0,1), -0.5pi))
         elseif placement in (:below, :down, :bottom)
-            translate!(mp, Vec3f(o[1] + align * w[1], o[2] + w[2], o[3]))
+            translate!(mp, Vec3f(o[1] + align * w[1], o[2] + w[2], o[3] - 1f0))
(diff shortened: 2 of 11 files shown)


Kevin-Mattheus-Moerman:
@ffreyer thanks! This looks nearly perfect. 

Not sure what it is called, but it would be good to turn off the auto-rescaling/zooming that normally happens when one rotates. When zoomed in to inspect the surface I can tell that it is doing that re-fitting/zooming a bit. Can that easily turned off? 

https://github.com/user-attachments/assets/3c86d9dc-9918-4eee-9898-394276ace5d4

Next I noticed that if I translate the scene and then rotate that the result doesn't make sense and there is odd clipping happening: 

https://github.com/user-attachments/assets/c5e49d8a-afc7-4b95-98cf-f9d7e717e872

jkrumbiegel:
> Can that easily turned off?

That's the `viewmode` which can be set to `:fit` to avoid the zooming https://docs.makie.org/stable/reference/blocks/axis3#viewmode

Kevin-Mattheus-Moerman:
> > Can that easily turned off?
> 
> That's the `viewmode` which can be set to `:fit` to avoid the zooming https://docs.makie.org/stable/reference/blocks/axis3#viewmode

Thanks @jkrumbiegel that works. However I wonder, for this "unhinged" style rotation/zooming, it probably makes sense to always have `viewmode` set to `:fit`. Like there is nothing to "fit" if the axis can completely extend outside of what we are looking at anyway.

Kevin-Mattheus-Moerman:
@ffreyer @jkrumbiegel how can the clipping I mention above (i.e. when translating significantly first followed by rotation) be avoided? Is that a simple setting?

ffreyer:
The near/far calculation needs to be updated for that. Iirc they act as multipliers `clip_plane_origin = eyeposition + mult * (lookat - eyeposition)` and should optimally place clip planes at the near and far end of the axis bbox

Kevin-Mattheus-Moerman:
@ffreyer, okay. I think after that clipping is fixed. It looks like this finally completed. Very exciting, this is a breath of fresh air for me once finished :)

PR #4303, opened 2024-09-02
[WIP] Try to fix contour memory leaks

### LINKED ISSUE #4302: Memory associated only with non-primitive plots is not garbage collected after empty!(axis)
- [x] what version of Makie are you running?  Makie 0.21.9 and GLMakie 0.10.9
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)  Yes
- [x] What platform + GPU are you on?  Kubuntu 24.04, Wayland, i5-4570, Mesa Intel HD Graphics 4600

The following program consumes all available memory, presumably because the large matrices are not being garbage collected. Note that `contour!` is used and that if it is changed to `heatmap!` then garbage collection works as expected and all is fine.

```julia
using GLMakie
fig::Figure = Figure()
ax::Makie.Axis = Axis(fig[1,1])
display(fig)
while true
    empty!(ax)
    big_matrix::Matrix{Float64} = zeros(10_000, 10_000)
    big_matrix[1,1] = 1
    contour!(ax, big_matrix)
end
```

asinghvi17:
# Description

Fixes #4302 

Specifically for `contour`, this PR associates all `lift` calls with the plot object, so that they can be correctly garbage collected.

## Type of change

### DIFF
--- a/src/basic_recipes/contours.jl
+++ b/src/basic_recipes/contours.jl
@@ -256,7 +256,7 @@ function plot!(plot::T) where T <: Union{Contour, Contour3d}
         transform_marker = false
     )
 
-    lift(scene.camera.projectionview, transformationmatrix(plot), scene.viewport,
+    lift(plot, scene.camera.projectionview, transformationmatrix(plot), scene.viewport,
             labels, labelcolor, labelformatter, lev_pos_col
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Thanks!

PR #4305, opened 2024-09-03
Fullbox

### LINKED ISSUE #2349: Add option to close an `Axis3`'s outline box

lukashappe:
# Description
Implemented option to close an Axis3's outline box. When the new field `fullbox = true`  for an Axis3-object the 3 remaining spines of the box are created in front of what is inside the space.
## Example:
```
using GLMakie
using FileIO

fig = Figure()
ax = Axis3(fig[1, 1], fullbox=true)

mesh!(ax, brain, color = :gray80)

fig
```
example

## Fixes  #2349 

 Add option to close an Axis3's outline box

## Type of change


src/makielayout/blocks/axis3d.jl
```diff
@@ -427,27 +427,60 @@ function add_gridlines_and_frames!(topscene, scene, ax, dim::Int, limits, tickno
...
+        to_topscene_big_z_2d.([p7, p8], Ref(scene))
     end
     colors = Observable{Any}()
     map!(vcat, colors, attr(:spinecolor_1), attr(:spinecolor_2), attr(:spinecolor_3))
-    framelines = linesegments!(topscene, framepoints, color = colors, linewidth = attr(:spinewidth),
+    framelines = linesegments!(topscene, framepoints_main, color = colors, linewidth = attr(:spinewidth),
         # transparency = true,
         visible = attr(:spinesvisible), inspectable = false)
-
+        if ax.fullbox[] == true
```
jkrumbiegel, line 457: You should always add the lines and turn visibility on and off with `fullbox`, because otherwise you cannot switch back and forth interactively. You can `&&` visibility with the `spinesvisible` attribute

lukashappe:
## Another question:
 When `set_theme!(theme_minimal())`  and `fullbox = true` are applied together (I mean, it doesn't really make sense to do so?!), what is the desired behavior?
### At the moment, the Result would be like this:
example1

jkrumbiegel:
The theme should set these new spines to invisible also.

Maybe fullbox is not the best name either because it doesn't say anything about the other spines, what about `front_spines` or so?

jkrumbiegel:
Hm now that I think about it, do we really need the separate Boolean keyword? Maybe it's enough to just have these three new visible attributes, and they're just false in all default themes.

lukashappe:
> Hm now that I think about it, do we really need the separate Boolean keyword? Maybe it's enough to just have these three new visible attributes, and they're just false in all default themes.

I see the point. Would you suggest creating a theme (called full_box for example?!) to set all spines visible then?

jkrumbiegel:
I don't think a theme is warranted for three booleans, if people like their plots this way they're going to call `set_theme!` or so once, it's not something you're likely to do manually on every plot.

SimonDanisch (approved):

PR #4310, opened 2024-09-03
show tooltip on NaN if nan_color is set

bjarthur:
title says it all.  specifically, for a heatmap in which some squares are NaN, if `nan_color` has been set (to other than the default `:transparent`), then show the tooltip.  otherwise, behave like before and don't show one for NaN values.

i suppose technically this is a breaking change?

does this need to be documented anywhere other than the CHANGELOG?

are there tests that can be written?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Show DataInspector tooltip on NaN values if `nan_color` has been set to other than `:transparent` [#4310](https://github.com/MakieOrg/Makie.jl/pull/4310)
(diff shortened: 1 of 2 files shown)


bjarthur:
could someone please review?

bjarthur:
here's some demo code to make reviewing easy:

```
julia> using GLMakie

julia> d = rand(3,3);

julia> d[2,2] = NaN;

julia> f,a,hm = heatmap(d)

julia> DataInspector();

julia> hm.nan_color[] = :fuchsia    # hover over the middle square before and after executing this line to see the difference
```

SimonDanisch:
Thank you! Merged in https://github.com/MakieOrg/Makie.jl/pull/4371

PR #4312, opened 2024-09-03
Fix absrect for select_rectangle

SimonDanisch:
Fixes https://discourse.julialang.org/t/glmakie-issue-with-select-rectangle/119008
Although I don't fully understand why xy isnt a tuple in that error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Ok, the mystery got solved, this was already fixed on master :D 
I think, I still want to merge the updated version, since i like the map version more.

Issue #3928, opened 2024-06-04
fontsize does not propagate into PolarAxis


pzabka:
The parameter fontsize does not propagate into PolarAxis

This works:
```julia
fig = Figure()
ax = PolarAxis(fig[1, 1]; rticklabelsize = 50, thetaticklabelsize = 50)
fig
```

This doesn't:
```julia
fig = Figure(fontsize = 50)
ax = PolarAxis(fig[1, 1])
fig
```

Tested with:
GLMakie v0.10.2 
CairoMakie v0.12.2

ffreyer:
Should be easy to fix by changing the relevant attributes to inherit from fontsize like Axis. See src/makielayout/types.jl

ffreyer:
Fixed by #4314

PR #4314, opened 2024-09-04
Added way for PolarAxis to pull fontsize attribute from Figure

NeunMonde:
# Description

Fixes issue #3928.

Adjusted the "rticklabelsize" and "thetaticklabelsize" properties in the "types.jl" file to fall back to inheriting the fontsize from the Figure, if the ticklabelsize is not defined.

## Type of change

## Checklist


src/makielayout/types.jl
```diff
@@ -1871,7 +1871,7 @@ end
         "The formatter for the `r` ticks"
         rtickformat = Makie.automatic
         "The fontsize of the `r` tick labels."
-        rticklabelsize::Float32 = inherit(scene, (:Axis, :xticklabelsize), 16)
+        rticklabelsize::Float32 = inherit(scene, (:Axis, :xticklabelsize), inherit(scene, :fontsize, 16))
```
asinghvi17, line 1874: This is a great idea!  Didn't think of that.

asinghvi17 (commented):
Looks good to me.  Do you want to add a reference image test to make sure this works?

To do so, simply add another `@reference_test` block at the end of this file.  https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/figures_and_makielayout.jl

asinghvi17 (approved):
Looks good to me.  Do you want to add a reference image test to make sure this works?

To do so, simply add another `@reference_test` block at the end of this file.  https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/figures_and_makielayout.jl

asinghvi17:
Reference image changes are because the default font size in the polar axis is now inheriting from the default 14, not 16 as it was when the recipe was written.  Given that this restores consistency with `Axis`, IMO this is for the better.

asinghvi17:
Could you also add a short entry in the CHANGELOG.md about the PR?  Thanks!

ReferenceTests/src/tests/figures_and_makielayout.jl
```diff
@@ -340,3 +340,15 @@ end
     translate!(vl, 1, 0, 0)
     f
 end
+
+@reference_test "PolarAxis fontsize from :Axis" begin
+    fig = Figure()
+    ax = PolarAxis(fig[1, 1], rticklabelsize = 35, thetaticklabelsize = 65)
+    fig
+end
```
asinghvi17, line 348: I don't think this is testing PolarAxis fontsize from Axis - to do that, you would do something like:
```suggestion
@reference_test "PolarAxis fontsize from :Axis" begin
    fig = Figure(; Axis = (; xticklabelsize = 35, yticklabelsize = 65))
    ax = PolarAxis(fig[1, 1])
    fig
end
```

and make sure it works that way!  There are already tests for the direct syntax.

NeunMonde: Would it be wise to add a reference test for setting the fontsize via theme aswell? And furthermore, would it be a good idea to move the test for direct syntax to the other tests so that they are grouped together?

ffreyer: Actually I think this would be better as unit tests. There is already a refimg test for decorations that should confirm that changing r-/thetaticklabelsize has the desired effect. So here you just need to confirm that the PolarAxis is correctly inheriting values. Something like
```julia
fig = Figure(; Axis = (; xticklabelsize = 35, yticklabelsize = 65))
ax = PolarAxis(fig[1, 1])
@test ax.thetaticklabelsize[] == 65
@test ax.rticklabelsize[] == 35
```
should be enough for that.

Btw, should we switch this around to be x -> theta and y -> r? Originally we had (r, theta) as the default argument order and matched that to (x, y), but now it's (theta, r).

asinghvi17: I think we may as well switch it then, so things stay consistent.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added option `label_position = :center` to place labels centered over each bar [#4274](https://github.com/MakieOrg/Makie.jl/pull/4274).
 - Fixed Boundserror in clipped multicolor lines in CairoMakie [#4313](https://github.com/MakieOrg/Makie.jl/pull/4313)
 - Fix float precision based assertions error in GLMakie.volume [#4311](https://github.com/MakieOrg/Makie.jl/pull/4311)
+- Added way for `PolarAxis` to pull fontsize attribute from `Figure()` [#4314](https://github.com/MakieOrg/Makie.jl/pull/4314)
```
asinghvi17, line 13:
```suggestion
- `PolarAxis` fontsize now defaults to global figure `fontsize` in the absence of specific `Axis` theming [#4314](https://github.com/MakieOrg/Makie.jl/pull/4314)
```

Just to make the verbiage more specific :)

test/PolarAxis.jl
```diff
@@ -149,4 +149,27 @@
...
+        @test ax.rticklabelsize[] == 65
+    end
+
+    @testset "PolarAxis fontsize from Theme()" begin
+        fig = Figure()
+        ax = PolarAxis(fig[1, 1])
+        fontsize_theme = Theme(fontsize = 10)
+        set_theme!(fontsize_theme)
+        @test ax.rticklabelsize[] == 10
+        @test ax.thetaticklabelsize[] == 10
```
asinghvi17, line 173: I don't think this should work - the figure is constructed before the theme is set, meaning that the theme doesn't apply to the figure.  Plus, `set_theme!` sets the global theme which will affect every plot after this test.

Instead, you could use `with_theme`:
```suggestion
	fontsize_theme = Theme(fontsize = 10)
	with_theme(fontsize_theme) do
        	fig = Figure()
        	ax = PolarAxis(fig[1, 1])
        	@test ax.rticklabelsize[] == 10
        	@test ax.thetaticklabelsize[] == 10
	end

CHANGELOG.md
NeunMonde: Can I resolve conversations like this when making a change that should fix the problems mentioned or is that something usually reserved for the reviewers?

asinghvi17: Feel free to resolve if the comment is addressed!  If we feel it's not, we can always reopen the comment.

NeunMonde: Alright, thank you!

ffreyer (approved):
Refimages just fail due to different default fontsize (inherited)

Issue #215, opened 2018-11-07
Cannot pass keyword to custom recipe


wsrinin:
Hello,

I think I sort of get the basic gist of how to create my own custom recipe as suggested in the new tutorial. However, I cannot seem to pass the keyword arguments down. That is

Let says I have
```julia
function plot!(p::Plot(MDSimulation); markersize = 1)
         simulation = to_value(p[1])
        scatter!(simulation.positions, color = :blue, markersize = markersize )
#more stuff
end
```
I seem that the value of markersize is always set to 1 eventhough I change value of the keyword. Now if I naively implement

```julia
function visualize(simulation::MDSimulation; markersize = 1)
   scatter!(simulation.positions, color = :blue, markersize = markersize )
   #more stuff
end
```
everything works fine. I have two questions

1. How can I make the kwarg pass down to the function body?
2. Is it necessary to use the syntax suggested in the tutorial? Can I just use the naive version?

SimonDanisch:
> 2 Is it necessary to use the syntax suggested in the tutorial? Can I just use the naive version?

Yes, you need to follow the syntax in the tutorial!

PR #4317, opened 2024-09-04
Introduce heatmap shader

SimonDanisch:
Which resamples a large image on zoom etc.
I first wanted to add it to `datashader(matrix)`, but realized we need really different attributes, so now it's `heatmapshader(matrix)`.
I also thought about adding it to heatmap directly, but that seemed to complicated, since then it would need to be implemented in the backends due to it being an atomic plot, without having a type to dispatch on.
This allows to plot really large heatmaps speedily, and may even work for out of RAM data (not tested, would need e.g. diskarrays as input).

### Changes in this PR:

* Resampling Updating scheme for as little data transfer as possible
* Implement interpolation based resampling scheme
* Implement a simple Gaussian Pyramid for higher quality resampling
* Documentation & tests

###  Since this was heavily tested with WGLMakie under non optimal connections, this PR includes a few more fixes:

* Debug why scroll is still really slow (It was DataInspector registering a very expensive event on scroll)
* Debugging why async/throttled data inspector is still slowing down scrolling (Bonito needs its own channel and message processing [#215](https://github.com/SimonDanisch/Bonito.jl/pull/215), to not hang on async data inspector calls (which call Bonito.evaljs_value, which sends a message to the browser and needs to wait on a reply))


jkrumbiegel:
Would it be better if it didn't have heatmap in its name then? I wonder how confusing it would be to have a fast and slow heatmap primitive

SimonDanisch:
Hm isnt that just hiding the obvious? I mean the fact that we have then another way of plotting a heatmap won't go away with another name, will it?
Maybe we could also do:
`heatmap(Resample(data))` or something like that...

jkrumbiegel:
Yeah ideal would be if this could be a part of heatmap, I agree.

SimonDanisch:
They don't, they differ from datashader ;)

asinghvi17:
The only unique attribute looks like `max_resolution = 1024`, so could that just be a kwarg to `Resample` instead of heatmap?

PR #4327, opened 2024-09-06
use InverseFunctions to invert functions

aplavin:
Getting both at once: simplify the code and support much more functions.
InverseFunctions.jl is very light, and quite likely to be installed in many environments anyways – it's a transitive dependency in thousands of packages.


SimonDanisch:
Thanks!

Issue #2694, opened 2023-02-18
wireframe decomposition makes duplicate interior edges


jkrumbiegel:
```julia
f, ax, w = wireframe(1:3, 1:3, ones(3, 3))
segs = w.plots[][1][]
tups = tuple.(segs[1:2:end], segs[2:2:end])
s1 = Set(tups)
s2 = Set(reverse.(tups))
s3 = intersect(s1, s2)

linesegments(collect(s3))
```

<img width="712" alt="Screenshot 2023-02-18 at 16 31 59" src="https://user-images.githubusercontent.com/22495855/219874387-887973e4-ecae-4b5e-ab2c-117b3f3a5e78.png">

ffreyer:
Maybe something for https://github.com/JuliaGeometry/GeometryBasics.jl/pull/173

I first thought this might be too expensive to just do since it sounds like an O(N^2) problem (for each line check with every other line) but this should be doable with a hashmap/Dict in O(N log(N)). E.g.
```julia
for face in faces
    for (i, j) in edge(face)
        i, j = minmax(i, j)
        hashmap[(i, j)] = Line(positions[i], positions[j])
    end
end
```

jkrumbiegel:
It would be easier to avoid this for rectangular meshes, so we could at least cover surfaces efficiently

PR #4332, opened 2024-09-06
Actually use `linestyle` in `triplot`

DanielVandH:
# Description

Noticed that `triplot` wasn't using the `linestyle` argument

## Type of change

## Checklist


ffreyer:
I tweaked another refimg instead of adding a new one to save a bit of CI time.

I also noticed that some lines are overlapping. I guess that's expected though?

DanielVandH:
> I also noticed that some lines are overlapping. I guess that's expected though?

Is there a way to view the new images from the CI without going into the code itself? I can double check.

ffreyer:
It's easier to see when you make the plot locally an rescale the window. I think all the central lines are duplicated. In this example the one to the top left appears solid, along with some others:

image

In your example the one straight down in the center and two diagonal ones connected to either endpoint of that line show it.

ffreyer:
Actually maybe that's just poly?

ffreyer:
Yea nvm, that's just #2694. I'll merge this once ci passes

PR #4338, opened 2024-09-09
support images with reversed axes

aplavin:
# Description

The whole pipeline almost-supported them already, here I just make a few minor fixes where the assumption of increasing values was baked in.

Major update to https://github.com/MakieOrg/Makie.jl/pull/3989 following recent image/heatmap changes.

## Type of change

## Checklist


ffreyer:
I fixed heatmap conversions, squashed the two refimages and added `surface` as well:
image

Also did a quick grep search for `to_linspace` in the Makie repo - it's only used in conversions.jl. So these changes shouldn't affect any odd code somewhere else. Within conversions.jl it also affects a PointBased conversion, but I think that's fine. Specifically `::Interval, ::Interval, ::Matrix` can now flip x/y if you supply reverse intervals.

ffreyer (approved):
`transform_func` seems to work fine with this too. Briefly checked with `p.transformation.transform_func[] = -`

jkrumbiegel:
Why do the corners look wonky here?

image

ffreyer:
Because that "cell" is bigger:
```julia
img = [
  2 0 0 3; 
  0 0 0 0; 
  1 1 0 0; 
  1 1 0 4
]
```

ffreyer:
Oh the corners - that's just FXAA doing its thing. It happens in other plots too

jkrumbiegel:
Hm but why is it so asymmetrical? I would expect all corners of one cell to show the same aliasing artifacts, but for example the top-right corner goes up and the top-left down.

image

jkrumbiegel:
Why does the green tile have pixels in row 3 at the top right (this must mean it intersects those pixels with its fragment) but not at the top left (where it doesn't even have fully opaque pixels in row 4)? That must mean the whole fragment is crooked, no?

SimonDanisch:
Oh, this is a regression from the plot attribute refactor I think, we always had `fxaa=false` for heatmap/image!

SimonDanisch:
>This is happening for the surface example

I just checked, you just dont see it because of `tightlimits` but heatmap/image actually have `fxaa=true` as a default now ...
Probably should be addressed in another PR.

PR #4343, opened 2024-09-10
Fix exception when WGLMakie is initalised after element removal

### LINKED ISSUE #4342: Exception thrown when WGLMakie containing element is removed
If we remove the element containing WGLMakie before its initailsed there is an exception:

```
using HTTP
using Bonito: App
using Makie
using WGLMakie


function main()
    HTTP.listen() do http::HTTP.Stream
        HTTP.setstatus(http, 200)
        HTTP.startwrite(http)
        write(http, "<html><body>")
        write(http, "<h1>Bonito element removal</h1>")
        app = App() do
            sleep(2)
            fig = Figure()
            return fig
        end
        io = IOBuffer()
        show(io, "text/html", app)
        body = take!(io)
        write(http, "<div id=\"bonito\">")
        write(http, body)
        write(http, "</div>")
        write(http, "<script>setTimeout(() => {document.getElementById('bonito').remove()}, 1000);</script>")
        write(http, "</body></html>")
    end
end

if abspath(PROGRAM_FILE) == @__FILE__
    main()
end
```

The exception is:

```
An exception was thrown in JS: TypeError: Cannot read properties of null (reading 'getContext')
Additional message: error initializing scene
Stack trace:
    TypeError: Cannot read properties of null (reading 'getContext')
        at threejs_module (http://localhost:9385/assets/435ec4531f8729b0dfad48036f91f9e1eab0aa39-wglmakie.bundled.js:23033:26)
        at Module.create_scene (http://localhost:9385/assets/435ec4531f8729b0dfad48036f91f9e1eab0aa39-wglmakie.bundled.js:23099:22)
        at eval (eval at <anonymous> (http://localhost:9385/assets/f1a88e34853c38ba45791ee4149df30a0c07525a-Bonito.bundled.js:3561:27), <anonymous>:5:30)
┌ Warning: error while processing received msg
│   exception =
│    Three object should be ready after init, but isn't - connection interrupted? Session: Session{Bonito.WebSocketConnection}:
│      id: 9d871a24-7426-425e-98a3-946845037797
│      parent: Nothing
│      connection: open
│      isready: true
│      asset_server: Bonito.ChildAssetServer
│      queued messages: 0
│    , initialized: false
```

But there could be legitimate reasons to have this happen, e.g. plot shown in a model element that the user immediately closes.

frankier:
# Description

Fixes #4342

This simple fix just checks if the element is still there before trying to initalise the canvas and bails if it's not there.

I can add a test/changelog entry if/when you agree this approach is ok.

## Type of change

## Checklist


SimonDanisch:
Sounds like a good idea!
It should not just return though, but instead push a value to init and send an error:
https://github.com/MakieOrg/Makie.jl/blob/5ccd0a169bf09f34a1acab2a8ddc89d5e59509d6/WGLMakie/src/three_plot.jl#L67-L71
I actually just started pushing the error to `done_init` in: https://github.com/MakieOrg/Makie.jl/pull/4317/files#diff-468ae987abf57692bfa58a73c15591f64a022fba56dd8a4b7fdee38d6a392970
So, after that got merged, it should likely also push a good error instead ;)

SimonDanisch:
I think this fix is fine, if it solves your problems ;)

frankier:
Okay great! Let's go with this for now, and if it turns out different behaviours are needed in different circumstances that can be tackled later.

PR #4354, opened 2024-09-13
Accessing created Blocks/Plots for SpecApi objects

SimonDanisch:
This adds a virtual `then` field to the specapi objects to interact with the created axis (plot/gridlayout?) object:
```julia
import Makie.SpecApi as S 
ax =  S.Axis(...)
ax.then() do actual_axis_object
    hidedecorations!(actual_axis_object)
    return on(events(actual_axis_object).mouseposition) do mp 
        println("mouse: $(mp)")
    end
end
```

The problem is, that we need to clean up any changes for when we re-use axes, and then re-apply the `then` functions.
Right now, any observable `ObserverFunction` needs to be returned, so e.g. two `on` invocations need to be done like this:
```julia
ax.then() do ax
    obs1 = on(f1, events(ax).keyboardbutton)
    obs2 = on(f2, events(ax).mousebutton)
    return [obs1, obs2]
end
```
`linkaxis!` is also supported, since we just clear the axis links when re-using any axis, so one can do:
```julia
ax_spec1 = S.Axis(...)
ax_spec2 = S.Axis(...)
ax_spec1 .then(ax1-> ax_spec2 .then(ax2-> linkaxes!(ax1, ax2)))
```

It would be nice if we could make this stricter, ideally it would be impossible to leak any mutation from `then`.
But I'm not sure how we could do that, without making the API a pain to use or much harder to implement.

I'm happy to hear alternative approaches for how to register e.g. interactions with axis objects etc.


SimonDanisch:
Some Benchmarks:
```julia
using WGLMakie, Bonito
import Makie.SpecApi as S

function to_spec(data)
    axes = map(enumerate(data)) do (i, data)
        S.Axis(plots=[S.Heatmap(data)], title="Axis: $(i)")
    end
    return S.GridLayout(permutedims(axes))
end
data = [rand(Float32, 10, 10) for i in 1:3]

x = Observable(to_spec(data[1:1]))

f, ax = plot(x)
display(f)
msgs1, msg1 = Bonito.collect_messages() do
    x[] = to_spec(data[1:3])
end;
msgs2, msg2 = Bonito.collect_messages() do
    x[] = to_spec(data[1:1])
end;
msgs3, msg3 = Bonito.collect_messages() do
    x[] = to_spec(data[1:3])
end;
msgs4, msg4 = Bonito.collect_messages() do
    x[] = to_spec(reverse(data[1:3]))
end;
```
## PR
```
Send 560 messages with a total size of 467.349 KiB
Send 68 messages with a total size of 9.737 KiB
Send 72 messages with a total size of 10.046 KiB # <---
Send 19 messages with a total size of 2.345 KiB
```
## Master 
```
Send 555 messages with a total size of 459.931 KiB
Send 57 messages with a total size of 8.833 KiB
Send 555 messages with a total size of 461.491 KiB
Send 4 messages with a total size of 1.186 KiB
```
The big improvement is obviously, when we go from 1 -> 3 axes that have been re-created previously.
The above is not yet perfect in all cases, since for e.g. reverse, it triggers some more changes, that we still need to get rid of.

SimonDanisch:
Another Benchmark for swapping axes:
```julia
using WGLMakie, Bonito, BenchmarkTools
import Makie.SpecApi as S
WGLMakie.activate!()
ax1 = S.Axis(; title="Title 1")
ax2 = S.Axis(; title="Title 2")
ax3 = S.Axis(; title="Title 3")
axis = [ax1, ax2, ax3]
gl = S.GridLayout(axis)
fig, _, pl = plot(gl)
display(fig)
msgs, msg = Bonito.collect_messages() do
    pl[1] = S.GridLayout(reverse!(axis))
end
@btime pl[1] = S.GridLayout(reverse!(axis))
```
### PR 
```
"Send 6 messages with a total size of 1024 bytes"
329.500 μs (2533 allocations: 152.05 KiB)
```
### Master 
```
"Send 14 messages with a total size of 2.472 KiB"
491.200 μs (2988 allocations: 182.78 KiB)
```

Issue #4344, opened 2024-09-10
`voronoiplot` default colours depend too much on the generator order


DanielVandH:
Consider

```julia
using DelaunayTriangulation, CairoMakie
points = [(0.005, 0.0), (1.0, 0.0), (1.01, 1.0), (-0.1, 1.0), (0.5, 0.5)]
w1 = [0.0, 0.0, 0.0, 0.0, -0.3]
w2 = [0.0, 0.0, 0.0, 0.0, -0.6]
fig, ax, sc = voronoiplot(voronoi(triangulate(points; weights = w1)))
ax2 = Axis(fig[1, 2])
voronoiplot!(ax2, voronoi(triangulate(points; weights = w2)))
fig
```
![image](https://github.com/user-attachments/assets/f38243b8-89a7-4692-84ef-5d34fa620df8)

Ideally the colour schemes should be the same, except the yellow tile in the middle would be gone. This discontinuity is because
```julia-repl
julia> each_generator(voronoi(triangulate(points; weights = w1))) # [5, 4, 2, 3, 1]
KeySet for a Dict{Int64, Tuple{Float64, Float64}} with 5 entries. Keys:
  5
  4
  2
  3
  1

julia> each_generator(voronoi(triangulate(points; weights = w2))) # [4, 2, 3, 1]
KeySet for a Dict{Int64, Tuple{Float64, Float64}} with 4 entries. Keys:
  4
  2
  3
  1
```

and my definition for the default colors here 

https://github.com/MakieOrg/Makie.jl/blob/ae8c1a455b20586febab682bc709a4ddd60d935f/src/basic_recipes/voronoiplot.jl#L176

is probably bad. Maybe it should be 

```julia
cs = [i for i in sort(collect(DelTri.each_generator(vorn)))] # or just cs = sort(collect(DelTri.each_generator(vorn)))
```

or (same as the above except without `sort`/`collect` since `each_point_index` is not derived from a `Dict`)

```julia
cs = [i for i in DelTri.each_point_index(DelTri.get_triangulation(vorn)) if DelTri.has_polygon(vorn, i)]
```

PR #4357, opened 2024-09-13
Change automatic color generation for Voronoiplot

DanielVandH:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/4344 by using a consistent method for generating colors. Using this approach the colors will be consistent between plots even if generators were removed or added. The old approach wasn't optimal since the `Set` defining `each_generator(vorn)` isn't ordered consistently across tessellations if the generator set changed.

## Type of change

## Checklist


DanielVandH:
This PR breaks

Voronoiplot for a centroidal tessellation with an automatic colormap.png
Voronoiplot after adding points.png                                                           
Voronoiplot for a tessellation with a custom bounding box.png        

because of the new color generation method. I've checked them myself and nothing else is broken, it's just the colors. I don't see a consistent way to define the colors that is consistent with the colors generated by the previous method, so is just regenerating the images sufficient?

ffreyer:
I think it's fine that the automatically chosen colors change, but it would mean that we consider this breaking and delay it until the next breaking release.

jkrumbiegel (approved):

SimonDanisch:
I'm wondering if we can tag this as a patch version, considering that it's not that close to core Makie and the colors are somewhat arbitrary.
I guess I'd let @DanielVandH decide this.

DanielVandH:
It would be nice to get it in earlier, but I don't mind waiting for another breaking release if you prefer . I would agree with you though that I don't think it's so breaking.

PR #4372, opened 2024-09-17
Add option `strip_zero` for MultiplesTicks

### LINKED ISSUE #4370: MultiplesTicks when multiples is 0
### Feature description

This may be an issue of personal taste and not very important, but right now using for example `MultiplesTicks(5, pi, "π")` leads to the ticks being `["−2π", "−1π", "0π", "1π", "2π"]`. 

I'd prefer to have just `"0"` instead of `"0π"`, which could be achievable for example with changing
https://github.com/MakieOrg/Makie.jl/blob/c463049102e775d4c31c4b891e9ac32adbca9891/src/makielayout/lineaxis.jl#L665-L671
to return something like
```julia
multiples .* m.multiple, map(x -> x[1] != '0' ? x : "0", showoff_minus(multiples) .* m.suffix)
```

Not sure if it would be worth it and whether it would be preferable for other people though.

tuncbkose:
# Description

Fixes #4370

Adds a new option `strip_zero::Bool` to `MultiplesTicks` and related logic to `get_ticks` that strips labels of the form `0x` to `0`.

## Type of change

## Checklist


tuncbkose:
I'm not sure how to add tests for these, even after reading `Contributing.md` etc. If someone gives me some instructions, I'd be happy to add some.

src/makielayout/lineaxis.jl
```diff
@@ -667,7 +667,13 @@ function get_ticks(m::MultiplesTicks, any_scale, ::Automatic, vmin, vmax)
     dvmax = vmax / m.multiple
     multiples = Makie.get_tickvalues(LinearTicks(m.n_ideal), dvmin, dvmax)
 
-    multiples .* m.multiple, showoff_minus(multiples) .* m.suffix
+    locs = multiples .* m.multiple
+    labs = showoff_minus(multiples) .* m.suffix
+    if m.strip_zero
+        labs = map(x -> x[1] != '0' ? x : "0", labs)
```
jkrumbiegel, line 673: I would check the numerical value, not the first char of the label. I don't really see lots of cases in which the char could break but it's simpler to go by the number.

src/makielayout/types.jl
```diff
@@ -66,13 +66,20 @@ that are multiples of pi, printed like "1π", "2π", etc.:
...
+will be set to "0".
 """
 struct MultiplesTicks
     n_ideal::Int
     multiple::Float64
     suffix::String
+    strip_zero::Bool
 end
 
+MultiplesTicks(n_ideal, multiple, suffix) = MultiplesTicks(n_ideal, multiple, suffix, false)
```
jkrumbiegel, line 81: This should get a keyword arg with default. At some number of arguments it's clearer to use keywords for the less common options. Then a reader of code has an easier time than with just a true or false

jkrumbiegel (changes requested):

jkrumbiegel (approved):

PR #4150, opened 2024-08-19
Refactor (GLMakie) render order (experimental)

ffreyer:
# Description

## Goals

The goal of this pr is to render scenes one by one in a depth-first order (render scene, then go deeper). This aims to:
1. make `scene.clear = true` work again (clear a section of the screen for the clearing scene)
2. fix/remove execution order dependence of rendering
3. provide a standard rendering order
4. allow layering based on scene order, e.g. allow scenes to act as backgrounds/overlays to another independent of plot/render object depth values (probably also separate `scene.children` into normal scenes and overlay scenes to make this usable)
5. maybe provide a transparent intermediate render buffer for saving transparent images, #3652
6. prepare for per-scene post-processing pipelines, which I hope will improve quality and perhaps also performance, and enable for more post-processors to exist in the future.

### Related Issues

Related issues:
- fixes issue #2650 (Axis3 ticks in front of menu)
- https://discourse.julialang.org/t/glmakie-overlay-plots-in-animated-3d-scene/116750
- #2995
- #604
- fixes #4153 (Box only if Box is used before LScene)

#### Fixed unreported (?) issues:

<details>
<summary>Empty scenes do not render if added after display</summary>

```julia
scene = Scene(backgroundcolor = :white, clear = true)
screen = display(scene)
s1 = Scene(scene, Rect2f(100, 100, 400, 300), backgroundcolor = :cyan, clear = true);
s2 = Scene(scene, Rect2f(50, 150, 100, 100), backgroundcolor = :purple, clear = true);
```

| master | pr |
| --- | --- |
| image | Screenshot 2024-08-19 150650 |

</details>

<details>

<summary>Scene order depends on (plot) insertion order</summary>

```julia
scene = Scene(backgroundcolor = :white, clear = true)
screen = display(scene)
s1 = Scene(scene, Rect2f(100, 100, 400, 300), backgroundcolor = :cyan, clear = true);
s2 = Scene(scene, Rect2f(50, 150, 100, 100), backgroundcolor = :purple, clear = true);
scatter!(s2, Point2f(0)) # trigger scene 2 insert
scatter!(s1, Point2f(0)) # trigger scene 1 insert
```

| master | pr (skipped plot) |
| --- | --- |
| image | Screenshot 2024-08-19 150650 |

</details>

## Changes

This pr mainly changes the scene/plot/renderobject storage infrastructure to more closely represent the scene tree from Makie and allow for rendering to follow a depth-first order.

<details>

<summary> Changes in more detail </summary>

[...]

PR #4380, opened 2024-09-18
add tooltips to blocks

bjarthur:
Adds tooltips to Textboxes which appear on hover.

Screenshot 2024-09-18 at 6 26 13 PM

sadly, despite specifying `overdraw=true` to `tooltip!`, they sometimes appear behind other blocks.  see for yourself by executing the code below and hovering over `tb21`.  will this be fixed by https://github.com/MakieOrg/Makie.jl/pull/4150?  is there a workaround in the meantime?

```
using GLMakie
f = Figure()
tb11 = Textbox(f[1,1], tooltip_text="hi\nthere", tooltip_enable=true, tooltip_placement=:below)
tb21 = Textbox(f[2,1], tooltip_text="okay\nhere", tooltip_enable=true, tooltip_placement=:below)
m31 = Menu(f[3,1], options = ["a", "b", "c"], default = "a")
m32 = Menu(f[3,2], options = ["a", "b", "c"], default = "a")
tb22 = Textbox(f[2,2], tooltip_text="okay\nhere", tooltip_enable=true, tooltip_placement=:below)
tb12 = Textbox(f[1,2], tooltip_text="hi\nthere", tooltip_enable=true, tooltip_placement=:below)
```

## Checklist

### DIFF
--- a/src/basic_recipes/tooltip.jl
+++ b/src/basic_recipes/tooltip.jl
@@ -103,6 +103,8 @@ function plot!(p::Tooltip{<:Tuple{<:VecTypes}})
             return Vec2f(l - align * (l + r), -o - t - ts)
         elseif placement in (:above, :up, :top)
             return Vec2f(l - align * (l + r),  o + b + ts)
+        elseif placement === :center
+            return Vec2f(l - align * (l + r),  b - align * (b + t))
         else
             @error "Tooltip placement $placement invalid. Assuming :above"
             return Vec2f(0, o + b + ts)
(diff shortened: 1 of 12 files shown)


jkrumbiegel:
My first reaction is that it feels like this should maybe be something that is implemented between `DataInspector` and `Block`s in general, instead of adding these attributes to `Textbox` itself? In principle, anything should be "tooltip-able", not just textbox, but it would be a lot of work adding this to every `Block` manually. @ffreyer do you think `DataInspector` would lend itself to this at all? I haven't really used it myself.

ffreyer:
DataInspector doesn't know about Blocks and blocks aren't reachable from a plot afaik. We could have blocks attach tooltip information to a scene and have DataInspector go up to `parent_scene(picked_plot)` instead. That would take priority over plto tooltips though. Not sure if that's a good idea...

bjarthur:
superseded by https://github.com/MakieOrg/Makie.jl/pull/4451

PR #4389, opened 2024-09-20
`viewmode = :free` prototype for Axis3 controls

ffreyer:
# Description

Another prototype for the suggestions from #4131 and #4294.

This one adds a new `viewmode = :free` which skips the complications of making the other viewmodes work with zooming & translation. Iirc that was the reason why I pulled in #3550 which is not needed here. 

The clipping issues from #4294 should be fixed here. Rotation should be centered on the axis bbox like in the matlab example from https://github.com/MakieOrg/Makie.jl/pull/4131#issuecomment-2320703045, though rotation is still restricted. ~~I can look into adding an option to unlock that too.~~ That would require managing the axis orientation differently, as two angles aren't enough to specify both the forward direction and up direction.

Main changes;
- add `ax.zoom_mult` and `ax.lookat[]` for passing information from Interactions to camera matrices
- move the basis transformation part of the view matrix to `ax.scene.model` so that rotations are centered on the axis bbox rather than lookat (center of 3D scene)
- move axis decorations to `ax.scene` to get the same viewport clipping as Axis3 content

TODO:
- [x] test various different combinations of viewmode, x/y/z-reversed, aspect and perspectiveness
- [x] refimg tests
- [ ] interaction tests
- [ ] docs

## Example

(Changes from #4294 are different scroll direction to zoom out, different filename and `ax1.viewmode[] = :free` instead of `ax1.zoommode[] = :cursor`.

```julia
using GLMakie
using FileIO

begin
    fig = Figure(size=(600,600))
    ax1.viewmode[] = :free
    fig
    record(fig, "camera_interaction.mp4", 1:150, fps = 30) do idx
        e = events(fig)
        if idx < 30
            e.mouseposition[] = (300, 300)
            e.scroll[] = (0.0, -0.5)
        elseif idx < 60
            if e.mousebutton[] != Makie.MouseButtonEvent(Mouse.right, Mouse.press)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.right, Mouse.press)
            end
            e.mouseposition[] = e.mouseposition[] .+ 3
        elseif idx < 120
            if e.mousebutton[] != Makie.MouseButtonEvent(Mouse.left, Mouse.press)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.right, Mouse.release)
                e.mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
            end
            e.mouseposition[] = e.mouseposition[] .- (4, 0)
        elseif idx < 150
            e.scroll[] = (0.0, 0.5)
        end 
    end
end
```

https://github.com/user-attachments/assets/7dbffa15-6cf3-42e9-94b0-8be818e9a304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,14 @@
 ## [Unreleased]
 
 - Added translation, zoom and limit reset interactions to Axis3 [4131](https://github.com/MakieOrg/Makie.jl/pull/4131)
+
+## [0.21.12] - 2024-09-28
+
+- Fix NaN handling in WGLMakie [#4282](https://github.com/MakieOrg/Makie.jl/pull/4282).
 - Show DataInspector tooltip on NaN values if `nan_color` has been set to other than `:transparent` [#4310](https://github.com/MakieOrg/Makie.jl/pull/4310)
 - Fix `linestyle` not being used in `triplot` [#4332](https://github.com/MakieOrg/Makie.jl/pull/4332)
+- Fix voxel clipping not being based on voxel centers [#4397](https://github.com/MakieOrg/Makie.jl/pull/4397)
+- Parsing `Q` and `q` commands in svg paths with `BezierPath` is now supported [#4413](https://github.com/MakieOrg/Makie.jl/pull/4413)
 
 ## [0.21.11] - 2024-09-13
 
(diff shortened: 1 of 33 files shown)


ffreyer:
I reworked how protrusions are handled so that they are not included in `ax.scene.viewport`, but in the projection matrix instead. That way all the axis content clips on the 2D area without protrusions, while the (initial) sizing still considers them.

## Added refimg test:

This shows the scene viewport in red/orange with rect lines showing the area with an without protrusions. Each groups varies `aspect = (:data, :equal, (1.2, 0.8, 1.0))` and `xreversed = (true, false)`. 

Screenshot 2024-09-28 021609

On master this is pretty close to the same, except for scene viewport, 2D clipping and x/y/z reversed being broken with `aspect = :data`:
Screenshot 2024-09-28 015910

<details>

<summary>Code</summary>

```julia
using FileIO, GLMakie
begin
    fig = Figure(size = (1200, 900))

    protrusions = (40, 30, 20, 10)
    perspectiveness = 0.0
    m = load(Makie.assetpath("cat.obj"))
    cs = 1:length(Makie.coordinates(m))
    
    for (bx, by, viewmode) in [(1,1,:fit), (1,2,:fitzoom), (2,1,:free), (2,2,:stretch)]
        gl = GridLayout(fig[by, bx])
        Label(gl[0, 1:3], "viewmode = $viewmode")
        for (x, aspect) in enumerate((:data, :equal, (1.2, 0.8, 1.0)))
            for (y, rev) in enumerate((true, false))
                ax = Axis3(gl[y, x], viewmode = viewmode, xreversed = rev, aspect = aspect, 
                    protrusions = protrusions, perspectiveness = perspectiveness)
                mesh!(ax, m, color = cs)

                ax.scene.clear = true
                ax.scene.backgroundcolor[] = RGBAf(1,0.8,0.6, 1.0)
                lines!(fig.scene, ax.layoutobservables.computedbbox, color = (:blue, 0.5), linestyle = :dash)
                fullarea = lift(ax.layoutobservables.computedbbox, ax.layoutobservables.protrusions) do bbox, prot
                    mini = minimum(bbox) - Vec2(prot.left, prot.bottom)
                    maxi = maximum(bbox) + Vec2(prot.right, prot.top)
                    return Rect2f(mini, maxi - mini)
                end
                lines!(fig.scene, fullarea, color = :red)
            end
        end
    end

    fig
end
```

</details>

PR #4319, opened 2024-09-04
Updates for GeometryBasics refactor

ffreyer:
# Description

#### Requires:

- https://github.com/JuliaGeometry/GeometryBasics.jl/pull/219
- https://github.com/JuliaIO/MeshIO.jl/pull/97
- https://github.com/SimonDanisch/ShaderAbstractions.jl/pull/28
- https://github.com/JuliaGraphics/RadeonProRender.jl/pull/22

Version bumps:
- https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/pull/90
- https://github.com/jkrumbiegel/GridLayoutBase.jl/pull/61
- https://github.com/Kolaru/MathTeXEngine.jl/pull/125
- https://github.com/JuliaGeometry/Packing.jl/pull/12
- https://github.com/JuliaGraphics/RadeonProRender.jl/pull/24

This is still in prototyping stage as I'm looking what things GeometryBasics is missing/should do.

#### TODO:

- [x] Makie
  - [x] compiles
  - [x] tests 
- [x] get GLMakie working 
  - [x] compile
  - [x] mesh(rect3d)
  - [x] meshscatter()
  - [x] poly(polygon)
  - [x] mesh(cat, color = texture) (with https://github.com/JuliaIO/MeshIO.jl/pull/96)
  - [x] tests pass
- [x] WGLMakie
  - [x] compiles
  - [x] tests pass
- [x] CairoMakie
  - [x] compiles
  - [x] tests pass
- [x] RPRMakie
  - [x] compiles
  - [x] tests pass 

All RPRMakie examples run with https://github.com/JuliaGraphics/RadeonProRender.jl/pull/22 except for:
- datashader-rpr (the code is broken regardless of this pr)
- material_x.jl (crashes julia)
- sea_cables.jl (fails to download resources)

I'm not familiar enough with RPRMakie to say they look correct with confidence, but most look good to me. I'm not sure what to make of `eat_topographie_sphere.jl` (that's not a sphere?) and "volume.jl", they might be wrong.

#### Test failures:

- FigureSpec step 5: This pr somehow generates a more transparent marker than the refimg, which is barely transparent. The code uses `(:red, 0.5)`, so this seems more correct to me here. I'm just confused why it's different here at all
- Scatter with FastPixel: red outline color looks more intense in 3D (maybe color change, maybe thickness change, maybe both
- Triplot of a constained...: a line shifted from one diagonal to the other
- ~~shading of catmesh changed, seams visible~~ fixed
[...]

PR #4397, opened 2024-09-22
Cleanup voxel clipping

ffreyer:
# Description

My intention for voxel clipping was to make it work based on voxel centers. I didn't bother with that in CairoMakie initially, but updated it in #4319 since mesh based clipping started looked pretty bad. After that there was still some discrepancy between the GL backends and CairoMakie. Turns out I was using voxel origins instead of centers. This pr fixes that and pulls in the CairoMakie changes. It also adjusts the test to avoid voxel centers sitting directly on the clipping plane, since that results in float precision based differences between backends.

This should make the voxel clip plane test fail, but be consistent between all backends.

## Type of change

- [x] Bug fix as it fixes unintended clipping behaviour

## Checklist


ffreyer:
Test failures are the expected changes to voxel clipping: 

## Before
image

## After
image

SimonDanisch (approved):

PR #4399, opened 2024-09-22
Improve and use `shared_attributes` for attribute passthrough

### LINKED ISSUE #2206: `attributes_from` breaking
It looks like `attributes_from` didn't make it to `0.17`
I was using it to forward attributes between recipes like here: https://github.com/JuliaPlots/Makie.jl/issues/1019#issuecomment-1059998192
Is there a recommended alternative ?

### LINKED ISSUE #2733: Toggling scatterlines
Currently, `scatterline`s and also `abline`s aren't able to be toggled, although scatters and lines can, and they don't have the visible property in the attributes.
Is it possible to enable it by including it in the recipe or is there a reason this isn't easy?

### LINKED ISSUE #3263: Recipes do not react to keyword arguments properly
- [X] are you running newest version (version from docs) (running on `sd/beta-20`)
- [X] can you reproduce the bug with a fresh environment ? (yes)
- [X] What platform + GPU are you on? (any)

I see that the `scatterlines!()` recipe does not pass through all keyword arguments that you might expect.  In particular, `inspector_label` gets ignored.  While this should not be too difficult to patch, is there an accepted way to do the spiritual equivalent of `kwargs...` within a plot recipe?

### LINKED ISSUE #4458: Visible option does not work with tricontourf!
- [ ] what version of Makie are you running? Makie v0.21.13
- [ ] can you reproduce the bug with a fresh environment ? Yes
- [ ] What platform + GPU are you on? Tried this on Linux and on Mac OS 

Here is my example:

```Julia


using Bonito
using WGLMakie

app = App() do session::Session
	visible = Observable(true)
	dropdown = Bonito.Dropdown(["visible", "invisible"]; index=1)
    on(dropdown.value) do value
        if value=="visible"
        	visible[] = true
        	@info "should be visible"
        else
        	visible[] = false
        	@info "should be invisible"
        end
    end

	f = Figure(size=(1000,200))
	ax1 = Axis(f[1,1])
	ax2 = Axis(f[2,1])

	x = randn(50)
	y = randn(50)
	z = -sqrt.(x .^ 2 .+ y .^ 2) .+ 0.1 .* randn.()

	vertices = [
	    0.0 0.0;
	    1.0 0.0;
	    1.0 1.0;
	    0.0 1.0;
	]
	
	faces = [
	    1 2 3;
	    3 4 1;
	]

	colors = [:red, :green, :blue, :orange]

	tricontourf!(ax1, x, y, z, visible=visible)
	mesh!(ax2, vertices, faces, color = colors, shading = NoShading, visible=visible)
	
	return DOM.div(dropdown,f)
end

```

Image

Tried this with GLMakie as well with the same result.

Greetz max

ffreyer:
# Description

We have `shared_attributes(parent, TargetType)` as a function that filters attributes from `parent` applicable to `TargetType` but we are not using it. This pr aims to extend the function and use it consistently in recipes. That should help fix a lot of the issues about attributes not being passed along/having no effect in recipes.

Some thoughts:
- Allowing `shared_attributes` to skip/drop a set of given attribute names means that some valid attributes will not get set. That sounds like something we should not encourage. Edit: It is quite useful in some cases, e.g. if you want to pass strokecolor to a poly but not a scatter or fxaa to a mesh but not lines.
- Renaming attributes is a common and valid task (e.g. `strokecolor => color`), so `shared_attributes` should handle that
- Replacing attributes is also common, but typically requires a multiline `map` (etc). Because of that I think it's preferable to have `attr[:name] = map(...)` instead of something like `obs = map(...); shared_attributes(..., :name => obs)`.
- There are some attributes that only make sense at the top level, e.g. the `inspector` ones. These should be automatically cleared.
- Warnings for missing valid attributes might be useful to find further issues with missing attribute passthrough. 

### First Version

The usage of the first version looked something like this:

```julia
# filters attributes applicable to TargetType from parent plot
# remove some top level/local attributes
# renames based on rename list
attr = shared_attributes(parent, TargetType, [:name_in_parent => :name_in_target, ...])

# removing attributes
pop!(attr, :name)

# replacing/adding
attr[:name] = new_attr
```

This has a couple of problems:
- External pop! and replacing/adding makes "warn_on_unset_attribute" pretty useless since a good chunk of attribute mutation happens afterwards
- Replacements invite mistakes as `setindex!(::Attributes, val, name)` updates Observables if val is not an Observable. E.g. `lines_attr[:fxaa] = false` would change the parent attributes fxaa value, which could also be used in another plot where it should remain true.
- The rename array is unintuitive. As a replacement it would be written in reverse order as `attr[:name_in_target] = parent[:name_in_parent]`

### Second Version ([c04f450](https://github.com/MakieOrg/Makie.jl/pull/4399/commits/c04f4502626dd9f796c90258bfd3dcb988d91924) and after)

```julia
attr = shared_attributes(
    parent, TargetType, 
    :name_to_delete, ...; 
    name_to_replace_or_add = value, ...
)
```

The second version solves the issues above:
- Everything can/should happen in the `shared_attributes` call so we can track defaulted attributes.
- Replacements (given as kwargs) are handled by `shared_attributes` to not change values of the parent.
- Renaming doesn't exist explicitly anymore. They are handled via replacements/kwargs, i.e. `new_name = parent[:old_name]`. 

Related Issues/prs
- Fixes #4458
- Closes #2206 (shared_attributes should be a preferable alternative)
- Fixes #2733
- Closes #3263
- Closes #4435

## Type of change

- Internal cleanup with potential for breaking changes in attribute passthrough

## Checklist


ffreyer:
This is more dangerous than I first thought. There are many recipes that use `mixin_generic_plot_attributes()` which defines `fxaa = true` which then wasn't passed along. With `shared_attributes` it gets picked up and passed. Some recipes define something used in multiple plot types, like `strokewidth`, but only pass it to one child plot. These then need to be explicitly excluded or replaced. Most recipes only define a subset of possible attributes. Before the undefined ones would be ignored, now they would get passed automatically, possibly in bad ways like `fxaa`. Also we don't actually have tests for all plot types in basic_plots. E.g. waterfall and rainclouds have none. 

I also noticed some issues with my `shared_attributes` approach. Because `Attributes` are set up to update Observables instead of replacing elements, doing `attr[:fxaa] = false` will affect the fxaa Observable (if it exists), which is the same (===) as `parent.fxaa` and the same as the fxaa observable passed to another plot. You have to use `attr[:fxaa] = Observable(false)` instead if you want to replace it. That seems like a source of errors to me.
The `warn_on_missing` that I added as a test is also pretty useless when replacements happen after the call.

ffreyer:
Some things I'm (re)considering;
- Should we base passthrough in `shared_attributes` on `attribute_names(parent)` rather than `attributes(parent)`, i.e. exclude everything that isn't defined in `@recipe`? This would make it less likely for attributes to get passed incorrectly, but increase the burden to include everything on the recipe writer 
- Should replacements be part of `shared_attributes` to avoid having to wrap with an `Observable` and to make `warn_on_missing` more useful?

ffreyer:
I'm marking this as breaking since it changes how attributes are passed along, i.e. it will make some attributes work that didn't work before. I hope that this is an improvement in most cases, but there will be some cases where an attribute is wrongfully included. I hope that I can catch most of these through CI, but I doubt it will be all of them.

ffreyer:
After some discussions with Simon we came to the conclusion that it's not a good idea to work on this or merge this before we work on the Plot update/Observable refactor #4360. Depending on how we do that refactor, the thing shared_attributes can do will change. E.g. if `update!(plot, attributes...)` becomes the recommended way to update attributes (and plot args), then shared_attributes shouldn't accept replacements, e.g. `shared_attributes(..., linewidth = calculated_linewidth)`. We may also drop this function entirely and make it part of `plot!(parent_plot, ....)` instead.

### Notes from Discussion

Current Attribute priority is: (high to low)
1. user supplied kwargs
2. theme (`scene.theme.<PlotType>.<attribute_name>`)
3. plot attribute defaults (i.e. what `@recipe` generates) (this may refer back to theme with `@inherit`

In the future we could have `parent_plot.attributes` as a source between user supplied and theme, which would replace the automatic pass-through `shared_attributes` (already) provides.

To take care of renames, e.g. `parent.strokewidth -> lines.linewidth`, we could add a `renames= [:strokewidth => :linewidth, ...]` arg or kwarg. As an arg this probably needs a new struct to be extracted correctly. That would also allow you to do `Inherited(linewidth = :strokewidth, ...)` which I find more intuitive.

Replacements work the same as on master in this framework, i.e. you just pass them as a plot kwarg. `plot!(..., linewidth = calculated_linewidth)`. 

---

To manage direct inheritance and renaming, we could add
```julia
struct InheritedAttribute
    name_in_parent::Symbol
end
```
as an option for Attributes. This would effectively replace `@inherit` and be usable at runtime. `shared_attributes` could generate these for attributes inherited from a parent, and for renamed attributes. These could be chained 
```julia
mesh.color = InheritedAttribute(:color)
# checks parent:
poly.color = InheritedAttribute(:patchcolor)
# checks parent:
scene.theme.patchcolor = some_color
```
and if this chain does not complete we error.

---

General Questions & Notes:
- Should we communicate that an attribute value comes from inheritance, or simply treat it as a value?
- I don't think we talked about it, but I still like the idea of having some tool to check which attributes as recipe covered.
- It seems likely that Observables will not be the preferred way to handle calculations in the future, and that the plot needs to exist for updates.

ffreyer:
Oh, this pr still exists...

I've more or less implemented `plot!(parent, parent.attributes, ...)` as a replacement in 0.24 (iirc). Anything that the child `plot!` can take from `parent.attributes` is extracted with that, and explicitly passing kwargs acts as overwrite. `shared_attributes` now also has a list of attributes to drop so you can use it in a similar way.

PR #4408, opened 2024-09-25
Reduce number of refimages by merging simple tests

ffreyer:
# Description

The purpose of this is to reduce the CI load a bit by merging (visually) simple (and similar) refimages.

The first commit merges a bunch of simpler refimage tests, typically into 2x2 groups. Nothing is deleted here, and only some minor changes have been made to merge tests nicely. (E.g. tweaking resolution, fontsize, placement, etc)

The second commit removes some tests I judged as duplicate or redundant:
- "levels": tests contour with ::Int levels, covered by "contour levels and colors" (new merged test)
- "position": tests simple scatter with a text plot so small it's unreadable. Covered by other scatter + text tests
- "rotation": tests text rotation, covered by "Text rotation"
- "Line Function": tests lines with range + vec input, covered by  "Standard deviation band" explicitly (probably implicitly through convert_arguments tests + lots of other lines tests)
- "scatter": scatter with single markersize, covered by "Plot transform overwrite" explicitly
-  ~~"Marker sizes": tests scatter with vector markersize and colors~~ NOT removed, but merged into "scatter inputs" later
- "reverse range heatmap": tests heatmap with reverse y range, covered by "Reverse image, heatmap and surface axes"

## Type of change

- Internal cleanup

## TODO

- [x] check reference images after CI run


ffreyer:
The file sizes you can see in the CI run summary for refimg tests are misleading. Every file there includes the full set of old reference images with a partial set of recorded images (or full set for merged). For GLMakie the extracted filesize changes are:
```
53.2MB -> 53.2MB reference (not changed until reference images get updated)
20.1MB -> 17.4MB recorded (86.5%, new recorded images)
```
So once the reference images get updated we should have a 10-15% reduction in file size from this. Test time is a little lower, maybe 5-10% on average:
```
               1.6, 1
   GLMakie: (26min, 16min) -> (25min, 13min)
  WGLMakie: (40min, 26min) -> (37min, 24min)
CairoMakie: (20min, 10min) -> (19min, 10min)
```

CairoMakie/src/utils.jl
```diff
@@ -595,7 +595,7 @@ function per_face_colors(_color, matcap, faces, normals, uv)
     elseif color isa Colorant
         return FaceIterator{:Const}(color, faces)
     elseif color isa AbstractVector{<: Colorant}
-        return FaceIterator(color, faces)
+        return FaceIterator{:PerVert}(color, faces)
```
ffreyer, line 598: Should probably point out that I changed this since it's the one thing that isn't just tests. This makes the "Colored Mesh" test work by forcing meshes to use per-vertex colors. 
Since the other backends don't do any special casing for colors that could be per face I don't really see why CairoMakie should. It also errors for 2D and 3D meshes (which is why the test was excluded) and isn't used outside of them, as far as I can tell.

WGLMakie/test/runtests.jl
```diff
@@ -19,19 +19,12 @@ import Electron
...
-    "image scatter",
-    # missing transparency & image
-    "Image on Surface Sphere",
-    # Marker size seems wrong in some occasions:
-    "Hbox",
-    "UnicodeMarker",
-    # Not sure, looks pretty similar to me! Maybe blend mode?
-    "Test heatmap + image overlap",
+    "Image on Surface Sphere", # TODO: texture rotated 180°
     # "heatmaps & surface", # TODO: fix direct NaN -> nancolor conversion
```
EdsterG, line 23: Does this comment need to be included here?

ffreyer: They still don't handle nan_colors the same as GLMakie. Iirc that causes this test to be different

SimonDanisch (approved):

PR #4413, opened 2024-09-27
Support parsing Q/q in BezierPath

DanielVandH:
# Description

Adds support for parsing `Q/q` in `BezierPath`s. Quadratic Bezier curves are constructed as a cubic Bezier curve using https://stackoverflow.com/a/3162732 

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Show DataInspector tooltip on NaN values if `nan_color` has been set to other than `:transparent` [#4310](https://github.com/MakieOrg/Makie.jl/pull/4310)
 - Fix `linestyle` not being used in `triplot` [#4332](https://github.com/MakieOrg/Makie.jl/pull/4332)
 - Fix voxel clipping not being based on voxel centers [#4397](https://github.com/MakieOrg/Makie.jl/pull/4397)
-- Parsing `Q` and `q` commands in svg paths is now supported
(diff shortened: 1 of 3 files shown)


jkrumbiegel (commented):
Looks good, just that I think to match the others, there should be another method taking two points

DanielVandH:
Where would the `(x0, y0)` come from in that case then? The cubic case doesn't need to know the initial control point but, with converting from quadratic to cubic, here we do right?

jkrumbiegel:
Ah sorry I meant `Point` as in signatures taking points and not single numbers. I didn't realize you were giving the previous point to it. Now that I think about it, you could just as well make a new `struct` and then you'd not need the previous point, it would be like with the other types that it only comes in when the path is created. Of course, that's a little more work to add that new type in all the necessary places.

Maybe the easiest way for now is to just leave it as it is, it's not exported anyway. I'd probably change the `QuadraticCurveTo` to `quadratic_curve_to` or so, as it otherwise looks like a type which it isn't.

DanielVandH:
I changed it to `quadratic_curve_to` as you suggest. I avoided adding the `Point` methods for now since, with the current implementation, they wouldn't be used anywhere for now. I looked into defining a new struct initially, just searching for instances of `CurveTo`, but I'm not at all familiar enough with all that to go through it all / properly test it 😅

jkrumbiegel:
Yeah it's fine, I thought at first it was a type, so I thought about public API, but since it isn't, it's not necessary.

jkrumbiegel (approved):

PR #4416, opened 2024-09-27
fix bug merging themes

haakon-e:
`deepcopy` correctly handles copying of non-toplevel attributes like `Theme(Axis = (; titlesize=10))`, otherwise leftmost argument to `merge` is modified.

# Description

Fixes the following bug in which `theme1` is being modified by `merge`:
```julia
julia> theme1 = Theme(Axis = (; titlesize = 10))
Attributes with 1 entry:
  Axis => Attributes with 1 entry:
    titlesize => 10

julia> theme2 = Theme(Axis = (; xgridvisible = false))
Attributes with 1 entry:
  Axis => Attributes with 1 entry:
    xgridvisible => false

julia> merge(theme1, theme2)
Attributes with 1 entry:
  Axis => Attributes with 2 entries:
    titlesize => 10
    xgridvisible => false

julia> theme1
Attributes with 1 entry:
  Axis => Attributes with 2 entries:
    titlesize => 10
    xgridvisible => false
```

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Show DataInspector tooltip on NaN values if `nan_color` has been set to other than `:transparent` [#4310](https://github.com/MakieOrg/Makie.jl/pull/4310)
 - Fix `linestyle` not being used in `triplot` [#4332](https://github.com/MakieOrg/Makie.jl/pull/4332)
 - Fix voxel clipping not being based on voxel centers [#4397](https://github.com/MakieOrg/Makie.jl/pull/4397)
+- Fix `merge(attr1, attr2)` modifying nested attributes in `attr1` [#4416](https://github.com/MakieOrg/Makie.jl/pull/4416)
 
 ## [0.21.11] - 2024-09-13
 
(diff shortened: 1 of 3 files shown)


ffreyer:
Thanks, this is something I've been unhappy with too.

I think we'll need to review this rather carefully since it has the potential to subtly change which attributes get picked. I'm not sure if I trust CI to pick up on that, though maybe I'm underestimating it.

We may also want to consider this breaking since users might already use `merge`

haakon-e:
> We may also want to consider this breaking since users might already use `merge`

That's fair. Though `merge` (without `!`) should promise to not modify input arguments, so I feel this is somewhat more clearly "bugfix" than "breaking feature".

In any case, I'd enjoy seeing this fixed as it was causing some reproducibility issues in my code, and in the end, this (+#4419) made it hard to combine `Theme`s in a reasonable way.

SimonDanisch:
I don't expect many people to use merge and I also don't expect anyone to rely on this behavior.
So I'm happy to merge this for the next patch release.

ffreyer (approved):
I looked through all the `merge` functions we call. As far as I can tell we always rely on `merge!()` for Attributes/Themes, or cast to Dict beforehand. So I think this shouldn't cause any issues

PR #4417, opened 2024-09-27
add unsafe_set!(::Textbox, ::String)

bjarthur:
# Description

a return value of `nothing` is clearly wrong for [this do-block](https://github.com/MakieOrg/Makie.jl/blob/master/src/makielayout/blocks/textbox.jl#L90) as the lifted observable is a Tuple of Points.
i came across this error when manually setting `displayed_string` in an app of mine.  ~happy to add a test if someone suggests how to do so.~

EDIT:  a 2nd commit now adds an unexported `unsafe_set!`, which like the extant `set!` method sets the `displayed_string` but does *not* throw an exception if the string does not pass the validator function.  the use case is to highlight the portion of the string that does not pass the validator by replacing it with the invalid portion.  was the most simple UI for an app of mine.

## Type of change

- [x]  New feature (non-breaking change which adds functionality)

## Checklist


bjarthur:
note that i have since changed the UI in the app of mine that needed this.  others might find `unsafe_set!` useful though, and certainly the return value`set!` should be fixed.

bjarthur:
not sure why the tests are failing.  the errors don't seem to be related to this PR.

```
┌ Error: obj file contains references to .mtl files, but none could be found. Expected: ["cat.mtl"] in /home/runner/work/Makie.jl/Makie.jl/assets.
└ @ MeshIO ~/.julia/packages/MeshIO/EA9dj/src/io/obj.jl:157
```

plus some of these which do not have textboxes:

```
  Expression: score <= threshold
   Evaluated: 1.18206787109375 <= 0.05
```

SimonDanisch (approved):

PR #4424, opened 2024-09-29
Convert powers in scientific notations to RichText.

### LINKED ISSUE #2467: Font of exponents sticks to default when the axis scales are regular.
Thank @jkrumbiegel for fixing the exponents displaying problem in #2321. It shows the right font when the axis scale is set to logarithmic. 

But for regular scales, the font of the exponents is still the default font after I changed the font. The MWE is as follows:
```julia
p=5000000 .+rand(40).*3000000
q=collect(1:40).*100000
f=Figure(fonts=(;regular="CMU"))
ax1=Axis(f[1,1],title="regular scales")
scatterlines!(ax1,q,p)
ax2=Axis(f[2,1],title="log scales",yscale=log10,xscale=log10)
scatterlines!(ax2,q,p)
f
```
test

himcraft:
# Description

Fixes #2467 

`Showoff.showoff()` selects `:scientific` notations when the range of tick values is too large or too small. However, the function returns the powers as Unicodes that are eventually rendered as the default font "DejaVuSans", which looks a little weird together with other fonts. 

The fix adds an additional conversion based on `showoff_minus`. It will first find any "x10" in the raw tick labels, and then convert the following Unicode characters into a superscript of the RichText type. 

However, as this is my first PR to `Makie.jl`, and the fix is rather rough. I am not sure if this fix is suitable to be merged. Any suggestions and guidance are welcomed :)

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed an issue with transformations not propagating to child plots when their spaces only match indirectly. [#4723](https://github.com/MakieOrg/Makie.jl/pull/4723)
 - Added a tutorial on creating an inset plot [#4697](https://github.com/MakieOrg/Makie.jl/pull/4697)
 - Enhanced Pattern support: Added general CairoMakie implementation, improved quality, added anchoring, added support in band, density, added tests & fixed various bugs and inconsistencies. [#4715](https://github.com/MakieOrg/Makie.jl/pull/4715)
 - Fixed issue with `voronoiplot` for Voronoi tessellations with empty polygons [#4740](https://github.com/MakieOrg/Makie.jl/pull/4740)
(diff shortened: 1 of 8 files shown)


himcraft:
### After fix.
test

### Before fix.
test_old

jkrumbiegel:
Oh yeah this has annoyed me for a long time, and the reason I haven't fixed it is exactly because manipulating the strings directly is not the prettiest solution.

I would suggest we do this right and add the necessary capability to PlotUtils first. Probably that would mean adding a function there that outputs these ticks as tuples   of the numeric components so we can do the formatting on those instead. This must already exist in there anyway because they have to get to the current strings somehow.

himcraft:
Yeah I agree that the current solution is not so elegant. I will look into Showoff to see how to expose those power numbers from the existing functions. Or perhaps we can dispatch a function that outputs the ticks as tuples especially for Makie types.

PR #4427, opened 2024-09-30
Legend overrides

jkrumbiegel:
Implements a mechanism to override legend element attributes that are normally set by copying them from the respective plot objects. ~Introduces the `LegendOverride` object, although I'm not 100% sure if we really need that as API, or should just use whatever key-value pairs we can get.~ Now uses any key-value object.

The overrides can either be paired to `label` keywords (as that's the legend interface for plotting functions), or to the plotting functions themselves in the `Legend` constructor (there I don't think it makes sense to pair to labels although it might be perceived as inconsistent?).

## Examples

```julia
f, ax, sc = scatter(
    cos.(range(0, 7pi, 100)),
    color = :black,
    markersize = 8,
    label = "cos" => (; markersize = 15)
)
scatter!(
    sin.(range(0, 7pi, 100)),
    color = :black,
    marker = :utriangle,
    markersize = 8,
    label = "sin" => (; markersize = 15)
)
Legend(f[1, 2], ax)
f
```

grafik

```julia
f = Figure()
ax = Axis(f[1, 1])
li = lines!(ax, 1:5, linestyle = :dot)
sc = scatter!(ax, 1:5, markersize = 10)
Legend(
    f[1, 2],
    [
        sc => (; markersize = 20),
        li => (; linewidth = 3),
        [li, sc] => (; color = :red),
        [li => (; linewidth = 3), sc => (; markersize = 20)],
    ],
    ["Scatter", "Line", "Both", "Both 2"],
    patchsize = (40, 20),
)
f
```
grafik

```julia
f, ax, bp = barplot(
    1:5,
    [1, 3, 2, 5, 4],
    color = 1:5,
    colorrange = (1, 5),
    colormap = :Set1_5,
    label = [label => (; color = i)
        for (i, label) in enumerate(["red", "blue", "green", "purple", "orange"])]
)
Legend(f[1, 2], ax)
f
```
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fix incorrect inverse transformation in `position_on_plot` for lines, causing incorrect tooltip placement in DataInspector [#4402](https://github.com/MakieOrg/Makie.jl/pull/4402)
+- Add `LegendOverride` object to override legend element attributes [#4427](https://github.com/MakieOrg/Makie.jl/pull/4427).
 
 ## [0.21.12] - 2024-09-28
 
--- a/ReferenceTests/src/tests/figures_and_makielayout.jl
+++ b/ReferenceTests/src/tests/figures_and_makielayout.jl
@@ -172,7 +172,7 @@ end
 
     li = lines!(
         1:10,
-        label = "Line" => LegendOverride(linewidth = 4, color = :gray60, linestyle = :dot),
+        label = "Line" => (; linewidth = 4, color = :gray60, linestyle = :dot),
     )
     sc = scatter!(
         1:10,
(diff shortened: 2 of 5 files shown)


asinghvi17:
This looks really useful, will probably let me move from manually constructing legends to simply specifying them in this way as well!  

I'd rather we support key-value pairs (maybe NamedTuples?) instead of an override object, just because it is a lot easier to learn and reason about than yet another custom object...

jkrumbiegel:
> I'd rather we support key-value pairs (maybe NamedTuples?

Yeah I guess I'll just convert internally to `Attributes`, I don't think the custom type is really necessary as long as we don't expand the `label => something` interface further so that one has to differentiate.

SimonDanisch (approved):

PR #4431, opened 2024-10-01
More improvements for ReferenceUpdater

ffreyer:
# Description

Changes:
- [x] pick latest CI run if multiple are present (e.g. due to closing and reopening a pr)
- [x] add tab for removing reference images of missing recordings (e.g. to clean up reference images for renamed or deleted tests)
- [x] Use latest reference images when updating, rather than the ones bundled with the CI run. This should prevent reference images from getting overwritten for outdated CI runs. ~~make ReferenceUpdater safe to use from an outdated CI run (currently this uses old reference images from the CI run, rather than the latest uploaded ones.)~~
- [x] use table layout for new (and missing) ref images
- [x] add some documentation to the page
- [x] add a "toggle all" for added and removed refimgs

## Type of change

- tooling

### DIFF
--- a/.github/workflows/reference_tests.yml
+++ b/.github/workflows/reference_tests.yml
@@ -198,9 +198,10 @@ jobs:
 
           # Loop through the directories and concatenate the files, and copy recorded folders
           for dir in WGLMakie CairoMakie GLMakie; do
-              # Concatenate scores.tsv and new_files.txt
+              # Concatenate scores.tsv, new_files.txt and missing_files.txt
               cat "${baseDir}/${dir}/scores.tsv" >> "./ReferenceImagesCombined/scores.tsv"
               cat "${baseDir}/${dir}/new_files.txt" >> "./ReferenceImagesCombined/new_files.txt"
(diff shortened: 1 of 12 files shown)


jkrumbiegel:
> make ReferenceUpdater safe to use from an outdated CI run (currently this uses old reference images from the CI run, rather than the latest uploaded ones.

Yeah that's currently not great that you can clobber already updated reference images with the old ones from a previous CI run. I did this at the time because I was focusing on showing the changes that CI picked up in that specific run (that's what the scores were computed for etc.). How did you want to change this, compute scores again against the freshly downloaded reference set?

ffreyer:
Updating should not download the lastest refimg set and merge with that, instead of using the one attached to a pr.

The page now looks like this: (at 80% zoom to fit more on a screenshot)
grafik

jkrumbiegel (approved):
Maybe take out the couple `console.log`s if they aren't needed, otherwise this looks good to me!

ffreyer:
Oh I forgot about those

PR #4433, opened 2024-10-01
Add reference image tests for untested plots/blocks

ffreyer:
# Description

Plots/Blocks that do not have (direct) reference images:
- [x] raincloud
- [x] series
- [x] stairs
- [x] stem
- ~~timeseries~~ (unlikely to be consistent in refimage tests due to dependency on sleep/time)
- [x] waterfall
- [x] boxplot
- [x] crossbar
- [x] qqplot
- [x] qqnorm
- [x] ecdfplot
- [x] Button
- [x] IntervalSlider
- [x] Slider
- [x] Toggle
- ~~LineSegmentBuffer, TextBuffer~~ tested via axis3d!, considered internal, may disappear in the future

Plots that may need more:
- [x] ablines
- [x] hspan, vspan
- [x] violin
- ~~spy~~ already has test testing multiple configurations
- [x] volumeslices
- [x] Textbox

## Type of change

- Internal Improvements / testing


EdsterG:
Is there any documentation about updating/creating ref image tests? Is the image pre-generated and uploaded or do you use some relative score between ref images dynamically generated during test time?

ffreyer:
#4108 added some notes to prs about it.
If you want to add a refernce image test, you just need to write a
```julia
@reference_test "some name" begin
    # your code...
    fig
end
```
block, like in this pr. CI produces an image for each of these blocks and compares it to a reference image if it exists. When a new reference test is added (or if one gets changed) the maintainers need to upload the new image.

ReferenceTests/src/tests/examples2d.jl
```diff
@@ -1496,6 +1496,33 @@ end
     fig
 end
 
+@reference_test "Violin" begin
+    fig = Figure()
+    
+    categories = RNG.rand(1:3, 1000)
+    values = RNG.randn(1000)
+    violin(fig[1, 1], categories, values)
```
jkrumbiegel, line 1504: I actually stopped doing reference tests where there's some kind of symmetry in the data. Like here, all groups have the same data basically, so if there was something wrong in the mapping from categories to groups, you wouldn't see it here. Instead I'd add the group values to the data maybe, so you'd see three shifted violins, or something to that effect.

jkrumbiegel (approved):

PR #4442, opened 2024-10-02
fix: reintroduce vmin, vmax into get_ticks; addresses #4441

### LINKED ISSUE #4441: get_ticks(ticks, identity, Makie.automatic, limits...) results in unintended results
- [ ] what version of Makie are you running? (`]st -m Makie`) - running v0.21.12
- [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) - yes

```
using Makie
ticks = 0.0:0.2:1.0
limits = (0.0, 0.6)
Makie.get_ticks(ticks, identity, Makie.automatic, limits...)
([0.0, 0.2, 0.4, 0.6, 0.8, 1.0], ["0.0", "0.2", "0.4", "0.6", "0.8", "1.0"])
```
**Expected behavior:**
I would have expected that I do not get ticks returned at 0.8 and 1.0, because I have given in the limits? 

**Probable source of bug:**
https://github.com/MakieOrg/Makie.jl/blob/59456d3f0f37094e95267c5a27405f1637afedbf/src/makielayout/lineaxis.jl#L631
This function ignores vmin and vmax - I would have expected that a filter gets applied.

thomvet:
# Description

Fixes #4441 
get_ticks now filters out values that are below vmin and vmax supplied to it. 

## Type of change
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- added appropriate test.
- other options seemed not relevant.

### DIFF
--- a/src/makielayout/lineaxis.jl
+++ b/src/makielayout/lineaxis.jl
@@ -628,8 +628,7 @@ get_tickvalues(lt::LinearTicks, vmin, vmax) = locateticks(vmin, vmax, lt.n_ideal
 
 Convert tickvalues to a float array by default.
 """
-get_tickvalues(tickvalues, vmin, vmax) = convert(Vector{Float64}, tickvalues)
-
+get_tickvalues(tickvalues, vmin, vmax) = filter(p -> p >= vmin && p <= vmax, convert(Vector{Float64}, tickvalues))
 
 # function get_tickvalues(l::LogitTicks, vmin, vmax)
 #     ticks_scaled = get_tickvalues(l.linear_ticks, identity, logit_10(vmin), logit_10(vmax))
--- a/test/makielayout.jl
+++ b/test/makielayout.jl
@@ -209,6 +209,11 @@ end
         @test get_ticks(numbers, func, xs -> string.(xs) .* "kg", 0, 5) == (numbers, ["1.0kg", "1.5kg", "2.0kg"])
 
         @test get_ticks(WilkinsonTicks(5), identity, automatic, 1, 5) == ([1, 2, 3, 4, 5], ["1", "2", "3", "4", "5"])
+
+        #Check that ticks outside supplied limits are filtered out
+        ticks = [-0.2, 0.0, 0.2, 0.4, 0.6, 0.8, 1.0]
+        limits = (0.0, 0.6)
+        @test get_ticks(ticks, identity, automatic, limits...) == ([0.0, 0.2, 0.4, 0.6], ["0.0", "0.2", "0.4", "0.6"])
     end
 end
 


src/makielayout/lineaxis.jl
```diff
@@ -628,8 +628,7 @@ get_tickvalues(lt::LinearTicks, vmin, vmax) = locateticks(vmin, vmax, lt.n_ideal
 
 Convert tickvalues to a float array by default.
 """
-get_tickvalues(tickvalues, vmin, vmax) = convert(Vector{Float64}, tickvalues)
-
+get_tickvalues(tickvalues, vmin, vmax) = filter(p -> p >= vmin && p <= vmax, convert(Vector{Float64}, tickvalues))
```
EdsterG, line 631:
```suggestion
get_tickvalues(tickvalues, vmin, vmax) = filter!(p -> p >= vmin && p <= vmax, convert(Vector{Float64}, tickvalues))
```

thomvet: Thanks. One allocation less!

thomvet:
Didn't expect this simple change to break something else; good thing the tests exist! I introduced another fix to address the issue (dont't use showoff on empty array). 

But now I am running into https://github.com/MakieOrg/Makie.jl/pull/4416, i.e., Theme merging doesn't work anymore? Not obvious to me why the changes on get_ticks should have caused this!?

jkrumbiegel:
The ticks are actually filtered later in the Axis code, as some tick algorithms return ticks outside the bounds, so you could either add that filtering to each algorithm, or just in a step after.

thomvet:
Hm, I stumbled over this while working on a code for a new type of axis (TernaryAxis - for ternary plots), where I didn't have the further steps implemented yet.

I took the basic code from PolarAxis, so it might be that the problem is in there as well. 

I'll check tomorrow.

src/makielayout/lineaxis.jl
```diff
@@ -628,8 +628,7 @@ get_tickvalues(lt::LinearTicks, vmin, vmax) = locateticks(vmin, vmax, lt.n_ideal
 
 Convert tickvalues to a float array by default.
 """
-get_tickvalues(tickvalues, vmin, vmax) = convert(Vector{Float64}, tickvalues)
-
+get_tickvalues(tickvalues, vmin, vmax) = filter!(p -> p >= vmin && p <= vmax, convert(Vector{Float64}, tickvalues))
```
t-bltg, line 631:
```suggestion
get_tickvalues(tickvalues, vmin, vmax) = filter!(p -> vmin ≤ p ≤ vmax, convert(Vector{Float64}, tickvalues))
```

PR #4447, opened 2024-10-04
Allow `width` to be set per box in `boxplot`

EdsterG:
# Description

Allows width to be set per box. Not sure if this is a bug fix or new feature, classified it as a bug fix since colors can be set per box.

This also fixes `boxplot` when `range` is zero. Before this PR nothing would get plotted (because the color vector was empty), after this PR the plot works correctly with zero range.

## Type of change

## Checklist


EdsterG:
This PR is good to go on my end. Would like to add a test but it overlaps with #4433. How would you like to coordinate it @ffreyer? Can ether wait for #4433 to get merged, OR if this PR is merged before, #4433 can easily be updated to test width, e.g. `WGLMakie.boxplot(fig[1, 1], categories, values, width = categories ./ 3)`.

EdsterG:
Why do the tests fail on scatter? I thought that was fixed and RNG stabilized?

EdsterG:
Nice, thanks for fixing these test issues and for adding all of the new tests.

EdsterG:
@ffreyer I've updated the ref image tests to include variable width per category, just needs an update to the ref images. If I understood correctly, only maintainers can update the ref images?

ffreyer:
Thanks for updating. I uploaded the new refimages and the remaining failure is unrelated to this pr so I'm merging this.

PR #4451, opened 2024-10-04
add tooltips to blocks

bjarthur:
supersedes https://github.com/MakieOrg/Makie.jl/pull/4380

adds tooltips to Textboxes which appear on hover.  API is a new `tooltip!(::Block)` method as proposed by @jkrumbiegel in https://github.com/MakieOrg/Makie.jl/pull/4380#issuecomment-2385576036

those blocks which do not already have a `hovering` field will need to be modified to have one.

currently only Textbox is known to work.  will get to the others shortly.

```
using GLMakie
f = Figure()
tb = Textbox(f[1,1])
tt = tooltip!(tb, "below", placement=:below)
```

## Type of change

## Checklist


jkrumbiegel:
I don't think it's necessary to add `hovering` fields to all blocks. Maybe the way this function works could be that it grabs the bbox of the block passed to it (that's an internal accessible via the layout observables already) and then just adds logic that checks if the mouse is somewhere in there, but also that it has stayed at some point for at least `t` seconds. I think that's the normal behavior for these tooltips, that they appear after the user has wondered for a short while what that element might be for.

bjarthur:
ready for review!

the tests only fail because i modified the tooltip reference image to include a centered tooltip for an axis.

bjarthur:
hovering fields have been removed from all blocks and mouseposition is compared to bbox as @jkrumbiegel suggested.  tooltips can now be always shown, never shown, or only shown when hovered.  i don't know how best to add a delay when hovering, and since this is a feature i wouldn't use, i chose not to pursue it.  if this is important, please suggest a way and i can add.

bjarthur:
here's how i use this PR:

https://github.com/user-attachments/assets/29fbb0cf-da3f-47e5-952c-9151a44be69b

```
using GLMakie

fig = Figure()

to_doc = Toggle(fig[1,1])
tooltip!(to_doc, "show the tooltip docs", visible = :hover)

tb = Textbox(fig[2,1], stored_string = "textbox")
tooltip!(tb, "i'm a textbox",
         placement = :right,
         visible = @lift $(to_doc.active) ? :hover : :never)

me = Menu(fig[3,1], options = ["option 1","option 2","option 3"])
tooltip!(me, "i'm a menu",
         placement = :right,
         visible = @lift $(to_doc.active) ? :hover : :never)

bu = Button(fig[4,1])
tooltip!(bu, "i'm a button", label = "push me",
         placement = :right,
         visible = @lift $(to_doc.active) ? :hover : :never)
```

in actual usage, i also save the state of the toggle using Preferences.jl.  when the user loads the package for the first time ever, the state is on and all tooltips are displayed.  if they then turn it off, this is remembered the next time and all subsequent times they load the package.

like this with GLMakie:

```
show_docs = @load_preference("show_docs", true)

to_doc = Toggle(fig[1,1], active=show_docs)

on(to_doc.active) do to_doc
    @set_preferences!("show_docs"=>to_doc)
end      
```

and this with Bonito:

```
c = filter(x->x.name=="show_docs", Bonito.HTTP.cookies(request))
show_docs = isempty(c) || c[1].value=="true"

to_doc = Toggle(fig[1,1], active=show_docs)

on(to_doc.active) do to_doc
    evaljs(session, js"document.cookie = \"show_docs=\"+$(widgets.to_doc.active).value;")            
end
```

jkrumbiegel:
@ffreyer in the video above you can see the rendering weirdness with tooltip I was talking about before. The stroke of the little triangle is halved and jagged for some reason

@bjarthur I like this implementation better already. I think it still needs a delay setting (which you of course can set to 0 if you prefer that) but it would get out of your way more if you didn't have to toggle it on and off, but could leave it on and tooltips would display smartly.

I'm also not quite sure if it's good to show the tooltip on the outside of the block, that can work ok for smaller ones but for bigger ones it might look a bit awkward, and also drags attention away from the user's cursor (which is where they're currently attending). So I'd probably mirror the more typical behavior in browsers where the tooltip appears right next to the cursor. Like this:


https://github.com/user-attachments/assets/192a9bfe-a7c6-46cb-a84a-af7c76f6d3ae

bjarthur:
delay to display and following the cursor now implemented in most recent two commits.  anything else?

bjarthur:
delay really slows down GUI.  need to refactor to minimize calls to `time()`.  hang tight...

ffreyer:
> @ffreyer in the video above you can see the rendering weirdness with tooltip I was talking about before. The stroke of the little triangle is halved and jagged for some reason

Yea and this is what I thought I fixed. And for me it is fixed in GLMakie:

image

EdsterG:
WGLMakie tooltips used to render correctly for me back in v0.21.8, and then broke a with one of the updates. I just pulled in the latest version v0.21.15 and the triangle is still wrong in WGLMakie.

EdsterG:
@ffreyer do you recall the commit in which you fixed the tooltip triangle? I wonder if that fix was OS specific for whatever reason.

bjarthur:
alright, it's really stripped down now!  let me know what you think.

bjarthur:
i think this can be made even more performant by turning `off` the observerfunction when visible is dynamically set to false.  give me a sec...

bjarthur:
ok, so latest commit removes the event handler to update the visibility if the observable is set to false, instead of having logic in the handler to not display the tooltip.  so the event loop shouldn't be bogged down unneccesarily.

i've also maybe the delay kwarg able to be an observable.

bjarthur:
battle testing this PR it seems to all work in GLMakie, but i just uncovered this strange behavior with WGLMakie.  i think this is a bug somewhere else in Makie, perhaps related to using `off` to disable observer callbacks.  or am i doing something wrong?  here is an MWE:

```
using WGLMakie     # with GLMakie it works fine

o_d = Observable(rand(10))
o_v = Observable(true)

f, a, s = scatter(o_d)
tt = tooltip!(a, "axis", visible = o_v)

# check that the tooltip is visible

o_v[] = false

# now it shouldn't be, or...
# maybe it still is and will now also be visible outside the axis bounds

o_d[] = rand(10)

# it probably/might still not be visible now, but..
# now zoom in or pan
# tooltip will now for sure reappear, even outside the axis
```

bjarthur:
with the most recent commit, the MWE that shows a difference between GLMakie and WGLMakie can be reduced to:

```
using WGLMakie

o_d = Observable(rand(10))
o_v = Observable(true)

f, a, s = scatter(o_d)
tt = tooltip!(a, "axis", visible = o_v)

# check that the tooltip is visible

o_v[] = false

# now it shouldn't be, but..
# now zoom in or pan
# tooltip will reappear, even outside the axis
```

bjarthur:
the MWE seems to be fixed with the latest commit!  my fault afterall.  sorry for the noise.

bjarthur:
uncovered another problem :(.    there is an as of yet unexplicable slow down when these tooltips are used simultaneously with DataInspector.  but again, only for WGLMakie and *not* GLMakie.  here's a MWE:

```
import GLMakie as G
using WGLMakie, Bonito, Sockets

app = App(threaded=true) do session::Session, request::Bonito.HTTP.Request
    f = Figure()

    to_doc = Toggle(f[1,1], active=true)
    tooltip!(to_doc, "to_doc")

    tb = Textbox(f[1,2])
    tooltip!(tb, "tb", visible = to_doc.active)

    me = Menu(f[1,3], options = string.(1:100))
    tooltip!(me, "me", visible = to_doc.active)

    sl = G.Slider(f[1,4], range = 1:100)
    tooltip!(sl, "sl", visible = to_doc.active)

    a, s = scatter(f[2,:], rand(10_000))
    tooltip!(a, "a2", visible = to_doc.active)
    
    DataInspector() 
    
    return DOM.div(f)
end;

Bonito.Server(app, string(getipaddr()), 8080)
```

the Toggle controls whether the block tooltips are displayed.  if initially set to true, and then one hovers over points in the scatter plot, the GUI becomes unresponsive and Timeout warnings are dumped to the REPL.  if instead you set the initial value of the Toggle to false, then hover over the scatter to "warm-up" DataInspector, and then turn the Toggle on to display the block tooltips, all is fine.

bjarthur:
CI tests pass and this PR is working well in production for me.  can someone please review again?

bjarthur:
i should've added to my last plea for a review that the slowdown with WGLMakie has been resolved with another commit.  @jkrumbiegel

bjarthur:
bump.  a review would be appreciated.

all tests pass except for this new "post benchmark gist".  the error there has sth to do with the GH_TOKEN environment variable not being set.

bjarthur:
bump

bjarthur:
runic passes now

bjarthur:
the failing CI is due to a security / permissions problem with the token:

```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
Error: Process completed with exit code 1.
```

otherwise tests pass.

bjarthur:
@jkrumbiegel could you please review again?  it's been months...

Makie/src/makielayout/blocks.jl
```diff
@@ -674,6 +674,104 @@ function REPL.fielddoc(t::Type{<:Block}, s::Symbol)
...
+
+julia> tt = tooltip!(b, "I'm a Button", placement = :below, enabled = e, delay = 1)
+Plot{Makie.tooltip, Tuple{Vec{2, Float32}, String}}
+
+julia> e[] = true
+:always
+```
+"""
+function tooltip!(b::Block, str::AbstractString; enabled = true, delay = 0, depth = 9.0e3, kwargs...)
+    _enabled = typeof(enabled) <: Observable ? enabled : Observable(enabled)
```
SimonDanisch: cant this just be convert(Observable, enabled) etc?

Makie/src/makielayout/blocks.jl
```diff
@@ -674,6 +674,104 @@ function REPL.fielddoc(t::Type{<:Block}, s::Symbol)
...
+    _delay = typeof(delay) <: Observable ? delay : Observable(delay)
+    _depth = typeof(depth) <: Observable ? depth : Observable(depth)
+
+    position = Observable(Point2f(0))
+    tt = tooltip!(b.blockscene, position, str; visible = false, kwargs...)
+    on(z -> translate!(tt, 0, 0, z), _depth)
+
+    function update_viz0(mp, bbox)
+        return if mp in bbox
```
SimonDanisch: i think these are runic artifacts and should move the return to the end, so it returns nothing!
Looks better, and can also be better for inference.

Makie/src/makielayout/blocks.jl
```diff
@@ -674,6 +674,104 @@ function REPL.fielddoc(t::Type{<:Block}, s::Symbol)
...
+            foreach(off, obsfun)
+            obsfun = nothing
+            tt.visible[] = false
+        end
+    end
+
+    notify(_enabled)
+    notify(_delay)
+    notify(_depth)
+    notify(b.blockscene.events.mouseposition)
```
SimonDanisch: Seems weird to me that we need all those notifies? 
Are they defensive, or really needed?

Makie/src/makielayout/blocks.jl
bjarthur: yes, they can!

Makie/src/makielayout/blocks.jl
bjarthur: moved!

Makie/src/makielayout/blocks.jl
bjarthur: _enabled and _depth seem to be necessary.  i've removed the other two.

bjarthur:
i've also fixed the delay kwarg.  previously the tooltip wouldn't show after they delay unless the mouse was moved within the widget.  now it is made visible even if the mouse is stationary.  here is a simple test harness:

```
using GLMakie
f = Figure()
b = Button(f[1,1])
tooltip!(b, "button")
b2 = Button(f[2,1])
tooltip!(b2, "button2", delay=1)
```

jkrumbiegel:
I've tested it a little bit recently and had a couple things I wanted to tweak. Will try to get back to it soon

bjarthur:
awesome, thanks!  note that tests pass now.

PR #4460, opened 2024-10-09
Update ecosystem.md

briochemc:
- Add URLs 
- Add swarms for Makie!

# Description

Fixes nothing

Adds SwarmMakie.jl to the ecosystem page in the docs + adds URLs to the ecosystem items.

I'm not sure what image should be added from SwarmMakie.jl.

## Type of change: Documentation


asinghvi17:
SwarmMakie has a logo, that also shows what it does I think.  But we could use something else.

PR #4461, opened 2024-10-09
Improve Relocatability for GLMakie and fix tests

SimonDanisch:
The tests didn't really test relocatability, since we only moved the package dir, but not the GLMakie source code.
This changes GLMakie to load all shaders on precompile, to make them part of the system image, much like what RelocatableFolder does it, but more explicitely (since we're caching + loading the shaders anyways).
Also deletes some dead code.


bjarthur:
+100 for all improvements to relocatability!  there is also a similar outstanding [PR](https://github.com/MakieOrg/Makie.jl/pull/3882) for WGLMakie.  still haven't found the time to finish that up.

PR #4463, opened 2024-10-09
Use polys for axis3

jkrumbiegel:
Currently, Axis3 doesn't work well with svg export because the meshes that are used for the side panels cause rasterization in Cairo. This breaks svg display for a lot of frontends, for example in VSCode:

grafik

This PR uses translated polys instead to avoid this.

grafik


src/makielayout/blocks/axis3d.jl
```diff
@@ -664,29 +664,42 @@ function add_panel!(scene, ax, dim1, dim2, dim3, limits, min3)
...
     end
 
-    faces = [1 2 3; 3 4 1]
+    plane = Symbol((:x, :y, :z)[dim1], (:x, :y, :z)[dim2])
 
-    panel = mesh!(scene, vertices, faces, shading = NoShading, inspectable = false,
+    panel = poly!(scene, rect, inspectable = false,
         xautolimits = false, yautolimits = false, zautolimits = false,
-        color = attr(:panelcolor), visible = attr(:panelvisible))
+        color = attr(:panelcolor), visible = attr(:panelvisible),
```
asinghvi17, line 689: Should we specify `strokecolor = :transparent` here, just in case we end up changing that default?

jkrumbiegel: Yes good idea

PR #4288, opened 2024-08-30
Test each backend for relocatability

jkrumbiegel:

PR #4467, opened 2024-10-10
WGLMakie relocatability

jkrumbiegel:
This PR adds only the necessary changes from https://github.com/MakieOrg/Makie.jl/pull/3882 to the recent GLMakie relocatability PR #4461. I confirmed locally that this allows to compile a relocatable WGLMakie app using the script that @SimonDanisch introduced in https://github.com/MakieOrg/Makie.jl/pull/4461

After both #4461 and this are merged, we can update https://github.com/MakieOrg/Makie.jl/pull/4288 so that CI tests CairoMakie and WGLMakie for relocatability as well. Last status there was that I had gotten it to work for CairoMakie but WGLMakie got stuck in CI, probably due to an incorrect setup with `xvfb` or Electron or so. As I tested locally that it can work, this should not be a fundamental issue.


bjarthur:
do your tests cover font handling?  i had trouble relocating without the explicit conversion to `String` when dealing with FreeTypeAbstractions.  please consider as necessary the changes i proposed in src/conversions.jl.

jkrumbiegel:
> do your tests cover font handling? i had trouble relocating without the explicit conversion to String

As mentioned in that other PRs comments, I think that was a side effect of the incorrect use of `@path`

Issue #4378, opened 2024-09-18
vertical toggle


bjarthur:
is there any reason that `Toggle` does not have a `horizontal` argument much like `Slider`?  should i take a stab at adding one?

jkrumbiegel:
No reason, if you want, go ahead :)

PR #4445, opened 2024-10-03
add vertical orientation option to Toggle

bjarthur:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/4378 by adding an `orientation` kwarg to `Toggle` that defaults to `:horizontal` and accepts `:vertical` too.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [x] Added unit tests for new algorithms, conversion methods, etc.
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #4471, opened 2024-10-10
add kwarg to rotate Toggle

bjarthur:
# Description

fixes https://github.com/MakieOrg/Makie.jl/issues/4378; supersedes https://github.com/MakieOrg/Makie.jl/pull/4445.

adds `orientation` kwarg to `Toggle` that specifies the angle in radians to rotate the toggle.  default of 0 is horizontal with active being to the right.

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/toggle.jl
+++ b/src/makielayout/blocks/toggle.jl
@@ -44,7 +44,8 @@ function initialize_block!(t::Toggle)
     button = scatter!(topscene, buttonpos, markersize = buttonsize,
         color = t.buttoncolor, strokewidth = 0, inspectable = false, marker = Circle)
 
-    onany(topscene, t.orientation, t.layoutobservables.computedbbox) do angle_rad, bbox
+    onany(topscene, t.orientation, t.layoutobservables.computedbbox) do or, bbox
+        angle_rad = or == :horizontal ? 0 : or == :vertical ? pi/2 : or
         center = Vec3f(bbox.origin[1]+bbox.widths[1]/2, bbox.origin[2]+bbox.widths[2]/2, 0)
         R = Makie.rotationmatrix_z(angle_rad)
         T = Makie.translationmatrix(-center)
--- a/src/makielayout/blocks/toggle.jl
+++ b/src/makielayout/blocks/toggle.jl
@@ -44,6 +44,14 @@ function initialize_block!(t::Toggle)
     button = scatter!(topscene, buttonpos, markersize = buttonsize,
         color = t.buttoncolor, strokewidth = 0, inspectable = false, marker = Circle)
 
+    onany(topscene, t.orientation, t.layoutobservables.computedbbox) do angle_rad, bbox
+        center = Vec3f(bbox.origin[1]+bbox.widths[1]/2, bbox.origin[2]+bbox.widths[2]/2, 0)
+        R = Makie.rotationmatrix_z(angle_rad)
+        T = Makie.translationmatrix(-center)
+        Tinv = Makie.translationmatrix(center)
+        topscene.transformation.model[] = Tinv * R * T
+    end
+
     mouseevents = addmouseevents!(topscene, t.layoutobservables.computedbbox)
 
     updatefunc = Ref{Any}(nothing)
--- a/src/makielayout/blocks/toggle.jl
+++ b/src/makielayout/blocks/toggle.jl
@@ -24,7 +24,8 @@ function initialize_block!(t::Toggle)
         costheta, sintheta = cos(theta), sin(theta)
         rotmatrix = GeometryBasics.@SMatrix [costheta -sintheta
                                              sintheta  costheta]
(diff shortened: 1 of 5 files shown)


src/makielayout/blocks/toggle.jl
```diff
@@ -44,6 +44,14 @@ function initialize_block!(t::Toggle)
     button = scatter!(topscene, buttonpos, markersize = buttonsize,
         color = t.buttoncolor, strokewidth = 0, inspectable = false, marker = Circle)
 
+    onany(topscene, t.orientation, t.layoutobservables.computedbbox) do angle_rad, bbox
```
asinghvi17, line 47: `orientation` should also support `:horizontal` or `:vertical` like the rest of them IMO.  Maybe just have a function that converts from the symbol to the angle, and invoke that within the statement?

src/makielayout/types.jl
```diff
@@ -1152,15 +1152,32 @@ const CHECKMARK_BEZIER = scale(BezierPath(
...
+```julia
+Toggle(fig_or_scene; kwargs...)
+```
+
+## Examples
+
+```julia
+t_horizontal = Toggle(fig[1, 1])
+t_vertical = Toggle(fig[2, 1], orientation = pi/2)
+```
```
asinghvi17, line 1169: Sweet!  Would also be nice to add a small `on` to demonstrate how it's done, if possible.

bjarthur:
Screenshot 2024-10-10 at 5 55 32 PM

bjarthur:
i just noticed that the vertical alignment is off when rotated:

Screenshot 2024-10-31 at 4 09 19 PM

i would've though the top edges of both toggles should be at the same height.  can someone please suggest how i can fix this?

```
julia> using GLMakie

julia> fig = Figure()

julia> scatter(fig[1,1], [1,2,3], axis = (; height=200))
Makie.AxisPlot(Axis (1 plots), Scatter{Tuple{Vector{Point{2, Float64}}}})

julia> Toggle(fig[1,2], orientation=:vertical, valign=:top)
Toggle()

julia> Toggle(fig[1,3], orientation=:horizontal, valign=:top)
Toggle()
```

bjarthur:
i've fixed the alignment issue in the previous comment by refactoring to do my own rotation instead of using @ffreyer 's method.  i couldn't figure out another way to do it.  everything seems to work, and i suspect tests will pass, but only because none of the reference images uses the `width` or `height` kwargs to `Toggle`.

there is a slight backwards incompatibility that might warrant a bump in Makie's minor version with this PR as the size of the toggle is now controlled by `length` and `markersize` instead of `width` and `height`.  with the new `orientation` kwarg i feel this is a much more natural parameterization.  let me know if you disagree.

bjarthur:
i'm confused about the CI failure for ref img "Button - Slider - Toggle - Textbox":

```
running 241: Button - Slider - Toggle - Textbox
Button - Slider - Toggle - Textbox: Error During Test at /home/runner/work/Makie.jl/Makie.jl/ReferenceTests/src/database.jl:51
  Test threw exception
  Expression: save_result(joinpath(RECORDING_DIR[], "Button - Slider - Toggle - Textbox"), result)
  No applicable_savers found for UNKNOWN
```

it works fine locally and nothing in this PR should effect that test.

bjarthur:
could you please review again?

ffreyer:
Tests fail because I merged the new refimg with an existing one. So no problem there.

I'm not sure about the (width, height) -> (length, markersize) change. I see why it's there but I wonder if width, height is supposed to be something users can always set. Right now it gets overwritten if I'm not mistaken.

src/makielayout/blocks/toggle.jl
```diff
@@ -2,30 +2,39 @@ function initialize_block!(t::Toggle)
 
     topscene = t.blockscene
 
-    markersize = lift(topscene, t.layoutobservables.computedbbox) do bbox
-        min(width(bbox), height(bbox))
+    onany(topscene, t.orientation, t.length, t.markersize) do or, len, ms
+        theta = or == :horizontal ? 0 : or == :vertical ? pi/2 : or
+        t.width[] = (len - ms) * cos(theta) + ms
```
jkrumbiegel, line 7: This is not correct. The automatically determined size goes into `t.layoutobservables.autosize`. You can check against `Button` or `Label` for example. This allows the determined values to be overridden by `width` and `height` (this would not affect the size of the toggle, just the space in the layout).

bjarthur: oh, thanks!  i'm still on the Makie learning curve.  fixed in the latest commit.

jkrumbiegel:
The boundingbox computation does not work correctly for all rotations yet which you can check like this:

```julia
f = Figure()
rot = Observable(0.0)
Box(f[1, 1], color = :tomato)
Toggle(f[1, 1], orientation = rot)
sl = Slider(f[1, 2], range = 0:0.01:2pi)
connect!(rot, sl.value)
f
```

bjarthur:
oof.  good catch.  fixed in latest commit.  i've added a ref image to test for this too.

Issue #2637, opened 2023-01-29
Documentation doesn't explain well the model matrix things


Zou-Bo:
Many plotting functions(e.g. heatmap) have this attribute:
"model::Makie.Mat4f sets a model matrix for the plot. This replaces adjustments made with translate! , rotate! and scale! ."
I don't understand this description and cannot find any explanations in detail or any examples about how to use this attribute.

ffreyer:
I would personally prefer if we pushed against using `model` as an attribute since working with a translation-rotation-scale matrix makes some things a lot easier internally than a completely general `Mat4f`. So maybe instead of documenting `model` better we should document translate!, rotate! & scale! better and push `plot.model` into a footnote.

bjarthur:
+1 for better documentation for `rotate!`.  specifically, if i render a Toggle in the middle of a figure, how do i rotate it about it's center?  see https://github.com/MakieOrg/Makie.jl/pull/4445#issuecomment-2402496785 for a failed attempt of mine.

EdsterG:
Another +1 for better documentation (and maybe even an updated implementation). Following up on #4445, I too can't figure out what the origin is for those toggles. I looked at all of the parent models and they are all identity matrices implying that the toggle should be drawn at the origin (0, 0), but it obviously isn't. So what transform is translating it to the middle of the screen?
```
t2.blockscene.transformation.model[]
t2.blockscene.parent.transformation.model[]
t2.blockscene.transformation.parent_model[]
t2.blockscene.transformation.parent[].model[]
t2.blockscene.transformation.parent[].parent_model[]
t2.parent.scene.transformation.model[]
```

As a suggestion to the translate/roatate API's, it would be helpful to have the following:
1) document the reference point
2) allow users to specify the reference point (see Unity API for example [Rotate](https://docs.unity3d.com/ScriptReference/Transform.Rotate.html)/[Translate](https://docs.unity3d.com/ScriptReference/Transform.Translate.html), both have a `relativeTo` parameter which defaults to self)

ffreyer:
Transformations affect plots, not scenes. At least not directly. Effectively Makie ends up calculating
```julia
plot.model * apply_transform(plot.transformation.transform_func, plot.converted)
```
where things default as
```julia
t = plot.transformation
plot.model = t.model = t.parent.model * transformationmatrix(t.translation, t.scale, t.rotation)
t.transform_func = t.parent.transform_func
```
[...]

PR #4472, opened 2024-10-10
Add origin to Transformation

ffreyer:
# Description

Related: #4445, #2637

This adds an `origin` to `Transformation` so that `rotate!()` and `scale!()` can be relative to something other than `Vec3(0)`. It can be set with `translate_origin!()` just like `translate!()`.

Example:

```julia
p = rand(Point2f)
rect = Rect2f(p .- 0.1, Vec2f(0.2))

fig = Figure()
sl = Slider(fig[1, 1], range = range(0, 2pi, length=101))
ax = Axis(fig[2, 1], aspect = DataAspect())
p1 = poly!(ax, rect, color = :orange)
p2 = lines!(ax, rect, color = :red)
xlims!(ax, -1.5, 1.5)
ylims!(ax, -1.1, 1.1)

on(sl.value) do angle
    center = to_ndim(Vec3f, p, 0)

    R = Makie.rotationmatrix_z(angle)
    T = Makie.translationmatrix(-center)
    Tinv = Makie.translationmatrix(center)
    p1.transformation.model[] = Tinv * R * T # Did we break setting plot.model?
    
    translate_origin!(p2, p)
    Makie.rotate!(p2, angle)
end

fig
```

## Type of change

## Checklist


docs/src/reference/generic/transformations.md
```diff
@@ -0,0 +1,146 @@
...
+You can set the transformation function of a plot by updating `plot.transformation.transform_func[] = new_func`.
+This will also change the transformation function of each child plot.
+Note that this will be reset when the plots parent transformation function changes, e.g. if `ax.xscale` is set in the example above.
+Another option is to create the plot with an explicitly given `Transformation` object, which can detach the plots transformation from its parent.
+
+## Scene Transformation
+
+The Scene also holds onto a `Transformation` object.
+It's `transform_func` acts as a default for any plot added to the scene.
+Its `model` transformation acts as a secondary transformation to any plot, i.e. it is applied aftr the plots own model transformation `scene.transformation.model[] * plot.transformation.model[]`.
```
bjarthur, line 63: after not aftr

docs/src/reference/generic/transformations.md
```diff
@@ -0,0 +1,146 @@
...
+For example, you may want to apply `transform_func` before running a triangulation algorithm in a recipe so that the triangulation doesn't get distorted.
+Or you may want to chain transformations of a series of plots to keep them relative to each other (see below).
+There are a few constructors which are helpful to know for this.
+
+If you want to fully detach a plot from its parents transformations, you can create it with `transformation = Transformation()`.
+If you want to remove only the `transform_func` but not model transformations, you can use `transformation = Transformation(parent, transform_func = identity)`.
+You can also pass different starting values for `translation`, `scale` and `rotation` to these functions.
+This will not affect whether the parents model transformations are considered.
+
+As an example, here is a two arms on a cart raising a box with a rope.
```
bjarthur, line 78: "here are two..." ?

bjarthur:
looking forward to this!

one other suggestion in addition to the typos:  i find the name `translate_origin!` a little confusing as the plot isn't actually getting translated.  how about `set_origin!` instead?  or maybe just `origin!`?  or, maybe `rotate!` and `scale!` could take an `origin` kwarg?

jkrumbiegel:
> maybe rotate! and scale! could take an origin kwarg?

I would prefer that I think, or maybe `around` so it's `rotate!(x, ..., around = p)`? Or `relative_to`

ffreyer:
Having origin as an argument in rotate!() and scale!() was also my first thought but I think with how transformations work it's better to pull it out. Transformations don't consider the order you call things in. You're effectively setting or changing the state with `translate!()`, `rotate!()` etc. To me `rotate!(..., origin = ...)` makes it seem like you're setting the origin for this specific call, while you would actually set it for the transformation as a whole. I think a separate function for origin reflects that better.

I'd be fine with renaming the function. I also had `set_origin!()` originally but switched when I implemented the `Accum/Absolute` interface since `set` sounds like just overwriting (no `origin = origin + arg` which Accum does)

PR #4474, opened 2024-10-11
Make textbox accept keypad enter as input confirmation as well.

thomvet:
I noticed that textbox reacts differently to "enter" that is next to the letters on the keyboard than to "enter" that is on the keypad (number) block. The normal enter serves as input confirmation, while the other one doesn't do anything.

I would have expected them to show the same behavior for the purpose of "confirming input" to the textbox.

This simple PR addresses the issue.


jkrumbiegel:
Thanks, looks good!

PR #4368, opened 2024-09-15
Make `Makie.mesh` compatible with MultiMesh (GeometryBasics refactor)

ffreyer:
# Description

This is test/prototype for using MetaMesh from https://github.com/JuliaGeometry/GeometryBasics.jl/pull/219 to render a mesh with materials loaded with https://github.com/JuliaIO/MeshIO.jl/pull/98. This is effectively a continuation of #4319.

With this branch (and the related MeshIO, ShaderAbstractions & GeometryBasics branches), using https://free3d.com/3d-model/girl-blind-703979.html:
```julia
filepath = "path/to/14-girl-obj/girl OBJ.obj"
metamesh = load(filepath)

# Fix filename, convert white -> alpha
img = map(FileIO.load("path/to/14-girl-obj/tEXTURE/FACE Base Color apha.png")) do c
    Makie.RGBA{Makie.N0f8}(c.r, c.g, c.b, 1 - c.r)
end
metamesh[:materials]["FACE"]["diffuse map"]["image"] = img

# this splits up the mesh, extracts material parameters, loads the remaining textures and generates multiple mesh plots
f, a, p = Makie.mesh(metamesh)
```

image

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #4477, opened 2024-10-14
Breaking release 0.22

### LINKED ISSUE #1369: Color uniform per element
Hello,

again... this is more a feature request but I generally use Mesh to plot fields of size "number of vertices" or "number of elements". It is easy to do the first one using Makie's Mesh function, but for the second one I had to create multiple polygons and call poly instead:

```
using GLMakie
import ColorSchemes.jet
triangles=[1 2 3;1 3 4]
x=[0. 1. 1. 0.]
y=[0. 0. 1. 1.]
ps = [Makie.GeometryBasics.Polygon([Point2(x[triangles[i,1]], y[triangles[i,1]]), Point2(x[triangles[i,2]], y[triangles[i,2]]), Point2(x[triangles[i,3]], y[triangles[i,3]])])
                             for i in 1:size(triangles,1)]
Makie.poly(ps, color = [1.;2.], colormap = jet)
```

Screen Shot 2021-10-06 at 4 07 20 PM

It works, but it would nice to have this done automatically by Mesh. I admit that I may be expecting too much of Mesh (e.g. to behave like MATLAB's patch function), with other features like edgecolor=true for example...

Thank you!
Mathieu

### LINKED ISSUE #2181: Order of decomposes coordinates is not counter/clockwise
The `Point2f`s that `decompose(Point2f, Rect(0, 0, 1, 1))` returns are not sorted in a clockwise order (nor counter clockwise). This is not ideal because plotting a decomposed rectangle doesn't work:
```julia
poly(decompose(Point2f, Rect(0, 0, 1, 1)))
```
tmp
nor does plotting polygons with a hole:
```julia
using GeometryBasics
p = Polygon(decompose(Point2f, Rect(0, 0, 1, 1)), [decompose(Point2f, Circle(Point2f(0.5), 0.25))])
poly(p)
```
tmp

Only specifying the coordinates manually:
```julia
p = Polygon(Point2f[(0,0), (1,0), (1,1), (0,1)], [decompose(Point2f, Circle(Point2f(0.5), 0.25))])
```
worked:
tmp

### LINKED ISSUE #3561: Some triangles are not rendered

Various platform info:
```
MacOS 14.2.1 (Apple M2)
Julia 1.10.0
GLMakie 0.9.5, GeometryBasics 0.4.9
```

Certain triangles - depending on the vertex ordering - are not visible when plotted with `GLMakie.mesh`.

For example, if the following is run interactively:
```julia
using GeometryBasics
using GLMakie
t_ok = Triangle(Point3(0, 0, 0), Point3(1, 0, 3), Point3(0, 3, 2))
t_bad = Triangle(Point3(1, 0, -1), Point3(-1, 0, 3), Point3(0, 3, 2))

mesh(t_ok)
```
Then a blue triangle is shown, as expected. 
image

But if instead `mesh(t_bad)` is called, then no triangle is visible.
image

I think it's something to do with vertex ordering, because:
```julia
t_permuted = Triangle(t_bad.points[2], t_bad.points[1], t_bad.points[3])
mesh(t_permuted)
```
then shows the expected triangle.
image

---

Tracing through what is happening inside the `mesh` call, the eventual problem seems to be that the following call to `decompose` in `GeometryBasics` return any faces (the first step at extracting the points works fine, though):
```julia
points = decompose(Point{3,Float32}, t_bad)  # This step is fine
decompose(GLTriangleFace, points)            # This returns an empty array
```

I'm not sure if the "bug" is in GLMakie for not re-ordering the vertices before calling the `decompose` function, hence I'm filing it here rather than in GeometryBasics -- apologies if this is incorrect.

### LINKED ISSUE #4344: `voronoiplot` default colours depend too much on the generator order

### LINKED ISSUE #4383: Scatter `marker_offset` is broken
I get this in CairoMakie and GLMakie with Makie v0.21.11, the dots should be separated.

```julia
scatter([1, 1, 1], [1, 1, 1], marker_offset = [(10, 10), (0, 20), (-15, -15)])
```

image

ffreyer:
# Description

PR for the next breaking release

Merge TODO:
- [x] #4319
- [x] #4357
- [x] #4496 (originally #4368)
- [x] #4564
- ~~maybe #4399~~ (waiting on plot update refactor)
- [x] #4594
- [x] #4606
- [x] #4131
- [x] #4615

To fix:
- [x] partially reversed 2x2 surface fails `surface(210..180, 80..110, rand(2, 2))` in WGLMakie (picking test)

Closes CompatHelper prs:
closes #4635, closes #4634, closes #4628, closes #4623, closes #4622, closes #4620, closes #4616, closes #4610, closes #4605, closes #4600

Actual Issues:
- fixes #3561
- fixes #2181 (2D primitives are now counter clockwise)
- closes #1369 (FaceViews offer a way to do per-face colors)
- fixes #4344
- fixes #4383


ffreyer:
One of the failures here is that render order of the Axis3 framelines has changed in CairoMakie:
| old | new |
| --- | --- |
| Screenshot 2024-11-14 195932 | Screenshot 2024-11-14 195920 |

This is the case for all Axis3 refimages and also on master. I updated this one refimage for now.

SimonDanisch:
>This is the case for all Axis3 refimages and also on master.

Wait isn't that a regression, that was only recently reported?
Seems to me like we should try to figure out what's going on instead of updating the reference images. Or am I missing something?

ffreyer:
Idk if it has been reported. I just updated the refimg here because it has nothing to do with the breaking release

ffreyer:
Reverted the refimg update after merging #4591

The display times show opposite behavior for me locally. With
```julia
using GLMakie
macro ctime(x)
    return quote
        tstart = time_ns()
        $(esc(x))
        Float64(time_ns() - tstart)
    end
end
fig = scatter(1:4; color=1:4, colormap=:turbo, markersize=20, visible=true);
GC.gc()
@ctime display(fig)
```
I get ~630ms on master and ~380ms on this branch. Number of stale instances also went down

SimonDanisch:
I can confirm this! Really weird, considering that I should be using the same Julia version and this regression popping up quite reproducible on the CI.
It's pretty suspicious, that the numbers on the CI seem to match my local numbers quite closely, but the other way around.
I did check the using times as well, but they seem to match pretty well with what the CI measures, so those aren't flipped...
Not sure what's going on here, I guess it would be interesting to upload the `tinv` flamegraphs for the compile time benchmarks, since they show most accurately what's going on.

ffreyer:
"Moving Plots in SpecApi" has different grid lines in the center left (empty) Axis. Seems more correct to me, and might not be related to this pr. This doesn't trigger a failure and I didn't update the refimg yet

SimonDanisch:
There is still at least one problem left here which needs fixing:
```julia
poly(Circle(Point2f(0), 1f0), strokewidth=1, strokecolor=:black)
```
image

I'm still going to merge this, to already start rebasing existing PRs against this.

PR #4485, opened 2024-10-15
fix low hanging fruits for render performance

SimonDanisch:
# Description
```julia
using GLMakie
using BenchmarkTools
f, ax, pl = scatter(1:5);
sc = display(f)
@btime GLMakie.render_frame(sc)
```
## Makie master
`118.770 μs (1193 allocations: 30.41 KiB)`
## With sorting change
`104.764 μs (729 allocations: 25.48 KiB)`
## With framebuffer_size optimization
`85.648 μs (727 allocations: 25.45 KiB)`


ffreyer:
This is failing because lines drop the `model` uniform (due it not being used) when linestyles are used

ffreyer:
Same benchmark code, different sorting options:

| Change/State | time [µs]| Allocations | Allocated KiB |
| --- | --- | --- | --- |
| merge master | 77.2µs | 727 | 25.45 |
| rollback sortby function | 90.8 | 1191 | 30.38 |
| typed sortby | 91.2 | 1191 | 30.38 |
| inline `transformationmatrix(plot)` | 79.6 | 758 | 29.81 |
| inline `plot.model[]` | 80.7 | 696 | 21.09 |
| using `zvalue2d` with `transformationmatrix()` | 84.6 | 820 | 23.03 |

Calling `zvalue2d` seems to have a significant overhead compared to calling what it does directly. Perhaps due to runtime dispatch? `plot.model[]` is also slower than `transformationmatrix()`, maybe due to type stability. I went with calling `transformationmatrix()` for now, which is pretty close to the original optimization.

Moving around some clip planes code to make `setup_clip_planes()` type stable got me 53.3µs, 272 allocation, 18KiB. Waiting on CI before I push that.

GLMakie/src/rendering.jl
```diff
@@ -31,7 +31,12 @@ function render_frame(screen::Screen; resize_buffers=true)
     ShaderAbstractions.switch_context!(nw)
 
     function sortby(x)
-        return x[3][:model][][3, 4]
+        robj = x[3]
+        plot = screen.cache2plot[robj.id]
```
SimonDanisch, line 35: The plot lookup was the expensive bit here, so would be nice if we could avoid it!

ffreyer: Dict lookups are O(1) and somewhere around 10ns. My benchmarks for the current solution only have 3% difference to the old one.
Using `robj[:model]` is also what made this pr fail, because lines with linestyles apply it on the CPU so it does not end up in uniforms. I would also expect this to subtly fail with f32 converts because those can change the model matrix (set it to I after applying it on the CPU)

PR #4505, opened 2024-10-21
Move favicon and logo to where DocumenterVitepress expects them

jkrumbiegel:


asinghvi17:
It should be searching in assets as well, looks like that broke somehow.  Will look into it today.

PR #4515, opened 2024-10-23
Make benchmark plots and post a gist with them

jkrumbiegel:


jkrumbiegel:
@SimonDanisch this works now, a link is in the `Benchmark Results` status under Details. For some reason it's still prone to false positives even in purely visual form, as there shouldn't be differences vs master on this branch here.

Also, it's not as visible as the comment system we had before, we could in principle bring the commenting back but link the images from the gist. I think that works, here's an example link, the format is `https://gist.githubusercontent.com/MakieBot/e8365e6c46abbd310446c463f3cc893d/raw/CairoMakie.svg`

PR #4521, opened 2024-10-24
fix pick_sorted for WGLMakie and hover over heatmap(Resampler(data))

SimonDanisch:


ffreyer:
I added another mesh to the picking image so I can more easily run pick_closest tests from different directions. So that is an expected failure.

Issue #219, opened 2018-11-08
RecordEvents missing


AshtonSBradley:
running example http://makie.juliaplots.org/v1.0/examples-heatmap.html#Mouse-Picking-1 leads to error:

```julia
julia> RecordEvents(scene, "./docs/media/add_and_change_points")
ERROR: UndefVarError: RecordEvents not defined
Stacktrace:
 [1] top-level scope at none:0
```
(on master)

AshtonSBradley:
ok, I see that the intention was not to supply that function. I just point out that in this example:
http://makie.juliaplots.org/stable/examples-slider.html#sliders-1

one has to extract the relevant code from `RecordEvents`, i.e. a runnable example would instead end with
```julia
hbox(p, vbox(s1, s2), parent = Scene(resolution = (500, 500)))
```

SimonDanisch:
I added an overload for display to `RecordEvents` and exported it, so this should work now

PR #4522, opened 2024-10-24
correct spelling

t-bltg:
# Description

This is a code quality PR (hand corrections) fixing spelling typos (mostly in comments).

There are two ambiguities to be discussed:
- [`Tesselation`](https://github.com/MakieOrg/Makie.jl/blob/445d7aea5035734299f265f22a3e794e68c415c5/src/basic_recipes/wireframe.jl#L18) should be spelled `Tessellation`, but since those are variables, I haven't corrected them in `Makie`;
- [`independant`](https://github.com/MakieOrg/Makie.jl/blob/445d7aea5035734299f265f22a3e794e68c415c5/docs/src/explanations/backends/wglmakie.md?plain=1#L109) should be spelled `independent`, but this relates to a [function](https://github.com/SimonDanisch/Bonito.jl/blob/c35fe958f0f45bf8ee20ea87b48a2cc94faae281/src/export.jl#L3) in `Bonito.jl` so I haven't modified it to avoid disruption.

## Type of change

This is a NFC.


EdsterG:
`Tesselation` is a stuct defined in GeometryBasics.jl. There's a big breaking refactor currently happening, related to GeometryBasics, so might be a good time to correct the spelling? See #4477, #4319, #219.

t-bltg:
> Tesselation is a stuct defined in GeometryBasics.jl. 

Thanks for the clarification !
I also think this is best to change it, but I prefer to leave it for another PR, since this is breaking.

ffreyer:
I cancelled the WGLMakie runs here because they were full of timeouts. I tried running CI on a different branch (https://github.com/MakieOrg/Makie.jl/actions/runs/11531218386/job/32101665605?pr=4472) and did not get timeouts, but rerunning here still produced them.
The changes here are just error strings and comments, right?

t-bltg:
> The changes here are just error strings and comments, right?

Except a few variables local renames here, https://github.com/MakieOrg/Makie.jl/pull/4522/files#diff-b49aa74de270f83f7d96f661d23e3e1044de710184415a64e5db70890a671841R225-R259, which have no impact elsewhere.

t-bltg:
> Seems like the only change in the WGLMakie source to me that could possibly have an effect on anything, if interpolation of that string into javascript has some escaping bugs.

Wow that was a nice catch !

jkrumbiegel:
@SimonDanisch looks like that single quote was the reason for the hangs, probably worth looking into the JS interpolation mechanism.

jkrumbiegel:
Thanks @t-bltg, we might also want to make typo catching more automatic in the future, I know other repos have added such CI workflows in the past and I guess you also used a tool rather than reading everything in the repo?

t-bltg:
Of course there were a lot of false positive, but I've used `codespell --skip='*.js,*.vert,*.ai,*.frag,*.geom' .` as a base here.

jkrumbiegel:
@t-bltg we can try that in a separate PR, depends on how well we can cut down the false positive rate. Maybe if detection is only run on the diff.

jkrumbiegel:
This is fine, one WGLMakie test is currently flaky, the rerun is not needed for me @t-bltg

SimonDanisch:
Is our ReferenceUpdater handling renaming correctly?

jkrumbiegel:
> Is our ReferenceUpdater handling renaming correctly?

When you rename plots they appear in both the missing refimages and the new refimages section, so yeah that works

PR #4531, opened 2024-10-25
Custom interactions

### LINKED ISSUE #3950: FR: Add ability to change the "reset zoom" binding
MacOS seems quite adamant about claiming “Ctrl+Left Click” for “Right Click” and I can’t find any way to alter that, so if I lose my place when zooming I have no (easy) way to get back to the starting position.  It would be great if the "reset zoom" binding could be changed to something else.  Apparently it is adjustable for PolarAxis already, so maybe that could be copied more broadly.

EdsterG:
# Description

Partially Fixes #3950

The goal of this PR is to allow users to change default interactions, by modifying a global registry.

- Define a global registry of default interactions for Axis and Axis3D.
- Update `RectangleZoom` to use the registration functions `registration_setup!` and `deregistration_cleanup!`.
- Update `RectangleZoom` deactivate when modifier key (if one is defined) is released before the mouse.
- Allow `LimitReset` to be customizable.

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/types.jl
+++ b/src/makielayout/types.jl
@@ -149,10 +149,10 @@ mutable struct LineAxis
     minortickvalues::Observable{Vector{Float32}}
 end
 
+
 struct LimitReset end
 
 mutable struct RectangleZoom
-    callback::Function
     active::Observable{Bool}
     restrict_x::Bool
     restrict_y::Bool
@@ -162,8 +162,8 @@ mutable struct RectangleZoom
     modifier::Any # e.g. Keyboard.left_alt, or some other button that needs to be pressed to start rectangle... Defaults to `true`, which means no modifier needed
 end
 
-function RectangleZoom(callback::Function; restrict_x=false, restrict_y=false, modifier=true)
-    return RectangleZoom(callback, Observable(false), restrict_x, restrict_y,
+function RectangleZoom(restrict_x=false, restrict_y=false, modifier=true)
+    return RectangleZoom(Observable(false), restrict_x, restrict_y,
                          nothing, nothing, Observable(Rect2d(0, 0, 1, 1)), modifier)
 end
 
@@ -190,9 +190,8 @@ function DragPan(reset_delay)
     return DragPan(RefValue{Union{Nothing, Timer}}(nothing), RefValue{Union{Automatic, Float64}}(0.0), RefValue{Union{Automatic, Float64}}(0.0), reset_delay)
 end
 
+struct DragRotate end
 
-struct DragRotate
-end
 
 struct ScrollEvent
     x::Float32
@@ -679,8 +678,8 @@ Axis(fig_or_scene; palette = nothing, kwargs...)
     end
 end
 
-function RectangleZoom(f::Function, ax::Axis; kw...)
-    r = RectangleZoom(f; kw...)
+
+function registration_setup!(ax::Axis, r::RectangleZoom)
     rect_scene = Scene(ax.scene)
     selection_vertices = lift(_selection_vertices, rect_scene, Observable(ax.scene), ax.finallimits,
                               r.rectnode)
@@ -694,17 +693,15 @@ function RectangleZoom(f::Function, ax::Axis; kw...)
         inspectable = false, transparency=true, overdraw=true, visible=r.active)
     # translate forward so selection mesh and frame are never behind data
     translate!(mesh, 0, 0, 1000)
-    return r
+
+    return ax
 end
 
-function RectangleZoom(ax::Axis; kw...)
-    return RectangleZoom(ax; kw...) do newlims
-        if !(0 in widths(newlims))
-            ax.targetlimits[] = newlims
-        end
-        return
-    end
+function deregistration_cleanup!(ax::Axis, r::RectangleZoom)
+    # TODO: Remove mesh?
+    return ax
 end
+
 
 """
 Create a colorbar that shows a continuous or categorical colormap with ticks
--- a/src/makielayout/types.jl
+++ b/src/makielayout/types.jl
@@ -150,7 +150,19 @@ mutable struct LineAxis
 end
 
 
-struct LimitReset end
+struct LimitReset
+    mouseevent::MouseEventTypes.MouseEventType # e.g. MouseEventTypes.leftclick, or some other mouse event to start limit reset.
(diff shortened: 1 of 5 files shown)


EdsterG:
Any high level thoughts on this @ffreyer?

EdsterG:
@SimonDanisch I think Makie is very close to becoming the goto library for interactive dashboard. In my opinion, allowing users to customize interactions will help solidify Makie as the best library for complex dashboard. Here's a simple use case which requires the default interactions to change. Right click to drag/pan, left click to control the vertical bar, modifier + left click to box zoom.

https://github.com/user-attachments/assets/a039a643-f960-4d0c-8262-f5e152532767

SimonDanisch:
Thanks! 
I think this should be part of the theme, shouldn't it?

EdsterG:
I think interactions can go both ways, part of the theme or not part of the theme. They aren't currently part of theme. What's your preference? If you prefer having them be part of the theme, how do you suggest they be added? Doing something like this `Axis3 = (interactions=Dict(...),)` feels a bit unnatural and clunky because it's bypassing the `register`/`deregister` interface.

EdsterG:
@SimonDanisch any suggestions for how to make this part of the theme? Or shall it be left as is for now?

src/makielayout/types.jl
```diff
@@ -162,8 +174,8 @@ mutable struct RectangleZoom
     modifier::Any # e.g. Keyboard.left_alt, or some other button that needs to be pressed to start rectangle... Defaults to `true`, which means no modifier needed
 end
 
-function RectangleZoom(callback::Function; restrict_x=false, restrict_y=false, modifier=true)
```
SimonDanisch, line 165: Why remove the callback? It's there to make custom RectangleZoom interactions...
Should be better documented, I suppose...

EdsterG: I noticed the interactions had the following unused interface:
```
function registration_setup!(parent, interaction)
    # do nothing in the default case
end

function deregistration_cleanup!(parent, interaction)
    # do nothing in the default case
end
```
So I updated `RectangleZoom` to use that interface, it makes dealing with interactions much easier. The callback seemed like a round-about thing.

The callback used to be defined here:
```
function RectangleZoom(ax::Axis; kw...)
    return RectangleZoom(ax; kw...) do newlims
        if !(0 in widths(newlims))
            ax.targetlimits[] = newlims
        end
        return
    end
```
and called inside `process_interaction`:
```
    elseif event.type === MouseEventTypes.leftdragstop
        try
            r.callback(r.rectnode[])
        catch e
            @warn "error in rectangle zoom" exception=(e, Base.catch_backtrace())
        end
        r.active[] = false
        return Consume(true)
```
From what I can tell it was just passing a reference to `ax` in a round-about way. This pattern is also error prone because there's some `ax` referenced at construction and another in `process_interaction`. It doesn't make sense for them to be different axes.

Now it just modifies `ax` inside `process_interaction` directly:
```
    elseif event.type === MouseEventTypes.leftdragstop
        newlims = r.rectnode[]
        if !(0 in widths(newlims))
            ax.targetlimits[] = newlims
        end
        r.active[] = false
        return Consume(true)
```

EdsterG: For what it's worth, rectangle zoom still works with the change in this PR.

EdsterG: Would you point me to a use case where a custom callback is required for RectangleZoom? Prior to this PR there was no way for users to customs RectangleZoom without modifying the source code (unless I'm missing something).

SimonDanisch: I introduced this to be able to register your own interaction for rectangle zoom:
```julia
f, ax, pl = scatter(1:4)
deregister_interaction!(ax, :rectanglezoom)
zoom = Makie.RectangleZoom(ax) do rect
    @show rect # do something with rectangle
end
register_interaction!(ax, :rectanglezoom, zoom)
```
I admit it's a bit awkward how it uses the interface.
Maybe we can keep how you refactored it and figure out a way how to get the best of both worlds.
Maybe an optional user callback, and otherwise it uses the axis from the setup?

EdsterG: Sure, I can add an optional user callback. Though I would prefer the callback to take an `ax` argument for the reason mentioned above, if you're ok with that.
```
f, ax, pl = scatter(1:4)
deregister_interaction!(ax, :rectanglezoom)
zoom = Makie.RectangleZoom(ax) do ax, rect
    @show rect # do something with rectangle
end
register_interaction!(ax, :rectanglezoom, zoom)
```

EdsterG: By the way, with this PR it'll be super easy for something to create `RectangleZoomWithSpecialFeature` and make that the new default behavior (or just register it to a specific plot as in your example).

EdsterG:
@SimonDanisch I finally got the chance to re-add the callback. You mind taking another look at this PR?

EdsterG:
@SimonDanisch I've merged in latest main and would like to work with you on any lingering concerns you may have about customizable interactions.

EdsterG:
@SimonDanisch any more thoughts on the direction of this PR?

SimonDanisch:
Looks fine from my side. @jkrumbiegel what do you think?

EdsterG:
@jkrumbiegel any thoughts on the direction of this PR?

EdsterG:
@jkrumbiegel hope you had a good weekend. Any thoughts you might have about the direction of this PR would be helpful.

jkrumbiegel:
I think I agree with Simon's earlier assessment that the default interactions should be part of the theme somehow, at least that would keep the global state we have more confined. There could always be helper functions to interact with the theme if somehow a more complex structure is required that is annoying to set up. In principle I like the idea that interactions can be predefined if someone prefers different ones from the default. Populating the first state from the theme also would not necessarily invalidate the registration interface, you can still register more on the fly and deregister the ones that have been populated. The theme just decides the initial set.

EdsterG:
Ok, would you please give me some guidance? I'm not immediately sure how best to incorporate this with the theme. I also had an open question and an unaddressed concern about the theme route:
> how do you suggest they be added?

> Doing something like this Axis3 = (interactions=Dict(...),) feels a bit unnatural and clunky because it's bypassing the register/deregister interface.

PR #4538, opened 2024-10-28
Improve closeall and renderloop task handling

SimonDanisch:
This makes sure, that `close_renderloop` can be accessed safely from different tasks and removes unnecessary loops.

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -661,11 +661,12 @@ Doesn't destroy the screen and instead frees it to be re-used again, if `reuse=t
 function Base.close(screen::Screen; reuse=true)
     @debug("Close screen!")
     set_screen_visibility!(screen, false)
-    empty!(screen)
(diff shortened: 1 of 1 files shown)


GLMakie/src/screen.jl
```diff
@@ -164,7 +164,7 @@ mutable struct Screen{GLWindow} <: MakieScreen
     shader_cache::GLAbstraction.ShaderCache
     framebuffer::GLFramebuffer
     config::Union{Nothing, ScreenConfig}
-    stop_renderloop::Bool
+    stop_renderloop::Threads.Atomic{Bool}
```
vchuravy, line 167: What is the minimum version of Julia you require? `@atomic` is nicer these days,

SimonDanisch: We still run CI on 1.6, which we may want to change at some point, but right now we still try to support 1.6 as long as it's not too bothersome.

PR #4552, opened 2024-10-31
try running benchmarks on self hosted

SimonDanisch:


SimonDanisch:
Two take aways:

1) It's suprising how similar the noise is for a complete dedicated machine, that has nothing else to do but run the benchmark on linux, even with the tips from BenchmarkTools applied : https://juliaci.github.io/BenchmarkTools.jl/stable/linuxtips/.
We probably still need to make sure that we avoid thermal throttling (it's a mobile chip).
2) It's crazy how much faster a dedicated machine is (from 25 min to 5min CI runtime), likely because the whole julia folder stays, without a lengthy up/download for the action cache and I guess a few other factors, like better CPU speed and GPU support.

I will close this for now, and come back to it later, when we have a dedicated machine for CI (still needs some containerisation + a stable location).

PR #4554, opened 2024-10-31
speed up apply_theme

SimonDanisch:
After [#S.Colorbar](https://github.com/MakieOrg/Makie.jl/pull/4520) sped up default_theme, we can save a bit more time by not creating attributes.
May be slightly breaking and could go into https://github.com/MakieOrg/Makie.jl/pull/4550


SimonDanisch:
This will be addressed in https://github.com/MakieOrg/Makie.jl/pull/4630

PR #4558, opened 2024-11-01
add backend and update_state kwargs to show

bjarthur:
as [discussed](https://discourse.julialang.org/t/correct-way-to-save-makie-plot-to-iobuffer/87793) `show` is the way to "save" a figure to an IOBuffer.  but if the figure was made with say WGLMakie and you want to save as say PDF then it's a bit cumbersome because you have to `CairoMakie.activate!` and then `WGLMakie.activate!` after `show`ing.  this PR make it more concise by adding a `backend` kwarg to `show`, just like `save` has one already.  so one can simply `show(...; backend=CairoMakie)` instead of activating and deactivating it.
 
## Type of change

## Checklist

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -244,7 +244,7 @@ function Base.show(io::IO, m::MIME"text/markdown", fig::FigureLike)
     throw(MethodError(show, io, m, fig))
 end
 
-function Base.show(io::IO, m::MIME, figlike::FigureLike; backend = current_backend(), update_state=true)
+function Base.show(io::IO, m::MIME, figlike::FigureLike; backend = current_backend(), update=true)
     if ALWAYS_INLINE_PLOTS[] == false && m isa MIME_TO_TRICK_VSCODE
         # We use this mime to display the figure in a window here.
         # See declaration of MIME_TO_TRICK_VSCODE for more info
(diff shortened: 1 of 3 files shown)


bjarthur:
do any other `show` methods defined by Makie need to have a backend kwarg added as well?

also, i notice that `show` is not tested at all.  should i bother to add one?

SimonDanisch:
Thank you, seems reasonable!

> should i bother to add one?

Yeah why not :)

bjarthur:
all i can think to test is for the presence of a `backend` kwarg to `show`.

ideally i'd like to test that specifying a different backend works, but that would require putting the test in one of the backend repos and adding a dependency to one of the other backends.  you don't want to go that far, do you?

bjarthur:
ready for review

bjarthur:
the failed test on WGLMakie  for julia 1.6 is a memory leak.  is that related to this PR?

bjarthur:
just added another kwarg to make reseting the limits optional when `show`'ing.

src/display.jl
```diff
@@ -244,17 +244,16 @@ function Base.show(io::IO, m::MIME"text/markdown", fig::FigureLike)
     throw(MethodError(show, io, m, fig))
 end
 
-function Base.show(io::IO, m::MIME, figlike::FigureLike)
+function Base.show(io::IO, m::MIME, figlike::FigureLike; backend = current_backend(), update_state=true)
```
ffreyer, line 247: I think just `update` would be better since that's what we use in `display()`

bjarthur: oh, thanks!  i didn't notice that.  changed in a new commit.

PR #4563, opened 2024-11-04
HTML video size annotation

jkrumbiegel:
png images written out via inline html are already sized using their logical size (not final bitmap size) given `px_per_unit`. This PR adds the same for videos so that you get this look by default in environments like quarto:

image

and not this instead

image

### DIFF
--- a/src/recording.jl
+++ b/src/recording.jl
@@ -179,13 +179,18 @@ function Record(func, figlike, iter; kw_args...)
 end
 
 function Base.show(io::IO, ::MIME"text/html", vs::VideoStream)
+    scene = vs.screen.scene
+    if !(scene isa Scene)
+        error("Expected Screen to hold a reference to a Scene but got $(repr(scene))")
+    end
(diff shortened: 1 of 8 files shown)


src/recording.jl
```diff
@@ -179,13 +179,18 @@ function Record(func, figlike, iter; kw_args...)
 end
 
 function Base.show(io::IO, ::MIME"text/html", vs::VideoStream)
+    scene = vs.screen.scene
+    if !(scene isa Scene)
```
SimonDanisch, line 183: Why not `size(screen)`?

SimonDanisch (changes requested):

src/recording.jl
jkrumbiegel: seemed to return the framebuffer size in my test

jkrumbiegel: image

ffreyer:
Why change root_scene to scene? Isn't that just less descriptive?

PR #4570, opened 2024-11-05
Add `:solid_pattern` linestyle to allow switching between patterned and solid lines

### LINKED ISSUE #803: MethodError updating Observable linestyle
Looks like `linestyle` can't be updated at all. Example:
```
style = Observable(:dot)
scene = lines(rand(5), linestyle = style)
display(scene)
style[] = :solid 
# MethodError: Cannot `convert` an object of type Nothing to an object of type Array{Float64,1}
```
If `:dot` and `:solid` are swapped in this example, there's a different error:
```
MethodError: convert(::Type{Union{}}, ::Array{Float64,1}) is ambiguous
```

### LINKED ISSUE #3693: Cannot set linestyle to :solid with observable
When using an observable to change the linestyle of lines in an axis, it works for the :dot, :dash, :dotdash styles, but not for the :solid style (throws an error). Here is a short code that generates the error, with the error message, when running the code with Julia v1.10.2 and GLMakie v0.9.9.

```julia
using GLMakie

f = Figure()
ax = Axis(f[1, 1])
display(f)

ls = Observable{Symbol}(:dot)
lines!(ax, 1:100, rand(100), linestyle=ls)

ls[] = :dash  # ok
ls[] = :dashdot  # ok
ls[] = :dashdotdot  # ok
ls[] = :solid  # ERROR

ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type Vector{Float32}

Closest candidates are:
  convert(::Type{Array{T, N}}, ::StaticArraysCore.SizedArray{S, T, N, N, Array{T, N}}) where {S, T, N}
   @ StaticArrays ~/.julia/packages/StaticArrays/EHHaF/src/SizedArray.jl:88
  convert(::Type{Array{T, N}}, ::StaticArraysCore.SizedArray{S, T, N, M, TData} where {M, TData<:AbstractArray{T, M}}) where {T, S, N}
   @ StaticArrays ~/.julia/packages/StaticArrays/EHHaF/src/SizedArray.jl:82
   @ Base Base.jl:84
  ...

Stacktrace:
  [1] setproperty!(x::Observable{Vector{Float32}}, f::Symbol, v::Nothing)
    @ Base ./Base.jl:40
  [2] setindex!(observable::Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:122
  [3] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
  [4] #invokelatest#2
    @ ./essentials.jl:892 [inlined]
  [5] invokelatest
    @ ./essentials.jl:889 [inlined]
  [6] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
  [7] setindex!(observable::Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
  [8] (::Observables.SetindexCallback)(x::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:148
  [9] #invokelatest#2
    @ ./essentials.jl:892 [inlined]
 [10] invokelatest
    @ ./essentials.jl:889 [inlined]
 [11] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [12] setindex!(observable::Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [13] top-level scope
    @ ~/Desktop/test.jl:19

```

ffreyer:
# Description

Fixes #3693
Fixes #803

This linestyle generates a solid line pattern so it renders with shaders used for linestyles but never actually stops drawing. This enables switching between a solid pattern and other patterns (e.g. :dot, :dash, etc).

We could also adjust `:solid` to do that, though that would make `linestyle = :solid` slower than `nothing` (and potentially different if there are any bugs/peculiarities in the linestyle shaders).

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 - Allow plots to move between scenes in SpecApi [#4132](https://github.com/MakieOrg/Makie.jl/pull/4132).
 - Added empty constructor to all backends for `Screen` allowing `display(Makie.current_backend().Screen(), fig)` [#4561](https://github.com/MakieOrg/Makie.jl/pull/4561).
 - Added `subsup` and `left_subsup` functions that offer stacked sub- and superscripts for `rich` text which means this style can be used with arbitrary fonts and is not limited to fonts supported by MathTeXEngine.jl [#4489](https://github.com/MakieOrg/Makie.jl/pull/4489).
-- Added the `jitter_width` and `side_nudge` attributes to the `raincloud` plot definition, so that they can be used as kwargs [#4517]https://github.com/MakieOrg/Makie.jl/pull/4517)
+- Added the `jitter_width` and `side_nudge` attributes to the `raincloud` plot definition, so that they can be used as kwargs [#4517](https://github.com/MakieOrg/Makie.jl/pull/4517)
 - Expand PlotList plots to expose their child plots to the legend interface, allowing `axislegend`show plots within PlotSpecs as individual entries. [#4546](https://github.com/MakieOrg/Makie.jl/pull/4546)
 - Implement S.Colorbar(plotspec) [#4520](https://github.com/MakieOrg/Makie.jl/pull/4520).
 - Fixed a hang when `Record` was created inside a closure passed to `IOCapture.capture` [#4562](https://github.com/MakieOrg/Makie.jl/pull/4562).
+- Added `linestyle = :solid_pattern` to allow switching between non-solid and (effectively) solid linestyles. [#4570](https://github.com/MakieOrg/Makie.jl/pull/4570)
 
 ## [0.21.15] - 2024-10-25
 
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -351,7 +351,7 @@ Creates a connected line plot for each element in `(x, y, z)`, `(x, y)` or `posi
     "Sets the width of the line in screen units"
     linewidth = @inherit linewidth
     """
-    Sets the dash pattern of the line. Options are `:solid` (equivalent to `nothing`), `:dot`, `:dash`, `:dashdot` and `:dashdotdot`.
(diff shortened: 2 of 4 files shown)


SimonDanisch (approved):

jkrumbiegel:
Would it make sense to special case in the shader when the numbers appear that correspond to `:solid` or `:solid_pattern`? To me it seems bad to have both just because it's a peculiarity of our shader setup that you can't switch. Rendering doesn't have to be slow if it picks the fast branch right away?

ffreyer:
I'm not sure what you mean. Currently :solid converts to nothing and results in solid line shaders. :dot, :dash, etc converts to an array which results in patterned line shaders. :solid_pattern follows the latter conversion path. There is nothing else that plays into that decision.

We could set up some logic to convert :solid to an array when patterned line shaders are chosen, but that only works if you start with :dot etc then. If you start with :solid or nothing, we'd have to predict the future

jkrumbiegel:
I mean turn `:solid` into an array but check that array in the shader as a special fast path. Then the users don't have to pick "fast vs slow but exchangeable"

ffreyer:
Hmm, that could work too, but it would require padding position of solid lines to Point4f. (Because with linestyles the projections happen on the CPU and clipping won't work if you do `p4d[Vec(1, 2, 3)] / p4d[4]` beforehand.) I don't know how much of a performance impact that has...

ffreyer:
This seems rather convoluted to set up right now, compared to the plot update branch. I'll probably try it there instead

SimonDanisch:
Closing in favor of finding a solution in #4630

PR #4583, opened 2024-11-11
Only use automatic unit conversion for simple units (not compound units)

### LINKED ISSUE #3891: Disable Automatic Unit Conversion
This is half feature request and half bug...

I'd like a way to disable unit conversion in axes and instead just make the unit whatever is passed in. Also, the conversion seems to be converting to invalid units. An example for both is:
```julia
scatter((400:100:800)u"nm", (1:5)u"W/(m^2*µm*sr)") 
```
image

Feature: For the X axis, nm are converted to µm. I'd like to disable this and plot the passed in units.
Bug: For the Y axis, W/(m^2 µm sr) are converted to nm^-1, which is not a valid unit conversion.

### LINKED ISSUE #4063: Wrong units
```jl
  [13f3f980] CairoMakie v0.12.5
  [1986cc42] Unitful v1.21.0


julia> scatter(rand(5)u"m/s", rand(5)u"m/s")
```

The units are wrong.

image

ffreyer:
# Description

~~This pr removes code attempting to find the best units for the given user input/limits. As noted in #4063 this code (always?) leads to incorrect units for compound units, e.g. `u"m/s"`. It also seems like a difficult problem to determine how to distribute prefixes or if they are even wanted.~~

After a brief discussion with Simon I changed this pr to only affect compound units, i.e. units like `u"m/s"`. These units now no longer try to find an optimal prefix, which led to them becoming wrong. They instead get set once and then stay what they are. Simple units like `u"m"` continue to optimize their prefix.

The reason for the change in direction is that you can already turn off the prefix conversions by creating a UnitfulConversion with the unit you want. E.g.
```julia
uc = Makie.UnitfulConversion(u"m")
scatter(1:4, (10:10:40) .* u"km"; axis=(dim2_conversion=uc,))
```

Results:
- compound units are no longer simplified, e.g. `(1:10) .* u"nm/m^2` results in unit `u"nm/m^2` 
- prefixes of compound units are no longer adjusted, e.g. `1e6 u"nm/m^2"` keeps `u"nm/m^2"`
- if prefixes are changed from plot to plot, the initial prefixes are used for compound units. E.g. if the first plot uses `1200u"nm/m^2 "` and the second `1.3u"µm/m^2"`, `u"nm/m^2"` will still be used.

I also added some extra code to convert a unit string with exponents to rich text with superscript.

Added refimg:

```julia
begin
    # Don't swallow units past the first
    f, a, p = scatter((1:10) .* u"J/s")
    # Don't simplify (assume the user knows better)
    scatter(f[1, 2], (1:10) .* u"K", exp.(1:10) .* u"mm/m^2")
    # Only change prefixes of simple units, not compound units
    scatter(f[2, 1], 10 .^ (1:6) .* u"W/m^2", (1:6) .* 1000 .* u"nm")
    # Only change units/prefixes for simple units when adding more plots
    scatter(f[2, 2], (0:10) .* u"W/m^2", (0:10) .* u"g")
    scatter!((0:10) .* u"kW/m^2", (0:10) .* u"kg")
    f
end

```

image

Closes #3891
Fixes #4063

## Type of change

## Checklist


ffreyer:
Maybe slightly breaking for units with powers because of rich text addition. I.e. `u"m^3"` should now render as m³ instead of m^3

SimonDanisch:
>u"m^3" should now render as m³ instead of m^3

I wouldn't consider this as breaking

PR #4584, opened 2024-11-11
Don't restrict transformed point dim to input point dim in `poly_convert`

asinghvi17:
# Description

Fixes https://github.com/MakieOrg/GeoMakie.jl/issues/282

Previously, if you had 2D polygon and a 2D -> 3D transform func, the points would be transformed, and the resulting Point3 would be coerced back down to Point2.  This is likely also why a lot of the triangulation was wrong.

This PR removes the element type constraint on the array returned from `apply_transform`, thus allowing 3D points to stay 3D.

I have a suspicion that if we merge this, it will break plotting polygons in 3d.  It might need a very basic generic overload of `GeometryBasics.earcut_triangulate` for arbitrary dimensions, to restrict to 2D - at least for now.

## Type of change

## Checklist


src/basic_recipes/poly.jl
```diff
@@ -90,7 +90,9 @@ function poly_convert(polygon::Polygon, transform_func=identity)
     outer = metafree(coordinates(polygon.exterior))
     # TODO consider applying f32 convert here too. We would need to identify this though...
     PT = float_type(outer)
-    points = Vector{PT}[apply_transform(transform_func, outer)]
+    # Note that this should not be coerced to be a `Vector{PT}`,
+    # since `apply_transform` can change points from e.g 2D to 3D.
+    points = [apply_transform(transform_func, outer)]
     points_flat = PT[outer;]
     for inner in polygon.interiors
```
SimonDanisch, line 97: Some tests would be nice.
What about the `PT[outer;]`?

asinghvi17: That is fine, since points_flat contains the original points.

Will add some tests!

PR #4587, opened 2024-11-11
Don't force triangle mesh in crossbar with notch to avoid CairoMakie artifacts

jkrumbiegel:
CairoMakie has to draw triangle meshes with ugly seam artifacts

grafik

It seems to me the necessity to do that transformation must've been lost in some refactors as this code path is more than 4 years old. Just not creating the triangle mesh seems to work for me locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


asinghvi17:
Should we have a reference test here to insure against future changes?

PR #4589, opened 2024-11-14
Don't reset the axis limits / zoom for same Axis spec

SimonDanisch:
This tries to keep zoom levels etc, if the axes match exactly when re-using, since we assume the user intention is to move that axis, without resetting it:
```julia
p1 = S.Heatmap(rand(4, 4))
p2 = S.Heatmap(rand(50, 20))
ax1 = S.Axis(plots=[p1]; title="Axis 1")
ax2 = S.Axis(plots=[p2]; title="Axis 2")
grid = S.GridLayout(ax1, ax2)
f, ax, pl = plot(grid)
# swap axes while preserving zoom etc
pl[1] = S.GridLayout(ax2, ax1);
```

### DIFF
--- a/src/specapi.jl
+++ b/src/specapi.jl
@@ -305,8 +305,8 @@ function distance_score(at::Tuple{Int,GP,BS}, bt::Tuple{Int,GP,BS},
     (anesting, ap, a) = at
     (bnesting, bp, b) = bt
     scores = Float64[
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Closing in favor of https://github.com/MakieOrg/Makie.jl/pull/4643/

PR #4591, opened 2024-11-14
Fix Axis3 frameline z-sorting in CairoMakie

ffreyer:
# Description

#4485 changed `zvalue2d` to read `transformation.model` instead of `transformation.translation`, which causes it to include changes made to model directly. `Axis3` does that, so the results of plots under `ax.scene` changed. 

A few cases:
- Frame lines are part of the `ax.blockscene` so they don't see the model change from Axis3. Instead of getting translated, they have their input shifted. So before and after they have `zvalue2d() = 0`.
- Plots added to an Axis3 end up in `ax.scene`. Previously they had a default translation of `0`, now some other value based on the scenes model matrix. Since the scene centers its 3d boundingbox around 0, plots that have positive z values get shifted back and thus get a `zvalue2d < 0`
- Grid lines are also drawn to `ax.scene` and experience the same change as user plots. They end up at the same z value as them but earlier in the `scene.plots` array, so they draw first.

This pr should undo the change introduced in #4485 for CairoMakie by translating the frame lines plot instead of setting their data z-value to -10k.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 - Fix an error in `convert_arguments` for PointBased plots and 3D polygons [#4585](https://github.com/MakieOrg/Makie.jl/pull/4585).
 - Fix polygon rendering issue of `crossbar(..., show_notch = true)` in CairoMakie [#4587](https://github.com/MakieOrg/Makie.jl/pull/4587).
+- Fix render order of Axis3 frame lines in CairoMakie [#4591](https://github.com/MakieOrg/Makie.jl/pull/4591)
 
 ## [0.21.16] - 2024-11-06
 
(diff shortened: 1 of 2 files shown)


ffreyer:
Tests look good to me, can be merged

ffreyer:
Maybe worth noting that the z sorting we are doing only makes sense in 2D, where z translates to directly to depth. In 3D projectionview will generally rotate space for that to not be true. It also acts on a different scale and input data matters. (Even in 2D it sometimes does, see this pr.) 
The plot update refactor could help with this by making boundingbox caching cheaper and allowing us to fully transform that. #4150 could also help since it aims to render scene by scene, thus also making the z-sorting per scene. With that we could disable it in 3D.

SimonDanisch (approved):

PR #4601, opened 2024-11-19
improve performance of project_line_points

SimonDanisch:
```julia
x = rand(Point2f, 10^6)
f, ax, pl = lines(x);
@btime CairoMakie.project_line_points(f.scene, pl, x, :red, 1.0f0)
```
PR:
12.143 ms (16 allocations: 49.59 MiB)
Master:
2.643 s (86997922 allocations: 2.02 GiB)


ffreyer (approved):
As far as I can tell this just changes the generated function to a normal one, adds a couple of type annotations and an `@inbounds`. Should be fine

CairoMakie/src/utils.jl
```diff
@@ -190,245 +190,232 @@ function clip2screen(p, res)
...
-        # clip planes in clip space
-        clip_planes = if Makie.is_data_space(space)
-            Makie.to_clip_space(scene.camera.projectionview[], plot.clip_planes[])
-        else
-            Makie.Plane3f[]
-        end
+    space = (plot.space[])::Symbol
+    model = (plot.model[])::Mat4d
+    # Standard transform from input space to clip space
+    points = Makie.apply_transform(transform_func(plot), positions, space)::typeof(positions)
```
asinghvi17, line 204: That's not guaranteed for 2D to 3D or 3D to 2D transformations...is this type annotation necessary here?  Can we specify that it will be a Vector of VecTypes maybe?

SimonDanisch: This is required right now, since somehow it doesn't get inferred... This happens because transform_func doesn't infer. not sure what to do about it, I guess we could also put it behind a function barrier and extract transform_func before this function call and pass it as an argument.

jkrumbiegel:
I wonder if this has badly regressed at some point because I'm pretty sure I optimized the main drawing functions at some point so they didn't unnecessarily allocate. Could we add regression tests for this stuff?

SimonDanisch:
Well the benchmark seems to catch this really well... This may be the pr:
https://github.com/MakieOrg/Makie.jl/pull/3958
Which got merged with the 20% regression we see improved here.

SimonDanisch:
I think it'd be nice to slowly add more benchmarks and whole flamegraphs, which should catch function regressions pretty well.

SimonDanisch:
@asinghvi17 ah sorry, I forgot to address your comment.
Do you have time to look into it?

CairoMakie/src/utils.jl
jkrumbiegel: easiest would be to put everything after this behind another function barrier, `apply_transform` is already the function barrier for the dynamic `transform_func` result, but the stuff after is not guarded without `::typeof(positions)`

PR #4607, opened 2024-11-20
Introduce `max_auto_ticklabel_spacing!` for less jitter

SimonDanisch:
To not have jitter in the axis when panning/zooming, we've been recommending to use `tight_ticklabel_spacing!`, which doesn't fully work to avoid jitter.
1) It only considers the current ticks, so if the ticks grow, it will not allocate enough tickspace. This is solved by adding `n_tick_chars=automatic` to `tight_xticklabel_spacing!` where `automatic` replicated the old behavior. With set to e.g. 5, it will allocate enough space for 5 chars of ticks, e.g. "0.001".
2) Zoom and pan use `timed_ticklabelspace_reset` to reset the tickspace less often when zooming around. But it also resets the tickspace back to automatic (this may be a bug?). In any case, when we want truly static ticks, this should not run at all/ To achieve this, I added a  a new option to turn of the timed reset to `tight_ticklabel_spacing!` and `Axis`.

Open question: maybe this should be put in a new function, e.g. `static_tickspace!(ax, nchars)` ?
cc: @jkrumbiegel

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -1195,9 +1195,11 @@ end
 
 Sets the space allocated for the yticklabels of the `Axis` to the minimum that is needed and returns that value.
 """
-function tight_yticklabel_spacing!(ax::Axis; n_tick_chars=automatic)
-    space = tight_ticklabel_spacing!(ax.yaxis; n_tick_chars=n_tick_chars)
-    return space
+function tight_yticklabel_spacing!(ax::Axis)
+    spacing = tight_ticklabel_spacing!(ax.yaxis)
+    # Needed to not reset to automatic... But still doesn't keep the value somehow?
+    # ax.yticklabelspace = spacing
+    return spacing
 end
 
 """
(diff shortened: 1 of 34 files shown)


jkrumbiegel:
I'm not sure how much is gained here vs just doing `ax.yticklabelspace = some_value`? We already have an exact version with `tight_ticklabel_spacing!` but I'm not sure we need an inexact one. Also the nchars thing doesn't really make sense for both dimensions or if you consider ticklabel rotation.

jkrumbiegel:
> Well, as pointed out, the "exact" version is not really exact once you start zooming.

I didn't mean that it was useful for zooming specifically, just that it's purpose is an exact one (you wouldn't be able to pick the value it chooses for `yticklabelspace` easily yourself).

The `nchars` api feels weird to me because chars can have different sizes so it's not clear what that's supposed to mean, plus as I said, the meaning changes depending on rotation etc.

How about a mode like `:max_auto` where the tick spacing adjusts, but only upwards. So if you get longer tick labels, there will be room for them, but if they get shorter again, the spacing doesn't change?

src/makielayout/blocks/axis.jl
```diff
@@ -1198,8 +1196,10 @@ end
 Sets the space allocated for the yticklabels of the `Axis` to the minimum that is needed and returns that value.
 """
 function tight_yticklabel_spacing!(ax::Axis)
-    space = tight_ticklabel_spacing!(ax.yaxis)
-    return space
+    spacing = tight_ticklabel_spacing!(ax.yaxis)
+    # Needed to not reset to automatic... But still doesn't keep the value somehow?
```
jkrumbiegel, line 1210: what are these commented out parts for?

SimonDanisch: I started debugging `timed_ticklabelspace_reset` but couldn't figure out if it work as expected and what even expected means... setting ` ax.yticklabelspace = spacing` seems to help `timed_ticklabelspace_reset` to have the correct prev values, but still didn't seem to work correctly, so I reverted it, but left the comment to not loose the thought.

SimonDanisch:
Closing in favor of: https://github.com/MakieOrg/Makie.jl/pull/4642

PR #4618, opened 2024-11-22
Fix misaligned colors in colorbar?

### LINKED ISSUE #4121: Some of the `levels` are missing from `Colorbar`
- [x] what version of Makie are you running? (`]st -m Makie`): 0.21.5 (See below)
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`): Yes. I'm attaching an image created under this fresh env.
- [x] What platform + GPU are you on? ---macOS 14.6.1, Apple M1

When I specify the contour levels as `levels = [-5, -2, . . . ]`, some of the levels are missing from the colorbar even though those levels are plotted in `contourf`.

In the following example, levels "-2" and "2" are missing from the colorbar:
```julia
using CairoMakie

levels = [-5, -2, -1, -0.5, 0.5, 1, 2, 5]

func(x,y) = x

xs = -7.0:1.0:7.0
ys = 0.0:1.0:7.0
vals = func.(xs, ys')
fig = Figure()
ax = Axis(fig[1,1])
c = contourf!(ax, xs, ys, vals
              ,levels = levels
              ,extendlow=:auto, extendhigh=:auto)
Colorbar(fig[1,2],c; ticks=levels)
save("tmp.png", fig)
```

tmp

```julia
(@v1.10) pkg> status -m Makie
Status `~/.julia/environments/v1.10/Manifest.toml`
  [ee78f7c6] Makie v0.21.5

(@v1.10) pkg>
```

briochemc:
# Description

Fixes #4121? 

Currently the `Colorbar` built from a `contourf` may have misaligned colors. For a MWE see #4121 or
```julia
l = [1, 2, 5, 10, 20, 50]
x = 0:0.1:51
y = 0:0.1:51
z = [y for x in x, y in y]
fig, ax, plt = contourf(x, y, z; levels = l, extendhigh=:auto, extendlow=:auto)
cb = Colorbar(fig[1, 2], plt; tellheight = false)
```

which produces this misalignment (I edited the image in Preview to show the differences in colors):

test

The issue seems to be that the `continuous_pixels` (used to draw the heatmap for the colorbar) are not scaled properly. Simply adding the `colorrange` kwarg to that call fixes this issue for me, with properly matching colors:

Screenshot 2024-11-22 at 1 27 43 pm

However, I'm not if this breaks other things elsewhere, so hopefully someone can hold my hand to finish this PR?

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix an error in `convert_arguments` for PointBased plots and 3D polygons [#4585](https://github.com/MakieOrg/Makie.jl/pull/4585).
 - Fix polygon rendering issue of `crossbar(..., show_notch = true)` in CairoMakie [#4587](https://github.com/MakieOrg/Makie.jl/pull/4587).
 - Fix render order of Axis3 frame lines in CairoMakie [#4591](https://github.com/MakieOrg/Makie.jl/pull/4591)
+- Fix color mapping between contourf and colorbar [#4618](https://github.com/MakieOrg/Makie.jl/pull/4618)
 
 ## [0.21.16] - 2024-11-06
 
--- a/ReferenceTests/src/tests/figures_and_makielayout.jl
+++ b/ReferenceTests/src/tests/figures_and_makielayout.jl
@@ -342,6 +342,17 @@ end
     fig
 end
 
+@reference_test "Colorbar mapping to contourf" begin
+    l = [1, 2, 5, 10, 20, 50]
+    x = 0:0.1:51
+    y = 0:0.1:51
+    z = [y for x in x, y in y]
+    fig, ax, plt = contourf(x, y, z; levels = l)
+    cb = Colorbar(fig[1, 2], plt; tellheight = false)
+
+    fig
+end
+
 @reference_test "datashader" begin
     airports = Point2f.(eachrow(readdlm(assetpath("airportlocations.csv"))))
     # Dont use the full dataset, since WGLMakie seems to time out if it's too big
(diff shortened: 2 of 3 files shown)


ffreyer:
Seems fine to me. The plot colormapping code should always consider colorrange so I don't see why Colorbar shouldn't. With the ColorMapping type being constructed, the colorrange should also always be in a standard form.

I would say this can be merged after adding a release note to https://github.com/MakieOrg/Makie.jl/blob/master/CHANGELOG.md and adding a reference image test probably in https://github.com/MakieOrg/Makie.jl/blob/master/ReferenceTests/src/tests/figures_and_makielayout.jl.

briochemc:
I think I did what you asked:
- Add changelog item
- Add reference test

I also "fixed" formatting (spacing around kwarg `=` sign) and added the `colorrange` kwarg to the `heatmap` call in the continuous case, assuming it should be there as well.

Please let me know if I missed something!

briochemc:
BTW the reference test is a slightly simplified version of the MWE (removed the colorbar extensions):

| master | PR |
| :-- | :-- |
| Screenshot 2024-11-26 at 11 19 21 am | Screenshot 2024-11-26 at 11 19 31 am |

briochemc:
I'm not sure how/where to upload the reference image for the added reference test. Could you help with that?

ffreyer:
> I'm not sure how/where to upload the reference image for the added reference test. Could you help with that?

That's something Simon, Julius or I need to do.

ffreyer (approved):

ffreyer:
To things changed in `Colorbar for recipes`:

| old | new |
| --- | --- |
| reference_GLMakie_Colorbar for recipes | recorded_GLMakie_Colorbar for recipes |

On the center right, the Colorbar now has (slightly) different colors in the -0.5..-0.25 and -0.25..0.5 regions. These now match the colors in the plot exactly, so that's an improvement.

On the center left, the very wide Colorbar got stretched slightly in y direction. It comes from a 3D `contour()` plot, which lowers to a volume plot. I'm not sure how exactly the colors work here - but it seems like it's all handled early, i.e. in Makie through the ColorMapping object. Colorbar should be reading that more accurately now, so this should be an improvement?

Either way, I updated refimages. If the volume Colorbar is slightly off now it doesn't need to be fixed here.

PR #4629, opened 2024-11-26
Fix RPRMakie EnvironmentLight orientation

jkrumbiegel:
A normal exr or hdri image read in with `FileIO.load` will have to be transposed to match RPRs expected orientation, so I do that for convenience so that `EnvironmentLight(1, load("some_file.exr"))` looks as expected. Also, our coordinate system doesn't match that of RPRMakie which matters for the env light, so I exchange the y and z axes. The result looks ok to me although I'm not sure if the azimuthal rotation should be different. At least everything is the right way up and not squished weirdly.

Here's a simple scene with a camera looking at 0, 0, 0 with azimuth and elevation = 0 with a wide field of view:

## Before

grafik

## After

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix orientation of environment light textures in RPRMakie [#4629](https://github.com/MakieOrg/Makie.jl/pull/4629).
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Nice catch!

PR #4633, opened 2024-11-26
Fix CairoMakie heatmap offsets

### LINKED ISSUE #4053: heatmap slightly misaligned with respect to axis coordinates
I was plotting some analytic curves on top of a heatmap using CairoMackie, and I found out that possibly **all heatmaps are slightly misaligned with respect to the axis coordinates**.  This can be fixed by shifting the heatmap of a factor that I had to find by trial and error (and that  might vary!).  In the example below I generate heatmaps of random numbers, and then draw edges as vertical and horizontal lines. The contours are expected to match.

```julia
using Makie,CairoMakie
using StatsBase: midpoints

"""
  get_edges(v::AbstractVector{R}) where R<:Real

  Given a vector of values corresponding to center coordinates of size N,
  returns a vector of edges N+1.
"""
function get_edges(v::AbstractVector{R}) where R<:Real
  ret = midpoints(v)
  first_edge = v[1] - (ret[1]-v[1])
  last_edge = v[end] + (v[end]-ret[end])
  return vcat(first_edge,ret,last_edge)
end

N=50
fig = Figure(size=(700,300))
xedges = get_edges(1:N) 
yedges = get_edges(1:N)
ax1=Axis(fig[1,1];limits=(extrema(xedges),extrema(yedges)))

heatmap!(ax1,xedges,yedges, randn(N,N))
vlines!(ax1,xedges,linewidth=0.1,color=:white)
hlines!(ax1,yedges,linewidth=0.1,color=:white)
ax2=Axis(fig[1,2]; limits=(extrema(xedges),extrema(yedges)))
xedges_fix = xedges .+ N/500 # this fixes the problem by shifting the heatmap
yedges_fix = yedges .+ N/500 
heatmap!(ax2,xedges_fix,yedges_fix, randn(N,N))
vlines!(ax2,xedges,linewidth=0.1,color=:white)
hlines!(ax2,yedges,linewidth=0.1,color=:white)

colgap!(fig.layout,10)

save("/tmp/tmp.pdf",fig; pt_per_unit=1)

display(fig)
```
This is the plot I see.

image

When I zoom in, the left plot shows a clear misalignment between the heatmap edges and the lines in axis coordinates. In the right plot the misalignment is fixed by adding the `N/500` correction. 

image

I am using the latest version of Makie and the same problem appears consistently on Ubuntu 20.04 and Fedora 40 machines.

ffreyer:
# Description

Fixes #4053

We extend heatmap cells by half a pixel in every direction when `alpha == 1` to avoid gaps opening. This shifts the borders by half a pixel. I removed the padding for the side that covers other cells which fixes things for me. The other 0.5 padding is still there so this should still close gaps.

image

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix orientation of environment light textures in RPRMakie [#4629](https://github.com/MakieOrg/Makie.jl/pull/4629).
 - Fix uint16 overflow for over ~65k elements in WGLMakie picking [#4604](https://github.com/MakieOrg/Makie.jl/pull/4604).
 - Improve performance for line plot in CairoMakie [#4601](https://github.com/MakieOrg/Makie.jl/pull/4601).
 - Prevent more default actions when canvas has focus [#4602](https://github.com/MakieOrg/Makie.jl/pull/4602).
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -852,6 +852,8 @@ function draw_atomic(scene::Scene, screen::Screen{RT}, @nospecialize(primitive::
 end
 
 function _draw_rect_heatmap(ctx, xys, ni, nj, colors)
+    mini = min.(xys[1, 1], xys[end, end])
+    maxi = max.(xys[1, 1], xys[end, end])
     @inbounds for i in 1:ni, j in 1:nj
         p1 = xys[i, j]
         p2 = xys[i+1, j]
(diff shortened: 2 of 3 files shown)


ffreyer:
Widened the pad on the hidden side to a full pixel/unit since some 3D tests show gaps: (And fail because of it)
image
Screenshot 2024-11-26 223820

ffreyer:
Tests now fail because cells shifted. Heatmap 3D might be slightly worse in terms of coverage/gaps now, hard to tell. You can see the cell size being more accurate there though

SimonDanisch (approved):

PR #4663, opened 2024-12-13
add test for Billboard + transform_marker & fix CairoMakie scatter with markerspace = :data, rotation, transform_marker

### LINKED ISSUE #2291: 3D marker transformations wrong in CairoMakie
This is sort of a continuation of #2270.

With `transform_marker = true`, `rotate!` (as well as other transformations) applies not only to the position of the marker but also the quad the marker draws on. If the rotational axis is not parallel to the x, y or z direction (e.g. if it is `(1, 1, 0)`) this leads to a shear transformation. Since CairoMakie uses a 2D rotation, translation and scale for scatter, this can't be displayed correctly.

In text this was solved by using a transformation matrix instead:
https://github.com/MakieOrg/Makie.jl/blob/36ad29f180518fe87d09eba299db5dc31386ccd1/CairoMakie/src/primitives.jl#L474-L491

but it's going through FreeType (?) font matrices. Cairo should also have its own transform matrix (see https://www.cairographics.org/manual/cairo-Transformations.html#cairo-transform) but Cairo.jl does not seem to export it.

Note that this is also true for the `rotation` keyword in scatter, regardless of `transform_marker`.

As an example you can compare `scatter(rand(Point3f, 10), markersize = 20, rotation = Quaternionf(0.3, 0.4, 0.5, 0.4))` between GLMakie and CairoMakie.

### LINKED ISSUE #4609: show() to an IOBuffer with marker_space = :data results in an empty PDF
following this [discussion](https://discourse.julialang.org/t/correct-way-to-save-makie-plot-to-iobuffer/87793) on how to save a figure to an IOBuffer, i have found that setting `markerspace` to `:data` results in the marks being absent from the resulting PDF.  with `markerspace` unspecified, everything works.

MWE:

```
using WGLMakie, CairoMakie, Base64, Bonito 

app = App() do
    f = Figure()   
    a, s = scatter(f[1,1], Point3f.(rand(10), rand(10), rand(10)),
                   markersize = 0.03, markerspace = :data     # delete this line and it will work
                   )

    CairoMakie.activate!() 
    io = IOBuffer()
    show(io, MIME"application/pdf"(), f)
    
    io2 = IOBuffer()
    print(io2, "data:application/pdf;base64,")
    print(io2, base64encode(take!(io)))

    DOM.div(f,
            "Click to download ",
            DOM.a("scatter.pdf",
                  href=String(take!(io2)),
                  download="scatter.pdf"))
end 
```

screenshot of GUI:
Image

screenshot of PDF:
Image

the latest version(s) of Makie freshly installed into a temporary environment on an apple M2 max:

```
(jl_Us6BFN) pkg> st
Status `/private/var/folders/s5/8d629n5d7nsf37f60_91wzr40000gq/T/jl_Us6BFN/Project.toml`
  [824d6782] Bonito v4.0.0
  [13f3f980] CairoMakie v0.12.16
  [276b4fcb] WGLMakie v0.10.16
  [2a0f44e3] Base64 v1.11.0
```

ffreyer:
# Description

Adds tests for the different rotation types in different markerspaces with different transform_marker settings.

Also fixes `markerspace = :data`  not working correctly in CairoMakie and fixes a bunch of issues with CairoMakie not using rotation and transform_marker correctly.

Fixes #2291, fixes #4609

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) & test extension

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -356,7 +356,7 @@ function draw_atomic_scatter(
 
         ms = to_ndim(Vec3d, to_ndim(Vec2d, markersize, markersize), 1)
         p4d = transform * to_ndim(Point4d, to_ndim(Point3d, pos, 0), 1)
-        o = p4d[Vec(1, 2, 3)] ./ p4d[4] .+ model33 * to_ndim(Point3d, marker_offset, 0)
+        o = p4d[Vec(1, 2, 3)] ./ p4d[4] .+ model33 * to_ndim(Vec3d, mo, 0)
         proj_pos, mat, jl_mat = project_marker(scene, markerspace, o,
             ms, rotation, size_model, billboard)
 
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -367,8 +367,8 @@ function draw_atomic_scatter(
         # could be projected more accurately by projecting each point individually
         # and then building the shape.
 
-        # Can also skip if the x or y scale is ≈ 0
-        any(jl_mat * Vec2f(1) .≈ 0.0) && return
+        # Enclosed area of the marker must be at least 1 pixel?
+        (abs(det(jl_mat)) < 1) && return
 
         Cairo.set_source_rgba(ctx, rgbatuple(col)...)
         Cairo.save(ctx)
@@ -420,6 +420,7 @@ function draw_marker(ctx, marker::Char, font, pos, strokecolor, strokewidth, jl_
     cairo_font_face_destroy(cairoface)
 
     set_font_matrix(ctx, old_matrix)
+    return
 end
 
 function draw_marker(ctx, ::Type{<: Circle}, pos, strokecolor, strokewidth, mat)
@@ -432,7 +433,7 @@ function draw_marker(ctx, ::Type{<: Circle}, pos, strokecolor, strokewidth, mat)
     sc = to_color(strokecolor)
     Cairo.set_source_rgba(ctx, rgbatuple(sc)...)
     Cairo.stroke(ctx)
-    nothing
+    return
 end
 
 function draw_marker(ctx, ::Union{Makie.FastPixel,<:Type{<:Rect}}, pos, strokecolor, strokewidth, mat)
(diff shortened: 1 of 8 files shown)


ffreyer:
This grew into much more than I wanted to do...

Should now fix `markerspace = :data` and possibly some issues with rotation, billboarding and `transform_marker = true` for scatter in CairoMakie. Should fix "Scatter with FastPixel" and "scatter marker_offset 3D with rotation" tests.

I also extended the new test to cover all types of markers except FastPixel (Char, Rect, Circle, BezierPath, image). CairoMakie should match GLMakie for all of these now, up to marker distortion from perspective projection.

I'm not sure if this is even reachable, but there was also an order-of-operations difference with model and rotation for text, which is now fixed.

Fixes #2291

CairoMakie/src/primitives.jl
```diff
@@ -325,77 +325,88 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Scat
...
-        Cairo.set_source_rgba(ctx, rgbatuple(col)...)
+        # mat and jl_mat are the same matrix, once as a CairoMatrix, once as a Mat2f
+        # They both describe an approximate basis transformation matrix from
+        # marker space to pixel space with scaling appropriate to markersize.
+        # Markers that can be drawn from points/vertices of shape (e.g. Rect)
+        # could be projected more accurately by projecting each point individually
+        # and then building the shape.
+
+        # Enclosed area of the marker must be at least 1 pixel?
+        (abs(det(jl_mat)) < 1) && return
```
ffreyer, line 371: After some digging I figured out that CairoMakie CI failed specifically for `rotation = Quaternionf(0.6849482f0, 0.175505f0, 0.15756221f0, 0.6893606f0)` in "image scatter". This rotation causes `mat`/`jl_mat` (the same localized markerspace to screen projection matrix we have for text) to map the corner points of the marker on a line. This causes Cairo to give up on life.

I'm trying to catch this here by checking the determinant. In this case it's close to 0 (0.035). Since the det is related to area in 2D this should check that the marker has at least 1px^2 area. This seems to be fine with this specific rotation, even if I increase markersize until the value passes 1.

PR #4664, opened 2024-12-13
Fix poly Circle & Rect stroke

ffreyer:
# Description

The breaking pr changed `poly(Circle(...))` to use the Circle wireframe for stroke, which this pr fixes.

It also changes (fixes?) `Rect2` stroke to be continuous, i.e. avoid corners like this:

image

Also extends a refimg test to test these things

## Type of change

## Checklist

- Added an entry in CHANGELOG.md (for new features and breaking changes)


ffreyer:
> It also changes (fixes?) `Rect2` stroke to be continuous, i.e. avoid corners like this:
> 
> image

This is causing "poly stroke & array input" and "Axis3 viewmodes, xreversed, aspect, perspectivenes" to fail. This is also a real change from 0.21, so this pr should get a changelog entry

Issue #796, opened 2020-12-26
Feature request: contour that can handle curvilinear grids


Gyoshi:
`surface` currently accepts curvilinear grids. It would be nice if `contour` and `contourf` does too for plotting of parameterised values.

`contour`s would also use `VertexSurfaceLike` a la #748 
Similar to #675.

For example, the pyplot backend via Plots.jl does this:
```julia
import Plots; Plots.pyplot()
r1, r2 = 1.5, 1/2
r(u,v) = ((r1 + r2*cos(v))*cos(u), (r1 + r2*cos(v))*sin(u), r1/2*sin(u)+r2*sin(v)) # torus
us = vs = range(0, 2pi, length=25)
xs, ys, zs = [[pt[i] for pt in r.(us, vs')] for i in 1:3]
Plots.contourf(xs, ys, zs)
```
![torus_example](https://user-images.githubusercontent.com/41743877/103152104-eea55580-4784-11eb-8d87-9a816354346a.png)

ffreyer:
Closed as duplicate of #742

briochemc:
To track contours and heatmaps issues separately, what about reopening this issue while reducing the scope of #742 to heatmaps only? This issue is about contours only, while #742 is about contours _and_ heatmaps, but almost everything discussed in #742 is about heatmaps.

briochemc:
@ffreyer Sorry to bump it again, but I think this issue should be reopened: It remains unresolved and the scope is distinct from #742.

Surely I can't be the only one still wanting to plot contours for curvilinear grids with Makie, right? Unless there is a workaround that I don't know about?

asinghvi17:
It seems like the standard approach is to compute the contour lines of the `z` matrix directly, and then transform the coordinates according to the curvilinear grid (using bilinear interpolation).

It seems easy enough to do.  But it would be pretty slow.  

see also:
- https://observablehq.com/@recifs/contours-on-a-curvilinear-grid
- https://github.com/d3/d3-contour/issues/64

If you plot a contour or contourf into a GeoAxis or PolarAxis, this is exactly what happens already.

briochemc:
Yes I get this impression.

See also:
- https://github.com/kakearney/contourfcmap-pkg/blob/master/contourfcmap/contourfcmap.m

asinghvi17:
Yeah tricontourf is probably your best immediate bet for this.  It may even give you continuous contour lines.

[...]

PR #4670, opened 2024-12-15
Enable curvilinear `contour` plots

asinghvi17:
# Description

It turns out that Contour.jl could already handle curvilinear grids, even though it says it can't in the docs.  So we simply enable that here by overloading `contourlines`.

This PR also adds docs and tests.  Partially addresses #796, because this PR does not touch on `contourf`.

download-2

## Type of change

## Checklist


asinghvi17:
This is ready for review!

asinghvi17:
oops, forgot about the z level of the surface plot - thanks for catching that!

PR #4677, opened 2024-12-17
Allow GLMakie to change the title of a screen after its creation.

### LINKED ISSUE #4665: Allow changing window title after creation through GLMakie
### Feature description
Currently, Makie doesn't provide a way to change the title of a GLMakie window after its creation (screen.config.title = "some title" doesn't do anything), but it can be achieved by adding the GLFW package and using GLFW.SetWindowTitle(screen, "some title"). I think adding the GLFW package just to change a window title is a little too much, would it be possible to include this functionality in Makie?

lpaulino07:
# Description

Fixes #4665  

Previously, `GLFW.SetWindowTitle()` was needed, which required adding the GLFW package to the script. This change just wraps `GLFW.SetWindowTitle()` into `GLMakie.set_screen_title!()`, and modifies screen.config.title to store the new title.

## Type of change

## Checklist


SimonDanisch:
I took the freedom to rename it to `set_title!`... I did just notice, that it's also called `set_screen_visibility!`... Maybe that needs to get renamed too, but is rather an internal function in any case.
Anyways, I think the screen in `func(::Screen)` is redundant, so I prefer this name.

PR #4689, opened 2024-12-24
Rework postprocessor handling in GLMakie

ffreyer:
# Description

This pr aims to rework and streamline postprocesser handling. I would like to implement a system similar to Nodes in Blender or Unreal Blueprints to handle the render pipeline. I.e. describe the various operations in `render_frame()` (z sorting, rendering plots, running postprocessors) as Nodes with connectable inputs and outputs. These nodes and edges should be free of OpenGL, i.e. an abstraction of what the backend is doing. Resolving them should produce a set of necessary buffers for data transfer and a pipeline of backend tasks to run in `render_frame`. 

TODO/Plans/Changes:
- [x] extract OpenGL parts of GLFramebuffer and move them to GLAbstraction
- [x] rework PostProcessor type to better represent a renderloop task (generic abstract type with specialized implementations, try using dispatched function)
- [ ] add task for rendering
- [ ] add task for z sorting
- [ ] maybe add task for `setup!()`
- [ ] switch to one GLFramebuffer per task (as needed) instead of one global framebuffer
- [ ] if needed, implement GLFence for synching between framebuffers
- [ ] add a FramebufferManager which holds onto all attachments, adds new ones as required and distributes them to postprocessors
- [ ] ... (create abstracted Node type, buffer format type, format compatibility rules/judgements, edge type, resolution to backend components)

Note - I'm mainly creating a pr for this now to get a benchmark and test history going. 

## Type of change

... or maybe just internal rework for now?

## Checklist


ffreyer:
The test failure is very rarely reproducible by spam opening and closing windows. Maybe once every 5-10min with
```julia
using GLMakie, Colors
function make_scene()
    scene = Scene()
    color = map(tick -> HSV(tick.count % 360, 0.8, 0.7), events(scene).tick)
    scatter!(scene, (0,0), color = color, markersize = 100)
    display(GLMakie.Screen(render_on_demand = false, framerate = 10_000), scene)
end

running = Ref(true)
running[] = false
begin
    running[] = true
    @async while running[]
        try
            if rand() < 0.1
                GLMakie.closeall()
                make_scene()
            else
                make_scene()
            end
            yield()
            sleep(0.1)
        catch e
            GLMakie.closeall()
            rethrow(e)
        end
    end
end
```
which also randomly gives up on itself every so often. When trying to hunt down incorrect context with
```julia
is_current_context(window) = GLFW.GetCurrentContext().handle == window.handle

# in some function
@assert is_current_context(screen.glscreen) "Context incorrect"
```
I ended up in this block of code
https://github.com/MakieOrg/Makie.jl/blob/5eeb34d8c69a533974b4d2bba3cb00d8f5fac8bc/GLMakie/src/postprocessing.jl#L168-L171
which is exactly the same as before, including all the code it calls (afaik). I started that block with an assert and then a `switch_context!()` so the context should be lost somewhere in there. Or maybe assert itself is causing task switches and I was just chasing after that?
The latest commit still fails (without asserts, just running my test for ~10min) and I have not been able to reproduce it on master.

ffreyer:
CI seems to finish consistently now but I can still reproduce failures locally. Usually takes 5000+ spammed windows though

GLMakie/test/glmakie_refimages.jl
```diff
@@ -167,4 +167,124 @@ end
...
+        connect!(pipeline, fxaa,    color_tint, :color)
+        connect!(pipeline, render3, color_tint, :color)
+        connect!(pipeline, render3, display_stage, :objectid)
+        connect!(pipeline, color_tint, display_stage, :color)
+    end
+
+
+    GLMakie.activate!(render_pipeline = pipeline)
+    cow = load(Makie.assetpath("cow.png"))
+    f,a,p = image(rotr90(cow))
```
ffreyer, line 289: This adds a Sepia filter as a postprocessing stage: 

image

With 
```julia
GLMakie.activate!(render_pipeline = Makie.automatic)
fig = Figure(size = (800, 480))
ax = Axis(fig[1, 1], autolimitaspect = 1)
Makie.pipeline_gui!(ax, pipeline)
fig
```
you can get an ok-ish visualization of the pipeline:
Screenshot 2025-01-06 231749

GLMakie/src/glwindow.jl
```diff
@@ -10,182 +10,104 @@ end
...
-        error("Ran out of color buffers.")
-    end
 
-    glFramebufferTexture2D(GL_FRAMEBUFFER, next_color_id, GL_TEXTURE_2D, t.id, 0)
-    push!(fb.buffer_ids, key => next_color_id)
-    push!(fb.buffers, key => t)
-    return next_color_id
-end
+mutable struct FramebufferFactory
+    fb::GLFramebuffer # core framebuffer (more or less for #4150)
```
SimonDanisch, line 16: We need a better name for this... What's this doing exactly?

src/utilities/RenderPipeline.jl
```diff
@@ -0,0 +1,1044 @@
...
+    end
+    if fxaa
+        connect!(pipeline, _fxaa, display, :color)
+    end
+    connect!(pipeline, :objectid)
+
+    return pipeline
+end
+
+function test_pipeline_3D()
```
SimonDanisch, line 715: What are these functions for?

GLMakie/src/glwindow.jl
ffreyer: I read that GLFramebuffers are very cheap to build/use and it's generally advised to have many dedicated ones over a few everything-framebuffers.  So the design here is that the Makie side figures out which buffers need to be in a given render stage and the GLMakie side then builds them accordingly. This object is responsible for building them (which includes reuse of attachments) and keeping track of all the attachments and framebuffers for resizing and cleanup. 

Maybe calling this a factory is wrong and should be something like `FramebufferManager`? Though the main job/complexity is building GLFramebuffer objects.

GLMakie/src/render_pipeline.jl
```diff
@@ -0,0 +1,152 @@
+function Makie.reset!(factory::FramebufferFactory, formats::Vector{Makie.BufferFormat})
```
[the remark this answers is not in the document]
ffreyer: This frees and deletes all the GLFramebuffer objects, but tries to reuse attachments instead of freeing them. I.e. anything that the new RenderPipeline needs will be preserved and reordered so the GLFramebuffers can be built correctly. This is probably something I did to get better benchmarks.

I wouldn't want to call this free (or whatever) as is, because it does a lot more. But maybe this does too much/is over-engineered. Maybe it should just be `empty!(factory)` (deleting everything)  and `initialize!(factory, render_pipeline)` instead?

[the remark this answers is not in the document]
SimonDanisch: Yeah, that sounds like empty!

GLMakie/src/glwindow.jl
SimonDanisch: Yeah, sounds more like a manager....

Makie/src/utilities/RenderPipeline.jl
```diff
@@ -0,0 +1,1044 @@
...
+    end
+    if fxaa
+        connect!(pipeline, _fxaa, display, :color)
+    end
+    connect!(pipeline, :objectid)
+
+    return pipeline
+end
+
+function test_pipeline_3D()
```
ffreyer: I probably answered this in dms, but I might as well answer here. I played around with different pipeline configurations to see if some of them are better for specific purposes and also to check that the code works. These functions were leftover from that. I deleted them now

Makie/src/utilities/RenderPipeline.jl
```diff
@@ -0,0 +1,1044 @@
...
+
+"""
+    generate_buffers(pipeline)
+
+Maps the connections in the given pipeline to a vector of buffer formats and
+returns them together with a connection-to-index map. This will attempt to
+optimize buffers for the lowest memory overhead. I.e. it will reuse buffers for
+multiple connections and upgrade them if it is cheaper than creating a new one.
+"""
+function generate_buffers(pipeline::RenderPipeline)
```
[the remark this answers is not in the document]
ffreyer: There was a bigger test checking that the default pipeline gets handled correctly. I added a bunch of specific tests for various different cases now too.

ffreyer:
> I cant really tell from the code, but does this allow for disabling picking and allowing multisampling?

In theory the picking buffer could be removed by removing all the connections involving it. Then the Makie side of this won't attempt to generate it. In practice it's still hard-coded in `gl_render_pipeline!()`. I don't think that's necessary though...

I left multisampling out of this mostly because I'm not sure what requirements it brings. It's probably not that hard to add though. Iirc the edits needed are:
- an additional option for `BufferFormat` to declare it as an msaa buffer
- buffer compatibility logic ((how) can msaa combine with non-msaa?)
- buffer cost function (if msaa can combine with non-msaa, then how expensive is an msaa buffer compared to new buffer?)
- backend resolution to msaa texture
- maybe updates to GLAbstraction if that doesn't support msaa yet

Issue #4638, opened 2024-12-01
GLMakie errors in interactive mode after immediate use of a figure


jakubkaczor:
GLMakie errors if `Figure` is used immediately after creation in interactive mode.
```julia
using GLMakie
fig = Figure()
Axis(fig[1, 1])  # Raises the error

```
errors, but
```julia
using GLMakie
fig = Figure();
Axis(fig[1, 1])

```
does _not_.

<details><summary>Error output with the stacktrace</summary>
<pre>
ERROR: glGenBuffers returned invalid id. OpenGL Context active?
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] glGenBuffers
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLExtendedFunctions.jl:112
  [3] glGenBuffers
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLExtendedFunctions.jl:108 [inlined]
  [4] GLBuffer
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLBuffer.jl:11
  [5] #GLBuffer#13
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLBuffer.jl:57 [inlined]
  [6] GLBuffer
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLBuffer.jl:52 [inlined]
  [7] #_#9
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/AbstractGPUArray.jl:194 [inlined]
  [8] GPUArray
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/AbstractGPUArray.jl:193 [inlined]
  [9] #gl_convert#74
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLUniforms.jl:262 [inlined]
 [10] gl_convert
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLUniforms.jl:259
 [11] RenderObject
    @ ~/.local/share/julia/packages/GLMakie/xG4T9/src/GLAbstraction/GLTypes.jl:373
 [12] RenderObject
[...]

Issue #4457, opened 2024-10-07
`NO_CURRENT_CONTEX` error with GLMakie on WSL


hexaeder:
I am experiencing errors which are, unfortunately, quite hard to pin down/reproduce. My setup is an Ubuntu under WSL under Windows 11, on a Laptop with CUDA GPU. All on Julia 1.11-rc4

Relatively frequently, when I try to open a new Figure, I get the following error:

```julia
julia> fig = Figure()
┌ Warning: error in renderloop
│   exception =
│    GLFWError (NO_CURRENT_CONTEXT): Cannot set swap interval without a current OpenGL or OpenGL ES context
│    Stacktrace:
│     [1] _ErrorCallbackWrapper
│       @ ~/.julia/packages/GLFW/wmoTL/src/callback.jl:43
│     [2] SwapInterval
│       @ ~/.julia/packages/GLFW/wmoTL/src/glfw3.jl:779 [inlined]
│     [3] renderloop
│       @ ~/.julia/packages/GLMakie/NAHlI/src/screen.jl:1009
│     [4] #72
│       @ ~/.julia/packages/GLMakie/NAHlI/src/screen.jl:870
└ @ GLMakie ~/.julia/packages/GLMakie/NAHlI/src/screen.jl:1019
```

Sometimes it also happens when I try to draw stuff to an existing figure. Unfortunately, it is not entirely clear to me when this happens exactly. I can open 1000 figures in a loop, but sometimes, I cannot even open a second one. Is this some known problem? Do you have any ideas on what I could provide to help debug this? Admittedly, my setup is probably not the most tested out there 😬

hexaeder:
I might just found a reproducer which works for me at least.
Copy pasting the following in a reply without startup file all at once
```julia
using Pkg
pkg"activate --temp"
pkg"add GLMakie"
using GLMakie
fig = Figure()
ax = LScene(fig[1,1])
```
results in the error below. Its not exactly the same, but it also mentiones the inactive OpenGL Context?
I could *not* reproduce this problem under Julia 1.10.5 however.

```julia
ERROR: glGenTextures returned invalid id. OpenGL Context active?
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] glGenTextures()
    @ GLMakie.GLAbstraction ~/.julia/packages/GLMakie/NAHlI/src/GLAbstraction/GLExtendedFunctions.jl:132
[...]

PR #4699, opened 2025-01-02
Make cleanup of OpenGL objects explicit in GLMakie

ffreyer:
# Description

After some help from Simon we figured out that GC / `free()` was the cause of issues in #4689. Specifically because free() includes a context switch and GC can trigger it whenever via finalizers, context switches can occur randomly. To fix this I'm making all cleanup of OpenGL objects explicit here, i.e. remove the need for finalizers.

The current version should segfault if a finalizer deletes OpenGL objects. I'll let CI run with this so I can see if I missed anything

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/GLAbstraction/GLTypes.jl
+++ b/GLMakie/src/GLAbstraction/GLTypes.jl
@@ -21,16 +21,18 @@ cardinality(x) = length(x)
 cardinality(x::Number) = 1
 cardinality(x::Type{T}) where {T <: Number} = 1
 
-struct Shader
+mutable struct Shader
     name::Symbol
     source::Vector{UInt8}
     typ::GLenum
     id::GLuint
     context::GLContext
-end
 
-function Shader(name, source, typ, id)
-    new(Symbol(name), source, typ, id, current_context())
+    function Shader(name, source, typ, id, context = current_context())
+        obj = new(Symbol(name), source, typ, id, context)
+        finalizer(free, obj)
+        return obj
+    end
 end
 
 function Shader(name, source::Vector{UInt8}, typ)
@@ -453,6 +455,14 @@ function unsafe_free(x::GLProgram)
     GLAbstraction.context_alive(x.context) || return
     GLAbstraction.switch_context!(x.context)
     glDeleteProgram(x.id)
+    return
+end
+
+function unsafe_free(x::Shader)
+    x.id == 0 && return
+    GLAbstraction.context_alive(x.context) || return
+    GLAbstraction.switch_context!(x.context)
+    glDeleteShader(x.id)
     return
 end
 
--- a/GLMakie/src/GLAbstraction/GLTypes.jl
+++ b/GLMakie/src/GLAbstraction/GLTypes.jl
@@ -523,3 +523,7 @@ end
 function free(x::GLVertexArray)
     @assert x.id == 0 "Must be freed explicitly"
 end
+
+function free(x::Texture)
+    @assert x.id == 0 "Must be freed explicitly"
+end
--- a/GLMakie/src/GLAbstraction/GLTypes.jl
+++ b/GLMakie/src/GLAbstraction/GLTypes.jl
@@ -497,10 +497,10 @@ function unsafe_free(x::GLVertexArray)
     GLAbstraction.context_alive(x.context) || return
     GLAbstraction.switch_context!(x.context)
     for (key, buffer) in x.buffers
-        free(buffer)
+        unsafe_free(buffer)
     end
     if x.indices isa GPUArray
-        free(x.indices)
+        unsafe_free(x.indices)
     end
     id = Ref(x.id)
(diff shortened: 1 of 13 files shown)


ffreyer:
May fix:
- #4638 
- #4457

GLMakie/src/GLAbstraction/GLInfo.jl
```diff
@@ -97,3 +97,12 @@ function getProgramInfo(p::GLProgram)
...
 end
+
+const FAILED_FREE_COUNTER = Ref(0)
+
+function verify_free(obj::T, name = string(T)) where T
+    if obj.id != 0
+        FAILED_FREE_COUNTER[] = FAILED_FREE_COUNTER[] + 1
+        Threads.@spawn println(stderr, "Error: $name has not been freed.")
+    end
+end
```
ffreyer, line 108: This is used as the finalizer now, to "error" when something hasn't been deleted. I also added a counter to query in CI. We could also de/activate that with an environment variable but I figured incrementing and integer doesn't really matter.

ffreyer: TODO: 

`string(T)` seems to be pretty slow, should be moved so it only happens if the branch is true (dragged one of my benchmarks in #4689 from ~300µs down to 800µs)

SimonDanisch: Didn't we want to completely disable finalizers and only activate them in debug mode?

SimonDanisch: I think registering a finalizer is somewhat expensive, so if we only use it for debugging purposes we should just disable it via a global ( not an env var)

GLMakie/src/GLAbstraction/GLTypes.jl
```diff
@@ -397,14 +402,30 @@ function RenderObject(
...
+            end
+        end
+    end
 
     # remove all uniforms not occurring in shader
     # ssao, instances transparency are special for rendering passes. TODO do this more cleanly
     special = Set([:ssao, :transparency, :instances, :fxaa, :num_clip_planes])
     for k in setdiff(keys(data), keys(program.nametype))
         if !(k in special)
+            !haskey(buffers, k) && (data[k] isa GPUArray) && free(data[k])
```
ffreyer, line 428: This disallows optional shared resources. E.g. if we were to pass the same OpenGL texture to two renderobjects and one does not include a uniform for it, it would get deleted here.

SimonDanisch (approved):

ffreyer:
I deleted the try .. catch in `free(scene)` again because I don't think it makes much sense. The function was essentially:
```julia
try
    cleanup1()
    cleanup2()
    ...
catch e 
    safe_print(e)
end
```
If any of the cleanup() steps fails we either have a corrupt state, a memory leak or both. The try catch isn't doing anything to recover from that. In fact it makes things worse by replacing the error with a print. Deleting a plot or scene can also not be treated here. If that errors we have no idea how far cleanup progressed in the backend and in the called Makie functions. So these errors must be treated at a lower level.

GLMakie/src/postprocessing.jl
```diff
@@ -299,8 +309,6 @@ function to_screen_postprocessor(framebuffer, shader_cache, screen_fb_id = nothi
 end
 
 function destroy!(pp::PostProcessor)
-    while !isempty(pp.robjs)
-        destroy!(pop!(pp.robjs))
-    end
+    foreach(destroy!, pp.robjs)
```
SimonDanisch, line 312: This doesn't empty! the list?

GLMakie/src/precompiles.jl
```diff
@@ -49,7 +49,6 @@ let
             screen = Screen(Scene(), config; visible=false, start_renderloop=false)
             close(screen)
 
-            empty!(atlas_texture_cache)
```
SimonDanisch, line 52: Why was this removed? I guess it's correct, since it doesn't seem to error, but just want to make sure.

ffreyer: This would empty the dict without doing OpenGL or Makie cleanup

GLMakie/src/screen.jl
```diff
@@ -697,6 +737,12 @@ function closeall(; empty_shader=true)
         screen = pop!(ALL_SCREENS)
         destroy!(screen)
     end
+
+    if !isempty(atlas_texture_cache)
+        @warn "texture atlas cleanup incomplete: $atlas_texture_cache"
+        empty!(atlas_texture_cache)
+    end
+
```
ffreyer, line 745: That should clean up callbacks in the (Makie) texture atlas

GLMakie/src/precompiles.jl
SimonDanisch: I think that was the idea

PR #4711, opened 2025-01-09
Reorganize Makie test files

ffreyer:
# Description

I've been thinking for a while that organizing tests into a few folders would be helpful so I did that here. 

As of [8e09652](https://github.com/MakieOrg/Makie.jl/pull/4711/commits/8e09652a177052eebe03c40a1cf78261a38b3252):
- This does not change any test that was included before
- This does split up events.jl without any functional changes (1)
- This does change and include tests that were previously not running (zoom_pan.jl, timings.jl). (1) zoom_pan has been merged with the Axis tests from events.jl.
- This does change 3 lines of Makie code - 2 to catch BudgetedTimer callbacks erroring and 1 fixing `stop!(timer)` being unavailable in WGLMakie (how did this not error before? Does isopen(screen) not become false after closing a WGLMakie screen?)

## Type of change

- [x] Test/CI improvements


SimonDanisch (approved):

ffreyer:
Double checked the number of tests after the last merge, seems correct (54 more than master)

Issue #3734, opened 2024-03-25
`band` doesn't handle using `LinePattern` for `color` 


rafaqz:
- [x] are you running newest version (version from docs) ?
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x] What platform + GPU are you on? linux/nvidia

MWE:
```julia
using GLMakie
pat = Makie.LinePattern(; direction=Vec2f(1), width=5, tilesize=(20, 20), linecolor=:white, background_color=:black)
band(zeros(10), rand(10); color=pat)
```

```julia
ERROR: Wrong number of colors. Must be 10 or double.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] (::Makie.var"#469#472"{Observable{Vector{Float64}}})(c::Makie.LinePattern)
    @ Makie ~/.julia/packages/Makie/VRavR/src/basic_recipes/band.jl:49
  [3] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::@Kwargs{})
    @ Base ./essentials.jl:892
  [4] invokelatest(::Any, ::Any, ::Vararg{Any})
    @ Base ./essentials.jl:889
  [5] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
  [6] #map!#179
    @ ~/.julia/packages/Makie/VRavR/src/scenes.jl:153 [inlined]
  [7] map!
    @ ~/.julia/packages/Makie/VRavR/src/scenes.jl:146 [inlined]
  [8] plot!(plot::Plot{Makie.band, Tuple{Vector{Float64}, Vector{Float64}}})
    @ Makie ~/.julia/packages/Makie/VRavR/src/basic_recipes/band.jl:38
  [9] connect_plot!(parent::Scene, plot::Plot{Makie.band, Tuple{Vector{Float64}, Vector{Float64}}})
    @ Makie ~/.julia/packages/Makie/VRavR/src/interfaces.jl:260
 [10] plot!
    @ ~/.julia/packages/Makie/VRavR/src/interfaces.jl:265 [inlined]
 [11] plot!(ax::Axis, plot::Plot{Makie.band, Tuple{Vector{Float64}, Vector{Float64}}})
    @ Makie ~/.julia/packages/Makie/VRavR/src/figureplotting.jl:316
[...]

PR #4715, opened 2025-01-12
Cleanup Patterns/Hatching

ffreyer:
# Description

Fixes # (issue)

This pr aims to make Patterns/hatching a fully fledged feature.. This implies:
- [x] improving the line quality in patterns
- [x] adding unit tests for pattern code
- [x] adding reference image tests for relevant primitive plots (mesh, meshscatter, surface)
- [x] fixing related issues in GLMakie, WGLMakie
- [x] adding Pattern handling in CairoMakie
    - [x] merge new code with existing code for poly
- [x] adding reference image tests for (at least some) relevant recipes (barplot, hist, poly, density, pie, hvspan, band, ...?)
- [x] fixing related issues
- [x] updating documentation
- ~~adding exports~~ Pattern is already exported
- [x] cleaning up constructors/user interface if need be
- [x] consider anchoring patterns to model space instead of world space

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -979,7 +979,11 @@ function draw_mesh3D(
     meshpoints = decompose(Point3f, mesh)::Vector{Point3f}
     meshfaces = decompose(GLTriangleFace, mesh)::Vector{GLTriangleFace}
     meshnormals = normals(mesh)::Union{Nothing, Vector{Vec3f}} # note: can be made NaN-aware.
-    meshuvs = texturecoordinates(mesh)::Union{Nothing, Vector{Vec2f}}
+    _meshuvs = texturecoordinates(mesh)
+    if (_meshuvs isa AbstractVector{<:Vec3})
+        error("Only 2D texture coordinates are supported right now. Use GLMakie for 3D textures.")
+    end
+    meshuvs::Union{Nothing,Vector{Vec2f}} = _meshuvs
 
     if meshuvs isa Vector{Vec2f} && to_value(uv_transform) !== nothing
         meshuvs = map(uv -> uv_transform * to_ndim(Vec3f, uv, 1), meshuvs)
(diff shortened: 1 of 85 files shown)


WGLMakie/src/meshes.jl
```diff
@@ -32,7 +32,29 @@ function handle_color!(plot, uniforms, buffers, uniform_color_name = :uniform_co
...
         uniforms[:pattern] = true
-        uniforms[uniform_color_name] = Sampler(convert_texture(color); minfilter=minfilter)
+        img = convert_texture(map(Makie.to_image, plot, color))
+        uniforms[uniform_color_name] = Sampler(img; x_repeat = :repeat, minfilter=minfilter)
+        # different default with Patterns (no swapping and flipping of axes)
+        # also includes px to uv coordinate transform so we can use linear
+        # interpolation (no jitter) and related pattern to (0,0,0) in world space
+        converted_uv_transform = pop!(uniforms, :uv_transform)
+        scene = Makie.parent_scene(plot)
+        uniforms[:uv_transform] = map(plot,
```
SimonDanisch, line 42: Can we move these things into functions, that are easy to re-use?

ffreyer:
Should I add examples for all (some?) of the plots that are now tested with patterns?

ffreyer:
I switched anchoring to consider model so that patterns move with the object when using `translate!()`. This doesn't work correctly with float32convert. I'm not sure why and I don't think it's very important so I don't want spend time on it now. May also get easier to fix after the plot update refactor

ffreyer:
Hmm, seems like there is an offset between CairoMakie and other backends. I don't know where that comes from but I also don't think it matters.

Good to merge from my end

CairoMakie/src/primitives.jl
```diff
@@ -1036,12 +1072,46 @@ function draw_mesh3D(
         valid = Bool[]
     end
 
+
+    # Camera to screen space
+    ts = map(vs) do v
+        clip = projectionview * v
+        @inbounds begin
```
SimonDanisch, line 1079: Do we not have a function for this?

SimonDanisch (approved):

CairoMakie/src/primitives.jl
ffreyer: Not really. The CairoMakie projection functions drop the z component which we use for z sorting here. (And that needs to happen after projectionview.) Makie's project function do that as well except for the ones that include transform_func, model and f32convert.

SimonDanisch: Could we maybe improve project instead and use that?

ffreyer:
Fixes #3734

PR #4733, opened 2025-01-17
Did CI break?

ffreyer:


ffreyer:
After some local testing I'm fairly certain this is caused by ColorTypes 0.12 updating `rand()` code. Deleting packages holding back that update allowed me to reproduce this. (And just Colors did not). I'll just update the refimages.

PR #4738, opened 2025-01-18
Allow higher dimensional geometries in poly

ffreyer:
# Description

This reverts a change from the breaking release, allowing something like `poly(Rect3f[...])` again. (Do we want to allow poly to draw 3D shapes?)

## Type of change

## Checklist


jkrumbiegel:
I personally don't think it makes sense to have 3d polys. The only one that makes sense is a 2d poly positioned in some plane in 3D space. But I would only add a nicer interface to define that rather than actually accepting 3D points.

jkrumbiegel:
I would say the mesh + wireframe is just an implementation detail and in cairomakie it's not accurate either.

SimonDanisch:
>I would say the mesh + wireframe is just an implementation detail and in cairomakie it's not accurate either.

I meant, poly currently offers to draw a mesh + outline, which users rely on and use and which is something that's nice to have.
So I didn't mean the implementation detail ;)

SimonDanisch (approved):

PR #4742, opened 2025-01-19
Nudge Axis3 clip planes to avoid clipping at 0 margin

ffreyer:
# Description

Fixes #4741

Added Refimg:

| before change | after change |
| --- | --- |
| image | Screenshot 2025-01-19 182644 |

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #4727, opened 2025-01-16
fix NaN error

bjarthur:
# Description

when recording an animation using WGLMakie i sometimes get the following error in the browser console.  this PR fixes it.  my understanding of this part of the makie codebase is as superficial as it gets.  not sure if this breaks anything but it fixes my problem.  happy to add a test if desired.  have not taken the time to devise a MWE but could if you want.

```
Uncaught TypeError: n.toArray is not a function
    at as (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:9759:11)
    at yo.k_ [as setValue] (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:9979:13)
    at qi.upload (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:10152:39)
    at Fd (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:13117:628)
    at Ro.renderBufferDirect (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:12945:72)
    at tl (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:13073:450)
    at ks (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:13069:45)
    at el (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:13041:228)
    at Ro.render (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:13010:20)
    at render_scene (454d14251abd5406c633fb77841dd6097a73cd86-wglmakie.bundled.js:22949:18)
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added unit tests for new algorithms, conversion methods, etc.

PR #4772, opened 2025-02-10
Fix WGLMakie nan error

ffreyer:
# Description

Probably supersedes #4727

While merging #4742 I ran into an issue where Axis3 would stop working when rotated in vscode. In the Browser this produces the same chain of errors as #4727. The "stack trace" of this issue is:
- Plane produces a NaN vector from `normalize(Vec3f(0))`
- while handling an edge case in `to_clip_space`
- which is called in lines
- which come from the Axis3 background poly plots

## Type of change

## Checklist

### DIFF
--- a/src/utilities/Plane.jl
+++ b/src/utilities/Plane.jl
@@ -5,7 +5,10 @@ struct Plane{N, T}
     distance::T
 
     function Plane{N, T}(normal::Vec{N, T}, distance::T) where {N, T <: Real}
-        return new{N, T}(normalize(normal), distance)
+        n = norm(normal)
+        ϵ = 100 * max(eps.(normal)...)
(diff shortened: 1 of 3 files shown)


src/utilities/Plane.jl
```diff
@@ -5,7 +5,10 @@ struct Plane{N, T}
     distance::T
 
     function Plane{N, T}(normal::Vec{N, T}, distance::T) where {N, T <: Real}
-        return new{N, T}(normalize(normal), distance)
+        n = norm(normal)
```
SimonDanisch, line 8: I think a comment would be nice her!

SimonDanisch (approved):

SimonDanisch (approved):

Issue #4688, opened 2024-12-22
MacOS/GLMakie: NSException when displaying GLMakie figure


asinghvi17:
- [X] what version of Makie are you running? (`]st -m Makie`) master
- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) 
- [X] What platform + GPU are you on? Mac M1

I randomly get uncaught `SIGABRT`s when displaying figures on Mac M1 with GLMakie:
```julia
julia> using GLMakie

julia> fig = Figure()
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow should only be instantiated on the main thread!'
*** First throw call stack:
(
)
libc++abi: terminating due to uncaught exception of type NSException

[45503] signal 6: Abort trap: 6
in expression starting at none:0
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 14353417 (Pool: 14352381; Big: 1036); GC: 17

```

That's at the start of the session, no code run before it.  It's also sporadic - no guarantee if I will hit it or not.

asinghvi17:
I also occasionally see:
```julia


julia> fig = Figure()
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow should only be instantiated on the main thread!'
*** First throw call stack:
(
)
libc++abi: terminating due to uncaught exception of type NSException

[50043] signal 6: Abort trap: 6
in expression starting at none:0
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 18205594 (Pool: 18204515; Big: 1079); GC: 19
```

asinghvi17:
Hmm, this might be a thread juggling issue?  I can reliably replicate this with:
```julia
using GLMakie
_task = Threads.@spawn begin
       display(Figure())
end
fetch(task)
```

SimonDanisch:
We've discussed this in https://julialang.slack.com/archives/C688QKS7Q/p1741087503587829
The issue seems, that auto complete evals parts of the code in another thread, which causes problems on some platform.
This seems to be the reproducer:

![Image](https://github.com/user-attachments/assets/1733ff10-6f3b-4da6-bd82-5a2657e1ce81)
This needs to be fixed in Julia:
https://github.com/JuliaLang/julia/issues/57742

asinghvi17:
[...]

PR #4779, opened 2025-02-11
Use StableTasks to pin the thread of the renderloop on MacOS

asinghvi17:
# Description

In any Mac system, the window handler only allows the main thread of an application to handle threads.  For Julia, I assume that's thread id 1.

Every time I try to use GLMakie on Mac, on Julia versions where async tasks can migrate threads, I get a segfault:

```
julia> fig5 = Figure(size = (4000, 4000))
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow should only be instantiated on the main thread!'
*** First throw call stack:
(
)
libc++abi: terminating due to uncaught exception of type NSException

[80423] signal 6: Abort trap: 6
in expression starting at none:0
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 7437143 (Pool: 7436608; Big: 535); GC: 12
zsh: abort      julia
```

This PR uses StableTasks.jl to pin the GLMakie renderloop task to thread id 1, which seems to fix the issue on my system.  I've currently locked the change behind a `Sys.isapple()`, which should mean no changes for Windows and Linux users.  But this needs to be tested.

A consequence of using StableTasks is that the fields of the renderer have to change.  I can revert that if desired by unwrapping the StableTask.

Fixes https://github.com/MakieOrg/Makie.jl/issues/4688

No relation to the other NS* errors that were quite mysterious.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fixed issue with `WGLMakie.voxels` not rendering on linux with firefox [#4756](https://github.com/MakieOrg/Makie.jl/pull/4756)
 - Cleaned up surface handling in GLMakie: Surface cells are now discarded when there is a nan in x, y or z. Fixed incorrect normal if x or y is nan [#4735](https://github.com/MakieOrg/Makie.jl/pull/4735)
 - Cleaned up `volume` plots: Added `:indexedabsorption` and `:additive` to WGLMakie, generalized `:mip` to include negative values, fixed missing conversions for rgba algorithms (`:additive`, `:absorptionrgba`), fixed missing conversion for `absorption` attribute & extended it to `:indexedabsorption` and `absorptionrgba`, added tests and improved docs. [#4726](https://github.com/MakieOrg/Makie.jl/pull/4726)
+- Fixed issue with `GLMakie.Screen` not being on the main thread, causing rendering issues on macOS. [#4779](https://github.com/MakieOrg/Makie.jl/pull/4779)
 
 ## [0.22.1] - 2025-01-17
 


asinghvi17:
I have no clue - but when I start my Julia process with current master and 1 thread, there are no issues.  If I start it with 2 or more threads, then I get this.  

It may be that the parent process can't migrate but the _spawned_ process still can...but not 100% sure how that works.

asinghvi17:
There's also a persistent issue here where closing the window causes a deadlock somehow:
```julia

julia> ┌ Warning: error closing screen
│   exception =
│    deadlock detected: cannot wait on current task
│    Stacktrace:
│      [1] error(s::String)
│        @ Base ./error.jl:35
│      [2] wait(t::Task)
│        @ Base ./task.jl:367
│      [3] wait(t::StableTasks.StableTask{Any})
│        @ StableTasks.Internals ~/.julia/packages/StableTasks/3CrzR/src/internals.jl:15
│      [4] wait(x::GLMakie.Screen{GLFW.Window})
│        @ GLMakie ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:545
│      [5] #stop_renderloop!#83
│        @ ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:947 [inlined]
│      [6] stop_renderloop!
│        @ ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:939 [inlined]
│      [7] close(screen::GLMakie.Screen{GLFW.Window}; reuse::Bool)
│        @ GLMakie ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:753
│      [8] close
│        @ ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:739 [inlined]
│      [9] renderloop(screen::GLMakie.Screen{GLFW.Window})
│        @ GLMakie ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:1077
│     [10] (::GLMakie.var"#79#81"{GLMakie.Screen{GLFW.Window}})()
│        @ GLMakie ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:920
│     [11] (::GLMakie.var"#80#82"{StableTasks.AtomicRef{Any}, GLMakie.var"#79#81"{GLMakie.Screen{GLFW.Window}}})()
│        @ GLMakie ~/.julia/packages/StableTasks/3CrzR/src/internals.jl:107
└ @ GLMakie ~/.julia/dev/makie/Makie.jl/GLMakie/src/screen.jl:1079
```

SimonDanisch:
Since I don't have a Mac I can't debug this... Could you print the threadid from where it happens?
Want to make sure it doesn't hide another issue

asinghvi17:
Will do, Jameson also sent some info on what could be a better solution (to do what e.g. Gtk.jl does), will try to implement that this week.

Have to give a talk in an hour but will try to print the threadid tomorrow!

ffreyer:
Could this just be `Threads.@spawnat 1 renderloop()` or does that not keep the task on the master thread?

asinghvi17:
If we can get the GLFW PR in that would be good to at least stop segfaults on Macs for the time being.  Meanwhile I believe the fix is on the REPL end, since it was invoking `display` on a different thread, which caused this issue on MacOS specifically, and only when the paste was done with a newline in the clipboard.

PR #4783, opened 2025-02-12
Fix image name collision in docs

### LINKED ISSUE #4388: Apparent error in docs on "transparency"
In the docs "Problems with transparency" are described (https://docs.makie.org/stable/explanations/transparency#Problems-with-transparency). 

I do obtain the same plot as is shown (overlap of red and blue planes) with `GLMakie`, but not with `CairoMakie`. The `CairoMakie` plot looks like this:
cairo_makie
Which is very different from what is shown in the docs.

The error seems to be that the default in the docs has switched from `CairoMakie` to `GLMakie` at some point as the corresponding part in the markdown file does not specify `CairoMakie` as backend in the first case:
https://github.com/MakieOrg/Makie.jl/blob/899eeb591a151194605d5611232dd9a44d3fc46b/docs/src/explanations/transparency.md?plain=1#L54-L73

ffreyer:
# Description

Fixes #4388

## Type of change

### DIFF
--- a/docs/figure_block.jl
+++ b/docs/figure_block.jl
@@ -161,7 +161,8 @@ function Documenter.Selectors.runner(::Type{FigureBlocks}, node, page, doc)
     end)
     el.info = "@example $blockname"
 
-    id = string(hash(el.code), base = 16)[1:7]
+    backend = get(kwargs, :backend, :CairoMakie)
+    id = string(hash(backend, hash(el.code)), base = 16)[1:7]
     el.code = transform_figure_code(el.code; id, page = page.source, pagetitle = title, is_continued, kwargs...)
     Documenter.Selectors.runner(Documenter.Expanders.ExampleBlocks, node, page, doc)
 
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
I think the fix should rather be to hash the code after the transformation step a line below? 

I actually don't remember the reason for this design, but I wonder if one couldn't just use the page plus a running index to disambiguate. Using the transformed code there would still be collision problems when two examples have the same code, like a simple `fig` where some state is modified in between to show an effect on the visual output.

ffreyer:
The transformation step also adds `save($(id)...)` so the id needs to exist beforehand

jkrumbiegel:
Ah ok missed that on my phone, can you try the running index version instead, where the hash is `page, number_of_figure_block`? Not sure how to get the number, probably needs some kind of state sadly. But we already have state, maybe just reuse that. Could also just do `number_of_figure_globally` as we have a dict of all of them I guess?

SimonDanisch:
@jkrumbiegel does this look correct now? :)

PR #4787, opened 2025-02-12
Add documentation on mesh normal generation

### LINKED ISSUE #140: Document mesh connectivity triangle nodes need to be listed counter-clockwise
It took me a while to figure out why my mesh looked dark. Turns out the triangle coordinates need to be listed counter-clockwise.

ffreyer:
# Description

Fixes #140

Briefly documents that vertex order within a face affects normal generation and adds an example highlighting the difference.

## Type of change

- Documentation

### DIFF
--- a/docs/src/reference/plots/mesh.md
+++ b/docs/src/reference/plots/mesh.md
@@ -7,6 +7,10 @@ mesh
 
 ## Examples
 
+### Simple mesh plots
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Would probably help quick understanding to label the vertices with 1, 2, 3

PR #4788, opened 2025-02-12
Add a MacOS test runner to GLMakie

asinghvi17:
# Description

There are a lot of issues on Mac that we don't catch, this PR is a foot in the direction of catching those.
We may also want to set the JULIA_NUM_THREADS optionally in some runs to test if we have multithreading issues.

This is a CI change only.

### DIFF
--- a/.github/workflows/reference_tests.yml
+++ b/.github/workflows/reference_tests.yml
@@ -79,6 +79,9 @@ jobs:
           - ubuntu-20.04
         arch:
           - x64
+        include:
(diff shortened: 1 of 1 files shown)


asinghvi17:
Looks like we might need to implement the suggestions in https://developer.apple.com/library/archive/technotes/tn2229/_index.html#//apple_ref/doc/uid/DTS40008924-CH1-SECTION2 to get Mac working on CI.

Issue #4750, opened 2025-01-24
How should we handle transformations, projections and `space`?


ffreyer:
# Current Situation

## Pipeline (From user data to pixels)

(Conversions don't really matter here, they are just included for completion sake.)

1. conversion
	1. expand args - add automatic args, e.g. x, y for image
	2. dim_convert - extracting units (etc) annd/or merging with existing units
	3. convert_arguments - normalize data types and structure
2. transformation
	1. transform_func - apply (arbitrary?) function to converted data
	2. \* float32convert - make data Float32 compatible
	3. model - 4x4 transformation matrix: "model" to "world" space
3. projection
	0. \* float32convert - make data Float32 compatible
	1. view - 4x4 matrix:: "world" to "camera/eye/view" space (i.e. viewer centered coordinate system, applies translation, rotation, mirroring)
	2. projection - 4x4 matrix:: "camera/eye/view" to "clip" space (-1..1^3 box, applies scaling, perspective)
	3. viewport - transforms -1..1 space to px units (internal in W/GLMakie)

\* float32convert is an odd one out:
- 3.1 (+3.2) are derived from limits, and must be Float32 compatible. If they are not, the float-unsafe part gets extracted as a LinearScaling(scale, translation)
- 2.3 must also be Float32 compatible to run on the GPU. Thus the linear scaling is permuted with model. 
	- If this is not possible model is applied on the CPU after 2.1, then float32convert applies in 2.2 and identity is used on the GPU in 2.3
	- else float32convert applies after 2.1 and before model in 2.3
- For CPU projections float32converts get merged back into view/projectionview (i.e. apply in 3.0)

Ownership:
- (2.2, 3.1, 3.2, 3.3) float32convert, view, projection and viewport are owned by the scene/camera
- (2.1, 2.3) transform_func and model are owned by the plot but inherited from the scene (which may get its transformations from an Axis (x/yscale) or Axis3 (model), or another Block)
- (1.2) dim_converts are owned by the Axis and synchronized with plots
- (1.1, 1.3) convert_arguments is handled by the plot

## `space` Attribute

Originally `space` only affected projections:
- `:data` uses pipeline with default `view, projection` of camera
- `:pixel` use `view, projection` that map pixel units to clip space
- `:clip` use constant `view, projection = I`
- `:relative` use constant `view, projection` scaling 0..1 -> -1..1

Notes: 
[...]

PR #4792, opened 2025-02-14
Isolate `space` from transformations

### LINKED ISSUE #3796: Document `Makie.spaces()` values
### Feature description

`?scatter` says 

     •  space::Symbol = :data sets the transformation space for box encompassing the
       volume plot. See Makie.spaces() for possible inputs.

so 

```jl
julia> Makie.spaces()
(:data, :pixel, :relative, :clip)
```

but that function isn't documented

```jl
help?> Makie.spaces()
  No documentation found.

  Makie.spaces is a Function.
```

so I don't know what those things mean.

ffreyer:
# Description

Closes #3796

This pr removes `space` from transformations and adds more convenience inputs to the `transformation` attribute, as suggested in #4750. 

The goal of this is to give `space` a clear and well defined scope - that of "projections", i.e. coordinate spaces that could be handled by a camera. "Transformations", i.e. the model matrix and transform function (or scale in Axis) are completely cut off from `space` here. They would instead be handled through the `transformation` attribute which now accepts:
- `automatic, :auto, :automatic`: inherit transformations from parent based on `space` compatibility
- `:inherit`: inherit transformations regardless of space compatibility
- `:inherit_model`: inherit just the model matrix (i.e. rotate!(), scale!(), translate!()) regardless of space compatibility
- `:inherit_transform_func, :inherit_transform_function`: inherit just transform function regardless of space compatibility
- `nothing, identity, :nothing, :identity`: no inheritance, create new identity transformation

Previous options such as an explicit `Transformation()` and arguments for `transform()` (e.g. `:xy`, `(translate = (1,2,3),)`) continue to work. The latter assumes `automatic` for the inheritance setting. `(inheritance_option, transform_arg)` is currently also allowed.

TODO:
- [x] test transformation inheritance options
- ~~rename `space = :data` to `:world`~~ currently vetod by Simon
- [x] check/update/add space related tests
  - [ ] potentially test that every plot processes space, transformations correctly using #4804
- [x] document/update `space` documentation
- [x] update conversion/transform/projection pipeline docs
  - [x] clean up TODO in those docs
  - [x] fix nested list
- [x] document `transformation` attribute
- [x] document `transform!()` functions
- [x] document `spaces()`
- [x] check/update other space related docstrings (e.g. boundingbox, project, ...?)
- [ ] consider updating interface for functions other than plots (e.g. project(), mouseposition(), boundingbox(), position_on_plot(), ...) (this may be a lot of not strictly required work better suited for other prs)

Tangential changes:
- added convenience methods for `is_..._space()` to work with Observable and plot inputs
- cleaned up `space` handling to rely on `is_..._space()` more and assume `plot.space` exists when it should (i.e. for primitive plots mostly in backends)

## Type of change

## Checklist


SimonDanisch:
I'm for the following names:

* :automatic: inherit transformations from parent based on space compatibility
* :inherit: inherit transformations regardless of space compatibility
* :inherit_model: inherit just the model matrix (i.e. rotate!(), scale!(), translate!()) regardless of space compatibility
* :inherit_transform_func: inherit just transform function regardless of space compatibility
* :nothing : no inheritance, create new identity transformation

I'm inclined to use `automatic` and `nothing`, but I think it wins to have just one type for that field, especially once computegraph has more strictly typed inputs.
Since this shouldn't be set that often, I don't see a big argument for allowing more convenient alternative names.

ffreyer:
This pr has some niche edge cases where it changes behavior. For example if you do
```julia
plot!(..., transformation = Transformation(parent), space = :pixel)
```
and `parent` has a transform_func it would not be applied by `apply_transform()` on master because `space !== :data`. On this branch it would be applied. I would assume that is what you want if you explicitly inherit transformations like this though.

PR #4804, opened 2025-02-17
Generic plot tests

ffreyer:
# Description

I wanted to test that every plot correctly initializes and propagates `transformation` in #4792. I figured it would be good to do this in a more generalized way, so I worked on creating a function that produces a sample plot for everything we have. That should simplify writing tests a bit by making it easier to just get a plot for some random plot type.

Currently tests that every plot works in a `Scene()`.

Fixes h/vlines, h/vspan, ablines and datashader (?) not working in `Scene()` due to `projview_to_2d_limits` not working without a `Float32Convert`.

## Type of change

## Checklist

### DIFF
--- a/test/testplot.jl
+++ b/test/testplot.jl
@@ -1,0 +1,102 @@
+using SparseArrays
+using Random
+
+const ALL_PLOT_PRIMITIVES = [
+    Scatter, Lines, LineSegments, Makie.Text, Mesh, MeshScatter, Image, Heatmap,
+    Surface, Volume, Voxels
+]
+const ALL_PLOT_TYPES = vcat(ALL_PLOT_PRIMITIVES, [
+    # skipped Annotations, Axis3D
+    ABLines, Arc, Arrows, Band, Makie.BarPlot, Bracket, Contourf, Contour, Contour3d,
+    DataShader, Errorbars, Rangebars, HLines, VLines, HSpan, VSpan, Pie, Poly,
+    RainClouds, ScatterLines, Series, Spy, Stairs, Stem, StreamPlot, TimeSeries,
+    Tooltip, Tricontourf, Triplot, VolumeSlices, Voronoiplot, Waterfall, Wireframe,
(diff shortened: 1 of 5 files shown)


test/testplot.jl
```diff
@@ -0,0 +1,102 @@
...
+Any keyword arguments are passed to the plot function.
+
+This is meant to simplify creating throw-away plots for testing attributes.
+`ALL_PLOT_TYPES` may also be useful.
+"""
+function testplot!(scene, ::Type{PlotType}, kwargs...) where {PlotType <: Plot}
+    CT = Makie.conversion_trait(PlotType)
+    f = Makie.MakieCore.plotfunc!(PlotType)
+
+    if CT === PointBased() || PlotType <: Union{Makie.Text, ABLines, BarPlot,
```
asinghvi17, line 30: Can we make this multiple dispatch based, maybe so it returns aplotspec that can be decomposed to args and kwargs and then plotted?

That way it's easy to extend as we get more recipes, or for external recipes we may want to test.

ffreyer: Yea I guess multiple dispatch would be better if another library wants to reuse and extend the Code for testing. If it's just used here I don't think multiple dispatch matters much.

I'm not sure if PlotSpec is a good idea. It's calling `to_value` on kwargs/attributes and I'd assume it circumvents parts of the normal plot pipeline. It should probably be tested separately

asinghvi17: Makes sense, maybe just returning a tuple `(; args, kwargs)` might work?

ffreyer:
Not sure how merge worthy this pr is. As it is now, it only contains a sanity check that plots can be constructed in a plain scene and a small fix for h/vlines/span which only worked with float32convert present, i.e. in an Axis. Having a testplot constructor may be useful for other prs, but I doubt it has much general use. 
The point of this is really to enable checking things that every plot needs to have/needs to do, like common/core attributes. But these are currently inconsistent and it doesn't make much sense to work on them until after #4630. So maybe it's best to leave this pr until then, as a reminder?

PR #4817, opened 2025-02-21
resize wgl content using ResizeObserver

hexaeder:
# Description
Currently, each WGLMakie figure registers a `resize` event listener on the `window` to trigger `resize_to=:parent` updates. This checks for changes in viewport but not changes of the individual elements (i.e. the MWE below does not work correctly until you change the viewport size).

I propose to change this to a `ResizeObserver` on the parent itself.

I didn't add or change any tests since I don't know how something like this could be tested.
The [browser compatiblity](https://caniuse.com/resizeobserver) seems good but not great. Do we need some kind of fallback? Not sure about the WGLMakie project goals in that regard.

```julia
using Bonito, WGLMakie
WGLMakie.activate!(resize_to=:parent)

app = App() do session
    fig = scatter(rand(3))
    Card(fig; style=Styles(
        "height"=>"500px",
        "width"=>"500px",
        "resize"=>"both",
        "overflow"=>"hidden",
    ))
end;

isdefined(Main, :server) && close(server)
server = Bonito.Server(app, "127.0.0.1", 8082)
```

## Type of change

## Checklist

### DIFF
--- a/WGLMakie/src/wglmakie.bundled.js
+++ b/WGLMakie/src/wglmakie.bundled.js
@@ -24361,8 +24361,9 @@ function add_canvas_events(screen, comm, resize_to) {
     }
     if (resize_to) {
         const resize_callback_throttled = Bonito.throttle_function(resize_callback, 100);
-        window.addEventListener("resize", (event)=>resize_callback_throttled());
-        setTimeout(resize_callback, 50);
+        const resizeObserver = new ResizeObserver(resize_callback_throttled);
+        resizeObserver.observe(canvas.parentElement);
+        resize_callback();
     }
 }
 function threejs_module(canvas) {
--- a/WGLMakie/src/wglmakie.js
+++ b/WGLMakie/src/wglmakie.js
@@ -353,13 +353,14 @@ function add_canvas_events(screen, comm, resize_to) {
             resize_callback,
             100
         );
-        window.addEventListener("resize", (event) =>
-            resize_callback_throttled()
-        );
+
+        // use resize observer to watch parent size changes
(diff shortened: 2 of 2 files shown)


SimonDanisch:
I've tested this quickly with a complex app, and it makes some problems. Haven't had time to pin it down

hexaeder:
The two problematic things i can think of are
a) element exists without a parent, therefore the resizeObserver cannot get registered
b) the parent changes

I was not able to create a MWE for a). For b) I created the MWE below (Fig switches between two resizable Cards), but surprisingly it just works 🤔 
```julia
app = App() do session
    fig = scatter(rand(3))
    canvas = Bonito.jsrender(session, fig)
    btn = Bonito.Button("Switch Pane")
    style = Styles("height"=>"300px", "width"=>"500px", "resize"=>"both", "overflow"=>"hidden")

    cardA = Card(canvas; style, class="classA")
    cardB = Card(""; style, class="classB")
    switchjs = js"""
    const element = $(canvas);

    function switchCard(btn) {
        const parent = element.parentElement;
        const parentA = document.querySelector(".classA");
        const parentB = document.querySelector(".classB");
        if (parent.classList.contains("classA") && parentB) {
            parent.removeChild(element);
            parentB.appendChild(element);
        } else if (parent.classList.contains("classB") && parentA) {
            parent.removeChild(element);
            parentA.appendChild(element);
        }
    }

    $(btn.value).on(switchCard);
    """
    evaljs(session, switchjs)
    DOM.div(btn,cardA, cardB)
end;
isdefined(Main, :server) && close(server)
server = Bonito.Server(app, "127.0.0.1", 8081)
```

Issue #4074, opened 2024-07-31
menu with lots of options slows down keyboard entry into textbox


bjarthur:
on M2 apple silicon, a `Menu` with lots (say 10k) options slows down entering characters into a `Textbox`, but only on WGLMakie, not GLMakie.  like this:

```
using WGLMakie
        
f = Figure()
Textbox(f[1,1], placeholder="type quickly")
Menu(f[2,1], options = string.(1:10_000))
```

if you type quickly enough and length(options) is large enough, the characters will appear out of order in the textbox.

i'll agree with anyone who says a menu with 10k options is a bad UI, but still, why the performance problems in WGLMakie but not GLMakie?

bjarthur:
but just entering text into the box shouldn't trigger any menu related events.  so why does just having the menu there with lots of options slow the textbox down?

ffreyer:
This is more or less caused by layouting. Typing in the Textbox resizes it. Without setting tellwidth and tellheight to false, the resize propagates to the Menu block and then every plot inside. Allocations confirm this:

Writing "1" to an empty Textbox with your example code:
- both true (default): 6.4M allocs
- one true, one false: 2.6M allocs
- both false: 5.6k allocs

<details>
<summary>code</summary>

```julia
begin
    using WGLMakie

    f = Figure()
    Textbox(f[1,1], placeholder="type quickly", tellwidth = false, tellheight = false)
    Menu(f[2,1], options = string.(1:10_000), tellwidth = false, tellheight = false)
    f
end

begin
    str = "1"
    ks = map(collect(str)) do letter
        key = if letter == ' '
            Keyboard.space
        elseif letter == '.'
            Keyboard.period
        elseif isnumeric(letter)
            eval(:(Keyboard.$(Symbol(lowercase("_$letter")))))
        else
            eval(:(Keyboard.$(Symbol(lowercase(letter)))))
        end
    end

    @time begin
        events(f).mouseposition[] = (300, 250)
        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.release)

        for (letter, key) in zip(str, ks)
            events(f).keyboardbutton[] = Makie.KeyEvent(key, Keyboard.press)
            events(f).unicode_input[] = letter
            events(f).keyboardbutton[] = Makie.KeyEvent(key, Keyboard.release)
        end
    end
end
```
[...]

PR #4821, opened 2025-02-24
Widget optimizations

ffreyer:
# Description

This pr is somewhat in response of #4074 but not solely focused on it. There are mainly two types of changes here:

First is removing `pick()` from all Blocks (except Axis3 center-on-click interactions). It requires data to be copied from the GPU to CPU which is very expensive compared to bounding box checks, which should be sufficient for blocks. I also added some debug tracking for `pick()` calls and use those to test that blocks don't rely on them

The second type of changes is driven by message counts as reported by Bonito. Specifically, I ran code like this to check and compare message counts:
```julia
using Bonito, WGLMakie
begin
    f = Figure()
    t1 = Textbox(f[1, 1])
    m = Menu(f[2, 1], options = string.(1:100))
    display(f)

    # select TextBox
    events(f).mouseposition[] = (300, 250)
    events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
    events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.release)

    # emulate text input & track updates
    all_messages, summary_str = Bonito.collect_messages() do
        events(f).unicode_input[] = 't'
        events(f).unicode_input[] = 'e'
        events(f).unicode_input[] = 's'
        events(f).unicode_input[] = 't'
    end
    summary_str
end
```
Commits with notes like `32 -> 4` refer to reductions in those message counts. Changes made in this context can have heavy overlap with or be made redundant by #4630. For example text input often creates a lot of redundant updates because most text things trigger an update of the glyph collection which then triggers an update of everything that depends on the glyph collection. #4630 is partially meant to deal with that, so there's little point in optimizing it here.
Instead of that most of the message count reductions here are related to not doing unnecessary work. 

With respect to #4074 mainly e7c5450cf3a31818423226634bc69abd62108b28 and 55cd8a729d4d812669d42c6e2648c0f549bd10fa are relevant. These commits remove updates of hidden menu items caused by layouting changes. Before there were 72 messages per letter typed and the total size scaled with the number of menu items. Now there are 39 (just Textbox triggers 30 per letter) and no scaling with the number of Menu items.

TODO:
- ~~add message count tests~~
- [x] check other Blocks for excessive message counts

## Type of change

## Checklist

### DIFF
--- a/WGLMakie/test/message_counting.jl
+++ b/WGLMakie/test/message_counting.jl
@@ -1,0 +1,102 @@
+# These tests are relevant for general Makie performance but only tested here
+# because Bonito makes it easy to test. If the tests fail it is likely not
+# WGLMakies fault.
+# A lower number of messages maybe caused by optimizations or broken interactions.
+# A higher number could come from added functionality or performance regressions.
+
+
+@testset "TextBox with Menu" begin
+    f = Figure()
+    t1 = Textbox(f[1, 1])
+    m = Menu(f[2, 1], options = string.(1:1000))
+    display(edisplay, App(() -> f))
+
+    # trigger select
+    all_messages, summary_str = Bonito.collect_messages() do
+        events(f).mouseposition[] = (300, 250)
+        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
+        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.release)
+    end
+    @test length(all_messages) == 44
+
+    # type text
+    for (char, expected) in zip(collect("test"), [18, 39, 39, 39])
+        _key = eval(:(Makie.Keyboard.$(Symbol(char))))
+        all_messages, summary_str = Bonito.collect_messages() do
+            events(f).keyboardbutton[] = Makie.KeyEvent(_key, Keyboard.press)
+            events(f).unicode_input[] = char
+            events(f).keyboardbutton[] = Makie.KeyEvent(_key, Keyboard.release)
+        end
+        @test length(all_messages) == expected
+    end
+end
+
+@testset "Menu" begin
+    f = Figure()
+    m = Menu(f[1,1], options = string.(1:10))
+    display(edisplay, App(() -> f))
+
+    # open menu
+    all_messages, summary_str = Bonito.collect_messages() do
+        events(f).mouseposition[] = (300, 230)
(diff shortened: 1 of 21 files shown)


ffreyer:
After some discussion with Simon we decided to skip the message counting tests for now. #4630 should make it possible to handle them more easily and across backends. So we'll defer them to there instead of wasting time here trying to figure out why these tests don't work here.

WGLMakie/test/message_counting.jl
```diff
@@ -0,0 +1,102 @@
...
+    all_messages, summary_str = Bonito.collect_messages() do
+        events(f).mouseposition[] = (300, 250)
+        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.press)
+        events(f).mousebutton[] = Makie.MouseButtonEvent(Mouse.left, Mouse.release)
+    end
+    @test length(all_messages) == 44
+
+    # type text
+    for (char, expected) in zip(collect("test"), [18, 39, 39, 39])
+        _key = eval(:(Makie.Keyboard.$(Symbol(char))))
```
asinghvi17, line 24: Could this be getproperty instead of eval?

SimonDanisch: yes

src/interaction/interactive_api.jl
```diff
@@ -126,12 +140,23 @@ Return all `(plot, index)` pairs in a `(xy .- range, xy .+ range)` region
 sorted by distance to `xy`. See [`pick`](@ref) for more details.
 """
 function pick_sorted(scene::Scene, xy, range)
+    PICK_TRACKING[] && (_PICK_COUNTER[] += 1)
     screen = getscreen(scene)
     screen === nothing && return Tuple{AbstractPlot, Int}[]
+    if PICK_TRACKING[]
+        # if the Makie implementation is used we'd double count if we just increment here
```
SimonDanisch, line 147: Maybe we should let the backend do this and just have one counter in native_picking?

ffreyer: I generally prefer testing things that don't require a backend without the backend. I don't have a strong reason for it though. I guess it helps balance out test times so a full CI run doesn't get longer?
Backends also have multiple methods that would need tracking (point, rect, maybe closest, sorted) and we'd have to duplicate code across backends if we added it there.

PR #4832, opened 2025-02-27
Fix faulty font selection

### LINKED ISSUE #620: Allow `scatter` to accept keyword `font`
When using unicode characters as markers for a `scatter` plot, it is convinient to be able to provide a font to choose the characters from, since the default font may be missing some.

Following the discussion on Slack, Simon mentioned:

---

It tries to find the font in AbstractPlotting._alternative_fonts

https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/utilities/texture_atlas.jl#L145

that's the code figuring out, what font to use to render a particular unicode character in the scatter

I guess you could manually load a font with FreeTypeAbstration.FTFont and push it to that array

---

This process is a bit too low level and I think it would be more safe if this (or a similar) process was supported via a `font` keyword instead.

### LINKED ISSUE #4830: marker size different between backends
The same code results in very different marker sizes with Cairo and with GL:
```julia
scatter((0,0), marker=:circle, markersize=30, label="A")
scatter!((1,0), marker='∘', markersize=80, label="B")
scatter!((0,1), marker='◇', markersize=45, label="C")
axislegend()

save("tmp_gl.png", current_figure(), backend=GLMakie)
save("tmp_cairo.png", current_figure(), backend=CairoMakie)
```
Cairo:
Image
GL:
Image

- [x] what version of Makie are you running? (`]st -m Makie`) v0.22
- [x] What platform + GPU are you on? any

ffreyer:
# Description

Probably fixes #4830
Closes #620

The code from the first commit always picked the last font alternative because it was missing a `break`. CairoMakie picks the first working alternative, as it should. `find_font_for_char` does too, so this should equalize GLMakie and CairoMakie. (Both char markers from #4830 run into this issue.)

TODO:
- [x] add `font` to scatter (because CairoMakie, GLMakie and WGLMakie all suggest it should be there)
- [x] add tests
- [x] Figure out what to do with theme inheritance

## Type of change

## Checklist

### DIFF
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -320,7 +320,7 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Scat
     isempty(positions) && return
     size_model = transform_marker ? model : Mat4d(I)
 
-    font = to_font(to_value(get(primitive, :font, Makie.defaultfont())))
+    font = to_font(primitive.fonts, primitive.font[])
     colors = to_color(primitive.calculated_colors[])
     markerspace = primitive.markerspace[]
     space = primitive.space[]
--- a/CairoMakie/src/primitives.jl
+++ b/CairoMakie/src/primitives.jl
@@ -320,7 +320,7 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Scat
     isempty(positions) && return
     size_model = transform_marker ? model : Mat4d(I)
 
-    font = to_font(primitive.fonts, primitive.font[])
+    font = to_font(to_value(get(primitive, :font, Makie.defaultfont())))
     colors = to_color(primitive.calculated_colors[])
     markerspace = primitive.markerspace[]
     space = primitive.space[]
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -484,10 +484,8 @@ Plots a marker for each element in `(x, y, z)`, `(x, y)`, or `positions`.
     marker_offset = Vec3f(0)
     "Controls whether the model matrix (without translation) applies to the marker itself, rather than just the positions. (If this is true, `scale!` and `rotate!` will affect the marker."
     transform_marker = false
-    "Sets the font used for character marker. Can be a `Symbol` which will be looked up in the `fonts` dictionary or a `String` specifying the (partial) name of a font or the file path of a font file"
-    font = @inherit font
-    "Used as a dictionary to look up fonts specified by `Symbol`, for example `:regular`, `:bold` or `:italic`."
-    fonts = @inherit fonts
+    "Sets the font used for character markers."
+    font = automatic
     "Optional distancefield used for e.g. font and bezier path rendering. Will get set automatically."
     distancefield = nothing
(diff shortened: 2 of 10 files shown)


MakieCore/src/basic_plots.jl
```diff
@@ -484,6 +484,8 @@ Plots a marker for each element in `(x, y, z)`, `(x, y)`, or `positions`.
     marker_offset = Vec3f(0)
     "Controls whether the model matrix (without translation) applies to the marker itself, rather than just the positions. (If this is true, `scale!` and `rotate!` will affect the marker."
     transform_marker = false
+    "Sets the font used for character markers."
+    font = automatic
```
ffreyer, line 488: I only added `font` here with a `to_font(::Automatic) = defaultfont()` method in conversions.jl. So I didn't mirror the `fonts` interface from text. Not sure whether that's the right way to go about it.

ffreyer:
Failing test is "scatter rotation" which shows the inconsistency from #4830. With font handling fixed that test is now consistent across backends and also has different limits due to the marker metrics changing.

CairoMakie/src/primitives.jl
```diff
@@ -320,7 +320,7 @@ function draw_atomic(scene::Scene, screen::Screen, @nospecialize(primitive::Scat
     isempty(positions) && return
     size_model = transform_marker ? model : Mat4d(I)
 
-    font = to_font(to_value(get(primitive, :font, Makie.defaultfont())))
+    font = to_font(primitive.fonts, primitive.font[])
```
ffreyer, line 323: fonts end up as `::Attributes` so I've just passed them directly. Not sure if that's correct though

ffreyer:
If you don't change the theme this pr just makes GLMakie and WGLMakie do the same thing as CairoMakie does and did. I would call that a bug fix, since backends should do the same thing. If you do change the theme, then yes, this pr changes things.

Maybe we should take some more time to think about whether we should inherit scatter font(s) at all. I don't think I would expect markers to change if I set a font in theme. I would also not expect text.font and scatter.font to be the same, because they have different goals. (Text cares about multiple characters coming together while Scatter cares about the individual symbols and probably wants consistent thickness, symmetry, simplicity, ...) Perhaps it would be better to split text and scatter font(s) in the theme. And if we split them I would also drop the `fonts` interface for scatter, because, well, when is that ever going to be useful for scatter?

jkrumbiegel:
> I don't think I would expect markers to change if I set a font in theme. I would also not expect text.font and scatter.font to be the same, because they have different goals. (Text cares about multiple characters coming together while Scatter cares about the individual symbols and probably wants consistent thickness, symmetry, simplicity, ...) Perhaps it would be better to split text and scatter font(s) in the theme. And if we split them I would also drop the fonts interface for scatter, because, well, when is that ever going to be useful for scatter?

agreed, I would be fine with scatter just getting a working `font` selection and it doesn't need `fonts`

ffreyer:
I think I was done with this/waiting for review. The only ...

> Failing test is "scatter rotation" which shows the inconsistency from #4830. With font handling fixed that test is now consistent across backends and also has different limits due to the marker metrics changing.

So if we're happy with theming and font handling we can merge this

SimonDanisch (approved):

PR #4844, opened 2025-03-05
Add tutorial on pixel perfect image/heatmap plots

### LINKED ISSUE #4793: Saving image with campixel shifted/clipped by 1 pixel
- CairoMakie v0.13.1
- fresh temporary environment
- Windows 10

Using the code found [here](https://discourse.julialang.org/t/how-can-i-save-a-makie-heatmap-as-png-with-one-pixel-per-matrix-element/70579)
```julia
using CairoMakie
x = rand(30, 40)
scene = Scene(camera=campixel!, size=size(x))
heatmap!(scene, x)
save("out.png", scene, px_per_unit=1)
```
This first issue is that now we must use `px_per_unit=1` to get a png out with the same resolution as the image matrix, otherwise it is twice as big. The bigger issue however, is that one row and column of pixels is clipped, and replaced on the other side with what appears to be an interpolation with white. See this image (zoom in with paint.net or similar to preserve pixels without interpolation):

Image

This problem may in fact be related to the default `px_per_unit=2`, as in that case the image is clearly shifted by 1/2 "unit", aka one pixel:
Image

### LINKED ISSUE #4794: Issue with saving campixel png with GLMakie
- GLMakie v0.11.2
- fresh temporary environment
-Windows 10 with Nvidia GPU

```julia
using GLMakie
x = rand(30, 40); x[1:3,1:3] .= 1.0 # to help see what is happening
scene = Scene(camera=campixel!, size=size(x))
heatmap!(scene, x)
save("out.png", scene, size=size(x))
heatmap(x) # to compare
```

This issue is probably related to issue #4793 but the result is somewhat different, and specific to GLMakie. Here as well the png output looks clipped at the top and right edges, but even more importantly the values are all wrong and look like they were all interpolated from on row to the next...

left, png opened in paint.net, right, GLMakie viewer:
Image

ffreyer:
# Description

Adds a tutorial that explains how to perfectly align matrix data with the pixels grid using image and heatmap plots. Mainly talks about alignment of the plot and scene/cameras, but also mentions how FXAA may affect the output.

Closes #4793 and closes #4794

## Type of change

- [x] Documentation


asinghvi17:
It's also possible to do this with figure_paddi g=0 and `hidedecorations!` and `hidespines!` on the axis, although it is good to introduce people to scenes as well.

SimonDanisch (approved):

Issue #22, opened 2017-11-20
String as marker


mschauer:
It would be nice to use Strings as markers in ``scatter``. 
https://github.com/SimonDanisch/Makie.jl/blob/cef849737a6833c2b96987025c1b90bf4ec1da25/src/plotsbase/converts.jl#L324 also seems strange. Cheers

SimonDanisch:
>It would be nice to use Strings as markers in scatter.

What would you expect that to do? Basically annotations?

`to_spritemarker(b, marker::Vector{Char}) = String(marker)`

That's an artifact from text rendering bein pretty much a scatter plot of chars.

mschauer:
> What would you expect that to do? Basically annotations?

Yes

> That's an artifact from text rendering bein pretty much a scatter plot of chars.

Ah, I see were the problem comes from.

SimonDanisch:
A nicer way of not converting the marker to a string, would be to allow `Union{String, Vector{Char}` in the actual rendering code in GLVisualize.... 

Annotations are pretty much implemented in GLVisualize, I just need to wrap the function and export it in Makie :)

mschauer:
Yes, allowing that in GLVisualize would be best. Still I think there is a bug here:
 https://github.com/SimonDanisch/Makie.jl/blob/2386fb3b8b9784b649c6aca2e5c90f527b43e314/src/plotsbase/converts.jl#L344 gives problems. If you have e.g. a vector of Symbols, that gets converted into a vector of chars, which then is converted into a string in L344, which does not make sense.

SimonDanisch:
Why not? Symbols are just a stand in for unicode characters, and a bunch of unicode characters are represented as a string ;)

mschauer:
The issue in short:
```
 scatter(GeometryTypes.Point3f0[(1,0,0), (0,1,0), (0,0,1)], marker=[:x, :circle, :cross])
ERROR: MethodError: no method matching to_spritemarker(::Makie.Scene{:makie}, ::String)
```

SimonDanisch:
Ahhh, sorry, since you never mentioned that it actually errors, I was a bit slow to follow ;)

mschauer:
>  slow 

That was mutual, it took me also to figure out that it actually should work.

SimonDanisch:
Oh, well no wonder why i didn't see the problem with it - the conversion code is correct, but I don't have the scatter code setup to actually deal with strings for multiple markers :D

[...]

PR #4847, opened 2025-03-05
ComputePipeline docs & tests

ffreyer:
Adds documentation and tests to the ComputePipeline package.

TODO:
- [x] add docstrings to user facing functions
- [x] add documentation page to Makie docs (for now)
- [x] add full system test
- [x] cleanup docs
- [x] fix torrent of warnings and errors about input/computation overwrites in WGLMakie
- [x] check up on CairoMakie

Unclear TODO - add specialized unit tests:
- [x] test errors for `add_inputs!()`
- [x] test errors for `register_computation()`
- [x] test Graph access and manipulation interface
- [x] test io/show
- [x] test `delete!()`

Problems found:
- When inserting a node/Computed from a parent graph into a child graph, the node will keep its `name` as it's the same object. But name/key in `child_graph.outputs` can (and may need to) be different. This seems dangerous for ComputePipeline internals (Easy mistake to think `graph.outputs[key].name == key`.)

Code changes (so they don't get lost between docstrings)
- changed `add_input(conversion_func, ...)` to use an externally defined function wrapper to handle `conversion_function(key, value)`. This improves readability (no `#22()` etc) and error tracing
- restructured `add_input!()` to allow no-callback version to use `identity` instead of effectively creating an anonymous conversion_function version of identity that then wraps to work like identity
- changed `add_input!(graph, ... parent[:node]::Computed)` to never insert `parent[:node]` into the passed graph. Instead it always creates an edge hanging between two graphs. That way parent keeps sole ownership over its nodes. (This required breaking up `register_computation!()` a bit)
- added `setindex!(::Input, ...)` and adjusted `setindex!(::Computed, ...)` to call that if an input with the same name exists. This is a potentially temporary fix to `graph.name[] = new_value` updating the computed child of an input, which would reset on resolve.
- refactored `delete!()` to only be callable with `delete!(graph, key::Symbol)`. Disallowed deletion of nodes with children and siblings unless explicitly specified with `recursive = true` and `force = true` respectively.
- tightened errors:
    - Inputs may not be added more than once
    - ~~computations warn if the exact same computation is added a second time~~
    - computations now error if the only differences is in the inputs of the compute edge
- adjusted io:
    - added name generation for the input function wrapper
    - switched to :light_black for out-of-date compute nodes, values, edges
    - fixed error in empty ComputeGraph show()
    - changed to printing (shortened) values instead of types when tracing errors

Code changes outside of ComputePipeline:
- replaced second initialization of `model` with assert that model has been initialized to fix duplicate `add_input!()` of model. It might be better to remove `:model` as plot argument instead. (That overwrites transformations control over model, if it still works.)


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -6,6 +6,18 @@ using Base: RefValue
 
 abstract type AbstractEdge end
 
+"""
+    struct Computed ... end
```
asinghvi17, line 10:
```suggestion
    Computed
```

or

```suggestion
    struct Computed
```

are the "standard" ways of doing this IIRC

ffreyer: Fixed, thanks

GLMakie/src/new-primitives.jl
```diff
@@ -314,11 +314,16 @@ function draw_atomic(screen::Screen, scene::Scene, plot::Scatter)
     # We register the screen under a unique name. If the screen closes
     # Any computation that depens on screen gets removed
     atlas = gl_texture_atlas()
-    add_input!(attr, :atlas, atlas)
+    if haskey(attr, :atlas)
```
SimonDanisch, line 317: we'll need to do something like `:atlas_64_2048` for disambiguation with WGLMakie.
I already added this here:
https://github.com/MakieOrg/Makie.jl/blob/sd/compute-graph/src/utilities/texture_atlas.jl#L542
I guess we should normalize the adding of the texture atlas somehow... Maybe already in the the Plot constructor.
But I guess this doesn't need to happen in this PR.

GLMakie/src/new-primitives.jl
```diff
@@ -746,8 +751,8 @@ end
 function draw_atomic(screen::Screen, scene::Scene, plot::LineSegments)
     attr = generic_robj_setup(screen, scene, plot)
 
-    add_input!(attr, :px_per_unit, screen.px_per_unit)
-    add_input!(attr, :viewport, scene.viewport)
+    haskey(attr, :px_per_unit) || add_input!(attr, :px_per_unit, screen.px_per_unit)
```
SimonDanisch, line 754: related to the texture atlas comment, I think we may want to only add inputs in Makie.jl / Plot constructor.
ALthough screen couldn't be added there. 
But I still need to think about if we even should add the screen to the plot object.

SimonDanisch (approved):

PR #4782, opened 2025-02-12
fix UInt underflow in line indices

ffreyer:
# Description

Fixes an integer underflow in GLMakie as is happening in https://github.com/MakieOrg/Makie.jl/issues/4771#issuecomment-2652325708. May solve #4373 (but I have no way to confirm that)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [x] Added unit tests for new algorithms, conversion methods, etc.
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #4848, opened 2025-03-05
Fix GLMakie embedding support

JamesWrigley:
# Description

An exit handler was added to GLMakie in https://github.com/MakieOrg/Makie.jl/pull/4782 to ensure that all OpenGL resources are cleaned up, which will eventually call `close(::Screen)` and `destroy!(::Screen)`, which previously assumed that GLMakie owned `screen.glscreen` and would try calling GLFW functions on it. Now they will only do GLFW things if GLMakie owns the screen window. Also documented that embedders will need to implement `destroy!(::MyWindow)`.

Relevant prior PR: https://github.com/MakieOrg/Makie.jl/pull/4073

I don't have the bandwidth for it right now, but in the future I'd like to try writing some unit tests for embedding support.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed GLMakie embedding support [#4848](https://github.com/MakieOrg/Makie.jl/pull/4848).
 
 ## [0.22.2] - 2025-02-26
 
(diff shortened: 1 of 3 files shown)


JamesWrigley:
I believe the benchmark failure is unrelated, maybe it's because I'm making the PR from a fork?

PR #4849, opened 2025-03-06
add smart fillto to enable logscale y-axis stacked barplot

### LINKED ISSUE #4549: stacked `barplot` empty when using `yscale=log`
```
CairoMakie v0.12.15
```

## linear scale
```julia
julia> barplot([1,2,3,1,2,3], [1,2,3,1,2,3], stack=[1,1,1,2,2,2], color=[1,1,1,2,2,2], gap=0)
```
Image

## log scale
```julia
julia> barplot([1,2,3,1,2,3], [1,2,3,1,2,3], stack=[1,1,1,2,2,2], color=[1,1,1,2,2,2], gap=0; axis=(; yscale=log))
```
Image

Moelf:
fix #4549

related to #3004 and refactored a bit of code
```julia
barplot([1,2,3,1,2,3], [1,2,3,1,2,3], stack=[1,1,1,2,2,2], color=[1,1,1,2,2,2], gap=0; axis=(; yscale=log))
```

## Before
Image

## After
image


src/basic_recipes/barplot.jl
```diff
@@ -17,23 +17,28 @@ Returns a Tuple of new y positions and offset arrays.
...
+        _smart_log_clamp(ys)
     else
         return ys, offset
     end
 end
 
+"""
+    _smart_log_clamp(ys, ref=ys)::(ys_clamped, min_ref/2)
+
+Clamps the values of `ys` to a minimum value that is half the minimum non-zero value in `ref`.
```
asinghvi17, line 33: Is there a particular rationale for that value, or just arbitrary?

Moelf: This is heuristic from ROOT

Moelf: we could lower it to min_ref/5 or something, but I guess they felt /2 is more visually balanced in most cases.

Moelf:
Do I need to provide reference plots? is there anything else I should do in the meantime?

asinghvi17:
A mention in the docstring about the source of the heuristic would be nice, but we (maintainers) have to update the reference images

Moelf:
@asinghvi17 https://github.com/MakieOrg/Makie.jl/pull/3004#discussion_r1229596472 we discussed that here, I can't find the exact code now, ROOT does something depends on if minimum is >1 or <1, but approximately /2

Moelf:
maybe something like this https://github.com/root-project/root/blob/d1cfaefcbdefbfb966ed4e8ab694d0c766944b8e/js/modules/hist2d/TH1Painter.mjs#L144-L146

Moelf:
Bump

SimonDanisch:
Looks good to me! @jkrumbiegel, do you have any objections?

jkrumbiegel:
I don't know, it's kind of too arbitrary for me with these "smart" fillto values, I also didn't like the other log one. But I also don't use these kinds of stacked barplots so I don't really mind either.

Moelf:
>it's kind of too arbitrary for me with these "smart" fillto values

I agree it's arbitrary but I think we also agree it shouldn't be empty, because then it's unusable. 

A nonarbitrary choice is to fill down to `eps()`, but that in practice means users are now forced to set `ylims!()` for every single `yscale=log` plot, which I thought would be too anoying.

ffreyer:
> A nonarbitrary choice is to fill down to eps(), but that in practice means users are now forced to set ylims!() for every single yscale=log plot, which I thought would be too annoying. Thus smart fillto is a midway solution.

Since Axis relies on boundingbox() now and boundingbox() processes transformations you can have `-Inf`` in transformed data with sane limits by adding a method. If that's closer to the optimal solution I could update the pr for that.

Moelf:
does that mean user must set limits themselves? or you're saying you can implement this `/2` heuristics there too?

ffreyer:
The latter. You can implement the heurisitic for boundingbox so that Axis uses it for limits while using something like eps() for the actual data.

Moelf:
in that case feel free to do that, and I very much appreciate it

Moelf:
still want this, looks like #4911 stalled

PR #4857, opened 2025-03-08
Fix billboarding on compute branch

ffreyer:
# Description

Putting this up as a separate pr because I'm not sure if this is the right approach.

### Problem

On master, Makie derives `billboard = map(rot -> rot isa Billboard, plot.rotation)`. This reads the user input of `rotation`, so before `convert_attribute()` takes effect.

On the compute pipeline branch `convert_attribute` runs asap. The user input is saved in `Input`, which feeds it to `convert_attribute` and pushes the result into a compute node. Any further computations start from the compute node, which has removed the `Billboard` wrapper. It's possible (by design) to add any more computations and output nodes to an `Input`.

### Options

1. We could rip attributes like this apart, i.e. add `billboard` as an attribute to scatter. This would be breaking and may not always work (e.g. `colormap` currently also doesn't detect if it gets a `PlotUtils.ColorGradient`. Ripping the attribute apart would effectively mean not allowing this type anymore, because we can't extract categorical, colorscale etc)
2. We could explicitly look back at input, e.g. have this to read a pre-conversion value. Seems like bad design to me....
    ```julia
    register_computation!(graph, [:rotation], [:billboard]) do _, changed, cached
        return (graph.inputs[:rotation].value isa Billboard, )
    end
    ```
3. We could defer `convert_attributes()` until after the `Input -> Computed`. This could be done my making `convert_attribute(rot) = rot` (but that's breaking our conversion interface) or by setting the `Input -> Computed` callback to `identity`. In order to still get a converted value from using `graph[:rotation][]`, the compute node needs to have a different name from `Input` in either case.

This pr prototypes option 3 by adding `add_input!([func], graph, input_name, compute_name, initial_value)` to allow differently named compute targets, and by adding an option to defer conversions for certain keys in `add_attributes!()`. With that:
```julia
# defer creation of :rotation node, creating :anon_rotation instead
add_attributes!(Scatter, attr, user_kw, Set([:rotation]))

# derive :rotation w/ conversion and billboard from :anon_rotation
register_computation!(attr, [:anon_rotation], [:rotation, :billboard]) do (input,), changed, cached
    rot = convert_attribute(input[], key"rotation"(), key"scatter"())
    billboard = input[] isa Billboard
    if cached === nothing || cached[2][] != billboard
        return (rot, billboard)
    else
        return (rot, nothing)
    end
end
```


src/compute-plots.jl
```diff
@@ -560,7 +564,17 @@ end
 
 function compute_plot(::Type{Scatter}, args::Tuple, user_kw::Dict{Symbol,Any})
     attr = ComputeGraph()
-    add_attributes!(Scatter, attr, user_kw)
+    add_attributes!(Scatter, attr, user_kw, Set([:rotation]))
+    register_computation!(attr, [:anon_rotation], [:rotation, :billboard]) do (input,), changed, cached
+        rot = convert_attribute(input[], key"rotation"(), key"scatter"())
+        billboard = input[] isa Billboard
```
SimonDanisch, line 570: why not just `attr.rotation[]`?

ffreyer:
Fixed differently

PR #4865, opened 2025-03-12
Broaden draw_marker to AbstractMatrix

jariji:
Allows drawing to Cairo with other `AbstractMatrix` types, such as `PermutedDimsArray`, not just `Matrix`.


ffreyer:
Would be good to add refmage test for this and a changelog entry. This may also need to be handled earlier if the other backends can't deal with special array types

Issue #3323, opened 2023-10-29
Axis3 causes markers of a meshscatter to be transformed


ffreyer:
On master, beta-20 (pre lighting merge, #3113), lighting (#3246) (In GLMakie, but this should be the same across backends)

```julia
fig = Figure()
a = Axis3(fig[1, 1])
ps = decompose(Point3f, Sphere(Point3f(0), 1f0), 10)
arrows!(a, map(p -> Point3f(10, 10, 1) .* p .+ Point3f(0), ps), Vec3f.(ps))
fig
```

![Screenshot from 2023-10-29 14-40-32](https://github.com/MakieOrg/Makie.jl/assets/10947937/abdbada9-8889-40d9-b689-032e6abd4b78)

This is happening because the scaling factor derived from the positional argument is applied to the vertices of the scattered mesh as well. 

To fix this, we should probably add a `transform_marker` attribute to meshscatter and adjust Axis3 to not normalize to a unit box (in terms of internal coordinates, not visuals). I tried this a bit in the lighting pr but ran into issues with near/far. I'll try to look into this again after we get done with v0.20.

PR #4871, opened 2025-03-17
Add `transform_marker` to arrows and default to true in 3D

### LINKED ISSUE #4870: arrows doesn't start at base
- [x] what version of Makie are you running? (`]st -m Makie`) v0.22.2
- [x] What platform + GPU are you on? Arch Linux, integrated Intel graphics on laptop.

just updated, so:
Julia 1.11.4
GLMakie v0.11.3

the arrows drawn here don't start at 0
```julia
fig = Figure()
ax = Axis3(fig[1, 1]) # also with aspect = :data.
o = Point3f(0.0, 0.0, 0.0)
n = Vec3f(0, 0, 1) 
arrows!(ax,[o], [n])
```

Image
Image

ffreyer:
# Description

Fixes #4870

Reverts a change in defaults caused by #4606

Arrows assumes the scattered meshes to transform just like a normal mesh. I.e. to get the correct length all the vertices should effectively go through the full transformation pipeline. This is not the case with `transform_marker = false`, as that excludes model. 

## Type of change

## Checklist


SimonDanisch (approved):

ffreyer:
I'm wondering if we should revert this at the level of `meshscatter`. With `transform_marker = false` the model matrix only applies to meshscatter positions, but not the vertices of the scattered mesh. I.e. it won't affect the size, rotation or translation relative to meshscatter position. That seems like the wrong default to me now.

ffreyer:
This now reverts the change in the default behavior of meshscatter from #4606, i.e. it sets `transform_marker = true` by default for meshscatter. This means that translate!(), rotate!() and scale!() apply to the scattered mesh by default again.

SimonDanisch (approved):

ffreyer:
Would reopen #3323 if it were closed

PR #4876, opened 2025-03-18
fix billboarding, colormaping by adjusting convert_attribute() (compute branch)

ffreyer:
# Description

Alternative to #4857.

Changes some `convert_attribute()` methods to return tuples which get distributed to different nodes later in the pipeline.

The changes here leave `to_rotation` and `to_colormap` unchanged. That way anyone calling these function will continue to get a valid rotations/colormaps. It also safes a bit of work for meshscatter which is never billboarded.


ffreyer:
closes #4857

ffreyer:
Fixed differently

PR #4882, opened 2025-03-20
Fix Cairo marker with sizes < 1

### LINKED ISSUE #4768: hexbin fails for small cell sizes in CairoMakie
For sufficiently small hexbin sizes, the bin is not plotted and no hexes show up. MWE:

```
using CairoMakie
x=rand(300); y=rand(300)
hexbin(x, y, cellsize=0.01, threshold=0, strokewidth=0.01) # this works
hexbin(x, y, cellsize=0.001, threshold=0, strokewidth=0.01) # here, nothing shows up
```

This example is slightly contrived (nobody needs bins so small), but the bug is appearing in a separate usecase where the minimum bin size is much larger. What can I do to fix this, and what could be going wrong?

The issue does not arise in the GLMakie backend...

Running Makie 0.22.1, the error still arises in a fresh environment. Running on Ubuntu 24.04.

### LINKED ISSUE #4881: spy plot empty for matrices 461 x 461 (or higher) in CairoMakie.jl
- [X] what version of Makie are you running? (`]st -m Makie`)

CairoMakie v0.13.2

- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [X] What platform + GPU are you on?

Mac OS X 13.7.4, Radeon Pro 580 8 GB

Spy is broken for too large matrices:

```julia
julia> using CairoMakie

julia> spy(randn(460,460)) # works

julia> spy(randn(461,461)) # empty
```

ffreyer:
# Description

Attempt to fix #4768 and fix #4881 without bringing back the issue from #4663. 

Reproducer for #4663:
```julia
q = Quaternionf(0.6849482f0, 0.175505f0, 0.15756221f0, 0.6893606f0)
f,a,p = scatter(Point3f(0), rotation=q, markersize=20, marker=Makie.logo())
scatter!(
    1:10, 1:10, rand(10, 10) .* 10,
    rotation=normalize.(rand(Quaternionf, 10 * 10)),
    markersize=20,
)
f
```

This produces `Float32[17.774902 -0.46377563; -9.153168 0.24081421]` as the marker transformation matrix. Viewing this as a basis transformation gives two basis vectors that are very close to linearly dependent:
```julia
julia> M[:, 1] ./ norm(M[:, 1])
2-element Vector{Float32}:
  0.88904786
 -0.4578143

julia> M[:, 2] ./ norm(M[:, 2])
2-element Vector{Float32}:
 -0.88749003
  0.46082675
```

I believe that's the issue for Cairo.

## Type of change

## Checklist


ffreyer:
Subpixel rendering is quite different between Cairo and GL backends but I think that's expected. I would mainly view the new test as an alarm that something changed anyway.

image

The two issues are fixed in this pr and the issue from #4663 hasn't reappeared.

SimonDanisch (approved):

PR #4884, opened 2025-03-21
make `colorrange_scaled` computation safe for inverting / sort-order-breaking scales

### LINKED ISSUE #4883: `colorscale=-` results in a black image
Result of `image(randn(20, 20), colorscale=-)` is just an empty black image:
Image
I would expect the same visual result as `image(-A)`, with the only difference being labels for the colorbar.

- [ ] what version of Makie are you running? (`]st -m Makie`) latest, v0.22.2
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) yes

asinghvi17:
# Description

Fixes #4883

Wraps calls to `apply_scale(scale, colorrange)` in `extrema` so that the computed colorrange is robust to e.g. `colorscale = -`. 

Not sure this needs a reference test given it's a bug on the Makie end of things.

## Type of change

## Checklist


test/issues.jl
```diff
@@ -46,4 +46,12 @@
         @test p.plots[1].plots[1].transformation.model[][3, 4] == 10.0
         @test p.plots[1].plots[2].transformation.model[][3, 4] == 10.0
     end
+
+    @testset "#4883 colorscale that breaks sort" begin
+        data = Float32.(Makie.peaks())
+        f, a, p = image(data; colorscale = -)
```
ffreyer, line 52: Did that error without the Float32?

asinghvi17: It did, because I was comparing the extrema in Float64 to the extrema in Float32.  

```julia-repl
julia> mp = maximum(Makie.peaks())
8.075173545159231

julia> mpf32 = Float32(mp)
8.075173f0

julia> mp == mpf32
false
```

ffreyer: Ah so it's just for the tests below and not because of some internal problem

asinghvi17: Yeah, I guess in this case Julia type promotes both numbers to Float64, so the lossy conversion kicks in.  And I don't want to test approximately here.

SimonDanisch (approved):
Thanks!

PR #4886, opened 2025-03-22
update logo to MakieLogos version

jkrumbiegel:


SimonDanisch:
Could we maybe directly link the logo from MakieLogos? 
Then we just have to update it in one location in the future!

jkrumbiegel:
yeah I guess so, at least the svgs in the readme. The logo.png is used in tests and such

jkrumbiegel:
had already updated the refimages so merging this now, but can change to links later as well

PR #4898, opened 2025-03-30
fix hexbin using logscales

### LINKED ISSUE #4895: Log-Scale Axis within hexbin plot
- [x] what version of Makie are you running? v0.22.2
- [x] can you reproduce the bug with a fresh environment ? Yes
- [x] What platform + GPU are you on? Linux (x86_64-linux-gnu)

Currently, the `log`-scale axis does not work as expected within `hexbin` plots. The `yscale=log10` option will cause the plot to fail with the following error:
```
Display Error: ERROR: DomainError with -1.5204805247485638:
log10 was called with a negative real argument but will only return a complex result if called with a complex argument. Try log10(Complex(x)).
```

For example, the following code with the comment line uncommented will FAIL with the above error:
```julia
using Random; Random.seed!(0)
using CairoMakie
n = 10000
xs = randn(n)
ys = randn(n) .|> exp
hexbin( # Fig. 1
	xs,
	ys;
	# axis = (; yscale=log10), # THIS LINE DOES NOT WORK
)
```

| linear scale |
|:--:|
| Fig. 1: `hexbin` plot with default `identity` scale for y-axis (WITHOUT commented line) |

For those who are eager to use `log10` scale in `hexbin`, here's a workaround that works:
```julia
using Random; Random.seed!(0)
using CairoMakie
n = 10000
xs = randn(n)
ys = randn(n) .|> exp
log_ys = log10.(ys)
yticks = floor(Int, minimum(log_ys)) : ceil(Int, maximum(log_ys))
hexbin( # Fig. 2
	xs,
	log_ys;
	axis = (;
		yticks = (yticks, [rich("10", y|>string|>superscript) for y in yticks]),
	),
)
```
| log10 scale |
|:--:|
| Fig. 2: `hexbin` plot with `log10` scale for y-axis (workaround) |

I hope the non-identity transformation of the axis scale can be fixed in the future, without the need for workarounds.

t-bltg:
# Description

Fixes #4895.

Fix hexbin using logscales (`data_limits` would extend limits, which can be invalid when using logscales).

```julia
using CairoMakie
using Random

begin
  Random.seed!(0)
  xs = randn(10_000)
  ys = randn(10_000) .|> exp
  log_ys = log10.(ys)
  yticks = floor(Int, minimum(log_ys)) : ceil(Int, maximum(log_ys))

  fap = hexbin(
    xs,
    log_ys;
    axis=(;
      yticks=(yticks, [rich("10", y |> string |> superscript) for y ∈ yticks]),
    ),
  )
  display(fap)
  save("hb-workaround.png", fap)

  println("==  "^20)
  fap = hexbin(
    xs,
    ys;
    axis=(; yscale=log10),
  )
  display(fap)
  save("hb-fix.png", fap)
end
```
**workaround**
hb-workaround

**fix**
hb-fix

Note that for the fix, the y limits are **not** extended, as they would produce a `DomainError` on log scale.

Added reference image:
hexbin logscale

Becnhmark test failure seems unrelated.

## Type of change

## Checklist

### DIFF
--- a/src/stats/hexbin.jl
+++ b/src/stats/hexbin.jl
@@ -26,6 +26,10 @@ function spacings_offsets_nbins(bins::Tuple{Int,Int}, cellsize::Nothing, xmi, xm
     return xspacing, yspacing, xmi, ymi, bins...
 end
 
+# hardcoded scale factors
+xfact() = 2
+yfact() = 4 / 3
+
 function spacings_offsets_nbins(bins, cellsize::Real, xmi, xma, ymi, yma)
     return spacings_offsets_nbins(bins, (cellsize, cellsize * 2 / sqrt(3)), xmi, xma, ymi, yma)
 end
@@ -36,8 +40,8 @@ end
 function spacings_offsets_nbins(bins, cellsizes::Tuple{<:Real,<:Real}, xmi, xma, ymi, yma)
     x_diff = xma - xmi
     y_diff = yma - ymi
-    xspacing = cellsizes[1] / 2
-    yspacing = cellsizes[2] * 3 / 4
+    xspacing = cellsizes[1] / xfact()
+    yspacing = cellsizes[2] / yfact()
     (nx, restx), (ny, resty) = fldmod.((x_diff, y_diff), (xspacing, yspacing))
     return xspacing, yspacing, xmi - (restx > 0 ? (xspacing - restx) / 2 : 0),
            ymi - (resty > 0 ? (yspacing - resty) / 2 : 0), Int(nx) + (restx > 0), Int(ny) + (resty > 0)
@@ -111,11 +115,11 @@ function plot!(hb::Hexbin{<:Tuple{<:AbstractVector{<:Point2}}})
         xspacing, yspacing, xoff, yoff, nbinsx, nbinsy =
             spacings_offsets_nbins(bins, cellsize, xmi, xma, ymi, yma)
 
-        ysize = yspacing / 3 * 4
-        ry = ysize / 2
-
-        xsize = xspacing * 2
+        xsize = xspacing * xfact()
         rx = xsize / sqrt3
+
+        ysize = yspacing * yfact()
+        ry = ysize / 2
 
         bin_map = Dict{Tuple{Int,Int}, Float64}()
 
--- a/src/stats/hexbin.jl
+++ b/src/stats/hexbin.jl
@@ -21,55 +21,45 @@ end
 _hexbin_xfact() = 2
 _hexbin_yfact() = 4 / 3
 
-function _spacings_offsets_nbins(bins::Tuple{Int,Int}, cellsize::Nothing, xmi, xma, ymi, yma)
+function _spacings_offsets_nbins(bins::Tuple{Int,Int}, cellsize::Nothing, r::Rect2d)
     any(<(2), bins) && error("Minimum number of bins in one direction is 2, got $bins.")
-    x_diff = xma - xmi
-    y_diff = yma - ymi
-
-    xspacing, yspacing = (x_diff, y_diff) ./ (bins .- 1)
-    return xspacing, yspacing, xmi, ymi, bins...
+    return r.widths ./ (bins .- 1), r.origin, bins
 end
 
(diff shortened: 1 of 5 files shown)


src/stats/hexbin.jl
```diff
@@ -26,6 +26,10 @@ function spacings_offsets_nbins(bins::Tuple{Int,Int}, cellsize::Nothing, xmi, xm
     return xspacing, yspacing, xmi, ymi, bins...
 end
 
+# hardcoded scale factors
+xfact() = 2
+yfact() = 4 / 3
```
asinghvi17, line 31: Can these be `hexbin_xfact` or underscore methods?  People doing `Makie.x<tab>` will likely see this, might get a bit confusing...

src/stats/hexbin.jl
```diff
@@ -50,11 +54,22 @@ function data_limits(hb::Hexbin)
...
-    ms = 2.0 .* fn(hb.plots[1].markersize[])
-    nw = widths(bb) .+ (ms..., 0.0)
-    no = bb.origin .- ((0.5 .* ms)..., 0.0)
+    ms = 2fn(hb.plots[1].markersize[])
+    ox, oy = bb.origin .- ((0.5 .* ms)..., 0.0)
+    wx, wy = bb.widths .+ (ms..., 0.0)
 
-    return Rect3d(no, nw)
+    # do not extend in order to avoid log-scale DomainError
+    tfx, tfy = transform_func(hb)
```
asinghvi17, line 62: This destructuring is not always possible, e.g. for PolarAxis.  Better to have some general multiple-dispatch method to check if this has an error at zero, like `can_handle_negative_domain(transf, dim)` IMO...

This will also error if transform_func returns a single function:

```julia-repl
julia> x, y = identity
ERROR: MethodError: no method matching iterate(::typeof(identity))
The function `iterate` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  iterate(::LibGit2.GitBranchIter)
   @ LibGit2 ~/.julia/juliaup/julia-1.11.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LibGit2/src/reference.jl:343
  iterate(::LibGit2.GitBranchIter, ::Any)
   @ LibGit2 ~/.julia/juliaup/julia-1.11.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LibGit2/src/reference.jl:343
  iterate(::Pipe, ::Int64)
   @ Base stream.jl:1297
  ...

Stacktrace:
 [1] indexed_iterate(I::Function, i::Int64)
   @ Base ./tuple.jl:162
 [2] top-level scope
   @ REPL[9]:1
```

src/stats/hexbin.jl
t-bltg: Thanks, you're right, done in https://github.com/MakieOrg/Makie.jl/pull/4898/commits/6355b3a6d6c97c68f140c95a2b30aea2aa019e9f.

src/stats/hexbin.jl
t-bltg: `can_handle_negative_domain` implemented in https://github.com/MakieOrg/Makie.jl/pull/4898/commits/c3ef7c97261241c20748c300848eaba96af9aeb8.

asinghvi17:
Thanks for implementing those changes!  Otherwise this looks good from my perspective, just a minor comment which could go either way.

src/stats/hexbin.jl
```diff
@@ -18,22 +18,24 @@ Plots a heatmap with hexagonal bins for the observations `xs` and `ys`.
 end
 
 # hardcoded scale factors
-_hexbin_xfact() = 2
-_hexbin_yfact() = 4 / 3
+@inline _hexbin_xsize_fact() = 2
+@inline _hexbin_ysize_fact() = 4 / 3
+@inline _hexbin_xmarker_fact() = sqrt(3)
+@inline _hexbin_ymarker_fact() = 2
```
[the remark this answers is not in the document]
t-bltg: It's easier to develop using functions than `const`s, when using `Revise`, otherwise I don't mind setting these as `const`s.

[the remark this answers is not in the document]
asinghvi17: Makes sense, I don't have a strong feeling one way or another but was just curious.

PR #4911, opened 2025-04-05
Try fixing log-barplot with explicit transformation and boundingbox

ffreyer:
# Description

Alternative to #4849. I'm not confident this is a good/working solution so I'm separating it from the pr.

## Type of change

## Checklist


ffreyer:
In W/GLMakie I can zoom out until Axis ticks break with this (which happens around 2^-1000)
image

ffreyer:
This currently has two problems:
- GLMakie CI is more fussy about the coordinates than I expected. Locally limiting values to a few magnitudes less than `floatmax(Float32)` works fine but in CI it does not. I don't know why that is and it worries me for compatibility
- Limiting relative to `floatmax(Float32)` makes sense for what gets pushed on the GPU but it does not make sense with Float64 values/float32convert. It should be possible to make a barplot from -1e100 to 1e100. Just bumping the limit to `~floatmax(Float64)` will break plots that are not in a Float64 range, and applying float32convert manually doesn't work because you can't skip it later

Moelf:
do you intend to finish this 👀 ?

ffreyer:
redone in #5412

PR #4914, opened 2025-04-10
update voxels if underlying chunk array changes

### LINKED ISSUE #4912: voxels passed as UInt8 chunk do not update using Observables
In the following example, the axis stays empty even though I'd expect it to be updated upon updating `chunk`:

```julia
chunk = Observable(zeros(UInt8, 8,8,8))
f, a, p = voxels(chunk)
chunk[] = rand(0x00:0x01, 8, 8, 8)
f
```

Happens for GLMakie as well as WGLMakie. It works when using `Float32` instead of `UInt8`, but in that case zero values don't make voxels invisible and AFAICT I can't apply custom textures. This is using Makie 0.22.2

simeonschaub:
# Description

Fixes #4912

Don't disconnect the listener if data is passed as raw UInt8 chunk. Works locally for me, but didn't figure out how to write a test for this.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed issue with UInt8 voxel data not updating correctly when Observable input is updated [#4914](https://github.com/MakieOrg/Makie.jl/pull/4914)
+
 ## [0.22.3] - 2025-04-08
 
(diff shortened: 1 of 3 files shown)


ffreyer (approved):

simeonschaub:
Thanks @ffreyer for the test cases and change log, that was quick!

PR #4916, opened 2025-04-11
Release v0.22.4

asinghvi17:
fix the accidental removal of `apply_transform(f, data, space::Symbol)`

update all Makie and downstream versions


src/layouting/transformation.jl
```diff
@@ -351,6 +351,11 @@ end
 # PointTrans{N}(func::F) where {N, F} = PointTrans{N, F}(func)
 Base.broadcastable(x::PointTrans) = (x,)
 
+function apply_transform(f, data, space::Symbol)
+    Base.depwarn("`apply_transform(f, data, space::Symbol)` is deprecated, use `apply_transform(f, data)` instead.", :apply_transform)
+    space == :data ? apply_transform(f, data) : apply_transform(identity, data)
+end
```
ffreyer, line 357: We should probably also bring the Observable{Symbol} version back

asinghvi17: Did that return an observable?

PR #4755, opened 2025-01-25
Fix rare shader compilation error

ffreyer:
# Description

Fixes https://discourse.julialang.org/t/glmakie-loaderror-when-plotting-contour/124424

## Type of change

Delete options that do not apply:

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #4918, opened 2025-04-12
Corrections for non uniformly scaled markers and text

ffreyer:
# Description

~~Probably related:~~ Fixes #4755

When scatter (or text) is used with non-uniform markersize (fontsize), e.g. `Vec2f(50, 10)`, the marker (character) edge is washed out. For example:
```julia
scene = Scene(size = (250, 250))
p = scatter!(scene,
    [-0.5, -0.5, -0.5], [-0.5, 0.5, 0],
    marker = :rect, markersize = [Vec2f(50, 10), Vec2f(10, 50), Vec2f(50)]
)
scatter!(scene, 0, +0.5, markersize = (50, 10))
scatter!(scene, 0, 0.0, markersize = 50)
scatter!(scene, 0, -0.5, markersize = (10, 50))
scatter!(scene, 0.5, 0.5, marker = 'o', markersize = 50)
scatter!(scene, 0.5, 0, marker = 'L', markersize = 50, rotation = Quaternionf(0.3, 0.7, 0.5, 0.2))
screen = display(scene)
```
master

This is caused by the signed distance field not being scaled correctly. The sdf encodes the minimum distance between a point and the edge of a shape. When the space in which the point and shape are in is scaled by a (uniform/scalar) value we can pull that scaling out of the distance calculation. When the scaling is non uniform we can't because it can't be pulled out of `sqrt(scale_x * (x - x0) + scale_y * (y - y0))`. It also affects what the closest point (x0, y0) on the edge of the shape would be. Technically we would need to recalculate the sdf's in the texture atlas for each scaling aspect ratio.

When looking at examples for sdf rendering I fairly often see `fwidth(sdf)` being used as the anti-aliasing scale. It calculates `abs(dFdx(sdf)) + abs(dFdy(sdf))`, where `dFd*(sdf)` calculates the partial derivative of `sdf` with respect to fragments coordinates. I thought this was just an approximation to avoid calculating the correct to-pixel scaling factors, but I decided to give it a try:
fwidth

Turns out that corrects for non-uniform scaling as well. The signed distance field should be scaled so that moving 1px away from the nearest edge increases the sdf value by 1. When the sdf is distorted this is currently true along on axis direction, but not the other. `fwidth` more or less reads what the real change is and weighs the sdf against it, instead of 1. That's how I believe it corrects for distortion.

But this also has artifacts, see the rectangles. Depending on where exactly the rect draws, these little spikes can cover the whole quad. I believe this is because the fragment shader looks at 2x2 blocks of fragments for `fwidth`. We basically have:
```
sdf00  sdf10
sdf01  sdf11
```
from which the partial derivatives of sdf00 to sdf11 are calculated. I.e. sometimes they are biased to the left, sometimes right, up or down. If the left edge considers a block to the right instead of around the edge the results can be significantly different causing artifacts.

I tried pre-computing gradients on the CPU side which didn't go anywhere and eventually had another thought. The uv coordinates with which we sample the sdf texture scale linearly across the quad, meaning that their pixel derivative is constant. Here it shouldn't matter if which block `dFd*` hits. And from that we can calculate a centered gradient of the sdf and do what `fwidth` does without that bias. The result of that is:

abs_grad

and with `length(grad)` instead of the sum of absolutes:

norm_grad

This still seems to have a little bit of these corner artifacts, but much less `fwidth`. I figured maybe this is due to clamping of `f_uv +- df_uv`, but my attempt at correcting for it changed absolutely nothing.

I will need to do some more testing (px_per_unit, tiny markers and fonts, thin symbols) to see if this has any significant drawbacks but I think this could be a general improvement. This probably also needs some fiddling with prefactors to match the old anti-aliasing width. 

TODO:
- [x] check refimg tests for differences, and specifically:
  - [x] check rendering of rotated markers, especially ones that become thin in x or y direction
  - [x] check markers with thin lines
  - [x] check px_per_unit interaction with AA
  - [x] compare AA widths before and after
- [x] check if mipmap and anisotropic filter can be re-enabled
- ~~check if to-pixel scaling of sdf values can be simplified~~ We could probably drop f_viewport_from_u_scale and rely on sdf derivatives for shapes, but that just seems like moving complexity around.
- [x] cleanup unused code, add more comments to explain code
- [x] port to WGLMakie

## Type of change

## Checklist


SimonDanisch (approved):

GLMakie/src/gl_backend.jl
```diff
@@ -43,17 +43,9 @@ function get_texture!(context, atlas::Makie.TextureAtlas)
...
-            # TODO: Consider alternatives to using the builtin anisotropic
-            # samplers for signed distance fields; the anisotropic
-            # filtering should happen *after* the SDF thresholding, but
-            # with the builtin sampler it happens before.
-            anisotropic = 16f0,
-            mipmap = true
-        )
+        # anisotropic filtering sometimes creates artifacts with aspect/distortion
+        # corrected anti-aliasing radius, mipmap seems irrelevant
+        tex = Texture(context, atlas.data, minfilter = :linear, magfilter = :linear)
```
ffreyer, line 48: With anisotropic filtering: (dots between rows of text)
Screenshot 2025-04-17 172348

Without:
Screenshot 2025-04-17 172356

I think what this pr adds may act as anisotropic filtering itself, so it might be redundant now. In GLMakie I also see no difference at all between mipmap on and off. I don't see how mipmaps would help here either.

ffreyer:
@jkrumbiegel Do you have any opinions on the changes to the AA radius/sharpness in this pr? I.e. the differences in refimages or the examples in https://github.com/MakieOrg/Makie.jl/pull/4918#issuecomment-2810016040 ?

jkrumbiegel:
@ffreyer here on github I think 3 looks best to me, I think 4 and 5 look too sharp in an artifact-y way. But it's not super easy to see side-by-side.

PR #4924, opened 2025-04-14
Handle heatmap, image, meshscatter, mesh and surface in Legend

### LINKED ISSUE #1703: `Legend` fails on a 3D plot ?

### LINKED ISSUE #4066: meshscatter does not produce an element in the legend
I noted that `meshscatter` does not produce a symbol in the legend and trying to create a MWE I got the following:

```julia-repl
julia> fig = Figure()

julia> ax = Axis3(fig[1, 1])
Axis3()

julia> meshscatter!(ax, 1, 1, 1, label = "hello")
MeshScatter{Tuple{Vector{Point{3, Float64}}}}

julia> Legend(fig[1, 2], ax)
ERROR: `legendelements` returned an empty list for content element of type MeshScatter{Tuple{Vector{Point{3, Float64}}}}. That could mean that neither this object nor any possible child objects had a method for `legendelements` defined that returned a non-empty result.
Stacktrace:
```

ffreyer:
# Description

More or less inspired by #2348, this adds Legend elements for image, heatmap, meshscatter, mesh and surface.

Refimg/Example:
image

Fixes #1703
Fixes #4066

When this or #2276 is merged, the other will need to add plots to LegendElements.

## Type of change

## Checklist


asinghvi17:
Awesome, thanks for taking that forward!

src/makielayout/blocks/legend.jl
```diff
@@ -450,6 +519,9 @@ _renaming_mapping(::Type{PolyElement}) = Dict(
     :colormap => :polycolormap,
     :colorrange => :polycolorrange,
 )
+_renaming_mapping(::Type{MeshElement}) = Dict()
+_renaming_mapping(::Type{ImageElement}) = Dict()
+_renaming_mapping(::Type{MeshScatterElement}) = Dict()
```
ffreyer, line 524: What's the point of these? They just seem to map the simpler attribute names in a LegendElement to more complicated ones like `color -> polycolor`. But the attributes are already isolated there. Am I missing something?

SimonDanisch (approved):

src/makielayout/blocks/legend.jl
ffreyer: @jkrumbiegel Is it fine to leave these?

Issue #3531, opened 2024-01-05
Provide a method to plot a sample (or calibration) arrow


ryofurue:
### Feature description

In meteorology, oceanography, and climate science (and sometimes in fluid mechanics), it is customary to include a "sample arrow" or "calibration arrow" in arrow plots.  For example, the lengths of the arrows are proportional to the wind speeds they represent. Then, you want to know the correspondence of the arrow lengths on the paper (screen) to the wind speeds.  For this purpose, you plot a "sample vector" which looks like **"← 1 m/s"** which indicates that this particular arrow length corresponds to a wind speed of 1 m/s.   The sample vector is sometimes plotted within the `Axis` if there is sufficient empty space.

Below I cite just one example from a NASA site.

An alternative is to indicate the speed with the colors of the arrows, but this method cannot always be used because you often want to plot black vectors on a color `contourf` map of a different quantity (like sea surface temperature).

### For plot types, please add an image of how it should look like

https://science.nasa.gov/resource/which-way-the-wind-blows/

PR #4925, opened 2025-04-15
Refactor arrows

### LINKED ISSUE #743: vec2quaternion MethodError when plotting arrows:
I'm trying the following:
```julia
using Makie, Zygote
x = y = -2:0.1:2
f(x, y) = 4sin(x)*ℯ^(-(x^2+y^2))
surface(x, y, f.(x, y'))
arrows!(vec(Point3.(x, y', f.(x, y'))), (p->-Vec3(p..., 0.) / 2).(gradient.(f, x, y')); arrowsize = 0.03, linecolor = (:white, 0.6), linewidth = 3)
```

Which plots half the arrow and then fails with:
```
ERROR: MethodError: no method matching vec2quaternion(::Array{Quaternion{Float32},2})
Closest candidates are:
  vec2quaternion(::Vec{4,Float32}) at /home/keno/.julia/packages/GLMakie/Pway1/src/GLVisualize/visualize/particles.jl:58
  vec2quaternion(::StaticArrays.StaticArray{Tuple{4},T,1} where T) at /home/keno/.julia/packages/GLMakie/Pway1/src/GLVisualize/visualize/particles.jl:49
  vec2quaternion(::StaticArrays.StaticArray{Tuple{2},T,1} where T) at /home/keno/.julia/packages/GLMakie/Pway1/src/GLVisualize/visualize/particles.jl:51
```

### LINKED ISSUE #1200: `arrowtail` does nothing?
(version 0.14.2)

I cannot figure how to draw symmetrical arrows (like this: ←→; actually in 3d in my example, although 2d does not work either). [The documentation](http://makie.juliaplots.org/stable/plotting_functions/arrows.html) mentions an `arrowtail` property, but gives nothing about its possible values, and looking at the code (in `src/basic_recipes/arrows.jl`) I see that this property is not ever used (contrary to `arrowhead`). Am I correct in this interpretation?

### LINKED ISSUE #1206: 3d arrow does not point to the right coordinate
scene=Makie.Scene()
    scatter!(scene,rand(10,3)) # make it 3d
    arrows!(scene,[0],[0],[0],[0],[0],[2]) # arrow [0,0,0] to [0,0,2]

This produces an arrow with the *base* of the tip at the [0,0,2] coordinate:
a

This is unexpected; instead, the tip of the arrow should be at the given coordinate.

### LINKED ISSUE #1688: transparent arrows

### LINKED ISSUE #2472: Account for arrow head in arrow length

### LINKED ISSUE #2607: How to make 2D arrowhead points point to endpoint instead of being centered at endpoint?

### LINKED ISSUE #2655: arrows() doesn't have a colorrange attribute that controls linecolor/arrowcolor in CairoMakie

### LINKED ISSUE #3039: align does not work in arrows
```julia
(jl_znYjZm) pkg> st
Status `/tmp/jl_znYjZm/Project.toml`
  [e9467ef8] GLMakie v0.8.6

(jl_znYjZm) pkg> st -m Makie
Status `/tmp/jl_znYjZm/Manifest.toml`
  [ee78f7c6] Makie v0.19.6
```
```julia
julia> versioninfo()
Julia Version 1.9.1
Commit 147bdf428cd (2023-06-07 08:27 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen Threadripper 2950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver1)
  Threads: 1 on 32 virtual cores
```
The `:align` option in `arrows` doesn't seem to work:
```julia
fig = Figure()
ax = Axis(fig[1,1])
for (i, (color, align)) in enumerate(zip((:black, :red, :green, :blue, :cyan, :magenta), (:origin, :head, :lineend, :tailend, :headstart, :center)))
    scatter!(ax, [1], [i]; color = :white, markersize = 25, strokewidth = 1, strokecolor = :gray)
    arrows!(ax, [1], [i], [1], [0]; align, color, label=string(align))
end
axislegend(ax)
```
a

### LINKED ISSUE #3247: `arrows(points, directions; kwargs...)` doesn't work (CairoMakie backend)
- [✅] are you running newest version (version from docs) ?
CairoMakie v0.10.9
- [✅] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [✅] What platform + GPU are you on?
MacOS Ventura 13.4.1, Radeon Pro Vega 20 4 GB / Intel UHD 630 1536 MB

Maybe I am missing something, but a call to the [documented](https://docs.makie.org/stable/reference/plots/arrows/) method `arrows(points, directions)` with `length(points) == length(directions)` and `size(points[1]) == size(directions[1])` gives PlotMethodError: no arrows method for arguments `(::Vector{Vector{Float64}}, ::Vector{Vector{Float64}})`.

```julia
include CairoMakie

my_point = [0.5, 0.5]
my_direction = [1.0, 0.0]

CairoMakie.arrows([my_point], [my_direction])
```
or
```julia
include CairoMakie
include StaticArrays

my_point = SVector{2}(0.5, 0.5)
my_direction = SVector{2}(1.0, 0.0)

CairoMakie.arrows([my_point], [my_direction])
```
give the error. For the second one:
```shell
ERROR: PlotMethodError: no arrows method for arguments (::Vector{SVector{2, Float64}}, ::Vector{SVector{2, Float64}}). To support these arguments, define
  plot!(::Arrows{S} where S<:Tuple{Vector{SVector{2, Float64}}, Vector{SVector{2, Float64}}})
```

### LINKED ISSUE #4054: arrow! and transparency does not work
```julia
julia> fig = Figure()

julia> ax = Axis3(fig[1, 1])
Axis3()

julia> arrows!(ax, [Point3f(0,0,0)], [Vec3f(1, 1, 1)], alpha = 0)
Arrows{Tuple{Vector{Point{3, Float32}}, Vector{Vec{3, Float32}}}}
```

shows the arrow, I have tested this with GLMakie and WGLMakie.

what works is

```julia
julia> arrows!(ax, [Point3f(0,0,0)], [Vec3f(1, 1, 1)], color = (:black, 0.3))
Arrows{Tuple{Vector{Point{3, Float32}}, Vector{Vec{3, Float32}}}}
```

### LINKED ISSUE #4708: ElasticArrays input broken for arrows
Makie v0.21.18

```Julia
using ElasticArrays
arrows(ElasticArray(zeros(Point{3},10)),ElasticArray(zeros(Point{3},10)))
```
```Julia
ERROR: MethodError: Cannot `convert` an object of type 
  ElasticVector{Vec{3, Float32}, Vector{Vec{3, Float32}}} to an object of type 
  Union{Vec3f, Vector{Vec{3, Float32}}, GeometryBasics.Vecf{3}, Vec3{Float32}}
```

ffreyer:
# Description

Arrows has a bunch of issues which I want to fix in this pr. I'm starting this with just some planning so there is more time for people to interject if I'm going down the wrong track. 

First of all, I'm planning to split up arrows into 2D and 3D because they are pretty different in implementation and sometimes in their attributes. E.g. a tipwidth would be a tipradius, quality and lighting attributes are 3D only, etc. I plan to add some glue code to make `arrows` still work and if things end up more similar than I imagine I may merge them again.

## Conversions

Related Issues:
- #743
- #2612
- #3247, #3531 (single point + direction fails)
- #4708

Arrows should rely on `PointBased()` conversions and flatten matrices for convenience. That should make plotting a single arrow using a single position and direction possible (similar to how Scatter accepts a Point).

For #2612 we could interpret the second argument as directions if they are `Vec`s or as endpoints if they are `Point`s but I think that's a bit too specific. It also doesn't work with argument groups, i.e. if you pass (x, y, u, v). I'm currently thinking an attribute `arg_mode = :directions / :endpoints` would be better.

## Attributes

Issues:
- #2655
- #4054 (alpha)

These just need to be connected.

## Marker

Issues:
- #1200 (double headed arrows)
- #1688 (self overdraw of 2d arrows)

Both of these are more complex.

To fix #1688 naively, we would have to calculate the markerspace position where the triangle marker ends and the line should begin. This is complicated with generic markers because they typically are less than `markersize` units large. It will also leave some overlap or gap in GL backends because anti-aliasing happens before lines and scatter are composed. The full solution to this would be to either extend lines/scatter to natively support arrows (similar to `marker = Rect` or `linecap = :round`), or to generate a mesh for each arrow and render them all with poly. Option 1 is more performant, higher quality, more restrictive and requires backend changes that Simon isn't a fan of, so I'll go with Option 2.

To fix #1200 we'll need to consider arrows as three parts - a tail, a shaft and a tip. This is important for the main course:

## Alignment

Issues:
- #1206 (tip not considered in 3D length)
- #2472 (tip not considered in 3D alignment)
- #2607 (tip not considered in 2D length)
- #3039 (in 2D, also too many aliases)

PyPlot also reminded me that with a static pixel size for the arrow/head we can't draw arrows shorter than that size. I.e. at some point the arrow needs to scale as whole, rather than just shortening the shaft/tail.

### 2D

With #1200 in mind, I think our arrow should have 3 parts, each with a length and width:

arrow

From these user attributes we then derive effectively the same set of internal attributes: (all in markerspace)
1. Calculate the target arrow length from `target_length = lengthscale * norm(direction)` (or `lengthscale * norm(endpoint - startpoints)`)
2. If `shaftlength = automatic`, calculate an internal `_shaftlength = max(minshaftlength, target_length - taillength - tiplength)`. Otherwise `_shaftlength = shaftlength`
3. Calculate the total marker length as `marker_length = taillength + _shaftlength + tiplength`
4. Scale all lengths and widths by `target_length / marker_length`
5. Generate the arrow from the scaled lengths and widths

This should allow you to:
- make the arrow scale below some threshold defined by `minshaftlength` if `shaftlength = automatic`
- always scale the arrow if `shaftlength` is a value
- optionally add a `tail`

The tail, shaft and tip should probably be individually settable markers/meshes/point vectors. I will probably repeat the interface from #4879 here, where you either define a shape on a 0..1 x 0..1 range and it gets scaled internally, or you define a callback function that constructs it from the calculated scales.

With respect to `align` I'm planning to deprecate duplicate names and just have :tail, :center and :tip. I'm also planning to allow a float for fractional alignment.

~~I haven't really thought about `normalize` in 2D yet, but maybe it would make sense to have it set `target_length = taillength + (min)shaftlength + tiplength`. That would force all arrows to be the user defined size. Or maybe it should just not be there, and something like `scale = false` should have that effect.~~ Just do the obvious and `normalize(directions)`

### 3D

I think the scaling and alignment handling will generalize to 3D. Though I think it would be good to rename `width` to `radius`. Since 3D doesn't use pixel space we don't need to build a mesh for every arrow and can instead just apply scaling to a persistent mesh through `meshscatter`. In the callback case that will need to be rebuild. 

I'm planning to move the mesh generation bits, i.e. `_mantle` and `_circle` to GeometryBasics with this pr. They don't really belong here, and having GeometryPrimitives like `Circle3D`, `Cone` and `CylinderShell` (Tube? OpenCylinder?) sounds generally useful. 

## tl;dr goals:

- rework arrows to have a tail, shaft and tip to allow double-headed arrows (tail hidden by default)
- reduce align options to `:tail, :center, :tip` or a 0..1 float
- fix align to consider the whole arrow, from tail to tip
- scale the whole arrow if the shaft length drops below a `minshaftlength`, otherwise scale only shaft
- add a way to always scale the whole arrow
- add conversions to allow plotting one arrow from one Point + direction
- add `arg_mode = :directions / :endpoints` to allow defining arrows based on two points
- fix various issues with Attribute passthrough, missing argument conversions
- maybe split arrow recipe into 2D and 3D variant to improve attribute names (and allow 2D arrows in 3D?)

## PR State

- fixes #743
- fixes #3247 (StaticVectors)
- fixes #4708
- fixes #1200
- fixes #1688
- fixes #1206
- fixes #2472
- fixes #2607
- fixes #3039
- fixes #2655
- fixes #4054
- closes #2884

TODO:
- [x] 2D prototype
- [x] 3D prototype
- [x] add "global" scale factor that scales everything
- [x] add `maxshaftlength`
- [x] check normal gen for cones (flat face may not be correct?)
- [x] rework bbox-relative scaling for 3D arrows to only apply with `markerscale = automatic`
- [x] clean up data_limits and boundingbox (consider align)
- [x] clean up attribute passthrough
- [x] handle deprecations (forward arrows/arrows! to arrows2d!/arrows3d! with depwarns)
- [x] fix colormapping in 2D (needs to be done early to work with vertex colors)
- [x] add `(from, to)` arg mode
- [x] update docstrings
- [x] update main docs
- [x] update tests
- [x] Check/fix #2655, #4054
- [x] check docs, tests

## Type of change

## Checklist


ffreyer:
3D arrows, requires https://github.com/JuliaGeometry/GeometryBasics.jl/pull/257
```julia
r = Rect3f(-1,-1,-1,2,2,2)
ps = Makie.coordinates(r)
zdirs = [Vec3f(0,0, z) for (x,y,z) in ps]
f = Figure()
a = LScene(f[1,1])
p = Makie.arrows3d!(
    ps, 0.5 .* ps, color = :blue, align = :tip,
    tail = Makie.Cone(Point3f(0,0,1), Point3f(0,0,0), 0.5f0), taillength = 0.06
)
Makie.arrows3d!(
    ps, -zdirs, color = :red, align = :tail
)
linesegments!(r)
linesegments!(Rect3f(Point3f(-0.5), Point3f(1)))
f
```
image

ffreyer:
`argmode = :endpoint` & colormapping:

```julia
f = Figure(size = (1000, 500))

# 2D
ps = [Point2f(cos(a), sin(a)) for a in range(0, 2pi, length=21)[1:end-1]]
a = Axis(f[1,1], aspect = DataAspect())
p = Makie.arrows2d!(
    ps, [ps[2:end]..., ps[1]], color = (:blue, 0.5),
    align = :center, lengthscale = 0.5, argmode = :endpoint,
    tail = Point2f[(0, 0.5), (1, 0), (1, 1)], taillength = 8
)
Makie.arrows2d!(
    ps, ps, color = eachindex(ps), align = :tail,
    colormap = :rainbow, lengthscale = 0.5
)
scatter!(ps, marker = Circle, color = :transparent, strokewidth = 1)

# 3D
ps = [Point3f(cos(a), sin(a), 0) for a in range(0, 2pi, length=21)[1:end-1]]
a = Axis3(f[1,2], aspect = :data)
p = Makie.arrows3d!(
    ps, [ps[2:end]..., ps[1]], color = (:blue, 0.5),
    align = :center, lengthscale = 0.5, argmode = :endpoint,
    tail = Makie.Cone(Point3f(0,0,1), Point3f(0,0,0), 0.5f0), taillength = 0.4
)
Makie.arrows3d!(
    ps, [Vec3f(0,0,1) for _ in ps], color = eachindex(ps), align = :tip,
    colormap = :rainbow, lengthscale = 0.5
)
scatter!(ps, marker = Circle, color = :transparent, strokewidth = 1)

f
```

image

ffreyer:
I'm wondering if I should officially deprecate `arrows` for `arrows2d/3d` or if we want to keep it as a shortcut method, i.e. have `arrows` call `arrows2d/3d` based on the passed arguments. I'm leaning towards deprecating it. One reason is that the new methods differ in whether they create 2D or 3D markers, not in whether they can plot in 2D or 3D space. You can use arrows2d in 3D space and vice verse and maybe you want to sometimes.

```julia
fig = Figure(size = (800, 800))
ps2 = [Point2f(x, y) for x in 1:10, y in 1:10]
vs2 = [0.3 * Vec2f(cos(x)+sin(y), sin(x)+cos(y)) for x in 1:10, y in 1:10]
arrows2d(fig[1, 1], ps2, vs2)
arrows3d(fig[1, 2], ps2, vs2)
ps3 = [Point3f(x, y, z) for x in 1:5, y in 1:5, z in 1:5]
vs3 = [0.3 * Vec3f(cos(x)+sin(y), sin(x)+cos(y), sin(z)) for x in 1:5, y in 1:5, z in 1:5]
arrows2d(fig[2, 1], ps3, vs3)
arrows3d(fig[2, 2], ps3, vs3)
fig
```

image

Another reason is that I'm currently trying to reproduce the visuals of the old arrows in `arrows` which I would like to get rid off at some point. For example the "shrink arrow when the shaft goes below a threshold" feature is turned off there, as of https://github.com/MakieOrg/Makie.jl/pull/4925/commits/5985d2735191bab3d758605a0938bf73f5b6e49d.

asinghvi17:
Couldn't `arrows` be a convert-args-only recipe that returns a plotspec?  I feel like that would be less breaking than just removing arrows in general.

ffreyer:
> Couldn't `arrows` be a convert-args-only recipe that returns a plotspec? I feel like that would be less breaking than just removing arrows in general.

I'm not sure what you mean by convert args only recipe. There is a good number of attribute defaults and names that changed. I'm also not planning on removing arrows immediately. I have `arrows/arrows!` implemented as functions atm, which try to reproduce old arrows to a degree. I guess that could be a PlotSpec as well but I'm not sure if/how that would be better. 

I'm mostly just wondering if we want to have something like `arrows(...) = is3D(args) ? arrows3d() : arrows2d()` in the future, or if we want to push for users to pick themselves.

jkrumbiegel:
I personally always found the places where Makie recipes are switching just once based on input arguments sort of confusing. I like the separate-recipe route much better.

ffreyer:
Added two tests (for now):

| Test for how arrows2d interact with transformations (neutral, log10 yscale, translation, scale+rotation) | Test for scaling behavior with `minshaftlength` and `maxshaftlength` (both arrow types) |
| --- | --- |
| Screenshot 2025-05-12 223513 | Screenshot 2025-05-12 230246 |

I added `markerspace` for arrows2d (though I'm not sure how useful that is) and dropped `transform_marker` from arrows3d. I added that at some point to improve Axis3 arrows but it basically doesn't work anymore. Instead I fixed the behavior more explicitly - now arrows only apply transformation to their start and endpoints and never to their markers. With that:

image

Tests to add:
- arrows with tails (double headed)
- endpoint mode
- colors per arrow component & per arrow, colormapping
- changes in all arrow metrics and shapes
- updates

Also thinking about either removing or squashing the old tests since most of them just test default arrow plots rather than more complex situations. They should also all fail due to to the length changes (i.e. that tail + shaft + tip matches the user given length, rather than just the shaft)

ffreyer:
Not sure if we can consider this a (big) bug fix. It's certainly visually breaking.

Potentially breaking changes for `arrows()/arrows!()`:
- directions now set the size of the full arrow, not just the shaft (bug fix?)
- Arrows as a recipe is gone 
- arrowhead, arrowtail, arrowcolor, linecolor, linewidth, arrowsize are deprecated with forwarding (still work more or less, so not breaking?)
- linestyle, transform_marker are completely gone
- align no longer accept :lineend, :tailend, :headstart and :origin. Now only :head, :center, :tail or a Float (bug fix/cleanup?)

SimonDanisch (approved):
I think visually breaking with good warnings for how to fix things is good enough to release this.
We could also still make a breaking release, to be on the safe side.

PR #4932, opened 2025-04-19
Respect panbutton selection

alhirzel:
# Description

I was customizing a figure window and having trouble understanding the behavior of the [`panbutton` option for an Axis](https://docs.makie.org/dev/reference/blocks/axis#panbutton). As it turns out, I came to believe the option is being ignored (at least for windows created via `plot()` and `image()`). This PR changes the behavior of the following to what I expect:

```julia
plot([1, 2, 3, 4, 5], axis=(panbutton=Makie.Mouse.middle, ))
```

Now, this works as expected and allows me to use the middle mouse button to pan around.

I did not know how to add a test for this, but happy to.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fixed issue with UInt8 voxel data not updating correctly when Observable input is updated [#4914](https://github.com/MakieOrg/Makie.jl/pull/4914)
+- Fixed `Axis.panbutton` not working [#4932](https://github.com/MakieOrg/Makie.jl/pull/4932)
 
 ## [0.22.4] - 2025-04-11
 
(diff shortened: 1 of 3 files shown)


ffreyer:
This seems fine to me. Axis3 has its own version of this function and PolarAxis and LScene don't rely on interactions at all.

It would be good to add a short test to check this works. Maybe just extend
https://github.com/MakieOrg/Makie.jl/blob/8334603f36dc06fdf3d451be906ac7a4bb77f7be/test/interactivity/Axis.jl#L187-L202
to change the button, check that `MouseButtonEvent` with the old panbutton doesn't affect limits, but the new does?

Also a note that this has been fixed in https://github.com/MakieOrg/Makie.jl/blob/master/CHANGELOG.md would be good

alhirzel:
Thank you for finishing this PR!

Issue #4829, opened 2025-02-27
Better support for open scatter markers


jkrumbiegel:
Open markers are often recommended because they are more robust against overplotting than filled markers. Currently, all our default markers such as `:circle`, `:rect`, etc. are filled. Of course we can make any marker open by only stroking it and not filling it, via `strokecolor = :black, strokewidth = something_nonzero, color = :transparent`.

But this is often not very convenient when you want to expose a plotting function to a user and they are supposed to be able to switch between open and closed markers. For example, in AlgebraOfGraphics, if you set up a `mapping(color = ...)` then just to get open markers, you would have to switch that to `mapping(strokecolor = ...)`. This is possible in recent AoG versions, but it would again require quite a bit of internal bookkeeping in order to facilitate easy open/closed switching for a user.

I was just in the process of implementing a solution for this by adding a transformation function for `BezierPath` markers, in order to add a reversed downscaled version which effectively cuts a hole. I thought this might be a convenient way to turn closed into open markers. However, while this works for circles, rects and triangles, other shapes look pretty bad with this approach:

<img width="591" alt="Image" src="https://github.com/user-attachments/assets/f502a023-93fe-46a8-a33a-8806f79a13c8" />

Because we do in effect want a stroke, and cutting a downsized hole is not equivalent to that.
The other drawback of this technique is that, again because we're not using a stroke, the thickness of an open marker depends on the markersize.

So I've finally reached the conclusion that maybe there should be a simple option to treat any scatter plot as "stroke only" which would use `color` in place of `strokecolor` and don't fill in markers. We would still use `strokewidth` to determine stroke size, which would have a nice and even behavior when scaling markers up and down.

I see two options, either make this dispatch-dependent by adding let's say an `Open` type, so `Open(:circle)` would have this effect. Or maybe a keyword option like `scatter(...; open = true)`. The drawback of the latter is obviously that all forwarding recipes need to add that option for forwarding if they validate their keywords.

cc'ing people that were interested in open markers: @devmotion @pkofod

PR #4935, opened 2025-04-22
Open circle and square markers

jkrumbiegel:
Open markers are often recommended because they allow discriminating overlapping data points better than closed markers. Here I'm adding convenience functions that match `:circle` and `:rect` markers with adjustable inner radius. I've not really seen many other open markers in the wild but more could be added later. One can also use stroked markers for the same effect, but for example in AoG, one can then not use the default `color` mapping which is annoying when switching between markers.

```julia
data = RNG.randn(200, 2)
f = Figure()
scatter(f[1, 1], data, marker = open_circle())
scatter(f[1, 2], data, marker = open_circle(0.5))
scatter(f[2, 1], data, marker = open_square())
scatter(f[2, 2], data, marker = open_square(0.5))
f
```

image

### DIFF
--- a/ReferenceTests/src/tests/examples2d.jl
+++ b/ReferenceTests/src/tests/examples2d.jl
@@ -582,6 +582,16 @@ end
     current_figure()
 end
 
+@reference_test "Open scatter markers" begin
+    data = RNG.randn(200, 2)
+    f = Figure()
+    scatter(f[1, 1], data, marker = open_circle())
+    scatter(f[1, 2], data, marker = open_circle(0.5))
+    scatter(f[2, 1], data, marker = open_square())
+    scatter(f[2, 2], data, marker = open_square(0.5))
+    f
+end
+
 @reference_test "2D surface with explicit color" begin
     surface(1:10, 1:10, ones(10, 10); color = [RGBf(x*y/100, 0, 0) for x in 1:10, y in 1:10], shading = NoShading)
 end
--- a/src/Makie.jl
+++ b/src/Makie.jl
@@ -241,6 +241,7 @@ include("event-recorder.jl")
 
 # bezier paths
 export BezierPath, MoveTo, LineTo, CurveTo, EllipticalArc, ClosePath
+export open_circle, open_square
 
 # help functions and supporting functions
 export help, help_attributes, help_arguments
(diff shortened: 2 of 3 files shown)


lazarusA:
could we also do ?

```
scatter(f[1, 1], data, marker = open_circle(), color=data, colormap=:plasma)
```

lazarusA:
oh, nice! if it works we don't need a test 😄 !

jkrumbiegel:
Ok I've added a bunch of other markers, all implemented in the scaled cutout way currently. This is not perfect because the "stroke width" differs a bit depending on the angle of the lines, but I'm not motivated right now to do a better one. We can still do that later and I would consider the slight changes of widths suitable for a patch release then.

jkrumbiegel:
any idea why WGLMakie would fail here @SimonDanisch or @ffreyer?

SimonDanisch:
Seems like a similar problem as in: https://github.com/MakieOrg/Makie.jl/pull/4879

Testing here:https://github.com/MakieOrg/Makie.jl/pull/4938
Of course it doesn't reproduce locally -.-

ffreyer:
Isn't this what you mentioned in #4829 as not working well for general shapes?

Doesn't this approach mean that the strokewidth scales with markersize? I.e. if you wanted 2px strokewidth with N markersizes you'd need to produce N different markers? That sounds problematic with the finite amount of space in the texture atlas. Especially when strokewidth already exists. Maybe it would be better to do something like
```julia
if user_strokecolor === automatic
    backend_strokecolor = is_open_marker(user_marker) ? user_color : :black
else
    backend_strokecolor = user_strokecolor
end
```
after #4630?
If the roundness of stroke is a problem I'd imagine it could be improved by changing the sdf generation. Probably only to a point due to the limited size of the texture element though.

Another point - since all the other Bezier markers have a Symbol name associated with them, shouldn't these do too? I.e. `:open_circle` for the default, instead of, or rather in addition to a function call?

jkrumbiegel:
Ah yeah nobody really reacted to that earlier issue with implementation ideas. But if you think a stroke version would be feasible, that might be an option, too. I mean both variants have their benefits, but a stroke is just easier to reason about as it's independent of markersize. On the other hand, the markers become slightly larger due to the stroke, in the current implementation the open markers are exactly as large as their closed counterparts.

About the symbols, yeah I thought about that, too. I didn't do it because of the parameter, but in an earlier draft I had a way to specify the fraction via the name. Like `:open_circle_50` for 0.5

PR #4938, opened 2025-04-23
Test WGLMakie CI

SimonDanisch:


SimonDanisch:
image

Welp.... Seems like https://github.com/MakieOrg/Makie.jl/pull/4935 and https://github.com/MakieOrg/Makie.jl/pull/4879 fail independently ... Or its one more reference test that pushes it over some limit? The memory leak tests seem fine though, and the logs also look good, besides a few warnings that have been there before.

SimonDanisch:
Since both PRs add new shapes to the texture atlas, and at some point https://github.com/MakieOrg/Makie.jl/pull/4879 failed with blank for anything texture atlas related.

ffreyer:
My change reduces padding from 12px to 6px for the WGLMakie atlas, which should stop it from filling up faster than GLMakie's atlas. I uploaded a v8 atlas with the adjusted padding. If texture atlas overflow is the problem with tests they should pass now. But the issue is bound to return until we allow it to grow dynamically and/or split it by fonts and/or purpose (text vs scatter)

PR #4940, opened 2025-04-25
Fix automatic unit decision for Unitful axes

jusack:
# Description

Fixes the automatically determined units for Unitful axes that are symmetric.

Currently if an axis has the same absolute value for its positive and negative extrema, the determination of the best unit in `Makie.best_unit` gives the smallest possible prefix. E.g. `Makie.best_unit(-1.0u"mm", 1.0u"mm")` gives ym=yoctometer=10^-24 m because it only uses the value in the middle between min and max which is zero in this case.
This results in useless units for these types of axes:
```julia
lines(range(start=-1.0u"mm", stop=1.0u"mm", length=50), randn(50))
```
grafik
For a non-symmetric axis everything is fine
```julia
lines(range(start=0.0u"mm", stop=1.0u"mm", length=50), randn(50))
```
grafik

## Proposed fix
In this PR I changed the decision of the unit to not use the middle value but instead the total length of the axis, which should be a characteristic value based on which the correct unit can be determined:
grafik
grafik

There are cases where the automatically decided axis unit is different than before, but I do not think this would count as a breaking change.
Before:
```julia
Makie.best_unit(0u"mm",10u"mm") # mm
```
This PR:
```julia
Makie.best_unit(0u"mm",10u"mm") # cm
```

## Type of change

### DIFF
--- a/src/dim-converts/unitful-integration.jl
+++ b/src/dim-converts/unitful-integration.jl
@@ -91,10 +91,10 @@ function get_all_base10_units(x::Unitful.Unit{Sym, Unitful.𝐓}) where {Sym}
 end
 
 function best_unit(min, max)
-    middle = (min + max) / 2.0
-    all_units = get_all_base10_units(middle)
+    axis_length = max - min
+    all_units = get_all_base10_units(axis_length)
     _, index = findmin(all_units) do unit
-        raw_value = abs(unit_convert(unit, middle))
+        raw_value = abs(unit_convert(unit, axis_length))
         # We want the unit that displays the value with the smallest number possible, but not something like 1.0e-19
         # So, for fractions between 0..1, we use inv to penalize really small fractions
         positive = raw_value < 1.0 ? (inv(raw_value) + 100) : raw_value


ffreyer:
I'm not sure if the length would be a good metric either. Something like 1km .. 1km + 10cm would show up as 1000 00cm .. 1000 10cm then, right? Maybe `0.5 * (abs(low) + abs(high))` would be better?

jusack:
Hm, that´s a good point I have not considered yet. Axes where the length greatly differs from the order of magnitude of the values are in general difficult to plot. One could argue for both the larger or the smaller unit, depending on which property is more important in that axis, the difference between two samples, or the overall size.

Using your example these are basically the possible options:

```julia
x = range(1.0u"km", 1.0u"km"+10u"cm", length=50); y=randn(50);
fig = Figure()
ax1 = Axis(fig[1,1], dim1_conversion=Makie.UnitfulConversion(u"km"))
ax2 = Axis(fig[1,2], dim1_conversion=Makie.UnitfulConversion(u"m"))
ax3 = Axis(fig[2,1], dim1_conversion=Makie.UnitfulConversion(u"cm"))
ax4 = Axis(fig[2,2], dim1_conversion=Makie.UnitfulConversion(u"mm"))
for ax in [ax1,ax2,ax3,ax4]
    lines!(ax,x,y)
end
```
grafik

In this case I would manually choose the m scale as its a tradeoff between both, however to automate that we would have to further extend the logic from using a single value.

For axes with even more orders of magnitude between offset and length it gets more difficult:
```julia
x = range(1.0u"km", 1.0u"km"+10u"µm", length=50); y=randn(50);
fig = Figure()
ax1 = Axis(fig[1,1], dim1_conversion=Makie.UnitfulConversion(u"km"))
ax2 = Axis(fig[1,2], dim1_conversion=Makie.UnitfulConversion(u"m"))
ax3 = Axis(fig[2,1], dim1_conversion=Makie.UnitfulConversion(u"mm"))
ax4 = Axis(fig[2,2], dim1_conversion=Makie.UnitfulConversion(u"µm"))
for ax in [ax1,ax2,ax3,ax4]
    lines!(ax,x,y)
end
```
grafik

I would say its good enough to have a default thats not completely off and for more complex axes the user has to manually change the units, depending on the use case.

My preference for a default would be a unit based on the length of the axis, as then you can see the scale of the steps at the first glance.

icweaver:
Yea, I've also had mixed feelings about `best_unit` and friends (https://github.com/SymbolicML/DynamicQuantities.jl/pull/165#issuecomment-2764825640) lately

I was wondering how much this buys us over just keeping things explicit. Here's how that might look with DynamicQuantities over in that PR:

```julia
import DynamicQuantities as DQ

lines(range(start=-1.0*DQ.us"mm", stop=1.0*DQ.us"mm", length=50), randn(50))
```

display

icweaver:
Howdy, the PR experimenting with this approach with DynamicQuantities lives here now: #5280

icweaver:
Update: The DQ approach is in Makie v0.24.7 now, and looks like the unit handling with Unitful will be simplified too in #5323 along with a whole bunch of other really nice improvements =D

image

PR #4942, opened 2025-04-29
[WIP] move most of the texture atlas logic to JS, to be able to use bigger atlas

SimonDanisch:
Since we've been running into texture atlas size problems on CI and for users, and for https://github.com/MakieOrg/Makie.jl/pull/4630, it would be great to use the same texture atlas between GL and WGLMakie. I decided to make this refactor in a separate pr against master, to help with the next release and the CI hangs.

Another motivation is, that I suspect the former way of updating the texture atlas to be a bit racy, which may explain the CI time outs when adding new tests, which mutate the texture atlas.
Hopefully fixes: https://github.com/MakieOrg/Makie.jl/pull/4938
Status:
Most of the logic is in place, but there are still small problems in the size calculations on the JS side:
image
I also need to update the logic for scatter plots using the texture atlas.

This should send _much_ less data to JS when updating text, which should make zooming etc smoother for WGLMakie.

### DIFF
--- a/src/utilities/texture_atlas.jl
+++ b/src/utilities/texture_atlas.jl
@@ -309,6 +309,34 @@ function insert_glyph!(atlas::TextureAtlas, hash::UInt32, path_or_glyp::Union{Be
     end
 end
 
+
+function sdf_uv_to_pixel(atlas::TextureAtlas, uv_width::Vec4f)
+    tex_size = Vec2f(size(atlas.data)) # (width, height)
+    # uv: (left, bottom, right, top)
+    uv_left_bottom = Vec2f(uv_width[1], uv_width[2])
+    uv_right_top = Vec2f(uv_width[3], uv_width[4])
+
+    # reverse the normalization to get pixel coordinates
+    px_left_bottom = floor.(Int, uv_left_bottom .* tex_size .- 0.5)
+    px_right_top = ceil.(Int, uv_right_top .* tex_size .+ 0.5)
+
+    # create pixel ranges
+    x_range = (px_left_bottom[1] + 1):(px_right_top[1])
+    y_range = (px_left_bottom[2] + 1):(px_right_top[2])
+    return x_range, y_range
+end
+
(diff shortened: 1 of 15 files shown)


src/utilities/texture_atlas.jl
```diff
@@ -309,6 +309,33 @@ function insert_glyph!(atlas::TextureAtlas, hash::UInt32, path_or_glyp::Union{Be
...
+
+function sdf_uv_to_pixel(atlas::TextureAtlas, uv_width::Vec4f)
+    tex_size = Vec2f(size(atlas.data)) # (width, height)
+    # uv: (left, bottom, right, top)
+    uv_left_bottom = Vec2f(uv_width[1], uv_width[2])
+    uv_right_top = Vec2f(uv_width[3], uv_width[4])
+
+    # reverse the normalization to get pixel coordinates
+    px_left_bottom = floor.(Int, uv_left_bottom .* tex_size .- 0.5)
+    px_right_top = ceil.(Int, uv_right_top .* tex_size .+ 0.5)
```
ffreyer, line 321: We should probably avoid the +- 0.5 to avoid float precision roll-overs in floor and ceil. I.e. something like this:
```julia
julia> uvs = [(i + 0.5) / 1000 for i in 499:501]
3-element Vector{Float64}:
 0.4995
 0.5005
 0.5015

julia> idxs = floor.(Int, uvs .* 1000 .+ 0.5)
3-element Vector{Int64}:
 500
 500
 501
```
All the uvs are set to be pixel centered, i.e. uv * size is an integer + 0.5. Just dropping the -+ 0.5 should give the same results here without the risk for float precision issues. Same for the javascript version of this.

Also it might be nicer to do `ceil` for both here and drop the +1 below. ceil and floor rather naturally get your the 0-based and 1-based ranges here, imo.
```
0           1  uv
|__| ... |__| pixels
0  1    N-1 N  uv * size
```

SimonDanisch:
Down to one test failure:
image

The other failures are because fonts are rendered more accurately now!

PR #4945, opened 2025-05-02
compute text + name cleanup

ffreyer:
#4930 with name cleanup for backend attributes and refactor of layout_text/glyph collections for normal strings. Mainly opening to check if precompilation works and benchmarks run


WGLMakie/assets/sprites.vert
```diff
@@ -78,17 +78,17 @@ void main(){
...
     vec3 marker_offset = f32c_scale * tovec3(get_marker_offset());
 
-    vec2 bbox_radius = 0.5 * markersize;
+    vec2 bbox_radius = 0.5 * ms;
     vec2 sprite_bbox_centre = quad_offset + bbox_radius;
-    f_sprite_scale = markersize;
+    f_sprite_scale = ms;
     mat4 pview = projection * view;
-    mat4 trans = get_transform_marker() ? model : mat4(1.0);
+    mat4 trans = get_transform_marker() ? model_f32c : mat4(1.0);
```
ffreyer, line 89: I think this is wrong because trans applies to marker_offset which is independent of f32convert

PR #4960, opened 2025-05-08
Make axis cycling contingent on the PlotFunc only

### LINKED ISSUE #4957: Incorrect color cycling for mixed Float32/Float64 vectors
- [x] what version of Makie are you running? (`]st -m Makie`): `[ee78f7c6] Makie v0.22.4`
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`): 
- [x] What platform + GPU are you on?: Apple M4 Pro
- [x] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie): `GLMakie` and `CairoMakie`
- [x] Please provide a minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

```Julia
using CairoMakie # or GLMakie
fig = Figure()
ax = Axis(fig[1,1])
xs = Float32.(10:100)  # if Float64 here, everything is normal
# if only 2/3 y-vectors are same Float type, color cycler is wrong
lines!(xs, cumsum(randn(Float32, length(xs))); label = "1")
lines!(xs, cumsum(randn(Float32, length(xs))); label = "2")
lines!(xs, cumsum(randn(Float64, length(xs))); label = "3")  # if Float32 here, everything is normal
axislegend()
fig
```

Image

asinghvi17:
# Description

Previously it was specific to the concrete type of the plot, which we don't want, since it causes each different set of (post-conversion) args to be cycled individually.  This fixes that by lifting the plot function out of the `Plot` type and reconstructing `Plot{Func}`.

Fixes #4957

This causes three refimage tests to fail, but those failures seem expected (and kind of harmless?).  We can fix either by updating the refimages, or fixing the color in the refimages to be consistent (maybe Cycled(1) on all of them).

Screenshot 2025-05-08 at 7 44 59 PM
Screenshot 2025-05-08 at 7 45 31 PM
Screenshot 2025-05-08 at 7 45 57 PM

## Type of change

Potentially visually breaking?  But I think we can all agree this is correct behaviour.  I was personally working around it by manually assigning Cycled, I'm sure other people were too.

## Checklist

### DIFF
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -675,7 +675,12 @@ validate_limits_for_scale(lims, scale) = all(x -> x in defined_interval(scale), 
 palettesyms(cycle::Cycle) = [c[2] for c in cycle.cycle]
 attrsyms(cycle::Cycle) = [c[1] for c in cycle.cycle]
 
-function get_cycler_index!(c::Cycler, P::Type)
+function get_cycler_index!(c::Cycler, P::Type{Plot{Func, Args}}) where {Func, Args}
+    # We want to avoid indexing into the cycler by the arguments of the plot, 
+    # since those can change on input type (float32 vs float64, polygon vs multipolygon, etc)
+    # so instead, we re-create the plot type with only the function, not the arguments!
+    P = Plot{Func}
+    # Now we can index into the cycler using this abstract type.
     if !haskey(c.counters, P)
         return c.counters[P] = 1
     else
--- a/src/makielayout/blocks/axis.jl
+++ b/src/makielayout/blocks/axis.jl
@@ -675,11 +675,11 @@ validate_limits_for_scale(lims, scale) = all(x -> x in defined_interval(scale), 
 palettesyms(cycle::Cycle) = [c[2] for c in cycle.cycle]
 attrsyms(cycle::Cycle) = [c[1] for c in cycle.cycle]
 
-function get_cycler_index!(c::Cycler, P::Type{Plot{Func, Args}}) where {Func, Args}
+function get_cycler_index!(c::Cycler, P::Type{Plot{func, Args}}) where {func, Args}
     # We want to avoid indexing into the cycler by the arguments of the plot, 
     # since those can change on input type (float32 vs float64, polygon vs multipolygon, etc)
     # so instead, we re-create the plot type with only the function, not the arguments!
-    P = Plot{Func}
+    P = typeof(func) # note: func is the actual function as a type parameter of the plot.
     # Now we can index into the cycler using this abstract type.
     if !haskey(c.counters, P)
         return c.counters[P] = 1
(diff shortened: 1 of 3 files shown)


test/issues.jl
```diff
@@ -54,4 +54,38 @@
...
+            p3 = lines!(a, rand(Float64, 10))
+
+            palette_colors = a.scene.theme.palette.color[]
+
+            @test p1.color[] == palette_colors[1]
+            @test p2.color[] == palette_colors[2]
+            @test p3.color[] == palette_colors[3]
+
+            p4 = lines!(a, rand(Float32, 10))
+            @test p4.color[] = palette_colors[4]
```
asinghvi17, line 71:
```suggestion
            @test p4.color[] == palette_colors[4]
```

src/makielayout/blocks/axis.jl
```diff
@@ -675,7 +675,12 @@ validate_limits_for_scale(lims, scale) = all(x -> x in defined_interval(scale),
...
 attrsyms(cycle::Cycle) = [c[1] for c in cycle.cycle]
 
-function get_cycler_index!(c::Cycler, P::Type)
+function get_cycler_index!(c::Cycler, P::Type{Plot{Func, Args}}) where {Func, Args}
+    # We want to avoid indexing into the cycler by the arguments of the plot, 
+    # since those can change on input type (float32 vs float64, polygon vs multipolygon, etc)
+    # so instead, we re-create the plot type with only the function, not the arguments!
+    P = Plot{Func}
+    # Now we can index into the cycler using this abstract type.
     if !haskey(c.counters, P)
```
SimonDanisch, line 684: Should we just use Func for the dict?

asinghvi17: We could use `typeof(Func)` - this was just the simplest solution.

src/makielayout/blocks/axis.jl
```diff
@@ -675,7 +675,12 @@ validate_limits_for_scale(lims, scale) = all(x -> x in defined_interval(scale),
 palettesyms(cycle::Cycle) = [c[2] for c in cycle.cycle]
 attrsyms(cycle::Cycle) = [c[1] for c in cycle.cycle]
 
-function get_cycler_index!(c::Cycler, P::Type)
+function get_cycler_index!(c::Cycler, P::Type{Plot{Func, Args}}) where {Func, Args}
+    # We want to avoid indexing into the cycler by the arguments of the plot, 
+    # since those can change on input type (float32 vs float64, polygon vs multipolygon, etc)
+    # so instead, we re-create the plot type with only the function, not the arguments!
+    P = Plot{Func}
```
asinghvi17, line 682:
```suggestion
function get_cycler_index!(c::Cycler, P::Type{Plot{func, Args}}) where {func, Args}
    # We want to avoid indexing into the cycler by the arguments of the plot, 
    # since those can change on input type (float32 vs float64, polygon vs multipolygon, etc)
    # so instead, we re-create the plot type with only the function, not the arguments!
    P = typeof(func) # note: func is the actual function as a type parameter of the plot.
```

src/makielayout/blocks/axis.jl
asinghvi17: Done

PR #4961, opened 2025-05-09
improve wireframe poly of meshes significantly

SimonDanisch:
```julia
 poly(Sphere(Point3f(0), 1), color=:white, strokewidth=1, strokecolor=:black, shading=true)
```
image

### DIFF
--- a/MakieCore/src/basic_plots.jl
+++ b/MakieCore/src/basic_plots.jl
@@ -711,7 +711,10 @@ Plots polygons, which are defined by
     """
     Depth shift of stroke plot. This is useful to avoid z-fighting between the stroke and the fill.
     """
-    stroke_depth_shift = -1.0f-5
+    stroke_depth_shift = -1.0f-3
+    mesh_fxaa = automatic
+    stroke_fxaa = false
+    stroke_transparency = true
     mixin_generic_plot_attributes()...
     mixin_colormap_attributes()...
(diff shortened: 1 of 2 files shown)


SimonDanisch:
For full effect, I needed a much larger depth_shift.
@ffreyer, I guess the correct value is a bit scene dependent?

jkrumbiegel:
One problem might be if you have folds or creases in the mesh, then depth shift might pull the lines from the back through the front. That's why I still think mesh + wireframes should be a separate shader where the distance from the face edge determines whether you draw stroke color or face color

SimonDanisch:
>That's why I still think mesh + wireframes should be a separate shader 

I think so too, but its just so much work :D

SimonDanisch:
Well, I guess the reference images show we cant just up the depth_shift that much ^^

ffreyer:
> For full effect, I needed a much larger depth_shift. @ffreyer, I guess the correct value is a bit scene dependent?

Depth shift applies in clip space so no matter what your camera does, you're always acting on a -1..1 range (or was it 0..1, I always forget)

> One problem might be if you have folds or creases in the mesh, then depth shift might pull the lines from the back through the front.

Yea pretty much. You shift the lines forward so if your mesh is thinner in clip space than depth_shift the lines from the backside will poke through

PR #4965, opened 2025-05-09
unify compute plot

SimonDanisch:
Does the final refactor of using only one Plot type for everyting for https://github.com/MakieOrg/Makie.jl/pull/4630.

### DIFF
--- a/ComputePipeline/src/ComputePipeline.jl
+++ b/ComputePipeline/src/ComputePipeline.jl
@@ -198,13 +198,11 @@ end
 function get_observable!(attr::ComputeGraph, key::Symbol)
     return get!(attr.observables, key) do
         val = attr.outputs[key]
-        result = Observable(val[]; ignore_equal_values=true)
+        result = Observable(val[])
         on(attr.onchange) do _
-            if isdirty(val)
-                _val = val[]
-                if !is_same(result[], _val)
-                    result[] = _val
-                end
+            _val = val[]
+            if !is_same(result[], _val)
+                result[] = _val
             end
         end
         return result
@@ -229,7 +227,10 @@ function Observables.onany(f, args::Computed...)
     obsies = map(x -> x isa Computed ? get_observable!(x) : x, args)
     return onany(f, obsies...)
 end
-
+function Observables.onany(f, arg1::Computed, args::Union{Observable, Computed}...)
+    obsies = map(x -> x isa Computed ? get_observable!(x) : x, (arg1, args...))
+    return onany(f, obsies...)
+end
 function Observables.map!(f, target::Observable, args::Computed...)
     obsies = map(x -> x isa Computed ? get_observable!(x) : x, args)
     return map!(f, target, obsies...)
--- a/ComputePipeline/src/ComputePipeline.jl
+++ b/ComputePipeline/src/ComputePipeline.jl
@@ -328,7 +328,11 @@ function Base.getproperty(attr::ComputeGraph, key::Symbol)
     key === :inputs && return getfield(attr, :inputs)
     key === :outputs && return getfield(attr, :outputs)
     key === :onchange && return getfield(attr, :onchange)
-    return attr.inputs[key].output
+    if haskey(attr.inputs, key)
+        return attr.inputs[key].output
+    else
+        return attr.outputs[key]
+    end
 end
 
 function Base.getindex(attr::ComputeGraph, key::Symbol)
@@ -674,6 +678,13 @@ function register_computation!(f, attr::ComputeGraph, inputs::Vector{Symbol}, ou
     _inputs = Computed[attr.outputs[k] for k in inputs]
     unsafe_register!(f, attr, _inputs, outputs)
     return
+end
+
+function Base.map!(f, attr::ComputeGraph, input::Symbol, output::Symbol)
+    register_computation!(attr, [input], [output]) do inputs, changed, cached
+        return (f(inputs[1]),)
+    end
+    return attr
 end
 
 function unsafe_register!(f, attr::ComputeGraph, inputs::Vector{Computed}, output_names)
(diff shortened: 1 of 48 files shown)


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -194,10 +192,62 @@ struct ComputeGraph
...
+        val = attr.outputs[key]
+        result = Observable(val[]; ignore_equal_values=true)
+        on(attr.onchange) do _
+            if isdirty(val)
+                _val = val[]
+                if !is_same(result[], _val)
+                    result[] = _val
+                end
+            end
+        end
```
ffreyer, line 209: I don't think `isdirty` is safe to do here.

Consider a Graph a -> b -> c, where we first register an observable for c, then for b. If a is updated, a, b and c will be marked dirty. Then onchange triggers the c observable, which pulls c, marking a, b, c as resolved/not dirty. Next we trigger the b observable which sees b as not dirty and thus doesn't update.

ComputePipeline/src/ComputePipeline.jl
```diff
@@ -74,7 +74,8 @@ end
 A `ComputeEdge` represents the computation connecting one set of nodes to another.
 They are considered internal and should not be interacted with or created directly.
 """
-struct ComputeEdge <: AbstractEdge
+struct ComputeEdge{T} <: AbstractEdge
+    graph::T
```
ffreyer, line 78: I wonder if it's better to do something like 
```julia
abstract type AbstractComputeGraph end

struct ComputeEdge <: AbstractEdge
    graph::AbstractComputeGraph
    ...
end

struct ComputeGraph <: AbstractComputeGraph ... end
```
for these. Iirc we had a case where this was better for performance already, maybe with ComputeEdge?

src/basic_recipes/buffers.jl
```diff
@@ -17,9 +17,9 @@ end
 function append!(lsb::LineSegments, positions::Vector{Point{N, Float32}}; color = :black, linewidth = 1.0) where N
     thickv = same_length_array(positions, linewidth, key"linewidth"())
     colorv = same_length_array(positions, color, key"color"())
-    append!(lsb.args[1].inputs[:arg1].value, positions)
-    append!(lsb.args[1].inputs[:color].value, colorv)
-    append!(lsb.args[1].inputs[:linewidth].value, thickv)
+    append!(lsb[:arg1][], positions)
+    append!(lsb[:color][], colorv)
+    append!(lsb[:linewidth][], thickv)
```
ffreyer, line 22: Won't this push to the computed instead of the input, skipping conversions?

src/compute-plots.jl
```diff
@@ -349,29 +368,46 @@ function add_attributes!(::Type{T}, attr, kwargs) where {T}
...
+            attr[k].value = RefValue{Any}(value)
         end
 
         # Hack-fix variable type
         if haskey(abstract_type_init, k)
             attr[k].value = abstract_type_init[k]
         end
     end
+    if !haskey(attr, :model)
+        add_input!(attr, :model, Mat4f(I))
```
ffreyer, line 380: At this point it should still be a Mat4d. Also maybe not as an input but a Computed that gets updated by transformations

src/compute-plots.jl
```diff
@@ -389,18 +425,41 @@ function add_theme!(plot::T, scene::Scene) where {T}
...
+
+    P = Plot{Func}
+    attr = ComputeGraph()
+    add_attributes!(P, attr, user_attributes)
+    register_arguments!(P, attr, user_attributes, user_args)
+    converted = attr[:converted][]
+    ArgTyp = MakieCore.argtypes(converted)
+    FinalPlotFunc = plotfunc(plottype(P, converted...))
+    p = Plot{FinalPlotFunc,ArgTyp}(user_attributes, attr)
+    p.transformation = Transformation()
```
ffreyer, line 446: `handle_transformation!()` will replace `p.transformation` so maybe it's better to leave it undefined here? That way trying to do anything will fail instead of seemingly working

ComputePipeline/src/ComputePipeline.jl
```diff
@@ -322,13 +372,19 @@ end
...
     key === :inputs && return getfield(attr, :inputs)
     key === :outputs && return getfield(attr, :outputs)
     key === :onchange && return getfield(attr, :onchange)
-    return attr.inputs[key].output
+    key === :observables && return getfield(attr, :observables)
+    if haskey(attr.inputs, key)
+        return attr.inputs[key].output
+    else
+        return attr.outputs[key]
+    end
```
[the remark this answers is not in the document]
SimonDanisch: But does it have the right name?

ComputePipeline/src/ComputePipeline.jl
SimonDanisch: But we only have one type here - I think the performance improvement came from it having several types in one array, making iteration etc slow.

src/compute-plots.jl
SimonDanisch: Yeah, I should have left a TODO, this definitely needs to change.

ComputePipeline/src/ComputePipeline.jl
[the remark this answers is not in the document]
ffreyer: It would error here if it didn't:
https://github.com/MakieOrg/Makie.jl/blob/15cb9a8d1f5ab5116738d5bd455d2b4e8ffae5c6/ComputePipeline/src/ComputePipeline.jl#L545-L549

ComputePipeline/src/ComputePipeline.jl
ffreyer: Alright, I checked, it's significantly worse to do what I proposed.

PR #4966, opened 2025-05-09
Compute lighting

ffreyer:
# Description

Refactors lighting to be part of the scene compute graph. Currently is and probably will be breaking, because the Observables in Light structs are annoying keep and lights are a bit awkward in general

TODO:
- [x] prototype Makie code
- [x] get fast shading working in GLMakie
- [x] get multi light shading working in GLMakie
- [ ] update WGLMakie
- [x] update CairoMakie
- [x] probably make `plot.shading` a Bool again
- [x] make multi-light vs fast shading pickable per scene
- [x] add utility functions for updating lights
- [x] cleanup FastShading and MultiLightShading in tests
- [ ] update docs (maybe leave this for later when the interface changes had some time to solidify)

Deprecates:
- FastShading & MultiLightShading as a plot attribute

Breaking Changes:
- lights no longer have Observable fields
- `RectLight` transformation function are no longer inplace (they now generate a new RectLight and no longer uses ! postifx)

### DIFF
--- a/CairoMakie/src/new_primitives.jl
+++ b/CairoMakie/src/new_primitives.jl
@@ -621,14 +621,6 @@ function draw_mesh3D(
         align_pattern(per_face_col, scene, f32c_model)
     end
 
-    # TODO: assume Symbol here after this has been deprecated for a while
-    if shading isa Bool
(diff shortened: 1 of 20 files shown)


ffreyer:
Updates of the light_direction are still broken in WGLMakie. Seems like that code that is supposed to update it never runs, or only when the julia process gets detached

PR #4967, opened 2025-05-09
Add sources to all Project.tomls

asinghvi17:
This way, if you dev Makie, it will automatically dev the local MakieCore as well.  If you dev a backend it should dev both Makie and MakieCore.

Sources does not apply if you are not in the project.toml or not dev'ing the project.  So regular package downloads just get the normal published MakieCore.


GLMakie/Project.toml
```diff
@@ -40,3 +40,6 @@ Printf = "1.0, 1.6"
 ShaderAbstractions = "0.5"
 StaticArrays = "0.12, 1.0"
 julia = "1"
+
+[sources]
+Makie = { path = ".." }
```
t-bltg, line 45: Should ```
[sources]
ReferenceTests = {path = "../../ReferenceTests"}
``` be added to `test/Project.toml` (same remark for `CairoMakie` and `WGLMakie`) ?

I think `ReferenceTests = "d37af2e0-5618-4e00-9939-d430db56ee94"` can be removed from `GLMakie/Project.toml` (only used in tests).

PR #4979, opened 2025-05-15
Fix `eltype` of `entry_observer_funcs` inside `initialize_block!(leg::Legend; entrygroups)`

lcontento:
The vector `entry_observer_funcs` should contain `ObserverFunctions` instead of `Observables`.
Its elements are populated by
```julia
obsfunc = onany(visibilities...) do vis...
  ...
end
append!(entry_observer_funcs, obsfunc)
```
Here `obsfunc` is a vector of `ObserverFunctions` which at the moment get wrapped automatically by `Observables` when they are appended to `entry_observer_funcs`.
This leads to an error when
```julia
foreach(off, entry_observer_funcs)
```
is later called, since `off` does not work on `Observables` but only on `ObserverFunctions`.
Fixing the `eltype` of `entry_observer_funcs` solves the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fixed regression in the updating logic of `Legend` [#4979](https://github.com/MakieOrg/Makie.jl/pull/4979).
+
 ## [0.22.6] - 2025-05-17
 
 - Added `alpha` keyword to `density` recipe [#4975](https://github.com/MakieOrg/Makie.jl/pull/4975).
--- a/ReferenceTests/src/tests/figures_and_makielayout.jl
+++ b/ReferenceTests/src/tests/figures_and_makielayout.jl
@@ -313,10 +313,16 @@ end
 @reference_test "Legend update" begin
     fig = Figure()
     ax = Axis(fig[1, 1])
-    lines!(ax, [0, 1], [1, 1]; color = :blue, label = "lines")
-    scatter!(ax, [0], [0]; color = :red, label = "scatter")
-    legend = axislegend(ax; position = :rb)
-    reverse!(legend.entrygroups[][1][2])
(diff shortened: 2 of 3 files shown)


jkrumbiegel:
Ah the conversion behavior of Observables hides the issue... Would be good to have a test for Legend reconstruction then, even if it's not something that's particularly advertised. I think a reference test would be good, where two legends are made with the same multiple groups, then one of them has their entrygroups observable modified such that the two groups are in reverse order and also themselves reversed. Then we can check using the reference image if everything got correctly reconstructed.

lcontento:
> Ah the conversion behavior of Observables hides the issue... Would be good to have a test for Legend reconstruction then, even if it's not something that's particularly advertised. I think a reference test would be good, where two legends are made with the same two groups, then one of them has their entrygroups observable modified and notified such that the two groups are in reverse order and also themselves reversed. Then we can check using the reference image if everything got correctly reconstructed.

If we are comparing images such test should go into one of the backends test suite? Something like
```julia
using Test
using CairoMakie
fig1 = Figure()
ax1 = Axis(fig1[1, 1])
scatter!(ax1, [0], [0]; color = :red, label = "scatter")
lines!(ax1, [0, 1], [1, 1]; color = :blue, label = "lines")
axislegend(ax1)
fig2 = Figure()
ax2 = Axis(fig2[1, 1])
lines!(ax2, [0, 1], [1, 1]; color = :blue, label = "lines")
scatter!(ax2, [0], [0]; color = :red, label = "scatter")
legend2 = axislegend(ax2)
reverse!(legend2.entrygroups[][1][2])
notify(legend2.entrygroups)
@test colorbuffer(fig1) == colorbuffer(fig2)
```

jkrumbiegel:
I meant one of the reference tests in examples2d.jl or so, just to ensure that the Legend changes visually as expected when it does the reconfiguration

lcontento:
I added a test to the `ReferenceTests` folder as you suggested.

PR #4983, opened 2025-05-15
Add option to change label and label value fontsize in SligerGrid

### LINKED ISSUE #2377: Cannot set fontsize with SliderGrid - GLMakie
Hello,

By updating my Julia version, I just remarked that `labelslidergrid!` is deprecated. Thus, I had to use `SliderGrid`. However, there is no way to increase the fontsize of both the label and the current slider value. I tried the argument `textsize` such as with `Button`, but it didn't work. In addition, the [documentation ](https://docs.makie.org/stable/examples/blocks/slidergrid/)does not mention any fontsize tuning.

Making it possible would be super nice.

Thanks for reading me.

Regards,
Arthur

pmc4:
# Description

Fixes #2377 

* Adds `labelsize` attribute to `SliderGrid` block to change the fontsize of the slider labels.
* Adds `valuesize` attribute to `SliderGrid` block to change the fontsize of the slider label values.

Maybe `valuesize` is not the most appropiate name. Should `valuelabelsize` or something else be used instead?

For reference, this is how it looks like on the example of the docs:

example

I have exagerated the fontsize of both elements to perceive the difference. The code is the following:

```julia
using GLMakie


fig = Figure()

ax = Axis(fig[1, 1])

sg = SliderGrid(
    fig[1, 2],
    (label = "Voltage", range = 0:0.1:10, format = "{:.1f}V", startvalue = 5.3),
    (label = "Current", range = 0:0.1:20, format = "{:.1f}A", startvalue = 10.2),
    (label = "Resistance", range = 0:0.1:30, format = "{:.1f}Ω", startvalue = 15.9),
    width = 350,
    tellheight = false,
    labelsize = 8,
    valuesize = 30)

sliderobservables = [s.value for s in sg.sliders]
bars = lift(sliderobservables...) do slvalues...
    [slvalues...]
end

barplot!(ax, bars, color = [:yellow, :orange, :red])
ylims!(ax, 0, 30)

fig
```

## Type of change

## Checklist

### DIFF
--- a/src/makielayout/blocks/slidergrid.jl
+++ b/src/makielayout/blocks/slidergrid.jl
@@ -23,10 +23,10 @@ function initialize_block!(sg::SliderGrid, nts::NamedTuple...)
         range = nt.range
         format = haskey(nt, :format) ? nt.format : default_format
         remaining_pairs = filter(pair -> pair[1] ∉ (:label, :range, :format), pairs(nt))
-        l = Label(sg.layout[i, 1], label, halign = :left)
+        l = Label(sg.layout[i, 1], label, halign = :left, fontsize = sg.labelsize)
         slider = Slider(sg.layout[i, 2]; range = range, remaining_pairs...)
         vl = Label(sg.layout[i, 3],
-            lift(x -> apply_format(x, format), slider.value), halign = :right)
+            lift(x -> apply_format(x, format), slider.value), halign = :right,fontsize = sg.valuesize)
         push!(sg.valuelabels, vl)
         push!(sg.sliders, slider)
         push!(sg.labels, l)
--- a/src/makielayout/types.jl
+++ b/src/makielayout/types.jl
@@ -1031,6 +1031,10 @@ end
         alignmode = Inside()
         "The width of the value label column. If `automatic`, the width is determined by sampling a few values from the slider ranges and picking the largest label size found."
         value_column_width = automatic
+        "The font size of the entry labels."
+        labelsize = @inherit(:fontsize, 16f0)
+        "The font size of the entry value labels."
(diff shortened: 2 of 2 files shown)


pmc4:
Hi, I'd like to continue with this PR. I was waiting for some feedback, but maybe I messed up something with the Draft option of the pull request.

I need to update the files to the most recent makie version, of course. But apart from that, is the syntax alright?

SimonDanisch:
Sorry, we're slow with reviews, since there's sadly not as much time for it :( 
But looks like the right approach to me and something which should be simple to merge!

jkrumbiegel:
The current behavior is to send all keywords to Slider. Rather than add keywords for the labels specifically, it would be easier to add `labelkw = (;...)` and `valuekw = (;...)` options where one can pass anything down. At the time, I didn't expect people to do much adjustment to the individual labels as they should usually look uniform

pmc4:
> Sorry, we're slow with reviews, since there's sadly not as much time for it :( But looks like the right approach to me and something which should be simple to merge!

I know, sorry! I know there's a lot of work to do and I appreciate it. That's why if I can do something on my end to make this easier, I'd love to.


> The current behavior is to send all keywords to Slider. Rather than add keywords for the labels specifically, it would be easier to add `labelkw = (;...)` and `valuekw = (;...)` options where one can pass anything down. At the time, I didn't expect people to do much adjustment to the individual labels as they should usually look uniform

My idea for the change arose because I made a plot with a SliderGrid. But the fact that the label and the value share the same size made everything look ugly and uncompensated. That's why I wanted a more precise control.

I'll make some changes and upload it, thanks!

PR #4990, opened 2025-05-17
Pin to documentervitepress 0.1 for now

jkrumbiegel:


asinghvi17:
What was the issue?  Or are we looking at the json thing

jkrumbiegel:
Just running the new version without correct changes pushes the new folder structure, which just broke dev and stable because I didn't put compat in early enough. I set up Makie's docs to only upload minor versions at some point (because docs are already huge so having patch versions hang around is too much for the deploy process). But I would have to adjust the new process to that. Maybe DocumenterVitepress could have an option like "minor version only" for makedocs

asinghvi17:
Could also host them on a separate repo, if this is an issue

jkrumbiegel:
I don't really see a point with keeping every patch version's docs, it's fine if they overwrite each other

PR #4996, opened 2025-05-20
deprecate `attr[key] = Observable` (compute)

ffreyer:
This is meant to not completely break things like:
```julia
c.attributes[:_computed_levels] = lift(c, zs, c.levels, c.mode) do zs, levels, mode
    return _get_isoband_levels(Val(mode), levels, vec(zs))
end
```

and maybe also

```julia
plot[:my_attr] = Observable(...)
```

### DIFF
--- a/ComputePipeline/src/ComputePipeline.jl
+++ b/ComputePipeline/src/ComputePipeline.jl
@@ -224,7 +224,12 @@ function get_observable!(attr::ComputeGraph, key::Symbol)
         result = Observable(val[])
         on(attr.onchange) do _
             _val = val[]
-            if !is_same(result[], _val)
+            # Using !is_same(result[], _val) here results in StackOverflows
+            # from Observables repeatedly updating, e.g. tricontourf:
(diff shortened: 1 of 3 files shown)


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -224,7 +224,12 @@ function get_observable!(attr::ComputeGraph, key::Symbol)
         result = Observable(val[])
         on(attr.onchange) do _
             _val = val[]
-            if !is_same(result[], _val)
+            # Using !is_same(result[], _val) here results in StackOverflows
+            # from Observables repeatedly updating, e.g. tricontourf:
+            #   c.attributes[:_computed_levels] = lift(..., c, zs, c.levels, c.mode)
+            #   which calls add_input!(attr, :_computed_levels, obs)
+            # will infinitely update zs (_arg2) when levels get updated
+            if result[] != _val
```
ffreyer, line 232: This also fixed StackOverflows in Datashader AggMean and AggCount

PR #4999, opened 2025-05-22
hoist transform_func() calls out of loops for type stability and performance

aplavin:
# Description

Fixes # (issue)

I noticed a very slow `rangebars` plot, and found that the vast majority of time was spent in 
https://github.com/MakieOrg/Makie.jl/blob/e4b3b5b5ebde7de7b4f745b859f68ab27667e364/src/basic_recipes/error_and_rangebars.jl#L265-L267
Turns out, this is caused by type instability: the type of `plot` doesn't tell anything about the type of `transform_func(plot)`. Hoisting that call out of the `map` loop helps immensely.

I also did a search for `transform_func` calls in loops, and found only one in contours – hoisted that call out of the loop as well.

No behavior changes anywhere, just performance improvements.

## Type of change

## Checklist

### DIFF
--- a/src/basic_recipes/contours.jl
+++ b/src/basic_recipes/contours.jl
@@ -264,6 +264,7 @@ function plot!(plot::T) where T <: Union{Contour, Contour3d}
         transform_marker = false
     )
 
+    transfunc = transform_func(plot)
     lift(plot, scene.camera.projectionview, transformationmatrix(plot), scene.viewport,
             labels, labelcolor, labelformatter, lev_pos_col
         ) do _, _, _, labels, labelcolor, labelformatter, lev_pos_col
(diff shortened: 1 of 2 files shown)


src/basic_recipes/contours.jl
```diff
@@ -264,6 +264,7 @@ function plot!(plot::T) where T <: Union{Contour, Contour3d}
         transform_marker = false
     )
 
+    transfunc = transform_func(plot)
     lift(plot, scene.camera.projectionview, transformationmatrix(plot), scene.viewport,
```
asinghvi17, line 268: Better to lift on transform_func_obs(plot), since the transform func can change arbitrarily during display.

src/basic_recipes/error_and_rangebars.jl
```diff
@@ -261,8 +261,9 @@ function plot_to_screen(plot, points::AbstractVector)
     spvm = clip_to_space(cam, :pixel) * space_to_clip(cam, space) *
         f32_convert_matrix(plot, space) * transformationmatrix(plot)[]
 
+    transfunc = transform_func(plot)
```
asinghvi17, line 264: here, transfunc should be passed in to plot_to_screen, lifted from transform_func_obs(plot) here:

https://github.com/MakieOrg/Makie.jl/pull/4999/files#diff-66312870738e99091c1bc0a5c6577921970f5caabba3348970fefdef0b963369L212

asinghvi17 (changes requested):

PR #5001, opened 2025-05-22
make plotkey() zero-cost

aplavin:
A trivial change to make `plotkey()` zero-cost in runtime. This function is sometimes convenient to use in loops, and it should fundamentally be free to call.


SimonDanisch:
I guess that's a fair use of `@generated`! :)

PR #5003, opened 2025-05-22
rangebars performance: improve type stability, hoist transform_func() out of loop

aplavin:
supercedes https://github.com/MakieOrg/Makie.jl/pull/4999


SimonDanisch:
Thanks!

PR #5004, opened 2025-05-22
rangebars: don't show whiskers when not needed

aplavin:
I noticed these whiskers taking a significant chunk of my `profview`, even when not needed (width is set to zero, which is the default). This PR makes it so that whiskers aren't computed nor drawn when not needed.

What I'm not certain about, and looking for your advice, is the specific implementation. Now it checks the initial value of the whisker width only.

### DIFF
--- a/src/basic_recipes/error_and_rangebars.jl
+++ b/src/basic_recipes/error_and_rangebars.jl
@@ -209,7 +209,7 @@ function _plot_bars!(plot, linesegpairs, is_in_y_direction)
 
     scene = parent_scene(plot)
 
-    whiskers = lift(plot, linesegpairs, scene.camera.projectionview, plot.model,
+    whiskers = iszero(whiskerwidth[]) ? nothing : lift(plot, linesegpairs, scene.camera.projectionview, plot.model,
         scene.viewport, transform_func(plot), whiskerwidth) do endpoints, _, _, _, _, whiskerwidth
 
         screenendpoints = plot_to_screen(plot, endpoints)
@@ -222,7 +222,7 @@ function _plot_bars!(plot, linesegpairs, is_in_y_direction)
         return [p for pair in screenendpoints_shifted_pairs for p in pair]
     end
     whiskercolors = Observable{RGBColors}()
-    lift!(plot, whiskercolors, color) do color
+    iszero(whiskerwidth[]) || lift!(plot, whiskercolors, color) do color
         # we have twice as many linesegments for whiskers as we have errorbars, so we
         # need to duplicate colors if a vector of colors is given
         if color isa AbstractVector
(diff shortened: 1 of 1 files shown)


src/basic_recipes/error_and_rangebars.jl
```diff
@@ -209,7 +209,7 @@ function _plot_bars!(plot, linesegpairs, is_in_y_direction)
 
     scene = parent_scene(plot)
 
-    whiskers = lift(plot, linesegpairs, scene.camera.projectionview, plot.model,
+    whiskers = iszero(whiskerwidth[]) ? nothing : lift(plot, linesegpairs, scene.camera.projectionview, plot.model,
```
ffreyer, line 212: This would also mean you can't make whiskers visible after the plot call. I.e. `plot.whiskerwidth[] = 5`.

Once we're done with #4630 this can be handled more efficiently with the compute graph too. Probably by just making the whisker plot invisible, because that should stop rendering from requesting this calculation to run.

jkrumbiegel: yeah this was always a drawback of the observables design, that you pay for the whole graph most of the time, even if parts are unused/hidden. How many rangebars were you using to notice the slowness? You could also use `linesegments` instead of keeping a fork, that should be little refactoring work

SimonDanisch: Do you have an MWE?

SimonDanisch: Just so i can take a look at what is slow, maybe also in the compute branch.

SimonDanisch: I see..
For compute graph that exact MWE goes at least from 0.133978s for tagged to 0.08s on `sd/compute-graph`, which is something :) 
I guess if we move the recipe to fully use compute-graph, and disable calculations when invisible (which would be nice to add before merging), I could see this coming down to a similar timing as this PR.

jkrumbiegel: > and won't update whiskerwidth for this plot

I don't agree with that, I've tried hard to make as many things in Makie updateable as possible, and regard the places where it's not possible as "design warts". But if the compute graph helps to shut down specific branches if some conditions hold, that would be a very good way of dealing with this.

What could be done before that is to avoid the calculation of the whiskers if their width is zero, for example make all coordinates NaN or some constants. It wouldn't matter that they are in the wrong spot as long as they are not visible anyway.

ffreyer: In #4630 rendering pulls data in. If a primitive plot is not visible, the backend won't bother to pull. So if this stuff gets translated setting the whisker plot to `visible = false` is equivalent to never triggering this `lift`. Well, probably once on initialization in W/GLMakie and really never in CairoMakie.

Btw another way you can probably drastically cut the cost of this is to use something like `scatter(errorbar_endpoints, marker = Rect, markersize = Vec2f(whiskerwidth, whiskerlinewidth))`. That should help with the whisker-visible case too

ffreyer:
I'm closing this since it's quite outdated now and we updated errorbars/rangebars to use the scatter based solution.

Some benchmarks with

```julia
n = 10^5
x = Observable(0.0)
f,a,p = rangebars(rand(n), rand(n), (@lift $x .+ rand(n)));
```

- `x[] = 1` with no display runs at ~100µs
- `x[] = rand()` with no display ~300µs
- `x[] = rand(); boundingbox(p)` is about ~2ms (attempt to force updates)
- `x[] = rand(); colorbuffer(p)` is about ~6ms (forced render)

PR #5005, opened 2025-05-22
optimize some marker calculations

aplavin:
I noticed these functions taking a lot of time in my `profview`, as they do significant calculations and font lookups.
This PR is an attempt to optimize them.

Currently, it just applies the underlying calculations to the first element of the vector. This is probably not fully correct when e.g. different Char markers have very different sizes?
What would be the best way to handle it? Run `unique` first, and only perform font calculations for unique symbols?..

### DIFF
--- a/src/utilities/texture_atlas.jl
+++ b/src/utilities/texture_atlas.jl
@@ -452,7 +452,10 @@ primitive_uv_offset_width(atlas::TextureAtlas, x, font) = Vec4f(0,0,1,1)
 primitive_uv_offset_width(atlas::TextureAtlas, b::BezierPath, font) = glyph_uv_width!(atlas, b)
 primitive_uv_offset_width(atlas::TextureAtlas, b::Union{UInt64, Char}, font) = glyph_uv_width!(atlas, b, font)
 primitive_uv_offset_width(atlas::TextureAtlas, hash::UInt32, font) = atlas.uv_rectangles[atlas.mapping[hash]]
-primitive_uv_offset_width(atlas::TextureAtlas, x::AbstractVector, font) = map(m-> primitive_uv_offset_width(atlas, m, font), x)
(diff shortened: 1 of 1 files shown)


src/utilities/texture_atlas.jl
```diff
@@ -500,7 +503,9 @@ end
 rescale_marker(atlas, char, font, markersize) = markersize
 
 function rescale_marker(atlas::TextureAtlas, char::AbstractVector{Char}, font, markersize)
-    return _bcast(markersize) .* marker_scale_factor.(Ref(atlas), char, font)
+    msf = marker_scale_factor(atlas, first(char), font)
+    msfs = fill(msf, length(char))
+    return _bcast(markersize) .* msfs
```
ffreyer, line 508: These are not going to work with per element markers because they don't have a consistent size in the texture atlas. E.g.:
```julia
julia> Makie.marker_scale_factor.(Ref(Makie.get_texture_atlas()), ['A', 'b', '.', '|', '-'], Ref(Makie.defaultfont()))
5-element Vector{Vec{2, Float32}}:
 [1.0, 1.09375]
 [0.84375, 1.109375]
 [0.46875, 0.46875]
 [0.4375, 1.3125]
 [0.609375, 0.4375]
```

SimonDanisch: Do you have an MWE as well that i can run?

SimonDanisch:
I'm not merging this for now, since I want to take a closer look at it and dont have time right now.
A least `primitive_uv_offset_width` should already be cached, so I'm wondering where the speed up comes from there.
For `marker_scale_factor`, if it's really that expensive, we should actually add the cache to the TextureAtlas, I think.

SimonDanisch:
Merged in https://github.com/MakieOrg/Makie.jl/pull/5049

PR #5009, opened 2025-05-23
onchange propagation (compute)

ffreyer:
This changes how `onchange` is triggered and cleared so that multiple onchange observables can be triggered when multiple graphs are connected. This fixes issues where a child graph does not trigger obseravbles when the parent graph is updated.

The way this works is that it collects every unique `edge.graph.onchange` it comes across in `mark_dirty!(edge)` in a vector. After marking is done, every observable gets notified, which triggers all the registered observable updates. To make sure the the changeset in onchange gets cleared for child graphs, it now clears `on(onchange)` with minimal priority (i.e. at the end of a trigger)

### DIFF
--- a/ComputePipeline/src/ComputePipeline.jl
+++ b/ComputePipeline/src/ComputePipeline.jl
@@ -335,6 +335,7 @@ function resolve!(input::Input)
 end
 
 function mark_dirty!(input::Input, obs_to_update::Vector{Observable})
+    push!(input.graph.onchange.val, input.name)
     if !(input.graph.onchange in obs_to_update)
         push!(obs_to_update, input.graph.onchange)
     end
--- a/ComputePipeline/src/ComputePipeline.jl
+++ b/ComputePipeline/src/ComputePipeline.jl
@@ -220,12 +220,19 @@ struct ComputeGraph
 end
 
 function get_observable!(attr::ComputeGraph, key::Symbol)
+    # Because we allow output arrays to be reused it can be impossible to tell
+    # if the data has updated. In this case the data is marked as dirty/changed
+    # and added to the `onchange`. If this data is fed into an Observable which
+    # updates the graph it can lead to infinite loops.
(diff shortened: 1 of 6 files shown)


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -294,20 +297,24 @@ function isdirty(edge::ComputeEdge)
     return any(edge.inputs_dirty)
 end
 
-function mark_dirty!(edge::ComputeEdge)
+function mark_dirty!(edge::ComputeEdge, obs_to_update::Vector{Observable})
+    # Assumes this is the same graph as edge.outputs (for parent -> child graph edges)
+    g = edge.graph
```
ffreyer, line 302: TODO: add test to make sure this is true/doesn't get changed

ffreyer:
Doing something like this does not work:
```julia
function mark_dirty(edge)
    edge.dirty = true
    for dep in edge.dependents
        mark_dirty!(dep)
    end

    g = edge.graph
    g.onchange[] = edge.name
end
```

Consider:

```julia
g = ComputeGraph()
add_input!((k, x) -> Float32[x; 0], g, :a1, [0])
add_input!((k, x) -> Float32[x; 0], g, :a2, [0])
map!((x,y,z) -> Point.(x, y, z), g, [:a1, :a2], :b)
ComputePipeline.get_observable!.(Ref(g), keys(g.outputs))
```
updating both inputs will trigger:
1. update loops through inputs, setting their value and calling mark_dirty one by one
2. input a1 -> computed a1 is called, marking the a1 (computed) dirty and continues to
3. (a1, a2) -> b which is marked dirty and has no dependents so
4. (a1, a2) -> b triggers onchange, pulling b. Currently a2 is not marked, so only a1 is pulled. We now have unsynchronized updates

Also if this Problem does not trigger we still have a performance issue. In this example we propagate to b twice, once from a1, once from a2. So the b observable triggers twice. If there are more paths to b, the number of triggers will grow further

PR #5013, opened 2025-05-26
Adjust docs build to DocumenterVitepress 0.2

jkrumbiegel:


jkrumbiegel:
@lazarusA hm good point, shouldn't the build have failed without that being present?

lazarusA:
not really, it just doesn't deploy to the right paths / folders, as far as I know 😄 .

jkrumbiegel:
hm no, that path is really just needed for the version picker, but during the build phase it should error due to the missing variable which is not replaced

ahh it's because we have our own version picker, and don't pull in the one from the template, I think

PR #5022, opened 2025-06-02
Allow `hist` with an empty list

asinghvi17:
# Description

Allows `hist(Float64[])` to work by special casing if in .  Defaults to 0 +- 0.5.

## Type of change

## Checklist

### DIFF
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -148,7 +148,8 @@ end
 
 function pick_hist_edges(vals, bins)
     if bins isa Int
-        mi, ma = float.(extrema(vals; init = (0.0, 0.0)))
+        isempty(vals) && return 0:10 # corresponds to what we do for empty `Axis`
+        mi, ma = float.(extrema(vals))
         if mi == ma
             return (mi - 0.5):(ma + 0.5)
         end
--- a/src/stats/hist.jl
+++ b/src/stats/hist.jl
@@ -148,7 +148,7 @@ end
 
 function pick_hist_edges(vals, bins)
     if bins isa Int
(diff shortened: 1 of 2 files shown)


asinghvi17:
@Moelf any objections / better ideas?

Moelf:
what's the motivation for this? sounds like this just makes an empty canvas with arbitrary x-range?

I think it makes slightly more sense of `bins=...` is required, but even then idk what y-range it's suppose to be, 0..0?

BenChung:
I prompted this - the use case is that I'm making a histogram interactively with a "clear data" button, and this change makes that not error out horribly.

The other argument for it I'd make is that the ergonomics of the error caused by making a histogram of empty data being a stack trace into bin edge selection aren't great.

Moelf:
>I prompted this - the use case is that I'm making a histogram interactively with a "clear data" button, and this change makes that not error out horribly.

That makes sense, sounds like you conceptually want to separate histogram making and histogram plotting, from where I come from (particle physics). I share a lot of that mental model with you, which is why I almost always make a FHist.jl histogram object before plotting, this allows me to think of histogram binedges as defined, and data being "pushed into" it.

I guess if `scatter(Float64[], Float64[])` works already, `hist(Float64[])` should not error out, this is just Makie's contract of "defaults to 0..10, 0..10 when there's nothing to show", and I'm fine with that.

jkrumbiegel:
seems to me if there are no values the bar calculation should just be sidestepped entirely and there should be a barplot with zero bars. `barplot(Float64[], Float64[])` renders an empty axis just fine.

asinghvi17:
Thanks @ffreyer!

PR #5027, opened 2025-06-02
Bug/(tri)contourf colorrange

### LINKED ISSUE #4016: `colorrange` is an invalid attribute for `contourf`

MWE:
```julia
contourf(randn(10, 10); colorrange = (0, 1))
```

It looks like the color attributes are specified explicitly (instead of using the new `mixin` functions).  We could change this to use eg `MakieCore.mixin_colormap_attributes()...` instead.  The implementation would have to be updated to pass all of the color attributes through to the color determining function as well.
https://github.com/MakieOrg/Makie.jl/blob/c4e3b7b7271322330e2b4db424a5e096e5e9acf2/src/basic_recipes/contourf.jl#L7-L42

lanceXwq:
# Description

Fixes #4016 by using `MakieCore.mixin_colormap_attributes` for both `contourf` and `tricontourf`. Setting `colorrange`, `colorscale`, `lowclip`, `highclip`, and `alpha` all seem to work. Please let me know if anything needs to change. :smiley: 

## Type of change

### DIFF
--- a/src/basic_recipes/contourf.jl
+++ b/src/basic_recipes/contourf.jl
@@ -27,8 +27,6 @@ similar to how [`surface`](@ref) works.
     This can be used for example to draw bands for the upper 90% while excluding the lower 10% with `levels = 0.1:0.1:1.0, mode = :relative`.
     """
     mode = :normal
-    colormap = @inherit colormap
-    colorscale = identity
     """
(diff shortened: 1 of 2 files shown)


SimonDanisch:
I think we should merge this?

PR #5033, opened 2025-06-03
annotation: default textcolor to color

aplavin:
# Description

Now `color` adjusts both line and text color, unless a `textcolor` is passed explicitly.
Opening this quick PR following Julius advice from slack.

## Type of change

Delete options that do not apply:

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #5034, opened 2025-06-03
`Annotation` quick fixes

jkrumbiegel:
- **add scalar vectypes conversions**
- **make `textcolor` follow `color` by default, and `color` theme-compatible**

Supersedes #5033 

I guess after the release two hours ago I got quickly convinced by @aplavin that it would be more user-friendly if `textcolor` followed `color` by default. I also found playing with AoG that it's more natural to try `color` first, and there are no lines in a typical ggrepel style plot, or they are very small, which is confusing as then `color` has no effect. So I'm pushing this as a "quick fix" and tagging patch, even though technically breaking, practically it has no impact as the feature was just released two hours ago. Better get ahead of it before more people start using this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+## [0.22.10] - 2025-06-03
+
+- Quick fix for the just released `annotation`, `textcolor` now follows `color` by default [#5034](https://github.com/MakieOrg/Makie.jl/pull/5034).
+
 ## [0.22.9] - 2025-06-03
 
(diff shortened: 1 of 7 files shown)


aplavin:
Nice, thanks for a quick update!
I still wonder about PointBased though, which would make it more composable (as `text` is). I added this short snippet to `MakieExtra.jl`, and found it to work well: https://github.com/JuliaAPlavin/MakieExtra.jl/blob/3d41001dd8a696b033b548142b7d696332af614a/src/MakieExtra.jl#L330-L338.

Maybe it can be added to Makie proper?

jkrumbiegel:
Yeah we can try cleaning up / refactoring the converts in a separate PR

PR #5035, opened 2025-06-03
add stroke support to Band

aplavin:
# Description

Add stroke (strokewith + strokecolor) support to Band. By default, no stroke is shown.

After adding `bandstroke` recipe to MakieExtra.jl, I recognized that in ~80% of cases I want a stroked band. It's a small addition to the Makie Band recipe, just a few lines in this PR.

## Checklist

### DIFF
--- a/Makie/src/basic_recipes/band.jl
+++ b/Makie/src/basic_recipes/band.jl
@@ -79,7 +79,7 @@ function Makie.plot!(plot::Band)
     mesh!(plot, plot.attributes, plot.coordinates, plot.connectivity, color = plot.colors)
 
     map!(plot, :strokecolor, :linecolor) do strokecolor
-        if strokecolor isa Vector
+        if strokecolor isa AbstractVector
             return vcat(strokecolor, strokecolor[1:1], strokecolor)
         else
             return strokecolor
@@ -98,7 +98,7 @@ function Makie.plot!(plot::Band)
 
     lines!(
         plot, plot.attributes, plot.merged_points,
-        linewidth = plot.strokewidth, color = plot.strokecolor,
+        linewidth = plot.strokewidth, color = plot.linecolor,
         fxaa = false
     )
 


aplavin:
Ideally, strokecolor default should be "same as `color`, but 100% opaque". What would be the best way to achieve that? I'm a bit lost in the Makie color pipeline.

ffreyer:
Same as color seemed like a weird default to me since that makes the lines invisible if you don't change anything. And it also doesn't mesh well with color patterns, since lines don't support them.

If you wanted to strip alpha you should probably add a strokealpha separately and do something like this:
```julia
lattr[:alpha] = plot.strokealpha
lattr[:color] = map(plot, plot.strokecolor, plot.color) do strokecolor, color
    if strokecolor !== automatic
        return strokecolor
    elseif color isa AbstractMatrix
        return :black # some default for when color is not usable
    elseif color isa Union{Real, Vector{Real}}
        return color
    else 
        return RGBf.(to_color(color))
    end
end
```
I hope that catches all the cases...

Btw did you lock this branch? I can't push to it anymore, or resolve conflicts. I have some more changes lying around locally that I wanted to push before re-running CI

SimonDanisch:
I think copying polygon seems like the way to go here.

aplavin:
Can we make strokecolor to default to "color but without alpha"? This is probably the desired behavior in the vast majority of cases.
So that
`band(..., color=(:red, 0.3), strokewidth=1)`
and/or
`band(..., color=:red, strokewidth=1, alpha=0.3)`
have `:red` stroke and `(:red, 0.3)` fill.

The current behavior for `band([1,2], [3,4], [5,7]; strokewidth=3, alpha=0.5, color=:red)`:
image
Doesn't seem very expected or desired.

aplavin:
Btw, assuming this one gets finalized and accepted, I plan to make a corresponding PR for hspan/vspan as well.

SimonDanisch:
One problem with that is, that we've been planning to make the default color for meshes not transparent anymore.
Which seems like a good idea? 
We could also try to find a better stroke color default for band not defaulting to a transparent color.

aplavin:
comment removed: I originally understood "transparent" as `:transparent` (ie fully transparent), not partially

ffreyer:
The problem is that band will also default to a non-transparent color then, and if `strokecolor = strip_alpha(color)` you won't see stroke unless you set either color to something transparent or strokecolor to some other color. It would be better to already see a change from just setting `strokewidth`.

aplavin:
You mean, `band(...)` and `band(..., strokewidth=2)` should be different without any extra arguments? I don't have any strong opinion here, was mostly thinking about manually-specified colors. That is, "red semi-transparent band with opaque red stroke" should require specifying "red" only once.

PR #5037, opened 2025-06-03
simplify & generalize closest_index_inexact

aplavin:
# Description

Slider set_close_to! only worked for "plain" number – eg, not units.
This PR fixes it by simplifying the implementation (same behavior).

## Type of change

## Checklist


ffreyer:
> (same behavior)

Well not quite. If you have a range starting below 0 you now get the first value in the range where previously you got the closest to 0. Maybe it would be better to strip units before searching for a fitting value and keep that?

aplavin:
Now the default startvalue is closest to zero

ffreyer:
Great, I think we can merge this then

PR #4964, opened 2025-05-09
Handle `dim_conversions` in `Axis3`

fhagemann:
# Description

Fixes #4963.

This is my attempt on fixing the issue #4963 I created, namely handling `DimConversions` in `Axis3`. It solves the problem I reported in the issue, but I cannot judge whether this is a full solution, so maybe some expert should review and judge the completeness of this.

I looked into how `dim_conversion` is handled in `Axis`, and it looks like, in `Axis3`,
1. the dimensions passed to the `axis` were just not passed to the newly created `Scene` when initializing the block
2. the conversions were not passed to `get_ticks` when creating the ticknodes.

This should be fixed now in this PR.

Example 1: 
```julia
using CairoMakie, Unitful
fig = Makie.Figure()
ax = Makie.Axis3(fig[1,1], dim1_conversion = Makie.UnitfulConversion(u"m", units_in_label = true))
@show ax.scene.conversions[1]
fig
```

Before this PR:
`ax.scene.conversions` would ignore what was passed and the axis ticklabels would not show the unit.
```
ax.scene.conversions[1] = nothing
```
image

Using this PR:
The `ax.scene.conversions` has the `dim_conversion` passed to it and the axis ticklabels show units now.
```
ax.scene.conversions[1] = Makie.UnitfulConversion(Observable{Any}(m), false, Observable(true), Dict{String, Tuple{Any, Any}}())
```
image

Example 2:
```julia
using CairoMakie, Unitful
fig = Makie.Figure()
ax = Makie.Axis3(fig[1,1], dim1_conversion = Makie.UnitfulConversion(u"m", units_in_label = true))
Makie.scatter!(ax, rand(10)*u"m", rand(10), rand(10))
@show ax.scene.conversions[1]
fig
```

Before this PR:
`ax.scene.conversions` would "guess" the unit conversion (`automatic`), in this case `dm` (decimeter) but not display them --> BUG
```
ax.scene.conversions[1] = Makie.UnitfulConversion(Observable{Any}(dm), true, Observable(true), Dict{String, Tuple{Any, Any}}("656252" => (0.0911379250149652 m, 0.8551460469135354 m)))
```
image

Using this PR:
`ax.scene.conversions` would still use the `dim_conversion` defined when creating the `Axis3`
```
ax.scene.conversions[1] = Makie.UnitfulConversion(Observable{Any}(m), false, Observable(true), Dict{String, Tuple{Any, Any}}())
```
image

## Type of change

[...]

PR #5047, opened 2025-06-10
Enable dim_converts for Axis3

SimonDanisch:
Continuation of #4964  with reference test


fhagemann:
Thanks so much for taking this over :pray:

PR #5063, opened 2025-06-13
force second resolve compilation

ffreyer:
Helps with interaction lag, but maybe not the way to go

Note that this is based against #5052, i.e. compare with:
GLMakie


SimonDanisch:
Closed in favor of: 
https://github.com/MakieOrg/Makie.jl/pull/5093

PR #5066, opened 2025-06-13
a few improvements for inspector hover task

SimonDanisch:
This closes the loop more reliably


ffreyer:
This causes errors on hover (which dont trigger on master)
```julia
f,a,p = scatter(rand(10))
DataInspector(f)
display(f)
GLMakie.closeall()
f
```

PR #5067, opened 2025-06-13
Improve performance a bit

SimonDanisch:
* setproperty! is faster in batches for ComputeGraph
* RelocatableFolders is _really_ slow for joinpath and friends  (serialize_scene for 40 figures goes from 4s to 0.4s with these changes).


src/compute-plots.jl
```diff
@@ -539,14 +539,12 @@ function add_attributes!(::Type{T}, attr, kwargs) where {T <: Plot}
...
-# function gscatter end
-
-# const GScatter{ARGS} = Scatter{gscatter, ARGS}
-
 function add_theme!(::Type{T}, kw, gattr::ComputeGraph, scene::Scene) where {T <: Plot}
     plot_attr = plot_attributes(scene, T)
     scene_theme = theme(scene)
     plot_scene_theme = get(scene_theme, plotsym(T), (;))
+
+    updates = Dict{Symbol, Any}()
```
asinghvi17: Should this be a LittleDict?  Or is that not so relevant?

SimonDanisch: I think that might actually be worse! Not sure if worth a try regardless;)

ffreyer: We could also generalize a bit and allow `Vector{Pair}` in `update!()`

PR #5077, opened 2025-06-15
Cleanup some TODO comments

ffreyer:
This mostly removes outdated comments, but also fixes a couple easy things. See #5072 for an overview of the remaining TODO comments


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -9,20 +9,6 @@ deref(x) = x
...
-    result = Expr(:tuple)
-    for name in names
-        qn = QuoteNode(name)
-        expr = Expr(:(=), name, :(($(qn) === field ? value : getfield(nt, $(qn)))::fieldtype(T, $(qn))))
-        push!(result.args, expr)
-    end
-    return result
-end
-=#
-
```
ffreyer, line 25: I'm not sure what this is about - is this fine to remove?

GLMakie/test/unit_tests.jl
```diff
@@ -19,9 +19,6 @@ end
     display(screen, scatter(1:4))
     @test length(cache.shader_cache) == 18
     @test length(cache.template_cache) == 18
-    # TODO, change from 11 -> 12
-    # This comes from having two versions of mesh, with transparency=true/false leading to different shaders for fragment_output
-    # Why did this change?
```
ffreyer, line 24: These used to be at 11 so I think this comment is outdated

Makie/src/compute-plots.jl
```diff
@@ -639,7 +636,6 @@ function Plot{Func}(user_args::Tuple, user_attributes::Dict) where {Func}
 end
 
 function get_plot_position(scene::Scene, plot::Plot)
-    # TODO, this may not reproduce the exact same cycle index as on master
```
ffreyer, line 642: I'm guessing this is resolved since tests pass and docs look fine?

Btw can we rename this function/attribute? It seems easy to confuse with `plot.position`

Makie/src/compute-plots.jl
```diff
@@ -682,7 +678,6 @@ function connect_plot!(parent::SceneLike, plot::Plot{Func}) where {Func}
         end
     end
 
-    # TODO, do this for recipes?
```
ffreyer, line 685: This function (`connect_plot!()`) runs for recipes now, right? So this is outdated?

Makie/src/utilities/texture_atlas.jl
```diff
@@ -717,7 +717,7 @@ function get_glyph_sdf(atlas::TextureAtlas, hash::UInt32)
 end
 
 function glyph_boundingbox(::BezierPath, ::Makie.NativeFont)
-    # TODO, implement this
+    # TODO:, implement this
     # Main blocker is the JS side since this is a bit more complicated.
     return (Vec2f(0), Vec2f(0))
 end
```
ffreyer, line 718: Iirc this was added for WGLMakie but wasn't needed? Can we remove this? Or mark this as a dummy method if it's needed as one?

Makie/precompile/shared-precompile.jl
```diff
@@ -95,13 +95,13 @@ end
...
-    # Makie.PolarAxis(f)
+    Makie.PolarAxis(f)
     Makie.IntervalSlider(f)
     lines!(ax, 1:2, label = "Line")
-    # Makie.Legend(f, ax, "Default", nbanks = 2) TODO
+    Makie.Legend(f, ax, "Default", nbanks = 2)
     Makie.Box(f)
     Makie.Label(f)
-    # Makie.Textbox(f)
+    Makie.Textbox(f)
```
ffreyer, line 104: using time probably increased due to bringing these back

Makie/src/utilities/texture_atlas.jl
SimonDanisch: No this is needed.
It's called here:
image
And the result won't be used - but should be as soon as we have time to implement it.

Makie/src/compute-plots.jl
SimonDanisch: yes

ComputePipeline/src/ComputePipeline.jl
SimonDanisch: Yes

PR #5079, opened 2025-06-16
use map! more instead of register_computation!

SimonDanisch:


asinghvi17:
Ah nice!  This is pretty great.

PR #5083, opened 2025-06-16
Update mesh.md arrows has been deprecated

caesquerre:
It is throwing an error because arrows has been deprecated,  
```
arrows!(ax, vertices, p.converted[][1].normal, lengthscale = 0.1, arrowsize = Vec3f(0.05, 0.05, 0.1), color = :orange)
ERROR: MethodError: no method matching getindex(::Observable{GeometryBasics.Mesh{…}}, ::Int64)
The function `getindex` exists, but no method is defined for this combination of argument types."
```

arrows3d! works fine.
Removed kwd arrowsize as it is also deprecated

# Description

Fixes # (issue)

`arrows!` is deprecated, I replaced by arrows3d! which works and deleted kwd arrowsize as it is also deprecated.

## Type of change

## Checklist

### DIFF
--- a/docs/src/reference/plots/mesh.md
+++ b/docs/src/reference/plots/mesh.md
@@ -41,12 +41,12 @@ colors = [:red, :green, :blue]
 f = Figure(size = (800, 400))
 ax = LScene(f[1,1], show_axis = false)
 p = mesh!(ax, vertices, faces1, color = colors)
-arrows!(ax, vertices, p.converted[][1].normal, lengthscale = 0.1, arrowsize = Vec3f(0.05, 0.05, 0.1), color = :orange)
+arrows3d!(ax, vertices, p.converted[:][1].val.vertex_attributes.normal, lengthscale = 0.1, color = :orange)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Looks like a wrong fix to me.
@ffreyer, we tried to not break the current arrow API, so I guess there also goes something wrong?

ffreyer:
This doesn't error in either docs build. The getindex error is because with the compute refactor `plot.converted[idx][]` changed to `plot.converted[][idx]`, i.e. running 0.24 code with Makie 0.23

caesquerre:
Sorry, I made a mistake as ffreyer noted the error is generated by: 
`p.converted[][1].normal`, if it is changed to  `p.converted[:][1].val.vertex_attributes.normal` the code works.
The deprecation of `arrows!` only generates a warning message.

ffreyer:
I think we should still fix the warnings. I'll probably push some changes later to update the code for 0.24

SimonDanisch:
I'm going to close this, since we'll look into it! Thank you for bringing this to our attention with a PR!

PR #5088, opened 2025-06-17
change patchcolor to opaque colors

ffreyer:
Since it's part of the 0.24 TODO list...

### DIFF
--- a/Makie/src/theming.jl
+++ b/Makie/src/theming.jl
@@ -17,7 +17,7 @@ end
 
 const DEFAULT_PALETTES = Attributes(
     color = wong_colors(1),
-    patchcolor = wong_colors(0.8),
+    patchcolor = map(c -> lerp(RGBAf(1,1,1,1), c, 0.8), wong_colors(1)),
     marker = [:circle, :utriangle, :cross, :rect, :diamond, :dtriangle, :pentagon, :xcross],
     linestyle = [nothing, :dash, :dot, :dashdot, :dashdotdot],
     side = [:left, :right]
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Or would it make sense to just set these to the same as the normal palette colors? I'm not sure

Makie/src/theming.jl
```diff
@@ -17,7 +17,7 @@ end
 
 const DEFAULT_PALETTES = Attributes(
     color = wong_colors(1),
-    patchcolor = wong_colors(0.8),
+    patchcolor = map(c -> lerp(RGBAf(1,1,1,1), c, 0.8), wong_colors(1)),
```
SimonDanisch: wong_colors(1) seems better?

ffreyer:
I think it's good to keep patchcolors lighter. I could see them becoming backgrounds to something with `color`, like a band behind data showing the spread. Having different colors would help there. I also think it's more visually pleasing for big patches to have less contrast. And less breaking, since the colors are closer to master

jkrumbiegel:
Can we make them dependent on the backgroundcolor of the theme and the color palette? Blend them by default? Then it would work better for dark theme

asinghvi17:
What's the rationale for making patchcolors opaque?

ffreyer:
For me it's about avoiding 3D rendering issues. 3D mesh plots just looks bad with that default. I'm not sure if we had other reasons for it

jkrumbiegel:
I kind of got over half transparent barplots as well, looks weird if the grid lines shine through.

PR #5096, opened 2025-06-18
Try lower precision for transformation matrix to fix spurious giant Numbers

SimonDanisch:
Hopefully fixes:
image

### DIFF
--- a/Makie/src/utilities/texture_atlas.jl
+++ b/Makie/src/utilities/texture_atlas.jl
@@ -375,7 +375,7 @@ end
 
 function render(atlas::TextureAtlas, (glyph_index, _font)::Tuple{UInt64, NativeFont})
     # We copy fonts for rendering to avoid issues with Cairo mutating the font matrix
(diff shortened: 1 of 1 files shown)


asinghvi17:
🤣 ok, I have never seen this brand of Cairo insanity before...

do we have an mwe to test?  Or is this not easily replicable?

SimonDanisch:
>do we have an mwe to test? Or is this not easily replicable?

It only reproduces after running the full doc build...
After that it's at least stable to reproduce.

jkrumbiegel:
Does something in the docs maybe turn on fast math or so? I read that this can poison a whole process and cause numerical issues

asinghvi17:
Yeah any C library that is built with fast math will infect the rest iirc

ffreyer:
This can be reproduced locally with `julia makedocs.jl` (on linux) with just
```
axis3.md
button.md
colorbar.md
```
active. The plots will end up in "docs/build/.documenter/reference/blocks". I still have no idea what's wrong

PR #5107, opened 2025-06-23
fix readme

SimonDanisch:


asinghvi17:
There were also a bunch of things in assetpath.  Should we change those to lazy artifacts?  Like brain.nii.gz, etc.

SimonDanisch:
They are! As mentioned in the changelog, they're now only reachable via `Makie.assetpath` or `Makie.loadasset`, which was actually already the preferred way before that ;)

PR #5109, opened 2025-06-23
Small fixes

SimonDanisch:
* Fixes forwarding attributes in recipes as in https://github.com/thchr/Brillouin.jl/pull/40#issuecomment-2994280000 (@kbarros, @thchr)
* Makes inserting into a closed scene/session, as it can be the case for async tasks, not error anymore


thchr:
Thanks for making this work again 🙏

PR #5112, opened 2025-06-23
Fix relocatability CI run for (W)GLMakie

SimonDanisch:
Lets face it, these don't run anymore.
It's a bit of a riddle why they stopped working, but I assume OOM, due to a larger compilation workload.
I tried them locally on linux, where everything works fine,  but it did require quite a lot of ram (19gb), which made me think it's RAM related.
The tricks to cut down on RAM usage with env vars and disabling precompile workload was not successful, which is a bit weird, considering that it was working well some weeks ago, and I dont think our ram usage can have grown by that much.

### DIFF
--- a/.github/workflows/relocatability.yml
+++ b/.github/workflows/relocatability.yml
@@ -30,14 +30,14 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.11'
+          - '1.10'
         os:
           - ubuntu-latest
         arch:
           - x64
         backend:
-          #- GLMakie
-          #- WGLMakie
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Maybe a 1.11 change?

SimonDanisch:
Huh, 1.10 does indeed use much less ram and is quite a bit faster:
1.10: `[01m:05s] PackageCompiler: compiling incremental system image`
1.11: `[02m:16s] PackageCompiler: compiling incremental system image`

SimonDanisch:
Welp, that was too easy?!

PR #5117, opened 2025-06-24
dont convert dicts to attributes

SimonDanisch:
I think dicts are too general and are used for other attributes (E.g. in OSMMakie).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [0.24.1] - 2025-06-24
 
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
with that justification, isn't it easier to only allow `Attributes` for attributes then? NamedTuples seem as useful as Dicts

SimonDanisch:
This is rather for backwards compatibility... I'm currently thinking about a nice version on how to properly replace this.
The dynamic empty named tuple is also not what I'd like here.
Maybe `line_kw = defaults(Lines)` should be the official replacement going forward, and properly check and document those instead of just having a dynamic input of any kind of keyword argument.

PR #5119, opened 2025-06-24
Some more fixes for patch release

SimonDanisch:
-  fix for cycle in event handling. Somehow the priority set for adding an observable to a compute graph managed to trigger an event loop for e.g. `events.keyboardbutton`
- Not polling plots from invisible scenes anymore, skipping costly computations and updates.
- Hiding Blocks during constructions. With the above, this actually fixes a bug where creating a new Axis in an already displayed scene would get a text size missmatch error. This is the poor mans version of actually porting Axis to the new compute-graph (which would fix the error as well), but this is a good thing even after the compute-graph refactor.
- Fix a bug with `heatmap(Resampler(...))`, where it could get into a state where the limits where nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## [0.24.1] - 2025-06-24
 
+- Don't pull plots from invisible scenes and hide Blocks during construction [#5119](https://github.com/MakieOrg/Makie.jl/pull/5119).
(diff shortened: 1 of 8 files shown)


SimonDanisch:
Hm, this does seem to fix the darker outlines in  "OldAxis + Surface": 
image
But somehow gets the problem from v0.23 back in the colorbar:
image
And it does seem to not completely resize the axis in scatter marker_offset 2D:
image

PR #5120, opened 2025-06-24
only run runic on merge and commit formatting changes

SimonDanisch:
I'm already tired of runic formatting churn :D 
Also, we either need to force outside contributors to run runic formatting, or we need to do that work, which is even more annoying since we need to check out their branch.
Even worse, we're running into CI usage queuing quite quickly, and with every third commit being runic fixes this makes the waiting for CI much worse.
This is directly committing when fixes are found, minimizing the time of people branching off from an unfixed master branch, and therefore reducing merge conflicts.
We considered letting the gh action make a PR, but that would need to run for an hour until we can merge it, eating up more CI resources and increasing merge conflict potential.

This means though, that before tagging we should wait for the master merge CI to turn green, since in theory runic could introduce errors. 
I can't really think of a better trade off, but happy to hear better ideas.
Github hook also isn't a viable solution, since this would not solve the outside contributor issue.

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -1,22 +1,59 @@
-name: Formatting
+name: Auto-format on Master Merge
+
 on:
   push:
     branches:
-      - 'master'
-      - 'release-'
-    tags:
-      - '*'
-  pull_request:
+      - master
+      - main
+
 jobs:
-  runic:
-    name: Runic formatting
+  auto-format:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: julia-actions/setup-julia@v2
+      - name: Checkout code
+        uses: actions/checkout@v4
         with:
-          version: '1'
-      - uses: julia-actions/cache@v2
-      - uses: fredrikekre/runic-action@v1
+          # Use a PAT or the default token with write permissions
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # Fetch full history to avoid shallow clone issues
+          fetch-depth: 0
+
+      - name: Set up Julia (if needed for runic)
+        uses: julia-actions/setup-julia@v2
         with:
-          version: '1.4'
+          version: '1.10'
+
+      - name: Install runic formatter
+        run: |
+          julia -e 'using Pkg; Pkg.add("Runic")'
+
+      - name: Run formatter
+        run: |
+          julia -e 'using Runic; exit(Runic.main(ARGS))' -- --inplace .
+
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Could we make it so formatting is the first ci step and everything else is contingent on it? And ci would only run if formatting succeeds, so on a formatting failure, only the fix commit would be created which then triggers a new round

SimonDanisch:
well that just fixes one concern, not the one that it puts work on us or outside contributors.
I'd prefer a solution that fixes that too.

jkrumbiegel:
I meant in addition to your suggestions

SimonDanisch:
Ah I would disable the runic check for PRs?

hz-xiaxz:
Not my place to comment here, but in my using experience, it might be more annoying to get CI blocked by formatting problem. There's always some time when you did a bunch of important works and were eager to see how it works, then you pushed and took a cup of coffee, back and see, oops, a spelling problem. Personally I would be upset by that experinece. 

Another idea here is, is it possible to run runic periodically like a compathelper. So one can fix the spelling and formatting once they would like to, instead of always letting these problems block the way?

hz-xiaxz:
> Not my place to comment here, but in my using experience, it might be more annoying to get CI blocked by formatting problem. There's always some time when you did a bunch of important works and were eager to see how it works, then you pushed and took a cup of coffee, back and see, oops, a spelling problem. Personally I would be upset by that experinece.
> 
> Another idea here is, is it possible to run runic periodically like a compathelper. So one can fix the spelling and formatting once they would like to, instead of always letting these problems block the way?

this approach would definitely ruin the blame, though 🤣

asinghvi17:
You can put certain commit patterns in a file so git knows to avoid them, so blame is not an issue

hz-xiaxz:
> Not my place to comment here, but in my using experience, it might be more annoying to get CI blocked by formatting problem. There's always some time when you did a bunch of important works and were eager to see how it works, then you pushed and took a cup of coffee, back and see, oops, a spelling problem. Personally I would be upset by that experinece.
> 
> Another idea here is, is it possible to run runic periodically like a compathelper. So one can fix the spelling and formatting once they would like to, instead of always letting these problems block the way?

As a follow up, I found the ITensorAction workflow, https://github.com/ITensor/ITensorActions, which has workflow files for "one for simply verifying the formatting, one for additionally applying suggested changes, and one that makes a PR to the repository formatting the code in the repository." 
If their approach is acceptable, it would be pretty easy to adapt these to Makie ecosystem.

Issue #5114, opened 2025-06-23
Plots that need to be updated to use the new ComputeGraph


ffreyer:
This is a quick overview of which recipes still need to be translated to use the ComputeGraph (Makie 0.24)

Some recipes more or less just require replacing `map/lift/@lift/on/onany` with `map!`. The 0.24 blog post has/will have some instructions on how to do this, so it should be easy for anyone to do. If you want to contribute something relatively small feel free to update one (or more) and make a pull request. A rough list of easy plots/recipes:
- [x] arc #5165
- [x] band #5165
- [x] pie #5165
- [x] poly #5121
- [x] stairs #5165
- [x] stem #5165
- [x] tooltip #5165
- [x] wireframe #5165
- [x] density #5184
- [x] qqplot (distributions.jl) #5165
- [x] stephist #5184
- [x] hist #5184
- ~~timeseries~~ needs to rely on Observables so time stamps are created at set-time, not resolve-time
- [ ] series

Some recipes still use `register_computation!()` when they could use `map!()`. This should be a fairly easy cleanup/replacement as well.
- [x] tricontourf #5484 
- [x] spy #5165
- [x] scatterlines #5165

Some recipes may require some more refactoring, like breaking up an `onany` that updates multiple observables. They should still be fairly easy to translate (though I haven't checked in depth). `contourf` or `errorbars` might be an instructive example.
- [x] barplot #5412
- [x] contour #5121
- [x] triplot #5121
- [x] waterfall #5184
- [x] boxplot #5184
- [x] crossbar #5184
- [x] hexbin #5121
- [x] violin #5184

Some recipes apply transformations and projections. These should probably be delayed a bit, as we can/should extend what's happening for primitive plots to recipe plots here
- ~~hlines, vlines, hspan, vspan (+ limit tracking)~~
- [x] annotation (+ bbox handling) #5448
- [x] bracket #5121
- [x] streamplot  #5121
- [x] textlabel #5121
- [x] voronoiplot #5121
- [x] arrows #5121

Other:
- [ ] volumeslices (how do we want to handle the transformation system)

Not sure if done:
- [x] axis3d (maybe needs full rework?) #5484
- [x] datashader
- [ ] rainclouds

PR #5121, opened 2025-06-25
Recipe projections

ffreyer:
# Description

As noted in #5114 some recipes need to project positions before they can be processed an passed on to other plots. This pr aims to simplify this by:
- [x] making `register_positions_transformed/_f32c!` usable in recipes, i.e. with variable input and output names (if it's not already) 
- [x] adding a generic `register_camera_matrix!(plot, input_space, output_space)`
- [x] adding a generic `register_positions_in_space()` to transform positions from some space to some other, with or without transformations
- [x] maybe adding simplified helpers like `register_markerspace_positions()`

~~This pr will probably also touch backends to make use of these functions, i.e. cut out some redundancy.~~ I'm planning on leaving backend code alone for now, and do cleanup in another pr. This pr doesn't need to get delayed for that. I don't plan on updating `Makie.project()` and similar functions to keep this pr non breaking.

Things to figure out:
- Should `register_camera_matrix!()` replace the current system? (I.e. should camera nodes be added on demand for all plots?)
- How should camera matrices be named? Should we keep the current plot.preprojection etc, or switch to e.g. `space_to_markerspace` to reduce the need for explicit renaming & checks?
- Should we add something to bypass `is_same` for cases like `camera_trigger`? (I.e. for nodes used as triggers?)

TODO:
- [x] update recipes:
    - ~~h/vlines/span~~ Skipped, mostly unrelated to this pr
    - ~~annotation~~ Rather complicated. Probably better to separate from this
    - [x] arrows
    - [x] bracket
    - [x] contour
    - [x] poly
    - [x] streamplot
    - [x] text
    - [x] textlabel
    - [x] triplot
    - [x] voronoiplot
    - [x] hexbin
- [x] include clip_planes 
- ~~maybe refactor CairoMakie to rely on these function~~ Probably going to do this in another pr
- [x] fix performance
- [x] docs
- [x] maybe tests (since these functions are used in recipes they already get tested to some extent)

## Type of change

## Checklist


ffreyer:
We should probably discuss naming and structure for this, since it's user facing. At the moment projection computations are effectively:
- `register_projected_positions!()` which goes from converted -> `output_space`
  - `register_position_transforms!()` which does:
    - `register_positions_transformed()` which applies `transform_func`
    - `register_positions_transformed_f32c()` which applies `float32convert` and generates `model_f32c`
  - `register_positions_projected!()` which applies the model matrix and projects from `input_space` -> `output_space` and optionally `clip_planes`

`register_projected_positions!()` has:
- `apply_transform = input_space === :space` which enables/disables `transform_func`, `float32convert` and `model` application
- `apply_clip_planes = false` which enables/disables `clip_planes` application
- `yflip = false` which adds a y flip for CairoMakie in pixel space

If you wanted to just apply `transform_func` you'd have to go with `register_positions_transformed!()`. I'm wondering if we should lean more into this, i.e. "pick the right function", or fill out the options in `register_projected_positions!()` more, i.e. "pick the right options".

I'm also still unhappy with naming. `register_positions_projected!()` and `register_projected_positions!()` are confusing. One applies all the matrix related transformations (i.e. projections), the other does the whole transformation-projection pipeline. I originally had `register_positions_in_space!()` for the latter, maybe that is better?

Makie/src/compute-plots.jl
```diff
@@ -285,12 +343,34 @@ function register_positions_transformed_f32c!(attr)
...
 end
 
+function register_model_f32c!(attr)
+    map!(attr, [:model, :f32c, :space], :model_f32c) do model, f32c, space
+        trans, scale = decompose_translation_scale_matrix(model)
+
+        # is_rot_free = is_translation_scale_matrix(model)
+        if !is_data_space(space) || isnothing(f32c) || (is_identity_transform(f32c) && is_float_safe(scale, trans))
+            return Mat4f(model)
+        elseif false # is_identity_transform(f32c) && !is_float_safe(scale, trans)
```
SimonDanisch, line 359: Why do we leave the false paths in here?

SimonDanisch: Also there seem to be 3 other functions with the same pattern?
E.g. : https://github.com/MakieOrg/Makie.jl/blob/d2544bf6c1bb6793fe5c0078b6d7ed33a790190d/Makie/src/backend-functionality.jl#L59-L97

Shouldn't this be handled by one project function?

Makie/src/utilities/utilities.jl
```diff
@@ -230,6 +231,17 @@ The length of an attribute is determined with `attr_broadcast_length` and elemen
...
+ref_wrap_scalar_parameter(x) = is_vector_attribute(x) ? x : Ref(x)
+
+"""
+    makie_broadcast(f, args...)
+
+Like `broadcast` but treats types that Makie considers scalars as such, e.g.
+`VecTypes` (Point, Vec) or Quaternions. `is_vector_attribute` controls how
+arguments are considered.
+"""
+@inline makie_broadcast(f, args...) = broadcast(f, ref_wrap_scalar_parameter.(args)...)
```
SimonDanisch, line 243: I don't think Makie broadcast is very descriptive. 
Dont have a better name handy immediately, but maybe worth to think about it.

SimonDanisch (changes requested):
This looks great! 
I would like to consider a few names and make sure the `else if false` makes sense and isn't duplicated 3 times before merging.

Makie/src/camera/projection_math.jl
```diff
@@ -363,31 +363,92 @@ function _to_world(
...
+"""
+    _project([TargetType], matrix::Mat4, positions[, dim4 = 1])
+
+Projects one or multiple `positions` using the given `matrix`.
+
+`dim4` controls the fill value for 4th component of each positions (if they are
+not 4D positions). `TargetType` controls the final output type. If the target
+type is not 4 dimensional, each output position will be divided by `p[4]`.
+"""
+@inline _project(::Type{OT}, M::Mat4, p) where {OT} = _project(OT, M, p, 1)
```
SimonDanisch, line 376: Do we really want to have them internal with `_`? This seems pretty convenient for anyone needing to do their own projections?

Makie/src/compute-plots.jl
ffreyer: > Why do we leave the false paths in here?

We left them there in the refactor. I just fought the formatter here so it doesn't destroy alignment. Whether we bring the fast paths back or not isn't part of this pr for me.

> Also there seem to be 3 other functions with the same pattern?

I'm only aware of 3 total. The model -> model_f32c one, which you originally commented on, the positions -> positions_f32c one which is above it, and the (x, y) -> (x_f32c, y_f32c) one, which you linked. They all do the same thing, but act on different things.

I didn't functionally change anything here, I just extracted the model part. That was necessary to allow multiple positions to go through the pipeline. Without it, i.e. with a `map!(attr, [...], [:model_f32c, output_name])` you'd be creating `model_f32c` twice with different inputs which errored even before my pr.

Makie/src/utilities/utilities.jl
ffreyer: I agree. I would've liked to name it similarly to `broadcast_foreach` but since it cuts out the `foreach` there isn't anything left that could be a name

Makie/src/camera/projection_math.jl
ffreyer: I started with `_project` in the refactor so I didn't have to worry about breaking 
https://github.com/MakieOrg/Makie.jl/blob/d2544bf6c1bb6793fe5c0078b6d7ed33a790190d/Makie/src/camera/projection_math.jl#L454-L459 
and just kind of continued here. Replacing `project` would only break the return type in some cases as far as I can tell. `project` uses the input type, `_project` defaults to Point3{T}. (Which I think is better because an input being 2D doesn't mean it won't get a z component, e.g. from model)

Generally I was planning to do some breaking cleanup for the project stack in another pr. I don't want this to get delayed by any breaking changes. (And another for backend cleanup which would hopefully not be breaking?)

ffreyer:
It would be nice to get a review from more of a user perspective on `register_projected_positions!()`. Maybe @asinghvi17 or @jkrumbiegel could do that?

asinghvi17:
Hmm I could play around with this tomorrow, SwarmMakie could use that probably

ffreyer:
Just fyi, we're looking to merge and release this and #5145 this week

Makie/src/utilities/projection_utils.jl
```diff
@@ -0,0 +1,257 @@
...
+transform_func applied) and float32-converted positions (i.e. with float32convert
+applied).
+
+Optionally `output_type` can be set to control the element type of projected
+positions. (4D points will not be w-normalized, 1D - 3D points will be. This is
+to allow clip space clipping to happen elsewhere.)
+
+## Keyword Arguments:
+- `input_space = :space` sets the input space. Can be `:space` or `:markerspace` to refer to those plot attributes.
+- `output_space = :pixel` sets the output space. Can be `:space` or `:markerspace` to refer to those plot attributes.
```
jkrumbiegel: it's `:pixel` but can be `:space` or `:markerspace`?

CairoMakie/src/scatter.jl
```diff
@@ -4,7 +4,11 @@ function draw_atomic(scene::Scene, screen::Screen, plot::Scatter)
     Makie.add_computation!(attr, scene, Val(:meshscatter_f32c_scale))
     cairo_unclipped_indices!(attr)
     Makie.compute_colors!(attr)
-    project_to_markerspace!(attr)
+    Makie.register_positions_projected!(
```
jkrumbiegel, line 7: Why does a backend mutate the compute graph? Does that have side effects when switching back and forth between CairoMakie and GLMakie?

jkrumbiegel (commented):
in principle the new function seems sensible to me, maybe a little complicated with the newly generated names. Maybe the docstring could have an example which lists the newly created nodes with their names

CairoMakie/src/scatter.jl
SimonDanisch: It's thought to be just another "conversion".
We're prefixing backend specific conversions and try to re-use most conversions that aren't backend specific.

ffreyer: All these functions mutate because they add compute nodes and edges (if they aren't there already). Some computations overlap between backends and get reused, others are backend specific and do not. This is handled based on names.
When adding a computation, we check that either none of the outputs already exists or that all of them exist with a shared parent edge that has the same set of inputs as requested. So if you try to register different computations for the same output it will generally error. (We used to also check functions but currently that's behind the debug mode.)

The new function also relies on this collision detection. The different combined projection matrices all have fixed unique names so that you get collision errors if you try to register the same output with different settings, e.g.
```julia
Makie.register_positions_projected!(plot, input_name = :position, output_name = :projected, output_space = :pixel)
Makie.register_positions_projected!(plot, input_name = :position, output_name = :projected, output_space = :clip)
```
The auto generated names (other than output_name) also match when they can be reused. (Which means they will be reused.) But now that I think about it, `transformed_name` and `transformed_f32c_name` should probably not be settable to make sure this is the case.

Makie/src/utilities/projection_utils.jl
ffreyer: Updated the docs to mention static spaces to

ffreyer:
"FigureSpec/step-3" errored due to alignment issues. I'm pretty sure I've seen this before, randomly, so I assume that's so some of synchronization issue irrelevant to this pr

PR #5145, opened 2025-07-02
Consider transform_func when using axis_limits

### LINKED ISSUE #5129: `vlines` not spanning over the whole y-range in `log` scale in Makie@0.24
In `CairoMakie` and since `Makie@0.24`, `vlines` does not span over the whole range anymore when using `log` scale:
```julia
using CairoMakie
vlines([1,2,3], axis = (yscale = log10,))
```
`Makie@0.23.0`:
Image

`Makie@0.24.1`:
Image

Is this related to #5114, #5121 about missing recipe projections?

Where in the code might this arise from, so I might also have a look at this myself and see if I can help.

### LINKED ISSUE #5143: `vlines` are no longer draw correctly
CairoMakie 0.15.2

I had a figure that used to work fine that basically did:

```julia
let
	fig = Figure(size=(800, 800))
	ax = Axis(fig[1,1], xscale=log10, yscale=log10)
	ylims!(ax, 1e-20, 10e1)
	
	vlines!(ax, [sqrt(eps())], color="red")
	vlines!(ax, [cbrt(eps()/2)],  color="orange")
	fig
end
```

Image

I can achieve the same now with: 

```julia
let
	fig = Figure(size=(800, 800))
	ax = Axis(fig[1,1], xscale=log10, yscale=log10)
	ylims!(ax, 1e-20, 10e1)
	
	vlines!(ax, [sqrt(eps())], color="red", ymin=-1)
	vlines!(ax, [cbrt(eps()/2)],  color="orange", ymin=-1)
	fig
end
```

I think there is a confusion between the docs/expectation that vlines is using scene coordinates https://docs.makie.org/dev/reference/plots/vlines

and `ymin=0` not playing well with a log10 transform.

ffreyer:
# Description

Fixes #5143 and fixes #5129

## Type of change

## Checklist


ffreyer:
I think this fails because `poly` has not been updated yet. With poly having observables, the change in `transform_func` triggers in hvspan before the limit update, which means log10 may run on limits for identity. I am planning on updating poly in #5121, so I'll probably leave this until #5121 is done

PR #5154, opened 2025-07-04
fix issue with `default_attribute` for Attributes containing Attributes

simeonschaub:
Also discovered an issue with `propertynames` while writing tests for
this.

# Description

Fixes an error with `default_attribute` for Attributes containing Attributes

## Type of change

## Checklist

### DIFF
--- a/Makie/src/attributes.jl
+++ b/Makie/src/attributes.jl
@@ -243,5 +243,5 @@ function merge_attributes!(input::Attributes, theme::Attributes)
 end
 
 function Base.propertynames(x::Union{Attributes, AbstractPlot})
-    return (keys(x.attributes.inputs)...,)
+    return (keys(x.attributes)...,)
 end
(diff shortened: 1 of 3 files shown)


SimonDanisch:
Thank you, much appreciated!

simeonschaub:
Bump. Does this look like the right fix? Would be great if this could be released right away so I can transition my packages to 0.24

simeonschaub:
Bump :)

SimonDanisch:
Does this actually work for plots?

SimonDanisch:
Nope... I think we should separate those functions and add tests for both types.

simeonschaub:
How do I best construct an AbstractPlot? Any suggestions for a test?

SimonDanisch:
you could just do `f, ax, pl = scatter(1:4)` .
You can also directly construct it:
```julia
pl = Scatter((1:4,), Dict{Symbol, Any}())
```

SimonDanisch:
Thanks!

PR #5156, opened 2025-07-05
Fix toggle animation when being triggered programmatically 

### LINKED ISSUE #2785: Non-interactive toggle of Toggle?
Is there a way of programmatically altering a `Toggle`? For example, I have a GUI with a bunch of toggles and I'd like to disable all of them with a button press. Looking at the code it does not seem obvious how to change the appearance of the toggle when the `active` observable is altered. A minimal version would be something like this:
```julia
using GLMakie
fig = Figure()
t = Toggle(fig[1,1], active = true)
fig
```
Once drawn, modifying it does not change the Toggle and it remains "looking" active. Clicking it again after setting it to false does what it should (goes to true without any animation), though, so it is just a question of the updating the appearance.
```julia
t.active[] = false
```
As the appearance gets a bit out of sync but the GUI itself works as expected, it is hardly a critical issue... Thanks in advance!

tknopp:
# Description

Fixes #2785

Previously the animation was not triggered when changing toggle.activate[] programatically. This is now fixed.

- moved the animation into a dedicated function perform_toggle_animation
- call this function from onmouseleftclick and from the activate callback

## Type of change

## Checklist

not sure if this belongs to the changelog but I added one entry.


tknopp:
visual proof of the fix:

https://github.com/user-attachments/assets/106ce164-2be4-497d-85d7-fdac923275e4

Makie/src/makielayout/blocks/toggle.jl
```diff
@@ -109,10 +103,28 @@ function initialize_block!(t::Toggle)
...
+
+    onmouseleftclick(mouseevents) do event
+        if animating[]
+            return Consume(true)
+        end
+        animating[] = true
+
+        t.active[] = !t.active[]
+
+        perform_toggle_animation()
```
ffreyer: Can't the animation be skipped here so that t.active triggers it instead?

PR #5157, opened 2025-07-06
Support Interpolations v0.16

cgarling:
# Description
Support Interpolations.jl v0.16. Supersedes #4986 which now seems out of date given the file structure reorganization that occurred in the package.


SimonDanisch:
Thanks!

Issue #4404, opened 2024-09-24
Enhance the `DateTime`, `Date` and `Time` axis formatting


mathieu17g:
## Feature description
Extend the handling of `TimeType` type for axis ticks and labels by handling `Date` and `Time` and enable the use of a formatter specified as a `DateFormat` or any `Function` on ticks' values

Here are three light modifications in [`src/date-integration.jl`](https://github.com/MakieOrg/Makie.jl/blob/v0.21.11/src/dim-converts/dates-integration.jl) of v0.21.11, enabling those:

### 1. Add easy to define formatter with `DateFormat`

In [`src/date-integration.jl`](https://github.com/MakieOrg/Makie.jl/blob/v0.21.11/src/dim-converts/dates-integration.jl) add the following `get_ticklabels` functions:
```julia
# Default implementation for `get_ticklabels` with `DateFormat`.
# Used before DateTimeConversion is set to the axis via plots addition.
get_ticklabels(::DateFormat, values) = get_ticklabels(Automatic(), values)

function get_ticklabels(::Type{T}, df::DateFormat, values) where {T<:TimeType}
    return [Dates.format(number_to_date(T, v), df) for v in values]
end
```

### 2. Extend `get_ticks` for `DateTimeConversion` to use [PlotUtilsjl](https://github.com/JuliaPlots/PlotUtils.jl/tree/v1.4.1) function [`optimize_datetime_ticks`](https://github.com/JuliaPlots/PlotUtils.jl/blob/76ca7bee602ef04e864640ed9fbc98336732364b/src/ticks.jl#L560-L606) for `Date`  beyond `DateTime`. And use the `DateFormat` formatter or any formatter defined by a `Function` on ticks' values.

Add to function [`get_ticks(conversion::DateTimeConversion, args...)`](https://github.com/MakieOrg/Makie.jl/blob/c463049102e775d4c31c4b891e9ac32adbca9891/src/dim-converts/dates-integration.jl#L89-L115) 

A conversion of extrema values to DateTime corresponding values
```julia
<<< if T <: DateTime
>>> if T <: Union{Date,DateTime}
        if T == Date
            # Convert T float extrema values to DateTime corresponding float values
            vmin = date_to_number(DateTime, DateTime(number_to_date(T, vmin)))
            vmax = date_to_number(DateTime, DateTime(number_to_date(T, vmax)))
        end
```

[...]

Issue #3972, opened 2024-06-17
Apply `DateFormat` to `DateTime` axis ticks


haakon-e:
### Feature description

Is there a way to apply a custom `DateTime` format to axis ticks?
```julia
fmt = DateFormat("u YYYY")  # want to use this
t = Date(1995,1):Month(1):Date(2005,5)
ys = rand(length(t))
scatter(t, ys)
```
The above code formats datetimes as `yyyy-mm-dd` by default.
<img src="https://github.com/MakieOrg/Makie.jl/assets/45243236/6890e54e-f7e0-4b10-898e-2f6f0103b595" width=50% height=50%>

With the format, it should look something like this (here, manually created ticklabels):
<img src="https://github.com/MakieOrg/Makie.jl/assets/45243236/36c73a4f-b481-4ffa-982d-b0196e94d406" width=50% height=50%>

the code to achieve this would (presumably) look something like this:
```Julia
scatter(t, ys;
    axis = (;
        xtickformat = fmt,
        # or,
        xtickformat = dts -> Dates.format.(dts, fmt),
        # or?
        dims1_conversion = Makie.DateTimeConversion(fmt),
))
```

Similarly/alternatively: is there a way to pass a `DateTime` as ticks to the axis? e.g.
```julia
Axis(fig[1,1]; xticks = [Date(1995,5), Date(2000,1)])
```
-----
related to: #3959

nilshg:
I failed at fixing this problem, but have posted at least a very basic workaround which allows for passing ticks and formatting in simple cases in the linked issue #3959

Issue #3959, opened 2024-06-13
`xticks` is ignored with unit aware axes


nilshg:
```
scatter(rand(Date(2000):Day(1):Date(2001), 100), rand(100); axis = (; xticks = Date(2000):Month(3):Date(2001)))
```

will not change the `xticks` at all - it appears that the argument is completely ignored (even feeding in obvious nonsense does not yield errors, the plot simply remains unchanged).

jariji:
Related failure with units:

```jl
using CairoMakie, Unitful

julia> let x = rand(50)u"yr"
           hist(x; axis=(;xticks=(0:.1:1)u"yr"))
       end
ERROR: DimensionError:  and yr are not dimensionally compatible.
```

nilshg:
Turns out this is too complicated for me to address properly, but for anyone coming across this here's a monkey patch that at least makes it possible to pass `xticks` and `xtickformat` kwargs in the `axis` keyword for plotting functions.

```
import GLMakie.Makie as Mke
import GLMakie.Makie.get_ticks

function get_ticks(conversion::Mke.DateTimeConversion, ticks, scale, formatter, vmin, vmax)

    if scale != identity
        error("$(scale) scale not supported for DateTimeConversion")
    end
    T = conversion.type[]
    # When automatic, we haven't actually plotted anything yet, so no unit chosen
    # in that case, we can't really have any conversion
    T <: Mke.Automatic && return [], []

    if in((Date, DateTime))(eltype(ticks))
        tickformatter = formatter == Mke.MakieCore.Automatic() ? identity : formatter
        return Dates.value.(ticks), string.(tickformatter.(ticks))
    end

    if T <: DateTime 
        if ticks isa WilkinsonTicks
            k_min = ticks.k_min
            k_max = ticks.k_max
        else
            k_min = 2
            k_max = 3
        end
        conversion, dates = PlotUtils.optimize_datetime_ticks(vmin, vmax; k_min=k_min, k_max=k_max)
        return conversion, dates
    else
        # TODO implement proper ticks for Time Date
        tickvalues = Mke.get_tickvalues(formatter, scale, vmin, vmax)
        dates = Mke.number_to_date.(T, round.(Int64, tickvalues))
        return tickvalues, string.(dates)
    end
end
```

You'll have to adapt the first couple of lines to your backend. Note that for some reason this doesn't seem to work when building up plots like 
```
f = Figure()
ax1 = Axis(f[1, 1], xlabel = "Month")
[...]

PR #5159, opened 2025-07-07
Date ticks overhaul

jkrumbiegel:
This PR changes the way that ticks for datetime axes are located and formatted by default. The new tick formatting reduces the amount of redundant information by printing time on a separate line from date (if time is different from 0). This way, the date only has to be shown whenever it changes from tick to tick. Similarly, only the time component of the tick step size is usually shown, unless there's a step in the parent component. For example, seconds would be shown `:58`, `:59` but then as the minute changes, the full time is shown like `03:17:00` and then further `:01`, `:02`, etc.

Fixes https://github.com/MakieOrg/Makie.jl/issues/4404, fixes https://github.com/MakieOrg/Makie.jl/issues/3972, fixes https://github.com/MakieOrg/Makie.jl/issues/3959

## This PR

```julia
f = Figure(size = (600, 800))
data = cumsum(randn(12))
for (i, T) in enumerate([Year, Month, Day, Hour, Minute, Second, Millisecond])
    times = DateTime(2025, 06, 30, 17, 54, 56, 993) .+ T.(1:12)
    lines(f[i, 1], times, data)
    Label(f[i, 2], "$T", rotation = pi/2, font = :bold, tellheight = false)
end
f
```

image

## Master

image

This PR also adds a separate dispatch route for datetime ticks which fixes the bug that the ticks or the tick format could not be changed manually and were always hardcoded to the automatic tick finding algorithm. 

```julia
dates = Date(2025, 06, 30) .+ Day.(1:5)
data = cumsum(randn(5))

with_theme(Axis = (; yticksvisible = false, yticklabelsvisible = false, ygridvisible = false)) do
    f = Figure()
    lines(f[1, 1], dates, data)
    lines(f[2, 1], dates, data; axis = (; xticks = (dates[[1, 2, 3, 5]], ["A", "B", "C", "E"])))
    lines(f[3, 1], dates, data; axis = (; xtickformat = "e, d.m.yyyy"))
    f
end
```

image


jkrumbiegel:
One question for review is whether `get_datetime_ticks`, `get_datetime_tickvalues` and `get_datetime_ticklabels` are needed in addition to the existing `get_ticks`, `get_tickvalues` and `get_ticklabels`. The reason I added them is that the others work with float vmin vmax values and I assumed one could get ambiguities relatively easily if one suddenly started to dispatch on the vmin, vmax arguments, too. Currently, these are not used for dispatch as they are always float, some methods may annotate the types and some may not.

jkrumbiegel:
Ok I've decided to remove the date-specific functions after all and simply use dispatches on `vmin::DateTime, vmax::DateTime`. That seemed to work ok as well, let's see if CI surfaces any ambiguities.

I've also left in the `scale` even though it probably doesn't make sense to use it with a `DateTime` axis. But this way it's consistent with the other functions.

Makie/src/dim-converts/dates-integration.jl
```diff
@@ -73,39 +75,430 @@ function convert_dim_value(conversion::DateTimeConversion, attr, values, previou
...
+        collect(ticks_hour), :hour
+    elseif i == 5
+        collect(ticks_minute), :minute
+    elseif i == 6
+        collect(ticks_second), :second
+    elseif i == 7
+        collect(ticks_millisecond), :millisecond
+    else
+        error("unreachable reached")
+    end
```
ffreyer, line 281: I think it would be a bit cleaner to have something like this:
```julia
ideal_ticks, ideal_cost = _ticks(Year, start, stop, k_ideal)
ideal_type = Year
for type in [Month, Day, ...]
    ticks, cost = _ticks(type, start, stop, k_ideal)
    if cost < ideal_cost
        ideal_cost = cost
        ideal_ticks = ticks
        ideal_type = type
    end
end
return ideal_ticks, ideal_type
```
and swap out the Symbols in the functions below. Doing `type === Year` shouldn't be any more expensive than checking symbols. (Though `==` and `in` are)

If the cost is well behaved, i.e. drops until it reaches a minimum and then rises again, and if the cost was always finite (not Inf), then this could shortcut with `elseif cost > ideal_cost; break`.

Makie/src/makielayout/types.jl
```diff
@@ -230,6 +230,39 @@ struct KeysEvent
...
+- `WilkinsonTicks`, the default tick finder for linear ticks
+- `LinearTicks`, an alternative tick finder for linear ticks
+- `LogTicks`, a wrapper that applies any other wrapped tick finder on log-transformed values
+- `MultiplesTicks`, for finding ticks at multiples of a given value, such as `π`
+
+DateTime:
+- A vector of `DateTime`s
+- A tuple with two vectors `(datetimes, labels)`
+- `DateTimeTicks`, the default tick finder for datetime ticks
+"""
```
ffreyer: Should this also include `AngularTicks`? They try use steps that are fitting for angles, e.g. 90° instead of 100 and switch to linear ticks if the ticks span less than ~3°. Maybe that's also useful for Axis in some niche cases?

asinghvi17:
```suggestion
tickinfo = """
The following is a list of common tick objects.

Numeric:
- A vector of numbers
- A tuple with two vectors `(numbers, labels)` where `labels` can be any objects that `text` can handle.
- [`WilkinsonTicks`](@ref), the default tick finder for linear ticks
- [`LinearTicks`](@ref), an alternative tick finder for linear ticks
- [`LogTicks`](@ref), a wrapper that applies any other wrapped tick finder on log-transformed values
- [`MultiplesTicks`](@ref), for finding ticks at multiples of a given value, such as `π`
- [`AngularTicks`](@ref), for finding tick steps that are fitting for angles with base 12 numbering (90° instead of 100)

DateTime:
- A vector of `DateTime`s
- A tuple with two vectors `(datetimes, labels)`
- [`DateTimeTicks`](@ref), the default tick finder for datetime ticks
"""
```

Makie/src/dim-converts/dates-integration.jl
jkrumbiegel: I did this split because of type stability, `start:Minute(1):stop` is a different type than `start:Second(1):stop`. As I don't want to `collect` early, to avoid allocations just for comparisons, I've split these all out into separate variables. The return type is then `Vector{DateTime}, Symbol` for all of them.

nilshg:
This is amazing, thank you so much!

Issue #5085, opened 2025-06-17
volume shows objects from behind the camera


hsgg:
Versions: Makie v0.23.0, GLMakie v0.12.0

Can be reproduced with a fresh environment, `--startup-file=no`.

I see the same on MacOS M1, Fedora Intel integrated GPU, and Arch Linux AMD Radeon 7600 XT.

Inside the volume, objects from behind are shown.
Here is a minimal example:
```
using GLMakie
vol = rand(1000,50,50);
fig, scene, _ = volume(vol)

controls = cameracontrols(scene)

# position the camera at the long end:
controls.eyeposition[] = Vec3f(1300, 25, 25)
controls.lookat[] = Point3f(1200, 25, 25)
update_cam!(scene.scene)
```
Next, press `w` to enter the volume. Once inside notice how elements from behind seem to get mirrored across the direction of view, creating an effect where part of the volume is moving past you from behind:

https://github.com/user-attachments/assets/ff75f314-378f-456d-a1cc-9528be545219

This was previously discussed [here](https://discourse.julialang.org/t/how-to-make-fly-through-animations/129775/5?u=hsgg).

ffreyer:
I'm guessing the rays we produce in 
https://github.com/MakieOrg/Makie.jl/blob/f6544877b32d2530543c354e92b0c00fcddb076e/GLMakie/assets/shader/volume.frag#L348-L364
are only clamped to the volume box, and not also to the view box

hsgg:
I have managed to produce a fix for this in my fork: https://github.com/hsgg/Makie.jl/commit/0b95b5dc294e431b291ed2ea5e1d82fba4c9b369. This is my first attempt at GLSL programming, and first time contributing to Makie, so feedback is more than welcome.

I have not made a PR, yet, because I have questions:
1. How to/Should I add a unit test?
2. No changelog entry.

SimonDanisch:
Great! :) Feel free to make a PR, that's easier to review and help you with reference tests etc.

PR #5164, opened 2025-07-07
GLMakie: fix volume.frag showing objects from behind the camera

hsgg:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/5085.

The volume.frag shader was showing objects that are behind the camera with the `:mip` algorithm. This fixes this.

## Type of change

## Checklist


ffreyer:
In model space the volume is always in a 0..1 box, and you already know the far/back position of the box and the position of the camera in the same space. So you check if the camera is inside the box (0 < x/y/z < 1) and if it is you use those positions as is. Otherwise the same code as before finds the near/front position. That seems reasonable to me.

TODO:
- add a reference image test with the camera inside the volume
- copy those changes over to WGLMakie https://github.com/MakieOrg/Makie.jl/blob/master/WGLMakie/assets/volume.frag
- add a changelog entry https://github.com/MakieOrg/Makie.jl/blob/master/CHANGELOG.md

To prevent the camera updating on display you need to do something like this:
https://github.com/MakieOrg/Makie.jl/blob/0332186a6c0420f8010dfb7cedafd325ad9b6bc3/ReferenceTests/src/tests/examples3d.jl#L9-L10

ffreyer:
Test failures are significant. To me it looks like start and stop swapped roles
image

hsgg:
> Test failures are significant. To me it looks like start and stop swapped roles

Ah, yes, I should have caught that. The mip algorithm doesn't care, but not surprised that the others do. I've pushed a change to fix that. I still get failures locally (though fewer) running as `] test GLMakie` on MacOS. Not sure how to assess those, yet.

To create a reference image, I have mostly been testing dynamically by moving around to assess where objects are, in a volume with random values. I will try to come up with something better, but would that be acceptable otherwise?

ffreyer:
I copied the changes over to WGLMakie and added a refimg:

| old WGLMakie | new GLMakie |
| --- | --- |
| Screenshot 2025-07-14 163741 |  Screenshot 2025-07-14 163750 |

ffreyer:
This should be good to merge now

PR #5166, opened 2025-07-07
Add additional scales based on `ReversibleScale`

cgarling:
# Description

Implements additional scales based on `ReversibleScale`. These are mostly inspired by color scales implemented in the astronomy-focused astropy project. I am using these as `colorscale` keyword arguments to methods like `heatmap`, but other uses may also be possible. Based on the tests that I added, they seem to work as `xscale` and `yscale` parameters, though `LogScale` will not work for negative `x` values. I could add a symmetric implementation for negative `x` values like is used for `Symlog10` if you think it's important the `LogScale` support all real `x`.

## Type of change

## Checklist

- [ x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ x] Added or changed relevant sections in the documentation
- [x ] Added unit tests for new algorithms, conversion methods, etc.
Not sure what the last point is asking me to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+ - Added new scales based on `ReversibleScale` for use as `colorscale`, `xscale`, and `yscale` attributes. The new scales are `AsinhScale`, `SinhScale`, `LogScale`, `LuptonAsinhScale`, and `PowerScale`.
 
 ## [0.24.3] - 2025-07-04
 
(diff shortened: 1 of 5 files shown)


cgarling:
Sorry, looks like I screwed up the formatting again ... I just missed the `LogScale` arguments, so I can fix that if you want, but I don't quite know how to read the runic output for the `for scale in ( ...` part. Is it asking for those following lines to be double indented?

Let me know if you want me to make these changes -- since you're triggering the CI manually, I figure it might be easier for you to make these fixes since I think if I push again it'll interrupt the CI and you'll have to manually trigger again.

SimonDanisch:
@jkrumbiegel @ffreyer  anything speaking against merging this?

jkrumbiegel:
Not really, I haven't gone into it in utmost detail but it appears to be sensible

SimonDanisch:
Cool, thank you! :)

Issue #5161, opened 2025-07-07
`deepcopy` called on a Module with new recipe system


timholy:
This is probably my fault (sorry for the noise), but if not...

- [X] what version of Makie are you running? (`]st -m Makie`)

Makie v0.24.3

- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)

Yes (even with `--startup-file=no`)

- [X] What platform + GPU are you on?

Windows 11, GeForce RTX 5070

- [X] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)

GLMakie

- [X] Please provide an minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

```julia
using GLMakie
using IntervalSets

struct FuncBound
    f            # The function
    bound        # Another function that gives an error bound, the true value is within f(t) ± bound(t)
end

@recipe ErrorBand (t::Union{AbstractInterval, AbstractVector{<:Real}}, m::FuncBound) begin
    "The color of the error band. If `:auto`, the color is taken from the line plot."
    bandcolor = :auto
    "The alpha value of the error band."
    bandalpha = 0.25

    cycle = @inherit cycle [:color]
    linecolor = @inherit linecolor
    linewidth = @inherit linewidth
    linestyle = @inherit linestyle
    transparency = @inherit transparency false
    visible = @inherit visible
    # Makie.mixin_generic_plot_attributes()...
end

function Makie.plot!(eb::ErrorBand)
    map!(eb.attributes, [:t, :m], [:itppoints, :tvals, :lohi]) do t, m
        (; f, bound) = m
        itppoints = convert_arguments(Lines, t, f)[1]   # let Makie use defaults for `lines(t, f)`
        tvals = [pt[1] for pt in itppoints]             # extract the time points that Makie decided to use
        lohi = [f(t) - bound(t) .. f(t) + bound(t) for t in tvals]
        return (itppoints, tvals, lohi)
    end
    hl = lines!(eb, eb.attributes, eb.itppoints)
    map!(eb.attributes, [:bandcolor], :bcolor) do bandcolor
        return bandcolor === :auto ? hl.color : bandcolor
    end
    band!(eb, eb.tvals, eb.lohi; color=eb.bcolor, alpha=eb.bandalpha)
    return eb
end

m = FuncBound(sin, sqrt ∘ abs)

errorband(-2π .. 2π, m; linewidth=4)
```

Result:
```
[...]

PR #5170, opened 2025-07-08
Add compute node initialization verification

ffreyer:
# Description

This checks that a compute node is not initialized to contain a compute node, edge or graph. This should catch the problem from #5161 with this kind of error:
image

## Type of change

- Improved errors/safety

## Checklist


ffreyer:
Not sure what to make of these benchmarks. I thought they got worse because of the TypedEdge constructor has a lot of runtime dispatch but now CairoMakie is basically the same...

ffreyer:
Did some local benchmarking and `invokelatest` does seem to be useful here overall. Without it, this pr increases ttfp by about `+30 + 14 + 4 = 48`ms (using + first create + first display), with it about `46 - 12 - 3 = 31`ms.

PR #5173, opened 2025-07-08
Fix changing text types

### LINKED ISSUE #5168: Text types can't be changed anymore with compute graphs
```julia
using CairoMakie

f, ax, tx = text(1, 1, text = rich("hi"))
tx.text = "hi str"
f
```

```
ERROR: Failed to resolve positions_transformed:
[ComputeEdge] positions_transformed = (::MapFunctionWrapper(#433))((text_positions, transform_func, ), changed, cached)
  @ /Users/krumbiegel/.julia/dev/Makie/Makie/src/compute-plots.jl:251
[ComputeEdge] text_positions = #2801((text_blocks, positions, ), changed, cached)
  @ /Users/krumbiegel/.julia/dev/Makie/Makie/src/basic_recipes/text.jl:383
[ComputeEdge] glyphcollections, glyphindices, font_per_char, glyph_origins, glyph_extents, text_blocks, text_color, text_rotation, text_scales, text_strokewidth, text_strokecolor, linesegments, linewidths, linecolors, lineindices = #2798((input_text, fontsize, selected_font, align, rotation, justification, lineheight, word_wrap_width, offset, fonts, computed_color, strokecolor, strokewidth, ), changed, cached)
  @ /Users/krumbiegel/.julia/dev/Makie/Makie/src/basic_recipes/text.jl:333
[ComputeEdge] _positions, input_text = #2790((position, text, arg1, arg2, ), changed, cached)
  @ /Users/krumbiegel/.julia/dev/Makie/Makie/src/basic_recipes/text.jl:49
[Input] text = (::InputFunctionWrapper(:text, AttributeConvert{text, text}))(hi str)
  @ /Users/krumbiegel/.julia/dev/Makie/Makie/src/compute-plots.jl:490
Due to ERROR: MethodError: Cannot `convert` an object of type String to an object of type Makie.RichText
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
   @ Base Base.jl:126
```

ffreyer:
# Description

Fixes #5168. 

If a computation returns a `Ref`, that Ref is used directly in the compute node. That allows you to prevent the type getting fixed to what it is at initialization. Here we missed a convert_attribute() for rich text (and maybe the vector one too?)

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed `p.text = "..."` erroring with `p = text(..., text = rich(...))` [#5173](https://github.com/MakieOrg/Makie.jl/pull/5173)
+
 ## [0.24.3] - 2025-07-04
 
 - Fix empty plotlist [#5150](https://github.com/MakieOrg/Makie.jl/pull/5150).
--- a/Makie/src/basic_recipes/text.jl
+++ b/Makie/src/basic_recipes/text.jl
@@ -25,7 +25,8 @@ convert_attribute(align, ::key"align", ::key"text") = Ref{Any}(align)
 
 # Positions are always vectors so text should be too
 convert_attribute(str::AbstractString, ::key"text", ::key"text") = Ref{Any}([str]) # don't fix string type
-convert_attribute(x::AbstractVector, ::key"text", ::key"text") = vec(x)
+convert_attribute(rt::RichText, ::key"text", ::key"text") = Ref{Any}([rt])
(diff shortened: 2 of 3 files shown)


jkrumbiegel:
Looks good, but what's the motivation for not making all attributes changeable to other types by default? I would think as long as the backend can handle it?

SimonDanisch:
The backend code is much easier if we can assume that the types can't change.
So, being able to change the type is the exception right now - and there aren't many where the backends can actually handle changing types as far as I know.

PR #5179, opened 2025-07-09
Fix UV transform docs

asinghvi17:
I got extremely confused because the REPL preview just showed `:flipy` - because there was no backslashing, that underscore in `flip_y` was interpreted as italics.

This is just a doc change and doesn't merit a changelog IMO


SimonDanisch:
Thanks!

PR #5183, opened 2025-07-10
Improve default spacing of Axis3 with non-equal aspect ratio

ffreyer:
# Description

Previously the size of the axis was always assumed to be a 1:1:1 box, which meant that other aspect ratios would produce larger gaps than they needed.

```julia
f = Figure()
a = Axis3(f[1, 1], viewmode = :fit, aspect = :data)
mesh!(a, Rect3f(0,0,0,5,2,0.1))
```

| before | after |
| --- | --- |
| Screenshot 2025-07-10 195311 | Screenshot 2025-07-10 195246 |

## Type of change

- Enhancement

## Checklist


ffreyer:
I checked a bunch of aspect - viewmode combinations and they all seem fine. Basically just increases the size of the Axis3 a bit when they get small due to aspect ratios.

I guess this should be considered breaking because it causes visual changes?

jkrumbiegel:
Hm no I would say wasting less space is a bug fix, the assumptions before were just not correct.

SimonDanisch (approved):

PR #5165, opened 2025-07-07
Switch some recipes over to the new ComputeGraph

theabhirath:
# Description

This is an attempt to tackle some of the recipes listed in #5114. I tried to switch things over using the blog post for 0.24 and the documentation. This is my first time contributing to Makie, so I may have missed some things, I apologize in advance for that!

## Plots handled:

- arc
- band
- pie
- stairs
- stem
- tooltip
- wireframe
- qqplot
- spy
- scatterlines

## Type of change

Enhancement(?). Technically not a bug fix because nothing is broken, more like a good-to-have I suppose.

PR #5184, opened 2025-07-10
Migrate some more recipes to ComputeGraph

theabhirath:
# Description

This is an attempt to tackle some (more) of the recipes listed in #5114, following up on #5165. Tests pass, but I would definitely appreciate a look-see to ensure I haven't missed anything – I'm particularly nervous about the density plot, where I removed an entire block of `on` code that I don't think is needed anymore? It seems to work 😅

Edit: oops. Not all tests pass, it seems. Somehow I didn't run them all locally 😰 Trying to fix them now, but would also appreciate any guidance on the same! `waterfall` and `density` seem to be the offenders.

Plots tackled:

- `boxplot`
- `crossbar`
- `density`
- `hist`
- `stephist`
- `violin`
- `waterfall`

## Type of change

Enhancement.

## Checklist

- Added an entry in CHANGELOG.md (for new features and breaking changes)


Makie/src/basic_recipes/waterfall.jl
```diff
@@ -44,16 +44,16 @@ function Makie.plot!(p::Waterfall)
...
-        return (xy = xy, fillto = fillto, final = final)
+        return xy, fillto, final
     end
 
-    fromto = lift(stack_bars, p, p[1], p.dodge, p.stack)
+    map!(stack_bars, p, [:converted_1, :dodge, :stack], [:xy, :fillto, :final])
 
-    if p.show_final[]
-        final_gap = p.final_gap[] === automatic ? p.dodge[] == automatic ? 0 : p.gap : p.final_gap
+    if p.show_final
```
theabhirath: Figured out that this is where the error is. Earlier this was `p.show_final[]`, and that converted it to an Observable implicitly, I think.  Is there a way to check the value of the variable without doing that?

ffreyer: `p.show_final[]` doesn't convert to an observable, it just grabs the value. But I think for "bonus" plots it would be preferable to refactor it to use `visible`. I.e. code like
```julia
if parent.should_be_shown
    # some code...
    plot!(...)
end
```
should become
```julia
# some code...
plot!(..., visible = parent.should_be_shown)
```
because that allows the plot to be toggled interactively. Same with `show_direction`

---

```julia
final_gap = p.final_gap[] === automatic ? p.dodge[] == automatic ? 0 : p.gap : p.final_gap
```
This was kind wrong to begin with, since it's not interactive. It should be something like this now:
```julia
map!(p, [:final_gap, :gap, :dodge], :calculated_final_gap) do final_gap, gap, dodge
    return final_gap === automatic ? dodge === automatic ? 0 : gap : final_gap
end
```

Makie/src/stats/boxplot.jl
```diff
@@ -77,15 +77,18 @@ flip_xy(p::Point2f) = reverse(p)
...
+        plot, [
+            :x, :y, :color, :weights, :width, :range, :show_outliers, :whiskerwidth,
+            :show_notch, :orientation, :gap, :dodge, :n_dodge, :dodge_gap,
+        ],
+        [
+            :centers, :boxwidth, :boxmin, :boxmax, :medians, :notchmin, :notchmax, :t_segments,
+            :boxcolor, :outlier_indices, :outlier_points,
+        ]
+    ) do x, y, color, weights, width, range, show_outliers, whiskerwidth, show_notch,
+            orientation, gap, dodge, n_dodge, dodge_gap
```
ffreyer: This seems like something that should get broken up, for readability and performance. As is, there are lots of little attributes that would invalidate this whole block of code and recalculate everything unnecessarily. For example changing `plot.color` should optimally just run `getuniquevalue(color, idxs)` on stored `idxs`, and not call `compute_x_and_width()` etc again.

I'm not sure what the individual step should be. For me it's a bit of trial and error to see what can be pulled out nicely, what's more deeply entangled and what's unnecessary to cache/split. I'd guess:
- `compute_x_and:_width()` and `finduniquesorted()` should be step 1 (`x̂` isn't used anywhere else and "sorted" is probably not free)
- `weights` and `quantile()` is used for a lot of things that probably don't interact otherwise
- `notchmin` and `notchmax` seem like one output bundle
- outliers should probably be one step producing a corrected set of q1's and q5's for t_segments. `orientation` should probably apply to outlier_points here, and the indices should probably get used directly to produce colors here (i.e. not be an output)
- t_segments and boxwidths should probably be combined (also with orientation)
- color should be isolated
- centers should be isolated
- boxmin, medians, boxmax should be combined

Makie/src/stats/violin.jl
```diff
@@ -51,16 +51,14 @@ function getuniquevalue(v::AbstractVector, idxs)
...
-    ) do x, y, width, vside, scale_type, color, show_median, n, bound, bw, w, limits, max_density,
-            dodge, n_dodge, gap, dodge_gap, orientation
+    map!(
+        plot, [
+            :x, :y, :width, :side, :scale, :color, :show_median, :npoints, :boundary, :bandwidth,
+            :weights, :datalimits, :max_density, :dodge, :n_dodge, :gap, :dodge_gap, :orientation,
+        ],
+        [:vertices, :lines, :colors]
+    ) do x, y, width, vside, scale_type, color, show_median, n, bound, bw, w, limits,
+            max_density, dodge, n_dodge, gap, dodge_gap, orientation
```
ffreyer: This also seems like a function that should be split up. I'd imagine `compute_x_and_width()` and `finduniquesorted` + `KernelDensity.kde` are probably relatively expensive and should get triggered unnecessarily, i.e. be in two isolated steps. The color stuff can probably be separated again.

Makie/src/basic_recipes/waterfall.jl
```diff
@@ -44,82 +44,73 @@ function Makie.plot!(p::Waterfall)
             xy[inds] .= Point2f.(x[inds], fromto.to)
             final[inds] .= Point2f.(x[inds], fromto.final)
         end
-        return (xy = xy, fillto = fillto, final = final)
+        return xy, fillto, final
     end
 
-    fromto = lift(stack_bars, p, p[1], p.dodge, p.stack)
```
theabhirath, line 50: I don't think I'm handling this right. I thought `p[1]` should map to `converted_1`, but clearly `p[1]` is somehow a combination of x and y, while `converted_1` seems to be only x. What's the way to get them together?

SimonDanisch: The equivalent of p[1] should be `:arg1`

ffreyer: On 0.24+ `p[1]` is `p.converted_1` or whatever it's name is in the recipe macro. You can check that by just doing
```julia
f,a,p = waterfall(rand(10))
p[1]
# Computed: 
#     name = :converted_1
```

Makie/src/basic_recipes/waterfall.jl
```diff
@@ -44,82 +44,73 @@ function Makie.plot!(p::Waterfall)
...
+    )
+        xs = first(
+            compute_x_and_width(first.(xy), width, gap, dodge, n_dodge, dodge_gap)
+        )
+        MarkerType = promote_type(typeof(marker_pos), typeof(marker_neg))
+        DataType = eltype(xy)
+        shapes = MarkerType[]
+        xy = DataType[]
+        for i in eachindex(xs)
+            y = last(xy[i])
```
ffreyer: I'm guessing your comment was about this CI error:

image

which eventually points at this bit of code. You're getting an indexing here because of the `xy = DataType[]` overwriting the `xy` you pass to the function

theabhirath: Yep, this was the problem, changed the local variable name and fixed it!

Makie/src/stats/boxplot.jl
theabhirath: Have pushed an attempt at this – reference image tests for both waterfall and box plot seem to be failing, for reasons I have not dived into yet, but on the face of it I've tried to implement what you've suggested here 😅

theabhirath:
Only one test is failing, and that is the waterfall plot. The change seems to be that the markers on x-axis seem to be spaced further apart, and I'm not really sure why. Unless this is some sort of side-effect of using `visible` for one of the component plots, I can't really narrow that down 😅.

image

image

Makie/src/basic_recipes/waterfall.jl
```diff
@@ -44,82 +44,65 @@ function Makie.plot!(p::Waterfall)
...
-                end
+    function direction_markers(
+            xy, fillto, marker_pos, marker_neg, width,
+            gap, dodge, n_dodge, dodge_gap
+        )
+        xs = first(compute_x_and_width(first.(xy), width, gap, dodge, n_dodge, dodge_gap))
+        MarkerType = promote_type(typeof(marker_pos), typeof(marker_neg))
+        DataType = eltype(xy)
+        shapes = MarkerType[]
+        scatter_xy = DataType[]
```
asinghvi17:
```suggestion
        PointType = eltype(xy)
        shapes = MarkerType[]
        scatter_xy = PointType[]
```

Just because `DataType` is the supertype of all Julia types, this is less confusing

SimonDanisch:
Ah, yes, so I guess visible=false still goes into the boundingbox (I thought it didn't to be honest, i wonder if that's a regression. @ffreyer do you remember how this is supposed to be?).
Anyways, I guess we should just make show_final[] a branch again with a TODO to not deal with this for now, and maybe make it dynamic once we fix the boundingbox issue?

theabhirath:
Everything should pass now! 😁

SimonDanisch:
Thank you all!

PR #5198, opened 2025-07-17
specapi fixes

SimonDanisch:
I noticed that changing the plot type in an axis creates a new axis:
```julia
function to_spec(Typ, cmap=:viridis, ms=20, f=sin)
    ms = Typ == S.Scatter ? (; markersize = ms) : (; linewidth = ms)
    pl = Typ(
        xrange, f.(yrange);
        colormap = cmap,
        color=yrange,
        ms...
    )
    cb = S.Colorbar(pl)
    S.GridLayout(
        [S.Axis(plots = [pl]) cb]
    )
end
begin 
    f = sin
    xrange = 0:0.3:10
    yrange = f.(xrange)
    typ = Observable{Any}(S.Scatter)
    obs = map(to_spec, typ)
    f, ax, pl = plot(obs)
    display(f)
end
typ[] = S.Lines

```

The reason is, that `distance_score(ax1, ax2) == 100` for different plot types in the axis, which is way over the threshold so we dont match the axes, and need to create new ones.
I took out typ comparison in the PlotSpec distance score, and instead do the check if types match in `find_minimum_score` which should be a nice shortcuircit condition for performance, and means the comparison between plots of the same type come out much smaller.

### DIFF
--- a/GLMakie/test/Project.toml
+++ b/GLMakie/test/Project.toml
@@ -2,7 +2,6 @@
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
 GeometryBasics = "5c1252a2-5f33-56bf-86c9-59e7332b4326"
-ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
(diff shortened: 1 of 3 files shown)


asinghvi17:
Sounds good to me!

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 
+- Fixes and improvements for SpecApi [#5198](https://github.com/MakieOrg/Makie.jl/pull/5198).
```
jkrumbiegel: maybe say what was fixed? otherwise it's not so useful for the changelog I think

PR #5201, opened 2025-07-18
[BACKPORT v0.23] Fix GLMakie's surface impl. for 3D transform funcs

asinghvi17:
I end up needing this for the satellite demos, otherwise you cannot plot surfaces on GlobeAxis (from geomakie) since they become completely flat.

Tested locally that this works.  Not sure we want another refimage on a backport though, and I don't think reaching into the render object is a particularly useful test since it could randomly break.

e.g. [this video](https://julialang.slack.com/archives/C8RQUU2KH/p1752859613033219?thread_ts=1752682618.751979&cid=C8RQUU2KH) had to be done in WGLMakie since this is broken on GLMakie (it's just a thin line around the equator instead)


asinghvi17:
Bump, can we get this in please?  Thanks :D

PR #5206, opened 2025-07-20
Pass parent plot's attributes down to PlotSpecs in PlotList

asinghvi17:
# Description

This allows transformation to also be passed down, fixing:
- theming
- plotspecs not getting their parent's transformation

A simple test is also included.

## Type of change

## Checklist

- [X] Added unit tests


asinghvi17:
Hm it looks like this broke something with cycling?  Specifically https://github.com/MakieOrg/Makie.jl/blob/dc3ae0aec8c14fb5666323ad7ca87aa6cf3c0964/ReferenceTests/src/tests/specapi.jl#L112 and the ecdfplot test

PR #5223, opened 2025-07-28
hide GL icon in MacOS dock until a window is shown

### LINKED ISSUE #4278: Each Julia instance with loaded GLMakie results in a MacOS taskbar item
Whenever I have a Julia process or a notebook that uses GLMakie, another taskbar item in the MacOS taskbar gets created:
image
They appear even if I just save the figure or it is statically displayed inline in the notebook – not just when `display(fig)` is called. So, naturally, all these windows are just empty (unless displaying interactive figures), clicking on them doesn't open anything.

Can these "windows" be created only when actually displaying figures in GL windows?

- [x] what version of Makie are you running? (`]st -m Makie`) any
- [x] What platform + GPU are you on? macbook air m2

aplavin:
# Description

Fixes #4278 – a small but annoying bug which makes MacOS dock to show an icon per Julia process with GLMakie loaded.

Adds `show_gl_icon_in_dock(::Bool)` and calls it so that:
- By default, after loading GLMakie, the icon is hidden
- After `display()` is called, the icon appears

This should work intuitively for:
- scripts that export images/videos: no display, no icon
- notebooks that use GLMakie: no display, no icon
- scripts/repl that calls display(): with icon

The `show_gl_icon_in_dock` definitely works, I've used it in MakieExtra already. Not 100% sure if I put the calls in the right places though, could you please check? The GLMakie sequence of loading/activation/display is a bit convoluted for someone unfamiliar with it.

## Type of change

## Checklist

- [x] Added an entry in CHANGELOG.md


SimonDanisch:
@jkrumbiegel can you review this? Anyone else doesn't have a mac.

jkrumbiegel:
I normally have this thing in the dock once I opened the first GLMakie window:

image

With this PR, it's not there, even if I display a window (in VSCode, using the VSCode repl). I can then not access the window at all, unless I three-finger swipe up to get the overview of all open windows. So that's not great, yet.

In general, it's cool that we can hide the icon as it's annoying if it sticks around after all windows are closed. But maybe that should be done in the open/close logic of a given Screen then? It would be much better if there was one icon per window, which would actually have the Makie logo. But I'm not sure if that can be done.

jkrumbiegel:
Huh I had claude try to generate some code to change the icon and it worked on the first try, so that would maybe be another nice thing to do?

```julia
function set_gl_dock_icon(icon_path::String)
    Sys.isapple() || return
    try
        # Get NSApplication and NSImage classes
        nsapp_class = @ccall objc_getClass("NSApplication"::Cstring)::Ptr{Cvoid}
        nsimage_class = @ccall objc_getClass("NSImage"::Cstring)::Ptr{Cvoid}
        nsstring_class = @ccall objc_getClass("NSString"::Cstring)::Ptr{Cvoid}
        
        # Get selectors
        shared_app_sel = @ccall sel_registerName("sharedApplication"::Cstring)::Ptr{Cvoid}
        set_icon_sel = @ccall sel_registerName("setApplicationIconImage:"::Cstring)::Ptr{Cvoid}
        init_with_path_sel = @ccall sel_registerName("initWithContentsOfFile:"::Cstring)::Ptr{Cvoid}
        alloc_sel = @ccall sel_registerName("alloc"::Cstring)::Ptr{Cvoid}
        string_with_utf8_sel = @ccall sel_registerName("stringWithUTF8String:"::Cstring)::Ptr{Cvoid}
        
        # Get NSApplication shared instance
        nsapp = @ccall objc_msgSend(nsapp_class::Ptr{Cvoid}, shared_app_sel::Ptr{Cvoid})::Ptr{Cvoid}
        
        # Create NSString from path
        nsstring = @ccall objc_msgSend(nsstring_class::Ptr{Cvoid}, string_with_utf8_sel::Ptr{Cvoid}, icon_path::Cstring)::Ptr{Cvoid}
        
        # Create NSImage
        nsimage_alloc = @ccall objc_msgSend(nsimage_class::Ptr{Cvoid}, alloc_sel::Ptr{Cvoid})::Ptr{Cvoid}
        nsimage = @ccall objc_msgSend(nsimage_alloc::Ptr{Cvoid}, init_with_path_sel::Ptr{Cvoid}, nsstring::Ptr{Cvoid})::Ptr{Cvoid}
        
        # Set application icon
        @ccall objc_msgSend(nsapp::Ptr{Cvoid}, set_icon_sel::Ptr{Cvoid}, nsimage::Ptr{Cvoid})::Ptr{Cvoid}
    catch e
        @warn "Failed to set dock icon" e
    end
end
```

image

aplavin:
> With this PR, it's not there, even if I display a window

That's why I said in the first post

> Not 100% sure if I put the calls in the right places though, could you please check? 

The GLMakie sequence isn't very clear for me. The function to hide/show icon definitely works, but I'm not certain where are the right places to put its calls.

> Huh I had claude try to generate some code to change the icon and it worked on the first try, so that would maybe be another nice thing to do?

Maybe indeed. For me the main priority is to resolve the practical issue of many useless icons popping up – currently, they appear for each Julia process with GLMakie, without any actual windows present.

With `using MakieExtra`, it currently just hides the icon by default, while giving the option to manually show it if needed. This is much friendlier than crowding the taskbar, but ideally it should be automatic hide/show depending on windows being present. Thus this PR...

jkrumbiegel:
> For me the main priority is to resolve the practical issue of many useless icons popping up 

Yeah that's why I was suggesting the icon thing, because the use of an icon that looks like a terminal and says julia is also limited for users that don't expect this is how GLMakie appears.

I'm also not sure which is the correct point in the screen lifetime to do this, that's @SimonDanisch's area of expertise. I guess at every screen open or close it should be checked if any other screens are currently open, and the icon shown/hidden accordingly.

aplavin:
fixed – now it properly hides/shows icon when windows are hidden/shown
also, followed @jkrumbiegel suggestion – now dock show Makie icon instead of black terminal one

GLMakie/src/GLMakie.jl
```diff
@@ -93,12 +93,56 @@ include("gl_backend.jl")
 load_all_shaders(SHADER_DIR)
 WARN_ON_LOAD[] = true
 
+const MAKIE_ICON_PATH = normpath(joinpath(@__DIR__, "..", "assets", "icon.png"))
```
jkrumbiegel: this is not relocatable I think

jkrumbiegel:
Except the path issue, this seems to work well overall, thanks :)

GLMakie/src/GLMakie.jl
aplavin: Not sure I understand – do you mean it breaks if the icon gets moved to another location? I assumed the `assets` folder is the natural place to put it.

asinghvi17: It will break if you use PackageCompiler to generate a system image with Makie, specifically because the absolute path gets baked in.  I think we use RelocatableFolders to get around this?

ffreyer: Is `assetpath("icon.png")` safe to use?

asinghvi17: Yes, but only because it uses artifacts 😅 

https://github.com/MakieOrg/Makie.jl/blob/74e6f39ef9ff03859048d4a8167d2f7d58242908/Makie/src/Makie.jl#L89

jkrumbiegel:
Looks like the icon was the same as in the Makie asset artifact so I used that one.

jkrumbiegel (approved):

PR #5224, opened 2025-07-28
Architecture docs

jkrumbiegel:
At the Makie BoF at JuliaCon we discussed that there's currently not really a page in the docs that talks about the design of the library a bit more abstractly, how the different elements fit together. This is a start for that, explaining scenes, plots, figure and blocks with some flowcharts via graphviz.

### DIFF
--- a/docs/src/explanations/architecture.md
+++ b/docs/src/explanations/architecture.md
@@ -101,13 +101,16 @@ This sets up the following relationship of objects:
 ```@graphviz
 digraph {
     rankdir=TB;
-    compound=true;
     edge [dir=none];
-    
-    Figure;
-    C [label="content::Vector"];
-    GridLayout;
-    Scene [label="Root Scene"];
+
+    subgraph cluster_figure {
+        label="Figure";
+        style=filled;
+        fillcolor="#fcfcfcff";
+
+        GridLayout;
+        Scene [label="Root Scene"];
+    }
     
     subgraph cluster_axis {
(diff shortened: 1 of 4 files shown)


docs/src/explanations/architecture.md
```diff
@@ -0,0 +1,184 @@
...
+    S4 [label="Child Scene\nSize: 350x300\nOffset: 0, 0\n22 Plots"];
+    S5 [label="Child Scene\nSize: 100x200\nOffset: 50, 25\n3 Plots"];
+    
+    R -> S1;
+    R -> S2;
+    S1 -> S3;
+    S2 -> S4;
+    S2 -> S5;
+}
+```
```
asinghvi17, line 33: Should this be mermaid instead?  That is interactive

jkrumbiegel: I tried for a bit getting mermaid to work but dropped it after a while. If you want to spend time on it go for it, although I'm reasonably happy with graphviz's output here and I don't need interactivity (what would you want to interact with?)

asinghvi17: If the diagrams are sufficiently small it's fine to leave as is, yeah.   But for large diagrams i do like Mermaid's zoom/pan stuff.

jkrumbiegel: ah, yeah these are not going to get big, I want to clarify not confuse 😆

PR #5225, opened 2025-07-29
Ignore volume box z values

ffreyer:
# Description

Fixes https://discourse.julialang.org/t/unwanted-clipping-at-corners-of-glmakie-3d-plot/131053

Volumes are drawn in a box, where the backside of the box "renders". If fragments (pixels) of this box end up outside the near .. far range, OpenGL/WebGL will clip them. And if they get clipped, the entire ray trace for that pixel gets clipped.

To avoid this, the pr hard-codes clip space z coordinates to 0, which should always pass the clipping test `-w <= z <= w`. For volume plots that don't write z values during ray tracing (or turn it off with `enable_depth = false`) this may change the z ordering between volume plots and other elements. If I'm not mistaken volumes should end up in front of things more often. I don't think that matters much, because the z ordering was already wrong to begin with. If it does, we could use `z = w` to move them to the far plane or `z = -w` to move them to the near plane, instead of the center.

* Updated volume plots to use the ray start depth as their depth. This gives better depth ordering than a static z value, but could cause near-plane clipping. So far I haven't noticed any issues in LScene or Axis3 with it.

## Type of change

## Checklist


ffreyer:
"Depth Shift" test fails in WGLMakie because it now looks more like GLMakie. (note the bottom corner of the volume)
| before | after | GLMakie
| --- | --- | --- |
| image | image | image |

SimonDanisch (approved):

Issue #4446, opened 2024-10-03
DataInspector for PolarAxis should display tuples of (r, theta)


thomvet:
### Feature description
It would be nice if DataInspector() shows tuples of (r, theta) instead of (x,y,z) for a PolarAxis. 

Consider the following example:
```
f = Figure(size = (800, 400))

ax = PolarAxis(f[1, 1], title = "Theta as x")
lineobject = lines!(ax, 0..2pi, sin, color = :red)

ax = PolarAxis(f[1, 2], title = "R as x", theta_as_x = false)
scatobject = scatter!(range(0, 10, length=100), cos, color = :orange)
DataInspector()
```
![image](https://github.com/user-attachments/assets/a46cde43-16d1-4e1d-815b-677dc0e848ab)

Also, the (r = 0, theta = 0) point is shown as (x = 0, y = 1, z = 0). That's confusing? 🙂

EdsterG:
It seems like the data inspector isn't aware of the underlying `Axis` the plot is displayed on. Am I missing something @ffreyer, or is there an easy way to access the axis object from `show_data`? Would be even better if it was accessible from the `inspector_label` function, but when I tried `plot` didn't seem to have any references to the axis.

asinghvi17:
The scene tree (and DataInspector by extension) is unaware of any axis or layout related things.

However, the scene's transformation should be a `Makie.PolarTransformation` or `Makie.PolarAxisTransformation` (I don't remember the exact name, but it's something along those lines), which you could check via `scene.transformation.transform_func[]` or `plot.transformation.transform_func[]`.

EdsterG:
Does those transformation functions somehow hint at the value of `theta_as_x`? That's the tricky part, getting the value of `theta_as_x`.

asinghvi17:
Isn't that embedded into the transformation struct?

asinghvi17:
https://github.com/MakieOrg/Makie.jl/blob/e74adcb95ac57630e0d6fbdab582dcacf2f52234/src/layouting/transformation.jl#L461-L471

EdsterG:
Oh, thanks for pointing this out. I wasn't too familiar with how polar axis are implemented.

ffreyer:
I think this was already showing (r, theta). The screenshot is at leats showing (r = 0, cos(r) = 1). #5241 drops the x/y/z in the string as well which might make this a bit less confusing?

Issue #4440, opened 2024-10-02
Histogram is now missing the `inspector_label` attribute


Sagnac:
`Makie v0.21.12`

```julia
julia> using GLMakie # v0.10.12

julia> hist(rand(15); inspector_label = (_, _, _) -> "label")
ERROR: Invalid attribute inspector_label for plot type Plot{Makie.hist, Tuple{Vector{Float64}}}.
```
```julia
julia> :inspector_label in Makie.attribute_names(Hist)
false

julia> :inspector_label in Makie.attribute_names(Image)
true
```

I remember being able to use a custom `inspector_label` function with histograms using GLMakie v0.9.6

ffreyer:
It ~~shouldn't~~ doesn't need to be passed down. DataInspector should stop looking at the top level plot that defines these attributes

ffreyer:
I think I fixed this in #5294. 

> It ~shouldn't~ doesn't need to be passed down. DataInspector should stop looking at the top level plot that defines these attributes

This isn't/wasn't really true, but will be after #5241 gets merged

PR #5241, opened 2025-08-08
DataInspector rework

### LINKED ISSUE #962: Persistent Data Tips for Data Exploration and Comparison

### LINKED ISSUE #4706: make Checkboxes not inspectable
don't know why anyone would want DataInspector to show tooltips for a Checkbox.  for now i'm hacking this with `foreach(x->x.inspectable[]=false, cb.blockscene.plots)`.  maybe this should be done by default?  or at least have an `inspectable` property like plots do to make turning it off easier?

```
<img width="712" alt="Image" src="https://github.com/user-attachments/assets/67ae713d-38fe-4335-b6ee-66e83e8a78ed" />
using GLMakie
f = Figure()
cb = Checkbox(f[1,1])
DataInspector()
```
Image

### LINKED ISSUE #4713: `scatterlines.inspector_label` is broken
For the following example:
```julia
using GLMakie

fig = Figure()
ax = Axis(fig[1,1])

DataInspector(ax)

lines!(ax, [1, 2], [1, 2];
    inspector_label=Returns("This works")
)

scatter!(ax, [1, 2], [1.5, 2.5];
    inspector_label=Returns("This also works")
)

scatterlines!(ax, [1, 2], [2, 3];
    inspector_label=Returns("This doesn't work")
)

fig
```
only the lines and scatter plot show the correct label. However, the scatterlines plot shows the default label.

Image
Image
Image
Image

I'm on Makie v0.21.18 and julia v1.11.2 on Windows with an Nvidia GPU

- [x] what version of Makie are you running? (`]st -m Makie`)

### LINKED ISSUE #4866: DataInspector errors with band
- [x] what version of Makie are you running? (`]st -m Makie`)
        [ee78f7c6] Makie v0.22.2
  Arch Linux, laptop with integrated Intel graphics.

to reproduce:

```julia
using GLMakie

times = 0:-0.4:-10000
upr = sin.(times)
lwr = cos.(times)
f = Figure()
ax = Axis(f[1,1],
          xlabel = "Time (kyr)",
          ylabel = "Banana (-)")
band!(ax, times, upr,  lwr, label = "Hi!")
axislegend()
DataInspector() # doesn't play nice with band
```

then hover over the image and see the error messages start to appear:

```
Error in callback:
TaskFailedException
Stacktrace:
  [1] check_channel_state
    @ ./channels.jl:188 [inlined]
  [2] put!
    @ ./channels.jl:356 [inlined]
  [3] (::Makie.var"#1337#1341"{Channel{Nothing}})(::Tuple{Float64, Float64}, ::Tuple{Float64, Float64})
    @ Makie ~/.julia/packages/Makie/ux0Te/src/interaction/inspector.jl:296
  [4] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::@Kwargs{})
    @ Base ./essentials.jl:1055
  [5] invokelatest(::Any, ::Any, ::Vararg{Any})
    @ Base ./essentials.jl:1052
  [6] (::Observables.OnAny)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:420
  [7] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
  [8] invokelatest
    @ ./essentials.jl:1052 [inlined]
  [9] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [10] setindex!(observable::Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [11] (::GLMakie.MousePositionUpdater)(::Makie.TickState)
    @ GLMakie ~/.julia/packages/GLMakie/87u59/src/events.jl:6
 [12] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
 [13] invokelatest
    @ ./essentials.jl:1052 [inlined]
 [14] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [15] setindex!
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123 [inlined]
 [16] pollevents(screen::GLMakie.Screen{GLFW.Window}, frame_state::Makie.TickState)
    @ GLMakie ~/.julia/packages/GLMakie/87u59/src/screen.jl:547
 [17] on_demand_renderloop(screen::GLMakie.Screen{GLFW.Window})
    @ GLMakie ~/.julia/packages/GLMakie/87u59/src/screen.jl:1033
 [18] renderloop(screen::GLMakie.Screen{GLFW.Window})
    @ GLMakie ~/.julia/packages/GLMakie/87u59/src/screen.jl:1061
 [19] (::GLMakie.var"#79#80"{GLMakie.Screen{GLFW.Window}})()
    @ GLMakie ~/.julia/packages/GLMakie/87u59/src/screen.jl:922

    nested task error: DimensionMismatch: arrays could not be broadcast to a common size: a has axes Base.OneTo(2) and b has axes Base.OneTo(3)
```

### LINKED ISSUE #5229: LoadError: Failed to resolve inspector_label on 0.24
Code:
```julia
fig,ax,plt = scatter(rand(10))
@show plt.inspector_label[]
plt.inspector_label[] = (args...) -> "Custom label"
@show plt.inspector_label[]
```
On Makie <0.24, it works correctly and outputs:
```
plt.inspector_label[] = MakieCore.Automatic()
plt.inspector_label[] = var"#1#2"()
```
On Makie 0.24 it throws when setting `inspector_label`:
```
plt.inspector_label[] = Makie.Automatic()
ERROR: LoadError: Failed to resolve inspector_label:
[Input] inspector_label = (::InputFunctionWrapper(:inspector_label, AttributeConvert{inspector_label, scatter}))(#1)
  @ /Users/aplavin/.julia/packages/Makie/6zcxH/src/conversions.jl:920
Due to ERROR: MethodError: Cannot `convert` an object of type var"#1#2" to an object of type Makie.Automatic
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
   @ Base Base.jl:126

Stacktrace:
 [1] setproperty!(x::Base.RefValue{Makie.Automatic}, f::Symbol, v::Function)
   @ Base ./Base.jl:52
 [2] setindex!(b::Base.RefValue{Makie.Automatic}, x::Function)
   @ Base ./refvalue.jl:60
 [3] resolve!(input::ComputePipeline.Input{ComputePipeline.ComputeGraph})
   @ ComputePipeline ~/.julia/packages/ComputePipeline/SSXPs/src/ComputePipeline.jl:427
 [4] _resolve!(computed::ComputePipeline.Computed)
   @ ComputePipeline ~/.julia/packages/ComputePipeline/SSXPs/src/ComputePipeline.jl:652
 [5] resolve!(computed::ComputePipeline.Computed)
   @ ComputePipeline ~/.julia/packages/ComputePipeline/SSXPs/src/ComputePipeline.jl:644
 [6] getindex(computed::ComputePipeline.Computed)
   @ ComputePipeline ~/.julia/packages/ComputePipeline/SSXPs/src/ComputePipeline.jl:563
 [7] macro expansion
   @ show.jl:1232 [inlined]
```

 
- [ ] what version of Makie are you running? (`]st -m Makie`) 0.24
- [ ] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie) any

### LINKED ISSUE #5383: `access to undefined reference` with child plots
Setting (some?) attributes on a child plot throws `UndefRefError`s now in the new Makie 0.24.
The same code works fine as expected in Makie 0.22.

MWE:
```julia
@recipe MyRecipe (X,) begin
	Makie.mixin_generic_plot_attributes()...
end

function Makie.plot!(p::MyRecipe)
	image!(p, attributes(p), p.X)
end


fig,ax,plt = myrecipe(rand(10, 10))
# throws UndefRefError: access to undefined reference:
plt.plots[1].inspector_label[] = (args...) -> "Custom label" 
```

- [ ] what version of Makie are you running? (`]st -m Makie`) 0.24
- [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) yes
- [ ] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie) any

ffreyer:
# Description

Related Issues:
- fixes #5229
- fixes #4713
- closes #962 by adding the feature (though the hotkeys may need some adjustments)
- closes #4706 by restricting DataInspector to one scene each
- closes #5383 by avoiding the need to set `inspector_label` for child plots

## Goals

This aims to:
- fix DataInspector not always considering `inspector_label` (etc) of the top level plot (#3263, #4440, #4713)
- add persistent tooltips (#962)
- make DataInspector per-scene (#4706)
- (probably) fix issues with labels being post-transform (#3526, #3571, #4446)
- maybe make DataInspector easier to extend
- ~~may drop indicators~~

I think I fixed #4866 too, which I should probably extract...

## Problems

The previous version of DataInspector worked like this:
- `pick()` picks a primitive plot, like `lines`
- DataInspector steps to the root parent of the plot tree `lines` belongs to
- DataInspector steps back down, calling `show_data` at each step

As a consequence, `inspector_label` etc was only accessed for plots that implemented `show_data`, and would be given that plot. For example `scatterlines` did not implement a `show_data` method and thus its `inspector_label` was not considered. If it were simply passed down, it would be called with either the lines or scatter plot and a position and index relevant to that plot. This is problematic for more complex recipes where you may want the recipes attributes and an appropriate index to generate a label.

Implementing `show_data()` is probably rather difficult for people, as it does multiple things:
- it may map the (plot, idx) pair generated by `pick` to something appropriate for the given recipe plot
- it generates a screen/pixel space position (which needs to correctly transform and project)
- it may generate (and clean up) an indicator plot
- it generates a label
- it updates the tooltip

Persistent Tooltips have been difficult to implement because `DataInspector` worked in a global scene. That forced tooltips to be drawn in pixel space, either by transforming a data space position or by directly using pixel space mouse positions. A persistent tooltip should be stuck somewhere in data space, e.g. anchored to a scattered marker, which means it would need to connect to the camera and project frequently. With `show_data` doing everything a persistent tooltip would also need to extract relevant information from a dynamic tooltip.

## Plan

<details>

<summary>Plan</summary>

DataInspector being a global object has been more painful than useful, so I'm planning to make it per-scene. This will fix the general annoyances with it picking irrelevant plots in other scenes. It will also allow tooltips to exist in data space, as they don't have to work with different cameras anymore, which should simplify persistent tooltips.

I'm also planning to break up `show_data` into multiple pieces and restructure the infrastructure around it. As I see it, the main tasks of `show_data` are:
- map `pick()` data to something appropriate to a high level plot (typically an index or interpolation between indices)
- extract a position from a plot using the processed pick data
- extract a label from a plot using the processed pick data (and position)
- do whatever processing DataInspector needs and update the tooltip
- (generate/update an indicator plot using the processed pick data?)

For the first point I'm planning to add a new `PlotElement` type generated by a `pick_element()` function. The `PlotElement` should encapsulate the indexing or interpolation so that you can get whatever plot attribute you want for the picked element. The `pick_element` function should generate a `PlotElement` from `pick` data, attempting the root parent plot first and progressing down the to the picked primitive like `show_data`. The resulting `PlotElement` should always be (considered to be) relevant to the root parent though. This may also be useful in general as a higher level `pick()`. For example:

```julia
# Lines are continuous, so the picked element should represent an
# interpolated point on the line. `element.color` should thus return 
# a float value between 1 and 10
f,a,p = lines(rand(10), color = 1:10, linewidth = 10)
element = pick_element(scene, mouseposition_on_line)
element.color # e.g. 3.7
element.linewidth # constant, 10

# Scatter is discrete, so the picked element should encapsulate an
# index, making element.color an integer
f,a,p = scatter(rand(10), color = 1:10, markersize = 10)
p.color # e.g. 4
p.markersize # constant, 10
```

With that, getting positions and labels can hopefully be rather simple and generic:
```julia
function get_position(element)
    name = get_converted_args_name(parent(element))
    return getproperty(element, name)
end

function get_tooltip_label(element, pos)
    label = element.inspector_label
    if label isa String
        return Format.format(label, pos...)
    elseif label isa Function
        return label(element)
    elseif label === automatic
        return default_label(element)
    end
end
```

For persistent tooltips my current idea is to allow passing a `PlotElement` array to `tooltip()`. The tooltip plot can then be saved in DataInspector and elements can be added or removed. Since different plots can have different transformations, there needs to be a tooltip plot for each parent plot. To allow tooltips to update with changes to plot attributes we need some sort of `map!(..., parent_attributes, ...)`. I'm currently thinking of tracking which attributes are used when calling `get_position()` and `get_tooltip_label()` to collect the necessary attributes.

</details>

## Changes

DataInspector is now per scene. This simplifies persistent tooltips and using pre-transform/projection data because DataInspector no longer has to deal with multiple scenes. This also stops DataInspector from picking plots outside its designated scene, which was generally more of a hassle.

The whole structure of how data inspector gets from a picked plot to a tooltip + indicators has been reworked. Previously there were two parts. First a generic part, which traced the picked plot up to its root parent and dispatched calls to `show_data`. If no fitting method exists, the function would try again with the next lower level plot. The second part is the `show_data` method which did everything else, including providing an extension interface for recipe developers.
The new structure breaks up `show_data` to better separate concerns and through that simplify extension (I hope). It also reworks the fallback logic in a way that the plot trace doesn't disappear, allowing each step to do fallbacks independently. The steps now include:
1. (internal) tracing of the picked plot to the root recipe
2. (extendable) `get_accessor(plot, idx, trace)` methods for creating a `AbstractElementAccessor` which acts as an index or interpolation information for accessing the picked element of a plot
3. (internal) creation of a `PlotElement` which bundles the trace with the accessor returned by `get_accessor()`
4. (internal) manual search for an applicable `get_position(element)` method which successively removes parent plots from the trace in the PlotElement. This is used to identify the plot from which the position is pulled so that the appropriate transformations and space can be copied.
5. (extendable) `position = get_position(::PlotElement{PlotType})` which gets the picked position
6. (internal) `get_tooltip_label(formatter, ::PlotElement{PlotType}, position)` which produces the tooltip label string
    - (extendable) `get_default_tooltip_label(formatter, ::PlotElement{PlotType}, position` which produces the string if no attribute based string exists
        - (extendable) `get_tooltip_data(::PlotElement{PlotType}, position)` which returns data to be converted to a string (defaults to position) for the label
7. (internal) `update_indicator_internal!(inspector, ::PlotElement{PlotType}, position)` which updates transformations and space of an updated indicator
    - (extendable) `update_indicator!(inspector, ::PlotElement{PlotType}, position)` which sets the data of an indicator plot and makes it visible
    - (extendable) `construct_indicator_plots!(inspector, PlotType)` which produces a cached indicator plot for a given plot type. `get_indicator_plot(inspector, PlotType)` will call this function once to produce this plot and otherwise return the cached version.

Changes to tooltips:
- Mesh currently gives position information instead of boundingbox information
- formatting has been reworked (more consistent, probably shorter)

## TODO

- [x] add docstrings
- [x] review online documentation
- [x] switch tests to use new DataInspector
- [ ] add more specialized tests 
- [x] remove old DataInspector
- [ ] improve label construction interface (i.e. how formatting works, shorter function names?)
- [x] re-add missing settings
- [x] figure out how to efficiently and correctly trigger updates (e.g. when updates can be discarded with tick based updates)
- [x] add refimg test for persistent tooltips

General feature testing - correct behavior with previously implemented tooltips, no hard errors in general
- [x] all primtives work (excluding volume)
- [x] all explicitly implemented plots from old DataInspector work (BarPlot, Arrows3D, Arrows2D, Contourf, VolumeSlices, Band, Spy, HeatmapShader, DataShader)
- [x] no plot hard errors (this would crash the renderloop)

## Notes

- There are a bunch of plots that may need further changes to their tooltips but I'm sure what those should be:
    - stairs, stephist, ecdfplot should maybe display step values
    - hspan, vspan should maybe display limits
    - qqnorm, qqplot, boxplot, crossbar should probably display some stats-based things
    - stem should probably be discrete and show max heights
    - streamplot should probably display local position + direction
    - dendrogram, triplot, voronoiplot ... not sure what's useful to show here
- barplot is messy and incomplete due to not yet being updated to the compute graph
- discrete heatmap/image picking could be simplified by per scene rendering + rendering indicators without picking (#4150 + #4689 + WGLMakie versions)
- refimg tests cover everything except 
    - some plots I'm unsure about: boxplot, dendrogram, ecdfplot, qqplot, qqnorm, streamplot, triplot, voronoiplot
    - some plots that don't produce tooltips: annotation, bracket, textlabel, tooltip

## Type of change

Likely breaking because `show_data` will most likely be removed in its current state

## Checklist

### DIFF
--- a/ReferenceTests/src/tests/generic_components.jl
+++ b/ReferenceTests/src/tests/generic_components.jl
@@ -768,7 +768,12 @@ end
 
     # One per plot, in order
     mpos = [
-        (82, 325), (116, 325),
+        # WGLMakie had a bug where invisible tooltips were still pickable.
+        # (116, 310) is inside the previous tooltip, to check that this doesn't
+        # happen anymore.
+        # TODO: If/once picking can fully ignore the tooltip, move it to (116, 325)
+        # to check that that is working too (no flicker when tooltip covers picked area)
+        (82, 325), (116, 310),
         (152, 350), (192, 350),
         (243, 380), (243, 270), (243, 230), (243, 180),
         (290, 325), (331, 325),


ffreyer:
I went through all the plots that have docs examples. They all produce a tooltip without erroring now, and most produce something reasonable I think. Here's an overview with plots I'm unsure about in bold

Primitives:
- lines, linesegments: interpolated position on line segment
- scatter, meshscatter: discrete position
- text: discrete position (per string) at anchor point
- mesh, surface: triangle-interpolated position
- image, heatmap: discrete cell value or interpolated color value depending on interpolate
- voxels: discrete voxel position
- volume: discarded

Inherited tooltip from child plot
- Line-like: ablines, arc, hlines, vlines, **stairs**, **stephist**, **qqnorm**, **qqplot**, **dendrogram**, wireframe, **ecdfplot**
- scatter-line-like: series, scatterlines, **stem**, **streamplot**
- text-like: **annotation**
- barplot-like: barplot, hist, **waterfall**
- mesh-like: **boxplot**, **crossbar**, **hspan**, **vspan**, poly, **triplot**, **voronoiplot**
- image/heatmap-like: **datashader**, spy, volumeslices
- mixed: rainclouds

Special cases:
- arrows: shows (position, direction)
- band: shows (high position, low position)
- barplot: shows (bar position (x or y), bar height)
- contour, contour3d, contourf, tricontourf: shows discrete level of line or area
- density: shows (position, density value)
- errorbars: shows either +- delta or (+delta, -delta)
- hexbin: shows number of values in bin
- pie: shows section values
- rangebars: show value range (low .. high)
- violin: (TODO) like density
- Discarded: bracket, textlabel, tooltip

Notes on bold plots:
- annotation should also be discarded I think
- stairs should maybe discretize
- stem should as well and display max height
- stephist should probably behave like hist (discrete bar information)

Makie/src/makielayout/blocks/axis.jl
```diff
@@ -83,6 +83,12 @@ function update_axis_camera(scene::Scene, t, lims, xrev::Bool, yrev::Bool)
     )
 
     Makie.set_proj_view!(camera, projection, Makie.Mat4f(Makie.I))
+
+    # CairoMakie relies on eyeposition for mesh, surface, ...
+    # Technically the camera should be infinitely far away from the 0 plane
+    # because there is no perspective projection
+    center = 0.5 * Point2f(left + right, bottom + top)
+    scene.camera.eyeposition[] = Point3f(center..., 1e12)
```
ffreyer: This causes rendering changes to surface plots with shading in Axis. The camera-to-pixel vector is now basically parallel to (0, 0, -1), rather than changing across the surface. So the shading is also basically flat now. (The previous eyepositon value was defaulted to (1, 1, 1))

This fixes a niche issue in CairoMakie where a surface with a vertex at (1, 1, 1) gets blacked out faces. This is due to the camera-to-vertex vector becoming NaN via normalize. 
```julia
surface(0..1, 0..1, [1 0 0; 0 0 0; 0 0 1], axis = (type = Axis,))
```
image

Makie/src/compute-plots.jl
```diff
@@ -803,22 +803,48 @@ end
...
+            Conversion failed for $(P)$(conv_trait) with args:
+                $(typeof(args)) $(kw_str)
+            Got converted to:
+                $(typeof(converted))
+            $(P) requires to convert to argument types
+                $(types),
+            which convert_arguments didn't succeed in. To fix this overload \
+            convert_arguments(P, args...$(kw_convert)) for Type{<:$(P)} or $(PTrait) \
+            and return an object of the correct type.
+            $dim_converts_info
```
ffreyer, line 1058: I updated this to include some notes on dim converts. Example from a test failure:
```
  ArgumentError: 
  Conversion failed for Tooltip with args:
      Tuple{Vector{Makie.PlotElement{Scatter{Tuple{Vector{Point{2, Float32}}}}}}} 
  Got converted to:
      Tuple{Vector{Makie.PlotElement{Scatter{Tuple{Vector{Point{2, Float32}}}}}}}
  Tooltip requires to convert to argument types
      Tuple{AbstractVector{<:Union{NTuple{N, var"#s417"}, StaticArraysCore.StaticArray{Tuple{N}, var"#s417", 1}}} where {N, var"#s417"<:Real}, Union{Nothing, AbstractVector}},
  which convert_arguments didn't succeed in. To fix this overload convert_arguments(P, args...) for Type{<:Tooltip} or NoConversion() and return an object of the correct type.
  (Dim converts were not applied. This happens if `space = data` is not in data space or if the target type of the conversion is reachable without dim converts.)

```

ffreyer:
I think I fixed one of two problems with WGLMakie, the one visible in tests from b117521cbfd1a3d3ae01ad7b2851c06e5101196b. When multiple graphs are connected it can happen that `update!()` intercepts `resolve!()` and corrupts state because of how we lock graphs. With a Graph a connected to a Graph b, I believe the order of actions for this is:

1. `resolve!()` triggers, locking b
2. `resolve!()` recurses up, locking a
3. `resolve!()` evaluates edges in a, unlocking a once done
4. task switches to `update!()` in a, which locks (or waits for) a
5. `update!()` replaces the input in a and marks every dependent as dirty. This does not lock/wait for b
6. `resolve!()` continues, evaluating the remaining edges in b using old values, marking everything it resolves as clean/up-to-date/resolved and finally unlocks b

With this the nodes set in step 6 are marked as clean while they are outdated with respect to update happening in 5. I believe this is what happened in the test - WGLMakie started resolving `visible` for tooltip plots, DataInspector interjected an update of `visible` from the top and the state desynchronized.

I changed the locking logic of `resolve` to not unlock in step 3. Instead it now starts by locking every involved graph first, before doing any computation, and unlocks all of them together once the full resolve is done. This should fix the locking behavior, and locally fixed the leftover plots as far as I could tell.

---

The other issue here is that with WGLMakie some regions just refuse to generate a tooltip. As far as I can tell this happens when the cursor moves to an area that was previously occupied by a tooltip. The `pick_sorted` call in DataInspector then only sees the tooltip, even though it is set to `visible = false`. The `wgl_renderobject` also reports it as invisible and it is not rendered in this case. This happens repeatedly as it triggers `on(events.tick)`. Interestingly calling `pick_sorted` in the terminal with the same parameters as DataInspector resolves the issue (until it is recreated).

I also checked the javascript objects that picking returns. They also have `visible: false` (though this is at the end of pick_sorted. I guess asynchronous things could happen between rendering and that point, though I would expect flickering and inconsistencies if visibility toggles and on and off)

Makie/src/DataInspector/DataInspector.jl
```diff
@@ -1,7 +1,7 @@
 # Wrapper to allow and handle PlotElements with tooltips
 @recipe DataInspectorTooltip (elements::Vector{<:PlotElement},) begin
     documented_attributes(Tooltip)...
-end
+end false
```
SimonDanisch, line 4: that seems like an oversight?

ffreyer: No, I extended `@recipe` to take in another argument for skipping `export` statements, because I didn't want `DataInspectorTooltip` to be exported. I'm not sure if this is generally useful though, so I didn't put in work documenting this or making this particularly readable.

PR #5248, opened 2025-08-11
fix `KeyError: key :visible not found` in `getlimits(::Axis3)`

thchr:
This fixes an issue I see when using `Axis3` together with a `@recipe` specified in the format `@recipe MyPlot (x, ) begin ... end`, which apparently doesn't necessarily create a `.visible` attribute (unlike the `@recipe ... do Scene` version). Cf. Slack https://julialang.slack.com/archives/C8RQUU2KH/p1754910109166579.

The change follows the 2D axis approach in `getlimits(::Axis)`


SimonDanisch:
Thank you!

Issue #5245, opened 2025-08-09
Please document arrow rotation Quaternionf


chakravala:
I have been using `fig,ax,plt = streamplot(...)` with `plt.plot[2].rotation[]` for both 2D and 3D.

Previously, this was a `Vec3`, now it is a `Quaternionf` instead, and for 2D it's a scalar.

It's necessary for me to transform these, so it would be good if somebody could explain if the scalar is an angle in 2D and what exactly the components of the `Quaternionf` represent in the `rotation[]`.

The problem is, when I transform the `Quaternionf` the result is arrow pointing in the wrong direction, so clearly I'm not understanding how to transform these values or how they're used in Makie.

ffreyer:
`Quaternionf` was always the final target for rotations. You can get it from other rotation inputs with `to_rotation()`.

Currently you can still do `plt.plot[2].rotation[] = to_rotation(...)`, though we may change this in the future. I personally think it's rather hacky to mess with subplots like this and it can cause updates to get lost. (Specifically if you try to reset a compute node you have overwritten by setting it's input to the same value.) In this case you'd also have to update rotations every time you move the camera if you're plotting interactively. It would be much nicer to handle this directly from `streamplot` instead. What would you need for that?

chakravala:
What I need is to transform the coordinates of a streamplot (only once, initially when the plot is created). This problem is solved, except that the arrows will point in the wrong direction because they do not get transformed. Therefore, the arrows must be transformed separately. Specifically, the direction of the arrows must be obtained, transformed, and then updated into plot.

chakravala:
Here's an example using my software and Makie:
```Julia
using Grassmann, Cartan
using Makie, GLMakie
fun(x) = Chain(x[1]^4*exp(x[2]),exp(x[1])*cos(x[2]),sin(x[1])+x[2]^2*x[3])
streamplot(unitball(),fun.(unitball());gridsize=(11,11,11))
```
The streamplot is successfully transformed, into a `unitball()` but the arrows point in the wrong directions.

My packages extend the `streamplot` functionality using multivariable calculus. Until recently, I was able to transform the arrows as well, but since `Quaternionf` was introduced, I need to re-design how the arrow directions are handled.

Either that, or I need to create my own streamplot methods entirely from scratch to suit my advanced calculus purposes.

PR #5249, opened 2025-08-11
Fix inaccurate `register_projected_rotations_2d!()` with transform_func

ffreyer:
# Description

This fixes rotations not transforming correctly in 2D with more complex transform functions as seen in #5245. For 3D this adds `register_transformed_rotations_3d!()` for the same purpose.

```julia
f = Figure()
a = PolarAxis(f[1,1])
p = streamplot!(a, p -> Point2f(1, 0), 0..2pi, 0..5, gridsize = (10, 10))
f
```
| Before | After
| --- | --- |
| Screenshot 2025-08-11 184118 | Screenshot 2025-08-11 184103 |

In `streamplot`, projected rotations were calculated from direction vectors by taking their position `p` and generating a second position `p + dir`, projecting them and taking the difference afterwards. This was introduced to avoid aspect ratios from affecting scatter marker rotations. #5121 slightly tweaked this code to include transform functions and unified it with `contour` labels.

The problem with this code is that `dir` can be large relative to the rate of change in transform_func. In the example above angles are limited to a 0..2pi range. A normalized `dir = Vec2f(1, 0)` covers about 1/6 of that, which noticeably incorrect rotations.

Optimally one would get the local basis transformation matrix at the position where the rotation/direction applies, and use that to transform it. This can be done with `ForwardDiff.jacobian(p -> apply_transform(transform_func(plot), p), pos) * direction`. However `ForwardDiff` only works on Julia functions and calculating jacobians may in general be too expensive for what we need here. So instead this pr adds

```julia
function apply_transform_to_direction(f, position::VecTypes{D1, T1}, direction::VT, delta) where {D1, T1, D2, T2, VT <: VecTypes{D2, T2}}
    D = max(D1, D2)
    T = float_type(T1, T2)
    pos = to_ndim(Point{D, T}, position, 0)
    dir = to_ndim(Point{D, T}, normalize(direction), 0)
    p0 = apply_transform(f, pos .- delta .* dir)
    p1 = apply_transform(f, pos .+ delta .* dir)
    return normalize(to_ndim(VT, p1 .- p0, 0))
end
```

which effectively does what we did before, but centered and with a `delta` based on the data limits of the plot. If needed, this can be overwritten to use something more accurate for specific transform functions. Since this function uses a direction vector I also changed `register_projected_rotations_2d!()` to use a position + direction instead of two positions. 

## Type of change

Not sure, could be considered breaking because it removes `endpoint_name` or a bug fix because that was a problematic way of handling rotations.

## Checklist

- [x] add version for 3D (just transform_func, probably?)


asinghvi17:
> Jacobian of the transform_func at the given arrow positions using ForwardDiff

This sounds quite tough, I don't think e.g Proj (for GeoMakie etc.) will work with this, and ForwardDiff is pretty heavy as well I thought?

Seems like this will fail for any ccallable function as well as anything not written in a specific, differentiable way.

Instead of hardcoding could we encode some kind of preference system into functions, so that transform funcs can opt in to ForwardDiff support that lives in an extension?  Better yet would be an API to just get the jacobian of the transform func at some position, I am sure that we could encode that a lot more easily for many transform_funcs.

ffreyer:
Yea, fair points. I guess I should go back to numerical differentiation then, but make it aware of limits so it can choose reasonably close points. Maybe we can have a `local_transformation_matrix(func, pos, deltas)`?

chakravala:
> Jacobian of the transform_func at the given arrow positions using ForwardDiff

I don't have any issues with calculating my own Jacobian and controlling the transformation myself.

This `PolarAxis` thing is not what I want at all ... all I require is facilities for transforming the `Quaternionf` for the arrow directions, as that's the only main issue I am facing, I do not plan to use this `PolarAxis` feature. I only intend to transform the `Quaternionf` values myself, that's all I wanted to accomplish in my issue #5245

ffreyer:
As far as I could tell you apply your transformation through a transform_func and streamplot does not consider them correctly/at all for rotations. That's a general bug which I'm trying to fix here. PolarAxis also applies its transformation through a transform_func, so I used it as a Makie-only example.

For reference, this is your example with what I want to add in this pr (local changes):
image
I assumed that getting the arrows correctly aligned like this was your end goal. If not you should be able to use your old `Vec3` rotations by passing them through `to_rotation`. E.g.
 ```julia
f,a,p = streamplot(unitball(),fun.(unitball());gridsize=(11,11,11), arrow_size = 0.1)
p.plots[2].rotation[] = to_rotation([Vec3f(0,0,1) for _ in p.plots[2].rotation[]])
```

**Comment by chakravala** [comment]
@ffreyer thanks for explaining, that's great thank you.

>  If not you should be able to use your old Vec3 rotations by passing them through to_rotation. E.g.

Do you have an official method for converting `Quaternionf` to something like `Vec3`?

**Comment by ffreyer** [comment]
`q= to_rotation(vec)` calculates a rotation that maps `Vec3f(0,0,1)` to `vec`, so `q * Vec3f(0,0,1)` will invert that.

**Comment by chakravala** [comment]
Now, if you could also make `transform_func` work on `volume` plots or `contour` of volume spaces...

**Comment by chakravala** [comment]
by the way, as far as obtaining the Jacobian goes as a dependency, if Makie.jl depended on my packages, then you'd already have access to all that calculus by default without any extra dependencies besides my core Grassmann and Cartan packages, which also includes a geometric algebra implementation unifiying vectors and quaternions

**Comment by chakravala** [comment]
If you try the most recent commit of https://github.com/chakravala/Cartan.jl/commit/4e589d4d193207f551a8689dd337fa0e7490078b which I might release tomorrow, here's some test examples which already apply the `transform_func` but the graph does not get updated yet, not sure if your commit already fixes this:
```Julia
contour(unitball(),norm(unitball()))
Makie.volume(unitball(),norm(unitball()))
Makie.volume(unitball(),norm(unitball());algorithm=:iso,isovalue=0.3)
```

ffreyer:
For now I'm going to avoid using a jacobian. Doing `normalize(f(p + delta) - f(p - delta))` is pretty simple and fast and probably accurate enough to get a rotation that looks the correct. If there are issues with that we can always tweak it further in the future.

chakravala:
@ffreyer yes, `normalize(diff)` sounds appropriate and sufficient for these purposes. If `Makie` used my packages instead of `GeometryBascis`, then you wouldn't have to make these compromises.

SimonDanisch (approved):

Issue #5162, opened 2025-07-07
Position based Interpolation/Resampling API


ffreyer:
This is a planning issue for adding a interpolation/resampling API, that adds (removes?) or adjust points. This can be useful to improve the visual appeal of a plot and sometimes its correctness.

## Possible applications

- Smoothing functions, #5158 (wasn't there another issue for this?)
- maybe regularly spaced markers, #2563 (this might be difficult to fit into the interface)
- resampling based on transform_func (e.g. make a straight line in polar coordinates curved in Cartesian coordinates)

## Requirements

1. Resampling positions requires resampling of per-position attributes
2. Resampling may depend on plot type (e.g. smoothing scatter positions should probably do nothing?)
3. Resampling may depend on transform_func:
    - it may only act with a specific transform_func's (e.g. resample in polar coordinates but not Cartesian)
    - transform_func may affect sampling (e.g. log10 needs different point densities than linear transforms)
4. Resampling may include settings (e.g. smoothing weight)

## Thoughts

We could follow the `transform_func` API with this, i.e. add a `resampler` attribute and something like `apply_resampling()`. To satisfy 2+3, this would need to be:
```julia
apply_resampling(PlotType, resampler, positions, transform_func)
```
4 can be easily satisfied by allowing `resampler` to be a struct.

1 is more difficult to satisfy. We effectively need to save whatever interpolations the resampler generates so that we can apply them to other attributes. So instead of `apply_resampling` we'd need a `generate_resampling` as the user facing interface. If we assume interpolations to always be linear between two points, the output could be an array of `(index1, index2, interpolation_factor)`.
dendrogram has a resampling satisfying 1 implemented, but with a hard coded interpolation:
https://github.com/MakieOrg/Makie.jl/blob/eb1ad5b7dfef3b93106760e2b5a1ca59495c0dc9/Makie/src/stats/dendrogram.jl#L94-L136

Something I'm not sure about is whether transform_func-based resampling and user supplied resampling should combine somehow. For example, a smoothed line may not be sampled well for log or Polar coordinates. So maybe these should both happen, and maybe independently?

asinghvi17:
> whether transform_func-based resampling and user supplied resampling should combine somehow

Could the user have the option to do that in display space (post transform) or data space (pre transform)?  Providing that option should solve the issue IMO.  Then it's a question of what your transform is, which the user knows.

And who calls this...the backend or the plotting pipeline itself?

ffreyer:
[...]

PR #5252, opened 2025-08-14
implement a simple compression for value -> n_array_value -> value

SimonDanisch:
This mainly for text and WGLMakie, where we go from a single value (e.g. one color per text) to one color per glyph, which then is just one value repeated n times.
This is a "lazy" fix, since I'm just going from `n_values` -> `value` in WGLMakie specifically, where this has the biggest impact. Especially for text it has been difficult to not duplicate values per glyph, since a lot of the infrastructure expects arrays with value per glyph and we also can't easily switch between `T` and `Vector{T}`, which we allow for text specifically right now (other plots don't allow that).
The proper fix will require a rethink on how we handle attributes per X across plots, which we may tackle in https://github.com/MakieOrg/Makie.jl/issues/5162. 

This can be seen as a continuation of https://github.com/MakieOrg/Makie.jl/pull/4942, with the goal of making text more efficient for smoother axis updates in WGLMakie.

A quick benchmark of the improvements starting from before moving more of the text infrastructure to JS:

```julia
using WGLMakie, Bonito

rm(WGLMakie.WGL.bundle_file)
f, ax, pl = scatter(rand(Point2f, 10))
begin 
    limits!(ax, (0, 0.5), (0, 0.5))
    WGLMakie.poll_all_plots(f.scene)
    msgs, msg = Bonito.collect_messages() do 
        limits!(ax, (0, 2.2), (0, 2.2))
        WGLMakie.poll_all_plots(f.scene)
    end 
end
msg

# PR: 
"Send 8 messages with a total size of 3.197 KiB"
# 0.24.5:
"Send 10 messages with a total size of 4.223 KiB"
# 0.23 (with textureatlas in JS)
"Send 52 messages with a total size of 10.459 KiB"
# 0.22 (without js textureatlas) (WGLMakie v0.11.4 to be exact)
"Send 60 messages with a total size of 12.279 KiB"
```

### DIFF
--- a/Makie/src/basic_recipes/text.jl
+++ b/Makie/src/basic_recipes/text.jl
@@ -71,7 +71,6 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
     return
 end
 
-
 function per_glyph_getindex(x, text_blocks::Vector{UnitRange{Int}}, gi::Int, bi::Int)
     if isscalar(x)
         return x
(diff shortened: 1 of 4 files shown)


jkrumbiegel:
Maybe the per glyph thing is not the right abstraction anyway and this stuff should be range based. That would get you a single color entry for a single colored string automatically. For example, emoji wouldn't have colors at all, if we supported them. And some unicode characters together form other characters, in which case you can also not mix properties. But this is all the more complicated string formatting stuff that we've always ignored in Makie.

PR #5135, opened 2025-06-30
make LogTicks work with pseudolog10

Moelf:
fix #4856

```julia-repl

julia> using Revise, CairoMakie

julia> lines(-100:0.1:100,
           axis=(; yticks=LogTicks(-2:2), yscale=Makie.pseudolog10)
       )
```

image

PR #5268, opened 2025-08-22
fix negative sign in LogTicks with pseudolog10

Moelf:
fix bug in #5135

### DIFF
--- a/Makie/src/makielayout/lineaxis.jl
+++ b/Makie/src/makielayout/lineaxis.jl
@@ -640,7 +640,7 @@ function get_ticks(::Automatic, scale::LogFunctions, any_formatter, vmin, vmax)
 end
 
 # log ticks just use the normal pipeline but with log'd limits, then transform the labels
-function get_ticks(l::LogTicks, scale::Union{LogFunctions, typeof(pseudolog10)}, ::Automatic, vmin, vmax)
+function get_ticks(l::LogTicks, scale::LogFunctions, ::Automatic, vmin, vmax)
     ticks_scaled = get_tickvalues(l.linear_ticks, identity, scale(vmin), scale(vmax))
 
     ticks = Makie.inverse_transform(scale).(ticks_scaled)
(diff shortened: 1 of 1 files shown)


Moelf:
@jkrumbiegel this should fix the most offending visual bug. But now I'm wondering is the y-tick transformation bugged to begin with?
```julia
julia> fig = scatter(0:0.1:2;
           axis=(; yscale=Makie.pseudolog10)
       )
```

image

this seems, not linear between 0 and 1

Moelf:
I guess the semantics of LogTicks is not suitable if we have all of the following:
- data crossing `0`
- user wants to hit specific log ticks


Because `LogTicks()` technically can only express positive values....

Moelf:
Ok how about this, when user gives us pseudolog10 + LogTicks we will interpret it as "use it for positive log ticks, but mirror it into the negative if there's negative data", and add a zero tick by itself.

also cc @SimonDanisch

jaredjeya:
> this seems, not linear between 0 and 1

That's intended behaviour right? Pseudolog10 is defined as sign(x) * log10(x + 1), which is a smooth transformation that approaches log10(x) for large x but is linear for small x.

SimonDanisch:
@Moelf @t-bltg and @jaredjeya is this ready to be merged?
@jkrumbiegel is this solving your concerns?
I guess https://github.com/MakieOrg/Makie.jl/pull/5279 is independent?

Moelf:
this is not quite ready yet -- I'm not entirely sure the semantics of `LogTicks` work with negative numbers. The best we can do here is maybe mirror the ticks into negative? i.e. if users says `LogTicks(-2:1),` we would have `-10^1, -10^0, -10^(-1), -10^(-2), 0, 10^-2, 10^-1, 10^0, 10^1`?

and yes this is separate from #5279.

jaredjeya:
We might not necessarily want the same limits on positive and negative (e.g. something I might use it for is data which follows a log trend + noise, so it can fluctuate to or below zero, but not to large negative values). Could we make e.g. a DualLogTicks that takes two inputs - one for positive and one for negative ticks - or supply a tuple of two LogTicks? (I'm not familiar with the tick-placing code so I don't know how feasible that would be).

Separately, I think it might also make sense to have similar behaviour defined for Symlog10 and Asinh axes seeing as these are all different ways of accomplishing the same result (a log scale which goes smoothly and linearly through zero to represent negative values). I don't forsee that interacting with my PR #5279 since it looks like the code is agnostic of the fine details of the transformations (and yes, I can confirm it's separate to this PR).

Moelf:
>We might not necessarily want the same limits on positive and negative (e.g. something I might use it for is data which follows a log trend + noise, so it can fluctuate to or below zero, but not to large negative values)

I think this is fine with the mirroring approach, because `pseudolog10` is ~linear around ` 0`, we should see small fluctuation below zero just fine. Mirroring the log ticks to negative doesn't mean we will force the `ylims` to be there, just where we have ticks.

jaredjeya:
Just taking a look at the code - have you checked that the tick labels render correctly for negative exponents? e.g. $\pm 10^{-1}$? For `pseudolog10` this would be in the linear region but for my suggestion that this also cover `asinh` or `symlog10` scales (which should just be a one-line change to the method definition, I hope?) that could be in the logarithmic region.

PR #5274, opened 2025-08-26
docs: Add newline between "See also" and "Used for" in the docstrings

kunzaatko:
This makes it visually more distinct from the "See also" which is
intuitively a different matter that what the method is used for.

It could be also moved above the "See also". Perhaps this would be
better?


jkrumbiegel:
Yes, intuitively I think "see also" should be last

kunzaatko:
I reordered it so that `See also` is last and that the lines are not set apart by a paragraph break.

jkrumbiegel:
Thanks!

PR #5279, opened 2025-08-28
Fix Symlog10 to work with limits less than one

### LINKED ISSUE #3788: Symlog scale leads to weird results
_**UPD**: use MakieExtra.jl `SymLog()` for the same behavior as in matplotlib._

---

I’m trying to plot a 2d array with both positive an negative values, using a “symlog” colorscale in Makie.

This is what I get, looks weird – I cannot make any sense out of colorbar labels:

And this is what I expected it to be – result from matplotlib for the same array for comparison:

My issue here is with the colorscale specifically, not any other visual difference between Makie and matplotlib.

Full code for both:
```julia
# Makie:
let
	fig, ax, plt = image(A, colorscale=Makie.Symlog10(0.1), interpolate=false, colormap=:viridis)
	Colorbar(fig[1,2], plt)
	fig
end

# PyPlot:
let
	plt.figure()
	plt.imshow(A, norm=matplotlib.colors.SymLogNorm(0.1))
	plt.colorbar()
	plt.gcf()
end
```

- [x] are you running newest version (version from docs) ? – yes
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) – yes
- [x] What platform + GPU are you on? – Any backend, same result.

jaredjeya:
# Description

Fixes #3788

Corrects the Symlog10 function to work with hi or abs(lo) less than 1, and adds a new parameter which adjusts the size of the linear region. 

~~There aren't currently any unit tests for this functionality (or indeed any of the other axis transformations, as far as I can tell), so I haven't created any. Please let me know if you'd like that.~~

The nature of the change means that we can't recreate the old behaviour for limits > 1, and still have it remain consistent for limits < 1. Setting `linscale= (log10(hi) + log10(abs(lo)))/2` will obtain the previous behaviour exactly, but this breaks down when hi and abs(lo) are smaller than 1, which is why it's not the default. But it seems sensible to me that the linear region should remain a fixed width, if the user sets a small `hi` then there is presumably data smaller than `hi` that is of interest.

## Type of change

## Checklist


jaredjeya:
Changed my mind, added some tests for Symlog10. I put them in a testset for axis rescalings in case anyone adds more tests in the future.

jaredjeya:
Ah, I hadn't realised logrange was a Julia 1.11 feature! I'll fix the tests to avoid it.

jaredjeya:
I don't know why that test is failing. My code shouldn't even be touched by it, but the axis seems to have been moved slightly.

`reference/WGLMakie/deletion.png`
image

`recorded/WGLMakie/deletion.png`
image

Difference:
image


@simondanisch any ideas?

SimonDanisch:
I think we had some instability in the exact axis size in 1 of ~100 CI runs...
Lets just rerun the CI.

jaredjeya:
@SimonDanisch @ffreyer Can I get this looked over please?

Makie/src/layouting/transformation.jl
```diff
@@ -443,25 +443,41 @@ const pseudolog10 = ReversibleScale(
...
-        x
+"""
+    Symlog10([lo=hi,] hi; linscale=1)
+
+A scaling which is linear in the interval `[-lo, hi]` and logarithmic outside, thus representing both positive and negative values.
+
+The parameter `linscale` controls the half-width of the linear region (default: 1). This region is shifted so that zero maps to zero.
+
+If only one argument is given, `lo` is set to `-hi`, and the linear region is symmetric around zero.
+"""
```
ffreyer, line 491: `lo=hi` should be `lo=-hi` in the first line.

I think it would be good to pick some more descriptive variable names for `lo` and `hi`, at least for the docstring. Maybe `lower` and `upper` (bound/limit)? 

I also found it a bit hard to understand linscale. I think it would be helpful to clarify that `lo, hi` controls the input range and `linscale` the output range.

jaredjeya: I've taken some inspiration from the matplotlib documentation and made the description a bit clearer now, and renamed the variables as suggested.

jaredjeya:
There's now a random error popping up across the board in a test for the newly-added dynamic quantities feature (and it's not a failed reference test, but actually that the test itself throws an error). Again this isn't something my code interacts with at all! So I'm a little confused.

Edit: In fact, it's happening for everyone, so it's not me

PR #5291, opened 2025-09-09
Proof of concept/Prototype for rendering opaque objects in transparent volume plots

### LINKED ISSUE #5290: Transparency and depth issues when displaying a mesh and a volume
I'm trying to display a mesh together with a volume, see the following MWE. 

I cannot make the two display correctly together (here the sphere ends up either behind or in front of the volume, but not half inside and half outside as it should be). It's hard to see on the pictures, but try the MWE and rotate the figure to see what I mean. 

```
using GLMakie
using NIfTI

# From the example given in the mesh docs
brain = niread(Makie.assetpath("brain.nii.gz")).raw
mini, maxi = extrema(brain)
normed = Float32.((brain .- mini) ./ (maxi - mini))

fig = Figure(size=(1000, 450))
ax = Axis3(fig[1, 1])

# Make a colormap, with the first value being transparent
colormap = to_colormap(:plasma)
colormap[1] = RGBAf(0, 0, 0, 0)
volume!(ax, normed, algorithm=:absorption, absorption=4f0, colormap=colormap)

# Add a sphere (make it half inside, half outside the brain)
s = Sphere(Point3f(100, 100, 100), 100f0)

mesh!(ax, s)
wireframe!(ax, s)

fig
```

I guess it's due to the absorption algorithm that does not take the mesh into account. I tried playing with various options such as transparency or depth shift to no avail. Is this the expected behaviour? If so, is there a way to achieve this kind of visualization?

Image

Image

Thanks!

ffreyer:
# Description

Adds/Fixes #5290

Usually depth sorting works by comparing the depth of the currently drawn pixel with the recorded depth value in the depth buffer. If the new pixel is less deep it gets saved and the depth buffer updated, otherwise it gets discarded.

Volume plots render by sampling volume data on a ray. These samples then accumulate to generate the drawn color. Since there are multiple samples at multiple positions, there is no single depth value to output. So the normal depth testing does not work.

What we can do instead is have all the opaque geometry render first and then explicitly depth test each sample. If the sample position is less deep than the drawn geometry it contributes, otherwise sampling stops. This pr is a quick test for this.

```julia
using FileIO, GLMakie, NIfTI
brain = niread(Makie.assetpath("brain.nii.gz")).raw
mini, maxi = extrema(brain)
normed = Float32.((brain .- mini) ./ (maxi - mini))

begin
    # Necessary to get a valid depth buffer in volume
    # (likely wont render anything without)
    GLMakie.closeall()

    fig = Figure(size=(1000, 450))
    ax = Axis3(fig[1, 1])

    meshscatter!(
        Point3f.(50:50:300, 50:50:300, 0),
        marker = Rect3f(0, 0, 0, 1, 100, 400),
        markersize = 1,
        color = :green,
        shading = NoShading,
    )

    # volume must render after opaque geometry
    colormap = to_colormap(:plasma)
    colormap[1] = RGBAf(0, 0, 0, 0)
    vp = volume!(ax, normed, algorithm=:absorption, absorption=2f0, colormap=colormap)

    fig
end
```

Screenshot from 2025-09-09 02-44-08
Screenshot from 2025-09-09 02-44-28

TODO:
- [ ] adjust render order to draw volume plots after opaque plots
- [ ] figure out how to safely include the depth buffer in volume plots
- [ ] extend updates to other volume algorithms + WGLMakie
- [ ] maybe also try adding an order independent transparency version

## Type of change

## Checklist

### DIFF
--- a/GLMakie/assets/shader/volume.frag
+++ b/GLMakie/assets/shader/volume.frag
@@ -135,6 +135,24 @@ float rand(){
     return fract(sin(gl_FragCoord.x * 12.9898 + gl_FragCoord.y * 78.233) * 43758.5453);
 }
 
+#ifdef ENABLE_DEPTH
+
+uniform sampler2D depth_buffer;
+vec2 frag_uv = gl_FragCoord.xy / textureSize(depth_buffer, 0).xy;
+
+bool is_behind_geometry(vec3 sample_pos) {
+    vec4 frag_coord = projectionview * model * vec4(sample_pos, 1);
+    float sample_depth = 0.5 * (frag_coord.z / frag_coord.w + depth_shift + 1);
+    float recorded_depth = texture(depth_buffer, frag_uv).x;
+    return is_clipped(frag_coord) || (recorded_depth < sample_depth);
+}
+
+#else
+
+bool is_behind_geometry(vec3 sample_pos) { return false; }
+
+#endif
+
 vec4 volume(vec3 front, vec3 dir)
 {
     // The per-voxel alpha channel is specified in units of opacity/length.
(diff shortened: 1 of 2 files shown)


touste:
Thanks for the quick fix! As is, it already works for my simple case.

PR #5294, opened 2025-09-10
Document every plot attribute

### LINKED ISSUE #1112: boxplot attributes aren't fully documented
@piever @kimlaberinto

Compare https://github.com/JuliaPlots/Makie.jl/blob/master/src/stats/boxplot.jl#L31-L64 with http://makie.juliaplots.org/stable/plotting_functions/boxplot.html#Makie.boxplot
This should be a good beginner issue :)

### LINKED ISSUE #3829: Horizontal lines for boxplot.
### Feature description
The horizontal lines at IQR of the boxplot.

This might be a tiny thing but I found it affect how people percieve Boxplot and got 
confused easily.

### LINKED ISSUE #4209: Missing docs for `datashader`'s `transformation` attribute
In the current documentation there is no information about what is the `transformation` attribute of `datashader`. It currently just says:
> Defaults to automatic. No docs available.

Maybe a sentence can be added to explain what it does and what other values `transformation` can take, if any? (link to the docs: https://docs.makie.org/stable/reference/plots/datashader#transformation)

ffreyer:
# Description

This aims to add docstrings for every attribute that does not yet have one. 

Changes that aren't just docs:
- compute errors no longer double print source nodes
- added some functions for filtering `DocumentedAttributes`
- `mixin_colormap_attributes` now accepts `allow` and `exclude` to filter specific attributes
- `uv_offset_width` has been removes from `scatter` (and recipes that reuse its attributes) as the attribute has no effect
- `series` now passes `markercolor` when it's given
- `streamplot` function output type has been loosened to `VecTypes` to allow `Vec` outputs
- `timeseries` now forwards line attributes
- `waterfall` now reuses `barplot` attributes, adding label related attributes
- `boxplot` now reuses `crossbar` attributes which shouldn't change anything
- `boxplot` now converts `whiskerwidth === :match` to `1.0` which shouldn't change anything
- `crossbar` now forwards all applicable attributes to `poly` which adds some forgotten colormapping and generic attributes
- `qqplot/qqnorm` now have `markercolormap` and `markercolorrange` like `scatterlines`
- `hexbin` now includes generic attributes and passes all applicable attributes to scatter
- `stephist` now reuses `stairs` attributes which shouldn't change anything
- `hist` now reuses a subset of `barplot` attributes which should only change `label_size` to default to the themes fontsize
- `default_automatic` now returns `automatic` if both inputs are `automatic`

Fixes #4209
Fixes #1112
Closes #3829

## Type of change

- Documentation


Makie/src/basic_plots.jl
```diff
@@ -508,10 +564,16 @@ Plots a marker for each element in `(x, y, z)`, `(x, y)`, or `positions`.
     Sets the font to be used for character markers
     """
     font = "default"
-    uv_offset_width = (0.0, 0.0, 0.0, 0.0)
```
ffreyer, line 511: This isn't being used anymore so I removed it.

Makie/src/basic_recipes/tricontourf.jl
```diff
@@ -42,6 +46,7 @@ for specifying the triangles, otherwise an unconstrained triangulation of `xs` a
     If it's `nothing`, no band is added.
     """
     extendhigh = nothing
+    "Sets the color used for nan values in the generated contour."
     nan_color = :transparent
```
ffreyer, line 50: Is there any way to use `nan_color` in tricontourf? Adding a nan to z values either triggers an error or produces a broken plot.

Also `edges` doesn't seem to be used anywhere?

Makie/src/basic_recipes/series.jl
```diff
@@ -92,7 +122,7 @@ function plot!(plot::Series)
             scatterlines!(
                 plot, positions;
                 linewidth = linewidth, linecap = plot.linecap, joinstyle = joinstyle,
-                miter_limit = miter_limit, color = series_color, markercolor = series_color,
+                miter_limit = miter_limit, color = series_color, markercolor = markercolor,
```
ffreyer, line 125: Causes a refimg to fail because it makes `markercolor = :transparent` actually work:

Before:
image

After:
image

Makie/src/stats/hist.jl
```diff
@@ -99,41 +101,25 @@ Plot a histogram of `values`.
...
+    ))...
+    "Sets the gap between bars relative to their width. The new width is `w * (1 - gap)`."
+    gap = 0
+    "Sets the color of labels that are drawn outside of bars. Defaults to `label_color`"
     over_background_color = automatic
+    "Sets the color of labels that are drawn inside of/over bars. Defaults to `label_color`"
     over_bar_color = automatic
-    label_offset = 5
-    label_font = @inherit font
-    label_size = 20
```
ffreyer, line 131: I changed `label_size` to `@inherit fontsize` by reusing `barplot` attributes here. That causes another test failure:

Before:
image

After:
Screenshot from 2025-09-11 15-50-14

Imo this should be inheriting from the theme fontsize anyway, so it's not a test failure I should "fix"

PR #5313, opened 2025-09-25
Refactor video display to support more formats + VSCode

ztangent:
# Description

Refactors `Base.show` for `VideoStream` to:
- Support all web-mimes ("text/html", "juliavscode/html", etc.)
- Support GIF and WEBM display via "" and "<video>" tags
- Avoid unnecessary saving or conversion of video to MP4
- Fallback gracefully when size can't be determined from the scene

## Type of change

## Checklist


ztangent:
I realized while testing this on my own machine that previews of "webm" videos aren't supported in VSCode, even though the generated HTML for a webm video plays perfectly fine in a regular web browser. I could address this by adding a special case to convert from "webm" to "mp4" for display for the "juliavscode/html" MIME type, but that would make the code a little messier.

SimonDanisch:
>I could address this by adding a special case to convert from "webm" to "mp4" for display for the "juliavscode/html" MIME type, but that would make the code a little messier.

I wouldn't do that and leave it like this for now ;)

ztangent:
Cool! In that case I'm happy to merge if the maintainers are! (Seems like the failing tests are unrelated, as far as I can tell.)

PR #5317, opened 2025-09-30
Document transform_func interface

ffreyer:
# Description

This adds an explanation section for transformations, mainly explaining the `transform_func` interface. Also trims the reference sections on transformations a bit.

## Type of change

- documentation

### DIFF
--- a/docs/src/explanations/transformations.md
+++ b/docs/src/explanations/transformations.md
@@ -10,7 +10,7 @@ The latter is a matrix which encodes scaling, rotation and translation in 3D spa
 It can be set via `translate!()`, `scale!()`, `rotate!()` and `origin!()`.
 
 This page will discuss transformations from a developer perspective, i.e. how they work and how they can be extended.
-For a user perspective, see [transformations_reference_docs](@ref).
+For a user perspective, see the [Transformation](@ref transformations_reference_docs) reference docs.
 
 ## Inheritance
 


docs/src/explanations/transformations.md
```diff
@@ -0,0 +1,133 @@
...
+# are used to build a new bounding box.
+Makie.apply_transform(f::typeof(double), r::Rect3) = Rect3(f(origin(r)), f(widths(r)))
+
+# Defines the (1D) limits of an empty Axis using the transform function as the
+# xscale or yscale.
+Makie.defaultlimits(::typeof(double)) = (0.0, 10.0)
+
+# Defines the range of values that are valid for a given transform function in
+# an Axis.
+Makie.defined_interval(::typeof(double)) = Makie.OpenInterval(-Inf, Inf)
```
ffreyer, line 119: These two don't have default implementations. Should we add some?

asinghvi17: Do these methods exist for 2D -> 2D transforms (non separable)?

docs/src/explanations/transformations.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Defines how a direction vector at specific position gets transformed. By default
+# this transforms two points separated by `delta * direction` to calculate a new
+# post-transform direction. The result is normalized.
+# Used in `streamplot` and `contour` labels to rotate arrow markers and text
+function Makie.apply_transform_to_direction(f::typeof(double), position::VecTypes, direction::VecTypes, delta)
+    return direction
+end
```
asinghvi17, line 127: Does this need to preserve the norm of the direction vector, or should it return something normed to a unit length vector?

asinghvi17: Nevermind I somehow did not see that "the result should be normalized"

asinghvi17:
This looks great, thanks!  And good to know how to provide the direction vector.

docs/src/explanations/transformations.md
ffreyer: These are only used by Axis which requires separable transform funcs

asinghvi17: In this case I think it may be better to leave them out?  But it should error telling you what to do.

Issue #5193, opened 2025-07-15
StackOverflow error in GLMakie's heatmap used with a DateTime axis [Solved]


vlad-circadia:
The following error is observed in Makie v0.24.3, Julia v1.11.6. 

( `lines(t, rand(length(t))` works just fine. )
```
julia> using Dates, GLMakie

julia> t = range(now(), step=Second(1), length=100) |> collect;

julia> heatmap(t, 1:10, rand(10, length(t)))
ERROR: StackOverflowError:
Stacktrace:
  [1] steprangelen_hp(::Type{Float64}, ref::Tuple{Int128, Int128}, step::Tuple{Int128, Int128}, nb::Int64, len::Int64, offset::Int64)
    @ Base ./twiceprecision.jl:341
  [2] _linspace(::Type{Float64}, start_n::Int64, stop_n::Int64, len::Int64, den::Int64)
    @ Base ./twiceprecision.jl:730
  [3] range_start_stop_length(start::Float64, stop::Float64, len::Int64)
    @ Base ./twiceprecision.jl:660
  [4] _range
    @ ./range.jl:167 [inlined]
  [5] range
    @ ./range.jl:148 [inlined]
  [6] to_linspace
    @ ~/.julia/packages/Makie/aJUtI/src/conversions.jl:727 [inlined]
  [7] convert_arguments(P::CellGrid, x::LinRange{Millisecond, Int64}, y::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, z::Matrix{Float64}) (repeats 34820 times)
    @ Makie ~/.julia/packages/Makie/aJUtI/src/conversions.jl:362
  [8] convert_arguments(P::CellGrid, x::Vector{DateTime}, y::UnitRange{Int64}, z::Matrix{Float64})
    @ Makie ~/.julia/packages/Makie/aJUtI/src/conversions.jl:362
  [9] convert_arguments(::Type{Heatmap}, ::Vector{DateTime}, ::Vararg{Any}; kw::@Kwargs{})
    @ Makie ~/.julia/packages/Makie/aJUtI/src/conversions.jl:26
 [10] convert_arguments(::Type{Heatmap}, ::Vector{DateTime}, ::UnitRange{Int64}, ::Matrix{Float64})
    @ Makie ~/.julia/packages/Makie/aJUtI/src/conversions.jl:14
[...]

Issue #5187, opened 2025-07-12
vlines! not works for datetime


kongdd:
- `Makie`: v0.24.3

```julia
using Makie, GLMakie 
import Makie: date_to_number
using Dates

t = DateTime(2010):Hour(1):DateTime(2010, 1, 5)
# t = Date(2010):Day(1):Date(2010, 1, 30)

ntime = length(t)
y = rand(ntime)
fig = Figure(; size=(800, 400))
ax = Axis(fig[1, 1])
lines!(ax, t, y)
vlines!(ax, t[1]) # not work
# vlines!(ax, date_to_number(t[1])) # works

fig
```

```julia
Failed to resolve arg1:
[ComputeEdge] arg1 = compute_identity((points, ), changed, cached)
  @ [C:\Users\hydro\.julia\packages\ComputePipeline\zismN\src\ComputePipeline.jl:716](file:///C:/Users/hydro/.julia/packages/ComputePipeline/zismN/src/ComputePipeline.jl#line=715)
[ComputeEdge] points = (::MapFunctionWrapper(#987))((axis_limits, arg1, ymin, ymax, transform_func, ), changed, cached)
  @ [C:\Users\hydro\.julia\packages\Makie\aJUtI\src\basic_recipes\hvlines.jl:52](file:///C:/Users/hydro/.julia/packages/Makie/aJUtI/src/basic_recipes/hvlines.jl#line=51)
[Input] arg1
Due to ERROR: MethodError: Cannot `convert` an object of type DateTime to an object of type Float64
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  convert(::Type{T}, ::Unit) where T<:Number
   @ Makie [C:\Users\hydro\.julia\packages\Makie\aJUtI\src\units.jl:29](file:///C:/Users/hydro/.julia/packages/Makie/aJUtI/src/units.jl#line=28)
  convert(::Type{T}, ::Number) where T<:Number
   @ Base number.jl:7
  convert(::Type{T}, ::T) where T<:Number
   @ Base number.jl:6
  ...


Stacktrace:
  [1] macro expansion
    @ [C:\Users\hydro\.julia\packages\StaticArraysCore\7xxEJ\src\StaticArraysCore.jl:88](file:///C:/Users/hydro/.julia/packages/StaticArraysCore/7xxEJ/src/StaticArraysCore.jl#line=87) [inlined]
  [2] convert_ntuple
[...]

PR #4797, opened 2025-02-16
Add DateTime support for band

NAThompson:
# Description

Fixes #3932

Adds DateTime support for abscissas in `band!`.

## Type of change

Delete options that do not apply:

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [x] Added unit tests for new algorithms, conversion methods, etc.
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc.

Issue #4412, opened 2024-09-26
vspan and vlines do not support DateTime x-axis values


rbeeli:
- Makie v0.21.11
- Platform: Ubuntu 24.04.1 LTS, Julia 1.10.5

**MWE:**

```
using Makie, Dates

f = Figure();
ax = Axis(f[1, 1]);

# vlines!
xs = [DateTime(2020, 1, 1), DateTime(2020, 1, 2), DateTime(2020, 1, 3)];
Makie.vlines!(ax, xs);

# vspan!
x1 = [DateTime(2020, 1, 1), DateTime(2020, 1, 5), DateTime(2020, 1, 11)];
x2 = [DateTime(2020, 1, 2), DateTime(2020, 1, 7), DateTime(2020, 1, 12)];
Makie.vspan!(ax, x1, x2);

display(f);
```

**Errors:**
```
julia> Makie.vlines!(ax, xs);
ERROR: MethodError: Cannot `convert` an object of type DateTime to an object of type Float64

Closest candidates are:
  convert(::Type{T}, ::Unitful.Level) where T<:Real
   @ Unitful ~/.julia/packages/Unitful/GYzMo/src/logarithm.jl:22
  convert(::Type{T}, ::Unitful.Quantity) where T<:Real
   @ Unitful ~/.julia/packages/Unitful/GYzMo/src/conversion.jl:139
  convert(::Type{T}, ::Ratios.SimpleRatio{S}) where {T<:AbstractFloat, S}
   @ Ratios ~/.julia/packages/Ratios/FsiCW/src/Ratios.jl:51
  ...

Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/StaticArraysCore/7xxEJ/src/StaticArraysCore.jl:88 [inlined]
  [2] convert_ntuple
    @ ~/.julia/packages/StaticArraysCore/7xxEJ/src/StaticArraysCore.jl:84 [inlined]
  [3] Point{2, Float64}(x::Tuple{DateTime, Float64})
    @ GeometryBasics ~/.julia/packages/GeometryBasics/ebXl0/src/fixed_arrays.jl:21
  [4] StaticArray
    @ ~/.julia/packages/StaticArrays/MSJcA/src/convert.jl:173 [inlined]
  [5] (::Makie.var"#741#744"{…})(val::DateTime, mi::Int64, ma::Int64)
[...]

PR #5323, opened 2025-10-07
Improve dim converts recipe compatibility, usability

ffreyer:
# Description

This pr adds `argument_dims()`  as an overloadable function similar to `convert_arguments()` to allow recipes to define how their arguments map to plot dimensions. With that you can specify which dim_convert applies to first, second, ... argument by returning the index of the appropriate dimension/dim convert. This aims to fix:
- #3889
- #3896
- #3897, #4176
- #3932, closing #4797
- some of #3947
- #3973
- #4069, #4412, #5187
- #4409

This pr also aims to make some usability improvements:
- simplify units in axis labels by allowing dim_converts to generate label suffixes, fixing #3890
- add more dim convert related attributes to blocks to avoid needing to create dim converts manually

## Type of change

Likely breaking for dim_convert implementations since moving options to blocks means they need to be passed to `get_ticks`.

## Checklist

### DIFF
--- a/Makie/src/compute-plots.jl
+++ b/Makie/src/compute-plots.jl
@@ -440,42 +440,86 @@ function add_convert_kwargs!(attr, user_kw, P, args)
     end
 end
 
-function add_dim_converts!(attr::ComputeGraph, dim_converts, args, input = :args)
-    if !(length(args) in (2, 3))
-        # We only support plots with 2 or 3 dimensions right now
-        map!(attr, :args, :dim_converted) do args
-            return Ref{Any}(args)
-        end
-        return
+function add_dim_converts!(::Type{P}, attr::ComputeGraph, dim_converts, args, input = :args) where {P}
+    # Get dim of each argument. This needs to be reactive if we allow dynamic
+    # attributes that change dim-mapping, e.g. direction
+    kwarg_names = argument_dim_kwargs(P)
+    map!(attr, [input, kwarg_names...], :arg_dims) do args, kwargs...
+        nt = NamedTuple{kwarg_names}(kwargs)
+        return argument_dims(P, args...; nt...)
     end
 
-    inputs = Symbol[]
-    for (i, arg) in enumerate(args)
+    return add_dim_converts!(attr, dim_converts, args, input, attr.arg_dims[])
+end
+
+function add_dim_converts!(attr::ComputeGraph, dim_converts, args, input, dim_tuple)
+    error("`arguments_dims() must return a `Tuple` of integers or `Nothing` but returned $dim_tuple")
+end
+
+function add_dim_converts!(attr::ComputeGraph, dim_converts, args, input, dim_tuple::Nothing)
+    # Dimensions of arguments not defined, so we can't map them to dim_converts.
+    # TODO: Is this an error case? I.e. is this function required to apply
+    # dim_converts or could not applying them be valid?
+    map!(attr, :args, :dim_converted) do args
+        return Ref{Any}(args)
+    end
+    return
+end
+
+function add_dim_converts!(attr::ComputeGraph, dim_converts, args, input, dim_tuple::Tuple)
+    # TODO: Could this be valid? E.g. excluding some trailing args? Or returning
+    # more dims than needed (e.g. (1,2,3) for (xs, ys) and (xs, ys, zs))?
+    if length(args) != length(dim_tuple)
+        error("Number of dimensions ($(length(dim_tuple))) does not match number of arguments ($(length(arguments)))")
+    end
+
+    # This sets conversions per dimension if they have not already been set.
+    # If a recipe has multiple arguments for one dimension that dimension may
+    # be set multiple times here (but only the first one will actually be used)
+    for (i, arg) in zip(dim_tuple, args)
         update_dim_conversion!(dim_converts, i, arg)
+    end
+
+    # Add input containing Symbol(:dim_convert_, i) which triggers when the
+    # conversion changes. (One per dimension, so use unique on dim_tuple)
+    # Note that the order in dim_convert_names is important
+    dim_convert_names = Symbol[]
+    for i in 1:maximum(dim_tuple)
         obs = convert(Observable{Any}, needs_tick_update_observable(Observable{Any}(dim_converts[i])))
         converts_updated = map!(x -> dim_converts[i], Observable{Any}(), obs)
         add_input!(attr, Symbol(:dim_convert_, i), converts_updated)
-        push!(inputs, Symbol(:dim_convert_, i))
+        push!(dim_convert_names, Symbol(:dim_convert_, i))
     end
-    return register_computation!(attr, [input, inputs...], [:dim_converted]) do (expanded, converts...), changed, last
-        last_vals = isnothing(last) ? ntuple(i -> nothing, length(converts)) : last.dim_converted
-        result = ntuple(length(converts)) do i
-            return convert_dim_value(converts[i], attr, expanded[i], last_vals[i])
+
+    # Apply dim_convert
+    register_computation!(
+            attr, [input, :arg_dims, dim_convert_names...], [:dim_converted]
+        ) do (expanded, dims, converts...), changed, last
+
+        last_vals = isnothing(last) ? ntuple(i -> nothing, length(dims)) : last.dim_converted
+        result = ntuple(length(dims)) do i
+            # argument i is associated with the dim convert of dimension dims[i]
+            return convert_dim_value(converts[dims[i]], attr, expanded[i], last_vals[i])
         end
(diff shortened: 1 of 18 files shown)


Beforerr:
Hi, I am just wondering if this pr would fix #5193. If so I would like to try this pr.

ffreyer:
> Hi, I am just wondering if this pr would fix #5193. If so I would like to try this pr.

d4d8ba6470f7cbdeb4a91846e7cdd66f8c9e86fc should fix it

ffreyer:
I now added conversion target types for almost all recipes because I'm worried there will be issues otherwise. If you define a recipe that uses a plot that does not have them set, you will end up with `status = nothing` from `convert_arguments`. That leads into trying dim converts. If `argument_dims()` are set for the child plot, which is now usually the case, dim converts will be applied. If the recipe already applied dim converts, then the child plot will error as numeric arguments are not compatible (when dim converts are used/needed).

Makie/src/basic_plots.jl
```diff
@@ -361,7 +355,7 @@ const VecOrMat{T} = Union{AbstractVector{T}, AbstractMatrix{T}}
 Plots a surface, where `(x, y)` define a grid whose heights are the entries in `z`.
 `x` and `y` may be `Vectors` which define a regular grid, **or** `Matrices` which define an irregular grid.
 """
-@recipe Surface (x::VecOrMat{<:FloatType}, y::VecOrMat{<:FloatType}, z::VecOrMat{<:FloatType}) begin
```
ffreyer, line 365: Does z actually allow vectors? I didn't allow it in the conversion trait...

Makie/src/compute-plots.jl
```diff
@@ -440,61 +440,143 @@ function add_convert_kwargs!(attr, user_kw, P, args)
...
+    defaults = default_theme(nothing, P)
+    space = to_value(get(user_kw, :space, get(defaults, :space, :data)))
+
+    if !is_data_space(space)
+        # dim converts do not apply in relative, pixel or clip space
+        map!(attr, :args, :dim_converted) do args
+            return Ref{Any}(args)
+        end
+    elseif force_dimconverts && needs_dimconvert(dim_converts)
+        add_dim_converts!(P, attr, dim_converts, args, args_converted, user_kw)
```
ffreyer, line 593: The changes here are:
1. relative, pixel and clip space skip dim converts
2. `need_dimconverts()` now only triggers when `force_dimconverts == true` for the plot, which happens by default for plots to a scene but not for plots to another plot. This was necessary to avoid forcing dim converts in recipes, where the parent had already done the conversion. Previously this was decided based on whether the argument was a `Vector{<:Point}`, more or less. `Axis3` also needed this behavior to be overwritable for axis decorations, so you can pass `force_dimconverts = false` to plots like an argument.
3. the `force_dimconvert` and `status == nothing` branches now apply recursive/early convert_arguments like the `false` branch did. (This moved to `add_dim_converts`)

Makie/src/coretypes.jl
```diff
@@ -163,3 +163,6 @@ Base.broadcasted(f, a, b::EndPoints) = EndPoints(f.(a, b.data))
 Base.:(==)(a::EndPoints, b::NTuple{2}) = a.data == b
 # Something we can convert to an EndPoints type
 const EndPointsLike = Union{ClosedInterval, Tuple{Real, Real}}
+
+const RangeLike = Union{AbstractVector{T}, ClosedInterval{T}, Tuple{T, T}} where {T}
```
ffreyer, line 167: This type changed from
`Union{AbstractVector, ClosedInterval, Tuple{Real, Real}}`
i.e. the eltype is now settable instead of being partially fixed to Real

Makie/src/dim-converts/dim-converts.jl
```diff
@@ -66,17 +66,41 @@ end
...
+show_dim_convert_in_ticklabel(dc::Union{AbstractDimConversion, Nothing}, ::Automatic) = show_dim_convert_in_ticklabel(dc)
+show_dim_convert_in_ticklabel(::Union{AbstractDimConversion, Nothing}) = false
+show_dim_convert_in_ticklabel(::Union{AbstractDimConversion, Nothing}, option::Bool) = option
+
+# Should this trigger an error or just return ""?
+get_label_suffix(dc, format, use_short) = apply_format(get_label_suffix(dc, use_short), format)
+get_label_suffix(dc, format) = apply_format(get_label_suffix(dc), format)
+get_label_suffix(dc, ::Bool) = get_label_suffix(dc)
+get_label_suffix(dc) = error("No axis label suffix defined for conversion $dc.")
+get_label_suffix(dc::Union{Nothing, NoDimConversion}) = ""
```
ffreyer, line 94: I set this up to error when setting `ax.x_unit_in_label = true` for a dim convert that doesn't implement `get_label_suffix`. We could also have it return an empty suffix, but then we would probably want to not automatically apply the format. Because that results in `"[]"`.
And if this does error we should probably remove the `Nothing/NoDimConversion` case.

Makie/src/makielayout/types.jl
```diff
@@ -297,6 +297,33 @@ Axis(fig_or_scene; palette = nothing, kwargs...)
...
+        Formatter for the ylabel suffix generated from dim_converts. Can be a
+        Format.jl format string or a callback function acting acting on the
+        string or rich text generated from the dim convert.
+        Can also be a plain String replacing an active dim_convert label.
+        """
+        ylabel_suffix = "[{}]"
+        "Switches between short and long x units, e.g. \"s\" vs \"Second\""
+        use_short_x_units::Bool = true
+        "Switches between short and long y units, e.g. \"s\" vs \"Second\""
+        use_short_y_units::Bool = true
```
ffreyer, line 329: These probably need some renaming to fix with Axis attributes better? I guess just remove all the `_`?

Makie/src/dim-converts/argument_dims.jl
```diff
@@ -0,0 +1,111 @@
...
+`xs, ys` and `xs, ys, zs` vectors (or values) as well as vectors of `VecTypes`.
+The latter also allows multiple vectors with matching inner dimension `VecTypes{N}`.
+If included via `argument_dim_kwargs()`, `:direction` and `:orientation` are also
+handled by the default path in the 2D case. For this `direction == :y` and
+`orientation == :vertical` are considered neutral, not swapping dimensions.
+Examples marked `# default` above mirror the default path.
+
+Note that the `::Type{<:Plot}` methods take precedence over `::ConversionTrait`.
+"""
+function argument_dims(PT, args...; kwargs...)
```
ffreyer, line 45: Do we want to export `argument_dims` and `argument_dim_kwargs`?

SimonDanisch: Nope... Also, I think it would be nice to get rid of arguments_dim_kwargs!

jkrumbiegel:
As this is similar to what AoG needs to do, maybe we can compare the two. Does this implementation have something that can respect the direction keywords and similar attributes? Also in AoG I have a DeltaY scale for errorbar for example where the unit needs to match. For example a DateTime axis can't have a DateTime as an error magnitude.

ffreyer:
> As this is similar to what AoG needs to do, maybe we can compare the two. Does this implementation have something that can respect the direction keywords and similar attributes? 

`argument_dims` can include attributes as kwargs (without removing them and with reactiveness), so direction is handled there.

>Also in AoG I have a DeltaY scale for errorbar for example where the unit needs to match. For example a DateTime axis can't have a DateTime as an error magnitude.

Didn't do anything like that. This pr just tags args as "should convert with the n-th dim_convert` or no dim convert.

icweaver:
This is amazing! Thanks so much for these new features and QoL improvements, it's been awesome trying things out.

I'm excited to try and start updating bits on [DimensionalData.jl's Makie extension](https://github.com/rafaqz/DimensionalData.jl/blob/main/ext/DimensionalDataMakieExt.jl) to see how doable it would be to get, e.g., a fully labeled heatmap and colorbar going from a single `DimArray`.

Now that units are explicit for both Unitful and DQ, would it make sense to unify `UnitfulConversion` and `DQConversion` now that they're essentially doing the same thing?

https://github.com/MakieOrg/Makie.jl/blob/7feb429b0d3a374a6310058021379b9441e9d458/Makie/src/dim-converts/unitful-integration.jl#L97-L99

https://github.com/MakieOrg/Makie.jl/blob/7feb429b0d3a374a6310058021379b9441e9d458/Makie/src/dim-converts/dynamic-quantities-integration.jl#L25-L27

ffreyer:
Fixes #5458

ffreyer:
TODO:
Check if density is treated incorrectly here because of #4113

Works fine

Issue #5318, opened 2025-09-30
Updating an observable fails for rasterstack with missing values


tiemvanderdeure:
Rasters tests are currently failing because of the following example:

```
using Rasters, RasterDataSources, ArchGDAL, GLMakie
ENV["RASTERDATASOURCES_PATH"] = tempdir()

st = RasterStack(WorldClim{Climate}; month = 1)
stack_obs = Observable(st)
fig = Rasters.rplot(stack_obs;
    Colorbar=(; height=Relative(0.75), width=5)
) 
record(fig, "rplot.mp4", 1:12; framerate = 3) do i
    rs = RasterStack(WorldClim{Climate}; month = i)
    stack_obs[] = rs
end 
```

This fails because `is_same` seems to return `missing` on this raster, but I'm having a super hard time reproducing this. I can sometimes trigger it, but then this just runs in the same REPL session - very mysterious. I haven't been able to reproduce it with anything else than a RasterStack. 

Stacktrace (from just `stack_obs[] = rs`:
```
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
  [1] _setproperty!(attr::ComputePipeline.ComputeGraph, key::Symbol, value::Raster{…})
    @ ComputePipeline C:\Users\tsh371\.julia\packages\ComputePipeline\03tW7\src\ComputePipeline.jl:490     
  [2] #39
    @ C:\Users\tsh371\.julia\packages\ComputePipeline\03tW7\src\ComputePipeline.jl:500 [inlined]
  [3] lock(f::ComputePipeline.var"#39#40"{ComputePipeline.ComputeGraph, Symbol, Raster{…}}, l::ReentrantLock)
    @ Base .\lock.jl:232
  [4] setproperty!
    @ C:\Users\tsh371\.julia\packages\ComputePipeline\03tW7\src\ComputePipeline.jl:499 [inlined]
  [5] (::ComputePipeline.var"#57#58"{…})(new_val::Raster{…})
    @ ComputePipeline C:\Users\tsh371\.julia\packages\ComputePipeline\03tW7\src\ComputePipeline.jl:783     
  [6] #invokelatest#2
    @ .\essentials.jl:1055 [inlined]
  [7] invokelatest
    @ .\essentials.jl:1052 [inlined]
  [8] notify
    @ C:\Users\tsh371\.julia\packages\Observables\YdEbO\src\Observables.jl:206 [inlined]
[...]

PR #5327, opened 2025-10-10
use isequal instead of == in `is_same`

tiemvanderdeure:
closes https://github.com/MakieOrg/Makie.jl/issues/5318

### DIFF
--- a/ComputePipeline/test/unit_tests.jl
+++ b/ComputePipeline/test/unit_tests.jl
@@ -843,3 +843,15 @@ end
         @test !ComputePipeline.isdirty(out)
     end
 end
+
+@testset "is_same NaN and missing" begin
+    # vectors
+    v = [NaN, missing, 1]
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
I checked locally and this does seem to fix the issue with the Rasters docs. But I still haven't managed to make a more simple example consistently throw this error so I don't really have a good test to add.

tiemvanderdeure:
What should I do to move this one forward? Should I add a trivial test like `is_same([NaN, missing], [NaN, missing])`?

SimonDanisch:
>What should I do to move this one forward? Should I add a trivial test like is_same([NaN, missing], [NaN, missing])?


Yeah that sounds good. Maybe Dict/Vector/single?

asinghvi17:
This is needed for Rasters recipes to work.

SimonDanisch:
Thanks, merged this in https://github.com/MakieOrg/Makie.jl/pull/5398

PR #5328, opened 2025-10-10
[WIP] implement spawning tasks in compute graph

SimonDanisch:
This implements:
```julia
map!(compute_graph, inputs, outputs) do inputs 
   return @spawn longrunning(inputs)
end
# or with `register_computation!` the task needs to return a tuple
register_computation!(graph, [:a, :b], [:sum, :product]) do inputs, changed, cached
    return Threads.@spawn (inputs.a + inputs.b, inputs.a * inputs.b)
end
```

### DIFF
--- a/.github/workflows/reference_tests.yml
+++ b/.github/workflows/reference_tests.yml
@@ -153,7 +153,7 @@ jobs:
       - name: Run the tests
         continue-on-error: true
         run: >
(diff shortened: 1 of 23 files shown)


SimonDanisch:
The current implementation seems to be too complex with performance reduction for all plot types.
I like the API so we still may want to have this at some point - but maybe rather when we refactor the internals a bit to make this easier.

SimonDanisch:
Closing in favor of: https://github.com/MakieOrg/Makie.jl/pull/5417

Issue #4980, opened 2025-05-15
contourf ignore colorscale keyword (both in CairoMakie/GLMakie)


dlfivefifty:
- [X] what version of Makie are you running? (`]st -m Makie`) v0.22.5
- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [X] What platform + GPU are you on? Mac OS X 
- [X] Which backend are you using? GLMakie or CairoMakie
- [X] Please provide an minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

The following ignores the `colorscale` keyword:
```julia
using CairoMakie
x = range(-3,3,10)
errs = 10.0 .^ (-rand([1,2,3], 10, 10))
contourf(x, x, errs)
contourf(x, x, errs; colorscale=log10) # identical plot, colorscale is ignored
```

I would have expected it to be have like `heatmap`:
```julia
heatmap(x, x, errs)
heatmap(x, x, errs; colorscale=log10) # has different colors
```

asinghvi17:
Looks like `colorscale` and `alpha` are not passed through here (but the rest of the color attributes are):

https://github.com/MakieOrg/Makie.jl/blob/f200f3d573b4f892d81fa38f25eacec09bc36815/src/basic_recipes/contourf.jl#L213-L226

might be the same for contour as well.  Should be a quick PR?

dlfivefifty:
Was this fixed? I managed to make this figure:

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/a0825d53-f557-4b12-a334-873b5648e398" />

dlfivefifty:
oh wait... I think specifying the levels helped here.

asinghvi17:
Not fixed in contourf.  Someone had done it to tricontourf, though, it was the latest commit to that file so you could try and copy that change more or less.  It's just passing those attributes through to the plot call.

dlfivefifty:
I see. Having played with it a bit more I think its that my levels were close enough that the non-scaled colours were OK. If I made them range from 1E-16 to 1E-8 then all the very small numbers become the same colour. So definitely not fixed.

lanceXwq:
This issue should be addressed by #5027. Please take a look. 😃

Issue #4374, opened 2024-09-17
`contourf` should support standard colormap attributes


asinghvi17:
Currently, `contourf` only supports `colormap` and `colorscale`, but computes `colorrange` internally:

https://github.com/MakieOrg/Makie.jl/blob/65a718195a30fc4f3775f5457357e1ba35248745/src/basic_recipes/contourf.jl#L73-L79

and there's no support for explicit `highclip` or `lowclip` colors.

Should we compute the levels and colorrange in `calculated_attributes!(p::Contourf)`, and only create polygons in the recipe?  Then we can expose the `colorrange` attribute as well.

cf. https://julialang.slack.com/archives/C8RQUU2KH/p1726592043000659

PR #5330, opened 2025-10-13
Use `mixin_colormap_attributes` in `contourf `and `tricontourf`

### LINKED ISSUE #4016: `colorrange` is an invalid attribute for `contourf`

lanceXwq:
# Description

Pretty much the same as #5027, but this PR is based on the latest master.

Fixes #4016 (perhaps related to #4374 and #4980 as well) by updating `contourf` and `tricontourf` to use `mixin_colormap_attributes`. This PR should enable support for `colorrange`, `alpha`, and `colorscale` attributes in both `contourf` and `tricontourf`.

One reference test (Colorbar for recipes.png) is currently failing. However, based on manual inspection, the reference image may be incorrect. Specifically, the reference image was generated with `colorscale=sqrt` https://github.com/MakieOrg/Makie.jl/blob/10d65fd67bb269ea96c6748d754fd3544bb3051c/ReferenceTests/src/tests/figures_and_makielayout.jl#L624, but the resulting image was identical to one generated with `colorscale=identity`.

Current reference image (`colorscale=sqrt`):
image

With this PR (`colorscale=identity`):
image

With this PR (`colorscale=sqrt`):
image

It is not very clear to me why the `contourf` plot in the third screenshot only shows two colors while the colorbar has three. I would appreciate some hints if this is not an expected behavior

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@
 - Added new `annotation` recipe which can be used for labeling many data points with automatically non-overlapping labels, or for more bespoke annotation with manually chosen positions and connecting arrows [#4891](https://github.com/MakieOrg/Makie.jl/pull/4891).
 - Fixed precompilation bug in julia dev 1.13 [#5018](https://github.com/MakieOrg/Makie.jl/pull/5018).
 - Fixed screen not open assertion and `Makie.isclosed(scene)` in WGLMakie [#5008](https://github.com/MakieOrg/Makie.jl/pull/5008).
+- Fixed `(tri)contourf` such that `colorrange` is a valid attribute. `colorscale` also starts to work with this fix [#5027](https://github.com/MakieOrg/Makie.jl/pull/5027).
 
 ## [0.22.7] - 2025-05-23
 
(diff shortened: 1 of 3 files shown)


asinghvi17:
MakieCore no longer exists, could you rebase this PR on latest master and then just delete the `MakieCore.` stuff?  That is now in the Makie module directly and should get the PR working again.

lanceXwq:
> MakieCore no longer exists, could you rebase this PR on latest master and then just delete the `MakieCore.` stuff? That is now in the Makie module directly and should get the PR working again.

The PR should be working again, but I am not sure if everything is working as expected. Please see my updated comment above. Thanks!

thchr:
This would be great to have - especially being able to pass in `colorrange` to `contourf`. What needs to be done to merge?

PR #5332, opened 2025-10-14
`merge` precedence fix for `Attributes`

### LINKED ISSUE #1939: [inconsistency] `merge` on `Attributes` does not work in the same way as the one in the standard library
```
julia> at1 = Attributes(;c = 1)
> Attributes with 1 entry:
  c => 1

julia> at2 = Attributes(;c = 2)
> Attributes with 1 entry:
  c => 2

julia> merge(at1, at2)
> Attributes with 1 entry:
  c => 1
```
Where in the standard library as it states in the docs:
> Construct a merged collection from the given collections. If necessary, the types of the resulting collection will be promoted to
  accommodate the types of the merged collections. If the same key is present in another collection, the value for that key will be the
  value it has in the last collection listed. See also mergewith for custom handling of values with the same key.

And it works as stated.
```
julia> d1 = Dict(:a => 1)
> Dict{Symbol, Int64} with 1 entry:
  :a => 1

julia> d2 = Dict(:a => 2)
> Dict{Symbol, Int64} with 1 entry:
  :a => 2

julia> merge(d1, d2)
> Dict{Symbol, Int64} with 1 entry:
  :a => 2
```
Wouldn't it be easier just to use the `merge` on  the underlining `Dict` and keep it consistent with the expectation of people that are familiar with the `merge` function? If it is used somewhere deeper in the pipeline, it should not be an expansion of the `merge` method and be called something else... (You can assign me)

kunzaatko:
Fixes #1939
Fixes the functions themselves and changes the code to reflect it. Adds tests for the changes. This is a very breaking change for any user who works with themes and merges them together. It is better to make the change now rather than later. It should be warned about in the CHANGELOG and the release notes vehemently since I assume that some people will have to fix their code to reflect it.

There are 3 places that I documented in the comments, where I believe that the order of arguments in the current implementation is incorrect logically.
This requires somebody to confirm that the implementation is incorrect and then to update the reference images to make the tests pass.
In the current state, I just replaced the uses of `Base.merge!` previously used by `leftmerge!` which has the left precedence as is `Base.merge!` implementation in `master`.
`Base.merge!` is now implemented so that it has the correct right precedence and is not used in the package only overloaded for users in the correct way.
Should the places, where it can be done use `Base.merge!` implemented correctly now?


kunzaatko:
The places that `merge` of `Attributes` is used is small. Only
```
Makie/src/compute-plots.jl
557:            return merge(value, Attributes(Dict{Symbol, Any}(pairs(user_attributes[key]))))

Makie/src/recipes.jl
431:        return merge(default_theme(scene, T), default_theme(T))
```

`merge` however is implemented through `merge!` as  
```julia
Base.merge(target::Attributes, args::Attributes...) = merge!(deepcopy(target), args...)
```
at `Makie/src/attributes.jl:83`

kunzaatko:
The attribute merging is implemented in the function `Makie.merge_attributes!`
https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/attributes.jl?plain=1#L228-L243
which does not overwrite for values that are already present in the target. The function is used in the `Base.merge!` implementation which also has left precedence, which is inconsistent with the `Base` implementation of `merge!`.

kunzaatko:
The precedence used in `merge!`
https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/theming.jl?plain=1#L283-L289
is incorrect in the precedence since with the current precedence of `merge!` the keyword arguments are overriden by the theme.

> [!IMPORTANT] 
> Is this correct or was it truly a bug?

kunzaatko:
> [!IMPORTANT]
> Is the precedence here correct?

https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/makielayout/blocks/legend.jl?plain=1#L614-L625

If I would like to override the label with a keyword argument, should it be allowed? Now it is not...

> [!NOTE]
> It is never called in the package with keyword arguments passed... This change would only affect the potential external (user) calls of `LegendEntry`.

kunzaatko:
Here
https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/compute-plots.jl?plain=1#L554-L569
there was incorrect precedence. The default value is preferred before the user supplied if there are matching keys.

> [!IMPORTANT]
> Is it truly incorrect as I think?

ffreyer:
> The precedence used in `merge!`
> 
> https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/theming.jl?plain=1#L283-L289
> 
> is incorrect in the precedence since with the current precedence of `merge!` the keyword arguments are overriden by the theme.

Based on the docstring it seems like this is intended to be used with either attributes or kwargs, not both. But I would agree that this should prioritize kwargs over attributes.

---

> Is the precedence here correct?
> 
> https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/makielayout/blocks/legend.jl?plain=1#L614-L625
> 
> If I would like to override the label with a keyword argument, should it be allowed? Now it is not...

I don't think `LegendEntry` is user facing, since it has no docstring and no mention in the online docs either. Maybe the kwargs being there (and it being exported) is a leftover from an earlier version of `Legend`? Maybe @jkrumbiegel has some thoughts on this?

---

> Here
> 
> https://github.com/MakieOrg/Makie.jl/blob/705045bd5a7c636185761ffb0f5d5d395656b5f6/Makie/src/compute-plots.jl?plain=1#L554-L569
> 
> there was incorrect precedence. The default value is preferred before the user supplied if there are matching keys.

Hmm, I think this is more buggy than that. `(key, value)` comes from elements of `plot_attributes()` which returns a `Dict{Symbol, AttributeMetadata}`. So the values are always a `AttibuteMetadata`, never `Attributes`, and that code never runs. 
`add_theme!()` (in the same file) also has some unreachable branches likes this. It also can't fill `Attributes` passed to a plot that are inherited from a scene, e.g.
```julia
f, a, p = text(0, 0, text = "test", fonts = Attributes(regular = "Fira Mono"))
p.fonts[] # still just Attributes(regular = "Fira Mono") (missing the other defaults)
```

kunzaatko:
Do you suggest that we deal with the unreachable branches here @ffreyer, or should there be a completely new issue created? I believe that on your first point, the `merge` is not meant to be used on the combination of the two types, but some times it is in the other functions such as the quoted one. Maybe this should be type stabilized by using some `where` type dispatches?

ffreyer:
I'd say the first case would be good to fix. For the second one I'm leaning towards removing kwargs, but I'd want a second opinion first. The third one is something Simon or I should probably handle. Maybe even in #5482

PR #5340, opened 2025-10-18
feat: Multi-histogram plot

lenianiva:
# Description

This PR adds a multi-histogram plotting function `multihist!` like [seaborn's stacked histogram](https://seaborn.pydata.org/examples/histogram_stacked.html). Below is an example of stacked vs dodged histograms.

Vide https://discourse.julialang.org/t/stacked-histogram/133248

comparison

```julia
using CairoMakie

data1 = rand(100) .* 2.0 .- 1.0
data2 = rand(100) .* 2.0

fig = Figure(size=(600,300))

multihist!(
    Axis(fig[1, 1]),
    [data1, data2];
    colormap = :Set3_10,
    label = ["red", "blue"],
    positioning = :stack,
)
multihist!(
    Axis(fig[1, 2]),
    [data1, data2];
    colormap = :Set3_10,
    label = ["red", "blue"],
    positioning = :dodge,
)
save("/tmp/stacked.svg", fig)

```

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -46,7 +46,7 @@ jobs:
           DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24'
           julia --color=yes makedocs.jl
     - name: Upload site as artifact
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: Docs build
         path: ./docs/build
--- a/.github/workflows/compilation-benchmark.yaml
+++ b/.github/workflows/compilation-benchmark.yaml
@@ -39,7 +39,7 @@ jobs:
         run: >
           DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24' julia --project=./metrics/ttfp/ ./metrics/ttfp/run-benchmark.jl ${{ matrix.package }} 20 ${{ github.event.pull_request.base.ref }}
       - name: Upload plots as artifact
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: ${{ matrix.package }}
           path: ./benchmark_results
@@ -65,7 +65,7 @@ jobs:
 
       - name: Download artifacts
         if: steps.check_token.outputs.has_token == 'true'
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           path: ./images
           merge-multiple: true
--- a/.github/workflows/reference_images_site.yml
+++ b/.github/workflows/reference_images_site.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Download ReferenceImages artifact
         if: steps.pr_number.outputs.pr != ''
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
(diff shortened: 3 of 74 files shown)


lenianiva:
I also want to add an optional dodge positioning for the histogram, but this would require conditionally setting either `:stack` or `:dodge` based on another option `:positioning`, and `map!` does not seem to be able to do this.

```julia
    map!(plot, :points, [:flatbins, :stack]) do points
        flatbins = collect(Iterators.flatten(points))
        stack = [i for (i, bin) in enumerate(points) for _ in 1:length(bin)]
        return (flatbins, stack)
    end
```

lenianiva:
The `weights` argument should be a vector of vectors as well, but this doesn't work:
```julia
    map!(plot, [:values, :edges, :normalization, :scale_to, :weights], :points) do values, edges, normalization, scale_to, weights
        points = []
        for (i, dataset) in enumerate(values)
            if weights === automatic
                wgts = automatic
            else
                wgts = weights[i]
            end
            centers, weights = _hist_center_weights(dataset, edges, normalization, scale_to, wgts)
            push!(points, Point2.(centers, weights))
        end
        return points
    end
```
Giving the error
```
Triggered by update of:
  arg1, scale_to, weights, normalization or bins
Due to ERROR: MethodError: no method matching weights(::Float64)
The function `weights` exists, but no method is defined for this combination of argument types.
```

ffreyer:
I'd say this should be a change to `hist` instead, and follow the syntax we already have for barplot, violin, boxplot and crossbar

lenianiva:
> I'd say this should be a change to `hist` instead, and follow the syntax we already have for barplot, violin, boxplot and crossbar

so with an extra `categories` variable?

jkrumbiegel:
> so with an extra categories variable?

No, barplot for example has a `dodge` and a `stack` attribute and these take `Vector{Int}`s that assign groups for each.

ffreyer:
I can also refactor this if you're not already working on it

ffreyer:
I think this should be done now. 

You can now do either:
```julia
# established syntax:
hist(rand(200), stack = vcat(fill(1, 100), fill(2, 100))

# vector of vectors syntax:
hist([rand(100), rand(100)], stack = [1, 2])
```
where stack can be replaced with dodge. `weights` work if they match the syntax used for the argument. `color` can be `:stack` or `:dodge` to use those indices as color values, a per-group color vector (one color per stacked/dodged histogram), the existing `:values` (uses heights as color values) or the existing per-bin color vector. All of these work with both input styles.

lenianiva:
> I think this should be done now.
> 
> You can now do either:
> 
> ```julia
> # established syntax:
> hist(rand(200), stack = vcat(fill(1, 100), fill(2, 100))
> 
> # vector of vectors syntax:
> hist([rand(100), rand(100)], stack = [1, 2])
> ```
> 
> where stack can be replaced with dodge. `weights` work if they match the syntax used for the argument. `color` can be `:stack` or `:dodge` to use those indices as color values, a per-group color vector (one color per stacked/dodged histogram), the existing `:values` (uses heights as color values) or the existing per-bin color vector. All of these work with both input styles.

Thanks! Sorry I didn't have time to work on this.

Issue #4353, opened 2024-09-12
plotspecs don't Cycle


aplavin:
Regular plotting calls – the color cycles automatically:
![image](https://github.com/user-attachments/assets/28698e8a-8292-4ea7-8f2a-61f50d71c86b)

Plotspecs – always use the first color from the cycle:
![image](https://github.com/user-attachments/assets/a5c2a805-6ec3-4d90-aa93-14ca75c2a70f)

Same if returning a plotspec from `convert_arguments`:
![image](https://github.com/user-attachments/assets/4b507f99-725f-4720-adec-90ac3805b9f1)

The expected behavior is, of course, to run cycles the same way.

- [x] what version of Makie are you running? (`]st -m Makie`) v0.21.10, but same on earlier as well
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
- [x] What platform + GPU are you on? any

jariji:
I see cycling here

```
  [13f3f980] + CairoMakie v0.15.9
  [ee78f7c6] + Makie v0.24.9
```

```jl

julia> using CairoMakie

julia> let
           ecdfplot(rand(10))
           ecdfplot!(rand(10))
           ecdfplot!(rand(10))
           current_figure()
       end
```

<img width="1200" height="900" alt="Image" src="https://github.com/user-attachments/assets/7fecd5b5-e595-4014-92a4-69a8de9183bf" />


```jl

julia> let 
           plotlist(S.Lines(1:10, 1:10))
           plotlist!(S.Lines(1:10, (1:10).+1))
           plotlist!(S.Lines(1:10, (1:10).+2))
           current_figure()
       end
```

<img width="1200" height="900" alt="Image" src="https://github.com/user-attachments/assets/5b76e4a8-3951-4457-8594-d741b0d0bf03" />

asinghvi17:
Yep, this got fixed in a PR but not sure which one.

aplavin:
pretty sure it's https://github.com/MakieOrg/Makie.jl/pull/5541

Issue #4322, opened 2024-09-05
ecdfplot! doesn't cycle colors


jerlich:
```
using GLMakie
# using CairoMakie
# Both give same bug

foo() = begin
       f = Figure()
       ax = Axis(f[1,1])
       for i = 1:5
           ecdfplot!(ax, randn(1000))
       end
       f
end

foo()
```

Makie Versions:
CairoMakie 0.12.
GLMakie v0.10.9

- [X] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)

```
julia> versioninfo()
Julia Version 1.10.5
Commit 6f3fdf7b362 (2024-08-27 14:19 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 10 × Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```

SimonDanisch:
Nope, doesn't seem fixed on master :(

SimonDanisch:
Well, it seems to be a new issue, so no worries ;)

feanor12:
Related to https://github.com/MakieOrg/Makie.jl/issues/4353

aplavin:
Appears fixed by https://github.com/MakieOrg/Makie.jl/pull/5541 as well

PR #5364, opened 2025-10-28
Cycle plotlist

### LINKED ISSUE #4843: `plotlist!` does not Cycle using SpecAPI
Right now using `plotlist!` does not cycle through color like `lines!` would do.

MWE is attached below. The second `plotlist!` uses a new cycle instead.

```
begin
    f = Figure()
    ax = Axis(f[1, 1])

    plots = [S.Lines(1:4),S.Lines(2:5)]
    plotlist!(plots)

    plots2 = [S.Lines(3:6),]
    plotlist!(plots2)
    f
end
```

manuelbb-upb:
# Description
Plots created by `plotlist` don't cycle properly.
Computation of the `cycle_index` via `plot_cycle_index` can now also work with `PlotList` objects.
We also recurse `PlotList` objects in `scene.plots`.
In `diff_plotlist!` we have to keep in mind that the current `plotlist` is not yet in `scene.plots`, so compute an offset.

`plot_cycle_index` did not look quite right with the `specs, spec::PlotSpec, plot::Plot` signature, but I don't know how to test it.

Seems to fix #4843 #4353 #4322

## Type of change

## Checklist

### DIFF
--- a/Makie/src/compute-plots.jl
+++ b/Makie/src/compute-plots.jl
@@ -752,20 +752,29 @@ function Plot{Func}(user_args::Tuple, user_attributes::Dict) where {Func}
 end
 
 function plot_cycle_index(scene::Scene, plot::Plot)
+    ## delegating to `_plot_cycle_index`; 
+    ## we want to specialize `plot_cycle_index` for other types (e.g., `PlotList`) that
+    ## are not defined yet
     return _plot_cycle_index(scene, plot)
 end
+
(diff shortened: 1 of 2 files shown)


manuelbb-upb:
@SimonDanisch can you run the workflows again? On my first commits, there was some confusion because of what `Revise` had loaded and I was using `TestEnv` for the tests.
Now precompilation should work and the tests pass locally.

PR #5367, opened 2025-10-28
Stems in `stem()` now originate from trunk (Fixes #5366)

### LINKED ISSUE #5366: Make stem's linstyle consisten by starting from the trunk
For non continuum linestyles such as `:dash` or `:dashdot`, IMHO, it is more pleasant for the eye if the stems are originated on the trunk and not on the datapoint.

Current stem plot

Image

More consistent stems 

Image

iagobaapellaniz:
# Description

Fixes #5366

Stems now originate from trunk, making the plot more visually pleasant, specially if `trunkwidth = 0`. Images of the new feature are on the issue #5366.

## Type of change

## Checklist

Added reference image tests for new plotting functions, recipes, visual options, etc.

## Test 

```julia
using CairoMakie

stem(rand(20), stemlinestyle = :dash, trunkwidth = 0)
```


ffreyer:
Tests just fail because they catch the change: (GLMakie before, CairoMakie after)
image

Issue #5345, opened 2025-10-21
Makie `viewmode = :fit` the default in `Axis3`


juliohm:
Inspired by this thread on Discourse:

https://discourse.julialang.org/t/rotating-animation-of-3d-plot-changes-size-as-it-rotates-glmakie/133219

The `viewmode = :fit` options seems more appealing to most users.

Related: it is currently impossible to set the `viewmode` in Makie recipes via the `Makie.args_preferred_axis` function.

jkrumbiegel:
The problem is that `viewmode = :fit` can use quite a bit of less space in the plot which isn't a great default either. Of course the size pumping thing is also not great but in general it's just harder to make good defaults here. For example, in the start position you might want to have the axis take more space but if you then rotate it it clips into other content.

juliohm:
What is the default in LScene? It works very well in terms or space usage.

Also, is there a solution for recipe authors to choose a different default?

Em qua., 22 de out. de 2025, 01:13, Julius Krumbiegel <
***@***.***> escreveu:

SimonDanisch:
>Also, is there a solution for recipe authors to choose a different default?

We want to soon introduce "axis hints" where you can set some attributes of the axis when the call to your recipe is the one to actually create the axis (so, `fig, ax, pl = myrecipe(args...)`).

juliohm:
> We want to soon introduce "axis hints" where you can set some attributes of the axis when the call to your recipe is the one to actually create the axis

That is great. It would still be useful to choose the preferred axis as part of the recipe though.

In our case, we know the number of dimensions at compile time and could customize the axis for users:

```julia
if Dim === 3
    Makie.Axis3(fig[1,1], viewmode=:free, aspect=:data)
else
    Makie.Axis(fig[1,1], aspect=Makie.DataAspect())
end
```

juliohm:
After additional experimentation with different options, I really believe the `Axis3` defaults can benefit from a review. The following code is basically `LScene` plus nice axis updates and features:

```julia
 Makie.Axis3(fig[1,1], perspectiveness=0.5, viewmode=:free, aspect=:data)
```

juliohm:
Related to #5361

juliohm:
Now that #5375 is merged, users can modify the defaults in their recipes.

I still wonder if the default options for `Axis3` deserve a review. Most users end up setting `aspect=:data` after plot creation. If there are no plans to change the defaults, please feel free to close the issue.

Issue #5205, opened 2025-07-20
Ways for recipes to communicate back to axes


asinghvi17:
https://github.com/MakieOrg/Makie.jl/blob/339dc52db82231e9142a83d40fc259ba3dd62874/Makie/src/figureplotting.jl#L101-L114

That function seems overridable, and could also be used to add auto legending

SimonDanisch:
@jkrumbiegel suggested (I think), to overload something like:
```julia
function axis_defaults(plot_args...)::NamedTuple end
```
To specialize on the plot arguments. 
Maybe we should start collecting different ideas about different solutions and start comparing them.

asinghvi17:
The thing is that you want to know if you are plotting into an empty cell or not, you may plot slightly differently, or reuse computations

jkrumbiegel:
In my view this axis setting feature should only work if you're implicitly creating an axis, it should not work when mutating one so it shouldn't matter what's in the figure already or not. Unless I misunderstood your suggestion

asinghvi17:
> this axis setting feature should only work if you're implicitly creating an axis

Agreed

> shouldn't matter what's in the figure already or not

If I am plotting into an empty layout cell, I might want to do multiple axes, but if plotting into an existing axis with plot! I might want to do a single axis.  Or something similar.  So there is different behaviour based on whether there was already a thing or not.  I should not know what that thing was but I should know if there was a thing.

ffreyer:
If this feature only works when you're implicitly creating an axis it will always ignore `plot!(ax, ...)`. You'd only have the case targeting an empty layout cell then.

I always find it useful to have some concrete examples of what a new feature should achieve. Maybe we should collect some? I guess `plot(data)` resulting in a Colorbar or Legend + Axis would be one? And allowing a plot to change the default axis attributes would be another? (Can you already do this by adding `axis = (...)` to a recipe as default attributes?) Would `plot!(ax, data)` adding a Legend/Colorbar to an existing Axis be something we want?

asinghvi17:
I think axis is stripped out of kwargs earlier in the pipeline.

Let me get some examples together.  Probably the most common example is indicating that you want an autolegend, it's common enough that we may want some separate processing for it.

Other examples:
- Auto colorbar
- Auto x/y labels and limits
- Auto axis choice (already exists)
- Creating a grid layout (marginal histogram) when axis kwargs or axis are not provided, otherwise run a regular plot via convert_args.

[...]

Issue #1524, opened 2021-12-20
Default xlabel and ylabel in recipe


tkf:
How do I set default x and y labels in the full recipe? I'm trying to do something similar to `xlabel --> "default X label"` in RecipesBase.jl. Here's a code sample that didn't work with Makie v0.15.3

```julia
module MyRecipe

using Makie

struct MyPlotData
    x::Vector{Float64}
    y::Vector{Float64}
    xlabel::String
    ylabel::String
end

@recipe(MyPlot, data) do scene
    # Errors since `data` is not in scope:
    # Attributes(axis = (xlabel = data.xlabel, ylabel = data.ylabel))

    # But setting labels here seem to not work anyway:
    # Attributes(xlabel = "default X", ylabel = "default Y")
    # Attributes(axis = (xlabel = "default X", ylabel = "default Y"))

    Attributes()
end

function Makie.plot!(myplot::MyPlot{<:Tuple{<:MyPlotData}})
    data = myplot[1][]
    axis = (xlabel = data.xlabel, ylabel = data.ylabel)

    # scatter!(myplot, data.x, data.y; axis = axis)
    sc = scatter!(myplot, data.x, data.y)

    # These does not work:
    # sc.axis = axis
    # sc.xlabel = data.xlabel
    # sc.ylabel = data.ylabel
    # myplot.axis = axis
    # myplot.xlabel = data.xlabel
    # myplot.ylabel = data.ylabel

    # Throws a method error:
    # xlabel!(myplot, data.xlabel)
    # ylabel!(myplot, data.ylabel)

    return myplot
end

randomdata(n = 100) = MyPlotData(randn(n), randn(n), "X axis", "Y axis")

function demo()
    data = randomdata()
    myplot(data)
end

end  # module MyRecipe

import CairoMakie
MyRecipe.demo()
```

I looked at https://makie.juliaplots.org/stable/documentation/recipes/ but it is not clear if this is supported. The only "workaround" I've found was to bypass the recipe system and just define `Makie.plot(::MyPlotData)` directly.

I think it is either a documentation request or a feature request, depending on if this is already possible.

Krastanov:
I do not think there is yet a way to modify an axis attribute from inside a plot recipe (which only touches a set of plots inside of an axis without touching the axis itself). My understanding is that it is this way for now because it is not clear what should happen if two recipes that are used together both want to modify the axis.

This seems like a related issue https://github.com/JuliaPlots/Makie.jl/issues/379

Issue #379, opened 2019-08-07
Have a type recipe affect the axis


grero:
I'm sure this is very obvious, but I just can't wrap my head around it. I am following the molecular simulation recipe that create a plot function for my own type. My question is how do I, through the type recipe formalism, specify e.g. axis names? There doesn't seem to be a way to get the axis object from the AbstractPlotting.plot! function that I am overriding for my type. Naively, I would want to insert something like this in my function

```julia
function AbstractPlotting.plot!(p::Plot(MyType))
  p[Axis][:names, :axisnames][] = ("x","y")
end
```

but this gives me a `MethodError` saying 
```julia
MethodError: no method matching getindex(::Combined{Any, Tuple{MyType}}, ::Type{Axis})
```

What is the proper way of overriding Axis attributes for a custom type?

asinghvi17:
You can probably specify an `axis2d` and `axis3d` Theme in the `@recipe` theme itself.

grero:
Thanks for the reply. However, that doesn't quite seem to work. This is what I tried:

```julia
struct MyType
end

function AbstractPlotting.default_theme(scene::SceneLike, ::Type{<: Plot(MyType)})
     Theme( textsize = 6,
            axis2d = Theme(frame = Theme(linewidth= 1.5,
                                           frames = ((true, false), (false, false))),
                      grid = Theme(linewidth = (0, 0)),
                      names = Theme(axisnames = ("Time", "Activity")))
         )
 end

function AbstractPlotting.plot!(p::Plot(MyType))
    x = range(0.0, stop=1.0, length=10)
    y = x.^2
    lines!(p, x,y)
end

Q = MyType()
p = plot(Q)
@test p[Axis][:frame, :frames][] == ((true, false), (false, false))
```
The test fails with the following:

```julia
Test Failed at /Users/roger/Documents/programming/julia/NeuralPlots/test/runtests.jl:29
  Expression: ((p[Axis])[:frame, :frames])[] == ((true, false), (false, false))
   Evaluated: ((false, false), (false, false)) == ((true, false), (false, false))
```

asinghvi17:
Oh, it might be the scene attributes problem again.  There is a way to fix that but it's a tad hacky -basically you capture the relevant attributes and add them to the scene during the pipeline.




Also, any particular reason you're not using the `@recipe` macro?




Sincerely,




Anshul Singhvi.







[...]

PR #5375, opened 2025-10-30
Axis Hints

### LINKED ISSUE #5361: Allow customization of axis in recipes
The recipe system allows the selection of axis based on types:

```julia
Makie.args_preferred_axis(obj::Object1) = Makie.Axis
Makie.args_preferred_axis(obj::Object2) = Makie.Axis3
```

Unfortunately, many default options in `Axis` and `Axis3` do not meet expectations of users. It would be great if a new function `Makie.args_preferred_axis_defaults` was introduced to set the defaults within recipes. For example:

```julia
Makie.args_preferred_axis_defaults(obj::Object2) = (aspect=:data, viewmode=:free, perspectiveness=0.5)
```

ffreyer:
# Description

This pr adds

```julia
Makie.preferred_axis_attributes(AxisType, [plot], [args...])
```

as an extendable function for providing axis attributes to a generated axis. This is used with non-mutating `plot()` calls just like `preferred_axis_type() / args_preferred_axis()`. When `plot(..., axis = (...))` is used, both sources are merged with the attributes from the plot call taking priority . If a different axis type is set there, the new type will be used to generate default attributes. The method call order here is:

1. `preferred_axis_attributes(AxisType, plot, args...)`
2. `preferred_axis_attributes(AxisType, plot)`
3. `preferred_axis_attributes(AxisType, args...)`

The first one to return a non-empty collection gets to set axis attribute defaults. 

This pr also cleans up the dispatch for `args_preferred_axis()`. Before we had

1. `preferred_axis_type(::Plot)`
5. `args_preferred_axis(::Type, args)`
6. `args_preferred_axis(::Type)`
7. `args_preferred_axis(args[i])` (iterate arguments)
8. `args_preferred_axis(::Type, converted)`
9. `args_preferred_axis(::Type)`
10. `args_preferred_axis(converted[i])` (iterate converted arguments)

where the first method not returning `nothing` would set the axis type. We also had `args_preferred_axis(::AbstractVector, ::AbstractVector, ::AbstractVector, ::Function)` as a method which was unreachable.

After my changes we now have:

1. `preferred_axis_type(::Plot, input_args...)`
2. `preferred_axis_type(::Type{<: Plot}, input_args...)`
3. `preferred_axis_type(::Plot, converted_args...)`
4. `preferred_axis_type(::Type{<: Plot}, converted_args...)`
5. `preferred_axis_type(::Plot)`
6. `preferred_axis_type(::Type{<: Plot})`
7. `preferred_axis_type(input_args...)`
11. `preferred_axis_type(converted_args...)`
12. `preferred_axis_type(arg)` for each argument
13. `preferred_axis_type(converted)` for each converted argument

where the first valid method still sets the axis type. This is now a different priority order (e.g. args no longer beat plot type + converted). My thoughts here are to prioritize more specific information over less specific information. That way you can have a general "plot -> Axis" and a specific "plot, angle_args, radial_args -> PolarAxis`.

I've also chosen to soft-deprecate `args_preferred_axis`, meaning that they are not longer mentioned in docs but still work without errors or warnings. (Each `preferred_axis_method` defaults to trying `args_preferred_axis`.) Imo `preferred_axis_type` is a better name here, as there are a few methods that don't include args and many that use converted arguments. It also fits more nicely with `preferred_axis_attributes()`, though that's a name I chose. (If we want to actually deprecate this I would suggest doing that in breaking 0.25)

Closes #5361

Related: #5345, #5205, #1524, #379, #4174

TODO:
- [x] tests
- [x] docs

## Type of change

## Checklist


asinghvi17:
How would this fit in with an auto-legending or auto-colorbar request?

jkrumbiegel:
Can this be used to define axis attributes that update with the data? If so, should that be based on compute graph rather than observables for a later point when we might switch blocks to that infra as well? Because currently you'll have to convert to observables at the border but we might want to hide that from the user.

ffreyer:
> How would this fit in with an auto-legending or auto-colorbar request?

We had a discussion internally about axis hints and auto legends and whatnot a while ago. The conclusion was that axis hints and auto legends should be separate, with axis hints generating defaults based on one plot and auto legends being part of some higher level system that may include multiple plots and layouting. 
The dim converts pr #5323 also came from that discussion, because setting axis labels to units seems like a fairly common hint/default you'd want to give an axis.

> Can this be used to define axis attributes that update with the data? If so, should that be based on compute graph rather than observables for a later point when we might switch blocks to that infra as well? Because currently you'll have to convert to observables at the border but we might want to hide that from the user.

The plot you get in `preferred_axis_attributes()` is partially constructed. It already has attributes, arguments and converted arguments which you can react to. It doesn't have anything scene related yet, because the axis does not exist yet, and as a consequence it also didn't run `plot!(plot)` yet. (Because that can depend on the scene via projections for example).

I could try delaying the call to `preferred_axis_attributes()` until after the plot is fully initialized, but I'm not sure if that's worth the extra book-keeping and Observable (or future compute graph) surgery required. (If you need a computation that's defined in `plot!(plot)` you could also just copy it over to `preferred_axis_attributes()`. If the inputs, output and callback are the same the compute graph will just skip the second `map!/register_computation!`.)

juliohm:
@ffreyer what is the status of this PR?

ffreyer:
I think I was just waiting for review/feedback on this

juliohm:
> I think I was just waiting for review/feedback on this

It would be great if someone could review this PR.

juliohm:
ping @SimonDanisch

juliohm:
We are looking forward to this feature. Is there any chance that someone could review this PR in the following weeks?

SimonDanisch:
Maybe you can review it as a user and try it out if it covers your use cases? That would be most valuable!

juliohm:
Will do! Just thought that my review wouldn't matter for the merge decision.

Em sáb., 4 de abr. de 2026, 13:30, Simon ***@***.***>
escreveu:

SimonDanisch:
I think we're mainly unsure if this is a good user facing API ;)

juliohm:
@ffreyer the new function works like a charm. Could we also have a method that dispatches on the `args` like we do for `args_preferred_axis`? https://github.com/JuliaGeometry/Meshes.jl/pull/1357

If we could dispatch on the `args`, we could retrieve additional information about the object being plotted. For example, if we know that the object being plotted has coordinates in `cm` we can set the `xlabel` to `X [cm]` as opposed to the generic `X`.

juliohm:
@SimonDanisch do you have an opinion about this method with `args`? I believe that is the only missing detail for the uses cases we have in mind. We need to set default attributes based on the object being plotted, not the recipe type.

jariji:
A slightly modified API is proposed in https://github.com/MakieOrg/Makie.jl/pull/5581/changes

juliohm:
Thank you @jariji. It is good to see that other people already identified this need in other contexts as well.

ffreyer:
I merged the changes from #5581 though I also reordered the methods. Instead of

1. `preferred_axis_attributes(AxisType, plot)` 
2. `preferred_axis_attributes(AxisType, args...)`

it is now 

1. `preferred_axis_attributes(AxisType, plot, args...)`
2. `preferred_axis_attributes(AxisType, plot)`
3. `preferred_axis_attributes(AxisType, args...)`

That way you can have plot based defaults and still specialize to (plot and) argument based defaults. I'm basically thinking that the more detail you add to these methods, the priority you should get for setting attributes. 

I also did the same thing to `preferred_axis_type` as noted in the pr description.

juliohm:
Thank you @ffreyer for the update. I will be able to test it again on Wednesday or Thursday.

juliohm:
@ffreyer the latest changes work beautifully:

https://github.com/JuliaGeometry/Meshes.jl/pull/1357/changes

Looking forward to the next Makie.jl release with this new feature.

aplavin:
Eager to try using this to replace hacky axis attributes setup in packages like MakieExtra and AxisKeysExtra!
Some questions that I couldn't fully understand from the docs:
- `preferred_axis_type` is documented for both "original" and converted arguments, but `preferred_axis_attributes` doesn't explicitly say it supports both – is it just a docs omission? A useful natural piece of this feature.
- How to utilize these attributes for exclamation-suffixed plot calls, when plotting onto an existing axis? I find this quite a common pattern to create an axis first with some geometry/layout options, and then plot stuff onto it later:
```julia
ax = Axis(fig[1,1], width=200, height=300)
hidespines!()

...

image!(ax, data)
```
and automatic labels would be natural here, based on `data`.

juliohm:
> * How to utilize these attributes for exclamation-suffixed plot calls, when plotting onto an existing axis?

Does it make sense to overwrite the attributes of an already created axis with `preferred_axis_attributes` though? I think that the new function is designed for cases where we are creating the axis automatically from arguments. If the axis is already created, it already has attributes set. IMHO, it is dangerous to allow the plot recipe to set attributes of an already created axis.

aplavin:
@juliohm note that I haven't said that `plot!` calls should automatically set axis attributes – any other convenient way to utilize them would work just as well.
Although I personally don't see major downsides to setting axis attributes when they weren't set before.

That's how things work now:
```julia
data = [
       KeyedArray(1:5, time=10:10:50),
       KeyedArray(1:5, time=15:10:55),
]
ax = Axis(Figure()[1,1])

for d in data
       scatterlines!(d)
end
```
<details><summary>Image</summary>
<p>
image

</p>
</details> 

And I'd be happy to transition the implementation to `preferred_axis_attributes` if it could support this scenario in some way.
I find this pattern quite common, and it becomes inconvenient without `plot!` funcs.

ffreyer:
> Eager to try using this to replace hacky axis attributes setup in packages like MakieExtra and AxisKeysExtra! Some questions that I couldn't fully understand from the docs:
> 
> * `preferred_axis_type` is documented for both "original" and converted arguments, but `preferred_axis_attributes` doesn't explicitly say it supports both – is it just a docs omission? A useful natural piece of this feature.

Is it actually useful? I'd imagine axis related information is one of those things that `convert_arguments`is very likely to remove, if implemented. (And you don't need to extract dimensionality here anymore, since that's already encoded in the axis type choice)

> * How to utilize these attributes for exclamation-suffixed plot calls, when plotting onto an existing axis?

Well, you'd need to figure out what attributes the user set for that. I don't think diffing current with default attributes would be enough for that, because it would fail to catch the user setting an attribute to its default value to avoid plot-based overwriting. (Or maybe that's an acceptable shortcoming?) Another option would be to track the state of each attribute through construction and setindex/setproperty. Which seems kind of bloaty but maybe that's ok?

In general, when we discussed axis hints we were kind of doubtful about how useful they would be. For units you can use dim converts, probably even with something like KeyedArray. (0.25/#5323 adds functionality for having dim converts affect axis labels and also allows them to hint whether they should be in tick labels or axis labels.) For more complicated things 0.25/#5465 adds complex recipes. They also allow you to add automatic legends, colorbars or anything layout related which seems like the thing people actually want.

aplavin:
> Is it _(support for converted args)_ actually useful? 

Of course – if my `convert_arguments` converts `MyType` to `AnotherType`, then I kinda want the plotting result to be the same as if AnotherType was plotted. Why wouldn't this be useful?
I don't personally rely on this aspect _super_ often, but definitely do sometimes when converting to `KeyedArray`s.

> Well, you'd need to figure out what attributes the user set for that.

Would be nice, yes – but even some simple callsite opt-in would help.
Eg, `scatter!(data, axis_attributes=true)` / `with_axattrs(scatter!)(data)` / whatever else, these are just initial variants.

> Another option would be to track the state of each attribute through construction and setindex/setproperty. Which seems kind of bloaty but maybe that's ok?

IMO tracking "default" vs "explicitly assigned" could generally be useful, not just for this purpose.

---

Haven't read other recent PRs you linked in detail yet, do you suggest that those advanced recipes make these "axis hints" obsolete?

ffreyer:
> > Is it _(support for converted args)_ actually useful?
> 
> Of course – if my `convert_arguments` converts `MyType` to `AnotherType`, then I kinda want the plotting result to be the same as if AnotherType was plotted. Why wouldn't this be useful? I don't personally rely on this aspect _super_ often, but definitely do sometimes when converting to `KeyedArray`s.

Hmm, if you're converting between types that have extra information then that makes sense. 

I view `convert_arguments` as something that changes data into a plot specific format, so I would expect it to remove that information. In fact, for primitive plots that is required to a degree by type annotations in `@recipe` (or trait based target type). This will expand to non-primitive plots with 0.25, because we use that information to figure out when conversions are done, which in turn is used to figure out if dim converts have already applied and if the user input is valid. For `PointBased` plots this forces you to convert to `AbstractVector{<: Point{D, <:Real}}` where D is 2 or 3, for example. 

> > Well, you'd need to figure out what attributes the user set for that.
> 
> Would be nice, yes – but even some simple callsite opt-in would help. Eg, `scatter!(data, axis_attributes=true)` / `with_axattrs(scatter!)(data)` / whatever else, these are just initial variants.

An opt-in kwarg sounds good to me. That would skip all the extra complexity. Might also be nice to be able to opt-out in the non-mutating calls.

> Haven't read other recent PRs you linked in detail yet, do you suggest that those advanced recipes make these "axis hints" obsolete?

tl;dr for complex/block recipes is that `@Block` can now be used as some sort of layout recipe

```julia
@Block MyBlock (x, y) begin ... end

function Makie.initialize_block!(b::MyBlock)
    ax = Axis(b[1, 1])
    map!(eachindex, b, :x, :scatter_colors)
    p = scatter!(ax, b.x, b.y, color = b.scatter_colors)
    Colorbar(b[1, 2], p)
end

f, b = MyBlock(rand(10), rand(10))
```

and if you are creating the axis manually in such a recipe, then you can of course also set all the axis attributes there. It doesn't make axis hints obsolete, but it is another way to get there.

juliohm:
@ffreyer do you have a timeline for merging/releasing this feature? It is super handy for teaching.

aplavin:
`use_axis_hints` seems great, thank you @ffreyer! agree that it can be useful in both directions – basically, the difference in `plot` vs `plot!` in this regard would be the default, right?

aplavin:
> I view convert_arguments as something that changes data into a plot specific format, so I would expect it to remove that information. In fact, for primitive plots that is required to a degree by type annotations in @recipe (or trait based target type). This will expand to non-primitive plots with 0.25, because we use that information to figure out when conversions are done, which in turn is used to figure out if dim converts have already applied and if the user input is valid. For PointBased plots this forces you to convert to AbstractVector{<: Point{D, <:Real}} where D is 2 or 3, for example.

Seems quite involved... I guess I don't understand this part of Makie api at all :) Conversion vs dim conversions etc.

Just thinking from a user's PoV: it's totally natural to say "my `a::TypeA` should plot as `f(a)::TypeB`" in all aspects. Including axes and everything.

ffreyer:
> @ffreyer do you have a timeline for merging/releasing this feature? It is super handy for teaching.

If there are no more suggestions or complaints I will probably just merge this in the next few days.

> `use_axis_hints` seems great, thank you @ffreyer! agree that it can be useful in both directions – basically, the difference in `plot` vs `plot!` in this regard would be the default, right?

Yea. It should default to true for `plot` and false for `plot!`, and can be overwritten by passing it as a kwarg. I also decided to make this work on plots beyond the first, so you don't have to order plots just for this and you can have multiple plots contribute. I didn't at first to avoid repeatedly setting the same values in examples like yours, but that's probably a pretty minor performance difference.

theabhirath:
Hi, could you please confirm when this might be part of a Makie release version? The changelog says it was released as part of 0.24.10 but I can't seem to find it in that version.

ffreyer:
Oh, that should not be listed under 0.24.10. It will be part of 0.24.11 which should release today or tomorrow

theabhirath:
Gotcha, thanks for clarifying that!

PR #5381, opened 2025-11-02
fix AbstractArrays failing in some CairoMakie recipes

aplavin:
# Description

Fix some recipes (`band`, `image`) failing with `AbstractArrays` with CairoMakie.
MWE for `band`:
```julia
using CairoMakie, StructArrays

ps1 = Point2f[(1,2), (3,4)] |> StructArray
ps2 = Point2f[(1,3), (3,5)] |> StructArray

fig,ax,plt = band(ps1, ps2)
display(fig)  # failed before this PR
```
This PR fixes CairoMakie specifically – the same plot already works (and has worked for a long time) with GLMakie.

Don't know what's the best way to test for it. Maybe add a simple "smoke test" to CairoMakie?..

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,9 @@
 - Added option to replace Makie native widgets with HTML based widget for WGLMakie [#5285](https://github.com/MakieOrg/Makie.jl/pull/5285)
 - Fixed empty `Label` not updating [#5362](https://github.com/MakieOrg/Makie.jl/pull/5362).
 - Fixed `band` not working with StructArrays in CairoMakie [#5381](https://github.com/MakieOrg/Makie.jl/pull/5381)
+- Updated `scatterlines` to include all remaining `scatter` attributes and pass all applicable attributes to its subplots [#5388](https://github.com/MakieOrg/Makie.jl/pull/5388)
 - Adjusted `stem` so that dash and dot patterns of stems start at the trunk [#5367](https://github.com/MakieOrg/Makie.jl/pull/5367)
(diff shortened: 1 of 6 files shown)


SimonDanisch:
Yeah I guess testing that it doesn't error sounds good enough ;)

ffreyer:
Shouldn't we fix stuff like this through convert_arguments methods?

SimonDanisch:
I think we want to preserve array types wherever possible (like StructArrays)

aplavin:
As I understand, `convert_arguments` is executed prior to creating `Plot{...}` object, meaning that plots would only contain raw `Array`s if `convert_arguments` converts to `Array`s.
But the current situation in Makie makes total sense, I regularly find it convenient that arbitrary array types can go into `Plot`s.

A natural implication is that this Cairo bug should be fixed downstream of `convert_arguments` and `Plot{}` creation. And the fix seems straightforward :)

ffreyer:
> Don't know what's the best way to test for it. Maybe add a simple "smoke test" to CairoMakie?..

I just added a refimg test because every backend needs to be able to handle StructArrays if we let them get to the backend.

That's also more or less why I asked about fixing it in convert_arguments. If we only get Julia arrays from there, backends don't need to be compatible with e.g. StructArrays, and tests can check if they get converted correctly rather than checking that construction and updates of backend objects work (correctly).

aplavin:
Thanks for updating the PR and merging it! I didn't have the bandwidth for it for a while.

PR #5376, opened 2025-10-31
Enhance documentation for image and heatmap plotting functions

bjarthur:
Clarify the descriptions to specify that `x` and `y` can be 2-tuples for axis end points.

PR #5389, opened 2025-11-05
improve documentation

SimonDanisch:
First stab at improving the docs and having a more unified system for adding examples, attribute and argument docs.
Based on #5306

### DIFF
--- a/Makie/src/stats/distributions.jl
+++ b/Makie/src/stats/distributions.jl
@@ -35,7 +35,7 @@ Draw a Q-Q plot, comparing quantiles of two distributions.
     also be an abstract distribution or distribution type, e.g. `Normal(0, 1)` or `Normal`. The
     latter will be fit to the `y` sample.
 """
-@recipe QQPlot (points::VecTypesVector{2, <:Real}, line::VecTypesVector{2, <:Real}) begin
+@recipe QQPlot (points::PointVector{2, <:Real}, line::PointVector{2, <:Real}) begin
     filtered_attributes(ScatterLines, exclude = (:joinstyle, :miter_limit))...
     """
     The attribute `qqline` determines how to compute a fit line for the Q-Q plot.
@@ -69,7 +69,7 @@ See [`qqplot`](@ref) for more details on the `qqline` attribute and other option
 
 * `y::AbstractVector{<:Real}` is a sample to compare against the standard normal distribution.
 """
-@recipe QQNorm (points::VecTypesVector{2, <:Real}, line::VecTypesVector{2, <:Real}) begin
+@recipe QQNorm (points::PointVector{2, <:Real}, line::PointVector{2, <:Real}) begin
     documented_attributes(QQPlot)...
 end
 
@@ -107,10 +107,10 @@ argument_dims(::Type{<:QQPlot}, x, y) = (1, 2)
 argument_dims(::Type{<:QQNorm}, y) = (2,)
 
 function convert_arguments(
-        ::Type{<:QQPlot}, points::AbstractVector{<:Point2},
-        lines::AbstractVector{<:Point2}; qqline = :none
+        ::Type{<:QQPlot}, points::VecTypesVector{2, <:Real},
+        lines::VecTypesVector{2, <:Real}; qqline = :none
     )
-    return (points, lines)
+    return (convert_arguments(PointBased(), points)[1], convert_arguments(PointBased(), lines)[1])
 end
 
 function convert_arguments(::Type{<:QQPlot}, x′, y::RealVector; qqline = :none)


ffreyer:
I'm not sure what kind of  width we should be aiming for, but the function call documentation overflows the default width of an ubuntu terminal and also the width of my vscode terminal when I have vscode cover half the screen.
image
I'd say we should remove the comments there, or work them into the docstring below. I'd also suggest reverting `kw` back to `attributes` since we call them that in the remaining docstring.

---

If the call signatures don't specify the arguments, some of the trait docs need to be reworked. For example `image` has
image
which suggests that you can call `image(image)` or `image(x, y)`. The latter is nonsense. Previously this was clarified by having `image(x, y, image)` and `image(image)` as the call signatures. Without that, I'd say each list item needs to be a complete call signature. That probably also means it needs to repeat itself sometimes. E.g. here it should probably clarify that the image in `x, y, image` is still an AbstractMatrix of colors or Real values.

You can make a similar critique for the description of the plot. For example `scatter` currently has "Plots a marker at each position.", but the plot signatures doesn't mention positions anymore. It would probably be good to cut out that referential language to the not yet introduced positions and have something like `Plots a collection of markers." instead.

There's also the problem of trait-less recipes. I don't think you've worked on any yet, but I feel like they're also generally awkard. E.g. `hlines` currently looks like this:
image
Keeping the `hlines(ys; xmin = 0.0, xmax = 1.0, attributes...)` doesn't seem like a clean solution since it breaks the pattern. It should really be part of `Arguments`, which means we need some function to add argument docs based on the plot type. But having to document arguments by defining `argument_docs(::MyPlot) = [signature1, signature2, ...]` also seems rather cumbersome?

---

I personally think docstring should aim to be short and dense, so that they can be used to look up information rather than learn something from scratch. At least for feature-rich things like plots. So I'm happy to see the same attribute handling as Axis has, though I wish it had some grouping. Looking at a plain list of attributes makes it hard to find what you're looking for imo. Maybe we could separate color attributes, generic attributes and recipe attributes? If it's not already planned, the `?plot.attribute` thing should also have the auto-complete feature (hack?) that Axis has.

When it comes to examples I think it's probably good to have one in there. Though I'd restrict it to one and keep it short, so that it doesn't push out the argument and attribute docs. Maybe the example should even be moved before "Arguments", after the main description, to avoid that. I'd imagine the generic call signatures, plot description and example(s) will become useless once you have used a plot a couple times, while argument and attribute docs are something you look up again and again. So having those at the bottom seems preferable to me.

---

In case someone else wants to comment on this but doesn't want to checkout the branch, here's what `?scatter` looks like atm:

Screenshot from 2025-11-06 15-37-29

SimonDanisch:
Alright, I'm at a point where I want to give this the final finish. The biggest problem right now is how to structure the argument documentation.
Right now I have something like this:
image

I think adding the generic "how to use any plotting command" needs to be somewhere prominent where users look on first contact and I'm not sure where else that should be. Of course its a bit wasteful, but now that we automatically generate these, it wouldn't be too hard to move such things around.

My other big problem is that we have a mix of two documentation styles for arguments:
1)

```
plot(a, b)
plot(c, b) 

* a is this
* b is that
* c is something else
```
2)
```
* `a, b` a and b do this 
* `c, b` this signature takes c but also the known b
```
They both have their pro and cons.
Its kind of cool to have to just one time explain a variable in the arguments, but on the other hand it can be  a bit confusing and it's a bit redundant to the `f, ax, pl = lines(args...; kw...)` and repeating the whole plot call for all argument combinations seem a bit bloated.
I want to decide on one of these forms (or come up with an even better one?), and use them consistently across the code base.
I'm kind of leaning towards 2 where one can still say `just like the above a` or simply implicitely let the user make the connection that a variable with the same name does the same thing and needs to only get explained once.
CC: @ffreyer @jkrumbiegel

SimonDanisch:
Re, beginner vs advanced docs, I think `?plot` should give everything a beginner needs to know, with references to how to get more advanced documentations, without overwhelming the user, so everything should get pretty shortened.
The shortening is not complete yet, but right now this means only ONE example will be added, while attributes will also be heavily shortened to something like this:
image
With explanations on how to get more information.


On the other hand, we have the full docs which prints out every piece of example/attribute/argument code we have for a plot, which we put in the static docs.
Inbetween, I want multiple function to query more advanced docs.
```julia
attribute_docs(Plot, attribute) # Full documentation with all attribute examples
argument_docs(Plot) # Everything we have on arguments
search(attribute_docs, Plot, keyword) # maybe have a generic search function which can search in lists of infos we have (e.g. also arguments).
plot_examples(Plot) # All examples we have for a plot
attribute_examples(Plot) # All attribute examples
```

ffreyer:
image

I still don't really like this. I feel like the comments make it hard to see the function calls. Would also still advocate for `attributes` over `kw`.

> My other big problem is that we have a mix of two documentation styles for arguments:

Yea that's basically the second block in my earlier comment. I'm leaning to your option 2, i.e. every bullet point contains a valid and complete set of arguments and explains them. If these explanations repeat information, so be it.

We also still need to figure out what to do with trait-less recipes. They currently have an empty argument block. Maybe we can detect "## Arguments" in the `@recipe` docstring and use that if no trait docs exist. (or use it over trait docs?)

I think what I wrote in the third block also still applies. I would still prefer attributes to just be in a dense list to not overwhelm everything else, preferably with grouping. With the current bullet points, `scatter` fills my whole screen with attributes without even telling me what they do. The only benefit I see in that over a dense list is readability, which I think grouping would also improve. I also still think `docstring, example, arguments, attributes` is going to be a more practical order.

For reference, this is `scatter` now, after reducing my font size enough to fit the whole thing on screen:
image

ffreyer:
> I would still prefer attributes to just be in a dense list to not overwhelm everything else, preferably with grouping.

Just for reference, with grouping I mean something like this: (Scatter here for example)
```
plot: strokecolor, glowwidth, marker_offset, markersize, depthsorting, font, distancefield, strokewidth, rotation, marker, markerspace, glowcolor, transform_marker
color: alpha, lowclip, colormap, colorscale, highclip, nan_color, color, colorrange
generic: transparency, visible, space, clip_planes, ssao, overdraw, cycle, transformation, model, depth_shift, fxaa
DataInspector: inspector_clear, inspector_hover, inspector_label, inspectable
```
It would be neat if recipes could also define groups. E.g. `barplot` could separate out a bunch of label related attributes.

jkrumbiegel:
These attributes should also be sorted alphabetically, having them in random (I guess definition) order makes them much harder to skim.

I kind of like the idea of grouping although it can be hard to make good groups. For example `plot` above is pretty nondescript.

ffreyer:
I was thinking of the `color`, `generic` and `DataInspector` groups as groups that filter out attributes that most plots have. Like the basic colormapping attributes and color, but not something like strokecolor or strokecolormap. The `plot` group would then be everything else, which should be everything that's (relatively) unique to the plot.

It would also be nice if you could add more groups for recipes. E.g. something like barplot could benefit from a `label` group. Or maybe `scatter, lines, shared` for scatterlines?

ffreyer:
Rough TODO:
- [x] fix docs build
- [x] extend `help` or similar to act as a replacement for `?Plot.attribute` since that does not work pre 1.12.2
- [x] cleanup `help()`
- [x] autocomplete for `?Plot.attribute`
- [x] maybe try implementing grouped attribute lists
- [x] grouping for full docs too
- [x] check changed plot examples for missing content
    - [x] Restore a bunch of online docs that got deleted, like the 1.5 pages on `marker` that are now just 4 words (?)
- [x] check that all docstring examples are short
    - [x] remove "Dealing with outline artifacts in GLMakie" from the base lines docstring 
- [x] check every plot for duplicate documentation
- [x] move arrows docs into arrows2d & arrows3d (or merge them again)
- [ ] consider moving examples before argument docs again
- [x] document recipe docs

Makie/src/basic_recipes/band.jl
```diff
@@ -1,11 +1,17 @@
...
+
+* `xs, ys_lower, ys_upper` Plots a band from `ys_lower` to `ys_upper` along `xs`. `xs` is an
+    `AbstractVector{<:Real}` containing x-values, and `ys_lower`, `ys_upper` are
+    `AbstractVector{<:Real}` containing the lower and upper y-limits of the band.
+* `xs, lowerupper` Plots a band where `lowerupper` is an `AbstractVector{<:Interval}` containing the
+    lower and upper limits of the band as intervals.
+* `lower, upper` Plots a [ruled surface](https://en.wikipedia.org/wiki/Ruled_surface) between the
+    points in `lower` and `upper`. Both are `AbstractVector{<:Point{D, <:Real}}` containing the
+    (x, y) or (x, y, z) coordinates of the lower and upper limits of the band respectively.
+* Setting `direction = :y` will reinterpret (x, y) as (y, x). (2D only)
```
ffreyer, line 14: Makie tests fail because I separate out notes like this in a few places. I didn't change the test yet because I thought it would be better to get some feedback first

ffreyer:
Rough summary of code changes I made:
- `help(plot_type_or_func)` has been reworked to just return docstrings, removing a bunch of old code
- `help(plot_type_or_func, attribute)` now acts as an alternative to `?PlotType.attribute` and is referred to pre 0.12.2
- example loading has some extra steps in it to allow specifying different file paths or returning markdown examples directly (per plot type)
- `attribute_groups(PlotType)` was added for grouping attributes in docstring. A couple recipes implement their own variants. 
- The attribute docs part of recipe docstring has been reworked to work with `attribute_groups`
- the auto-generated call signatures moved to a function so they are overwritable per PlotType. (This is not advertized anywhere)

This pr closes #5376 and closes #5306

`timeseries` is currently excluded from online docs because its examples output videos which `@figure` blocks don't support. (They aren't included on master either)

PR #5391, opened 2025-11-06
Move clamp of color values to WGLMakie backend

### LINKED ISSUE #5384: `Inf` values confuse the default heatmap colorrange used for display
- [x] What version of Makie are you running? (`]st -m Makie`)
0.24.6
- [x] What is the last Makie version this feature worked on?
0.23.0
yes
- [x] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)
GLMakie and CairoMakie
Linux, Nvidia RTX A2000 8GB Laptop GPU

```julia
data = rand(10, 10)
data[1, 1] = -Inf

f, ax, hm = heatmap(data)
Colorbar(f[1, 2], hm)
display(f)
```

| Makie version | GLMakie output| CairoMakie output |
|--------|--------|--------|
| v0.24.6 | Image | Image |
| v0.23.0 | Image | Image | 

I think the normal behavior is that the `Inf` values are ignored by the automatic colorrange calculation. It seems to be the case when looking at the colorbar values, but not for the actual displayed heatmap. Feels like a missing update of the colorrange in the rendering process. This does not happen if the colorrange values are set manually as a kwarg.

egavazzi:
# Description

Fixes #5384

~~Try to remove the clamp~~ Move the clamp to WGLMakie vertex color computation, as suggested by @ffreyer.

Figure from the issue:
fixed

I guess we should add a new refimage test to catch this kind of regression?
## Type of change

## Checklist

### DIFF
--- a/docs/src/tutorials/getting-started.md
+++ b/docs/src/tutorials/getting-started.md
@@ -163,7 +163,7 @@ lines!(ax, seconds, exp.(seconds) .+ 7, color = :tomato, linestyle = :dash)
 f
 ```
 
-## Legend
+## [Legend](@id Legend_getting_started_example)
 
 The last element we're missing is the legend.
 One way to create a legend is by labelling plots with the `label` keyword and using the [`axislegend`](@ref) function:
@@ -207,4 +207,3 @@ nothing # hide
 ```
 
 You should now find the three files in your `makie_tutorial` folder.
-
--- a/docs/src/tutorials/layout-tutorial.md
+++ b/docs/src/tutorials/layout-tutorial.md
@@ -224,7 +224,7 @@ axtop.xticks = 0:3:9
 f
 ```
 
-### Legend
+### [Legend](@id Legend_layout_example)
 
 We have set the `label` attribute in the scatter call so it's easier to construct the legend. We can just pass `axmain` as the second argument to `Legend`.
 


egavazzi:
Mmmh, WGLMakie does not like that indeed.

ffreyer:
Since it's just WGLMakie that has problems with this I think we should just move this clamp to the backend. It might also just be a problem when color values end up as vertex data, so maybe clamping in here is enough:
https://github.com/MakieOrg/Makie.jl/blob/6def6575349bb2968e3695903903c7d8e62bc81e/WGLMakie/src/plot-primitives.jl#L55-L58
(the AbstractVector can contain colors or values)

egavazzi:
> Since it's just WGLMakie that has problems with this I think we should just move this clamp to the backend. It might also just be a problem when color values end up as vertex data, so maybe clamping in here is enough:
> 
> https://github.com/MakieOrg/Makie.jl/blob/6def6575349bb2968e3695903903c7d8e62bc81e/WGLMakie/src/plot-primitives.jl#L55-L58
> 
> (the AbstractVector can contain colors or values)

I see! I will try that. I don't like my "filtering" solution. It seems to works but feels a bit dirty. This would be better.

egavazzi:
It works! And is much cleaner :)

ffreyer:
The refimage I added looked like this locally (GLMakie):
Screenshot from 2025-11-11 22-45-20
but like this in CI:
Screenshot from 2025-11-11 22-39-30
I guess the Inf sections are turning invisible because some drivers interpret Inf as NaN, resulting in `nan_color = :transparent` being used. I moved the clamp back to Makie to see if that fixes it.

PR #5398, opened 2025-11-11
Is same isequal

SimonDanisch:
Continuation of: https://github.com/MakieOrg/Makie.jl/pull/5327


SimonDanisch:
The relocatability issue seems like an upstream regression and not related to the PR.

PR #5400, opened 2025-11-11
Replace HTML ordered list with Markdown

abhro:
# Description

Markdown does support list nesting, and the HTML wasn't picking up the backtick formatting.

## Type of change

## Checklist


ffreyer:
I'm pretty sure I tried this and it didn't generate the nested list correctly. When I open the relevant html file from the build it also shows me a flattened list with matching source code.

abhro:
Hmm. Yes, I remember Vitepress/DocumenterVitepress being weird about this, and the syntax finicky. It does support nested Markdown lists, but I've not been able to get it to completely work either. For now, settled on a hybrid approach so that the backticks at least render

PR #5415, opened 2025-11-18
CairoMakie: make `poly` respect `linecap`, `joinstyle` and `miter_limit`

manuelbb-upb:
# Description

Currently, `CairoMakie` intercepts `Poly` plots as atomic.
Whilst the recipe would draw filled polygons and outline strokes separately, `CairoMakie` uses backend primitives to do so in one go.
The outline properties `linecap`, `joinstyle` and `miter_limit` are not respected by the overrides.
This pull should fix that.

We mainly just parse the poly attributes to corresponding `Cairo` constants.
A more opinionated change is in `create_shape_path!` to match recipe behavior for open Bezier paths.

Below reference image is "basic polygon shapes" with `joinstyle=:round` for fork and master:

fork_basic_polygon_shapes
master_basic_polygon_shapes

## Type of change

Could be considered "breaking" though, because some reference images might look different.

## Checklist

### DIFF
--- a/Makie/src/basic_recipes/text.jl
+++ b/Makie/src/basic_recipes/text.jl
@@ -8,6 +8,10 @@ struct RichText
     end
 end
 
+Base.:(==)(a::RichText, b::RichText) = a.type == b.type && a.children == b.children && a.attributes == b.attributes
+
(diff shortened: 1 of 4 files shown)


ffreyer:
I did some cleanup, added a couple reference image tests, fixed some cases where `linestyle` was also ignored and updated the changelog. Was there anything else to do here?

PR #5431, opened 2025-11-24
Improve window resize performance in GLMakie

### LINKED ISSUE #3746: figure resizes for multiple seconds after the window was resized
Hi, I am observing GLMakie to build up some backlog for resizing a figure's content which leads to some weird resize wobbling even long after the figure's window has been resized. This seems to be a somewhat recent change. Is there some way to reduce these artefacts?

https://github.com/MakieOrg/Makie.jl/assets/28118027/acf926c0-26bd-49af-ae94-c5d4a7c60fb8

- [x] are you running newest version (version from docs) ? I think so.
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) Yes
  ```
   (jl_zl0v06) pkg> status GLMakie
   No Matches in `/tmp/jl_zl0v06/Project.toml`
   (jl_zl0v06) pkg> status Makie
   Status `/tmp/jl_zl0v06/Project.toml`
     [ee78f7c6] Makie v0.20.8
   ```
   ArchLinux, amd64, Nvidia GPU

```
(@v1.10) pkg> status GLMakie
Status `~/.julia/environments/v1.10/Project.toml`
  [e9467ef8] GLMakie v0.9.9

julia> versioninfo()
Julia Version 1.10.0-beta3
Commit 404750f8586 (2023-10-03 12:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, skylake)
  Threads: 17 on 12 virtual cores
Environment:
  JULIA_LOAD_PATH = /usr/share/gmsh/api/julia/:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 12
```

egavazzi:
# Description

Fixes #3746 

This PR does two things:
1. It removes the `refreshwindowcb` function. 

That function was creating quite some lag on some platforms as it was queuing up calls to `poll_updates` on window resize. Removing it helps with the lag, and does not seem to break anything as the actual rendering seems to wait for the render loop anyway which also calls `poll_updates` and others, making it redundant. 

2. It computes the new window area only once per frame.
  
Similarly, on some platforms, resizing the window calls `windowsizecb` a lot, which calculates
```julia
area[] = Recti(xs, ys, widths(area[]))
```
for each call. The rendering waits for the render loop though, so these intermediate calculations are unnecessary. 
This PR changes that by adding a `WindowResizeUpdater` mutable struct holding a `new_size::Tuple{Cint, Cint}` and a `need_update::Bool` fields. We still rely on the `glfwSetWindowSizeCallback` listeners to change the `new_size` and `need_update`, but the actual recalculation/update of the area is done only once per `render_tick` (similar to how the `mouse_position` update works). We also check for `need_update` before applying the resize. That way, the update should be completely avoided if glfw didn't report a change.

[master.webm](https://github.com/user-attachments/assets/5053f9c1-be4f-42c1-a090-d2f6c89401bd) 

[PR.webm](https://github.com/user-attachments/assets/e605e4f3-9009-4e57-bace-29c1933ebe74)

## Type of change

not really a bug fix? but an improvement?

## Checklist


GLMakie/src/events.jl
```diff
@@ -55,7 +55,8 @@ function Makie.window_area(scene::Scene, screen::Screen)
         end
         winw, winh = round.(Int, (width, height) ./ winscale)
         if Vec(winw, winh) != widths(area[])
-            area[] = Recti(minimum(area[]), winw, winh)
+            screen.pending_resize = (winw, winh)
+            screen.requires_resize = true
```
ffreyer: This should use the `render_tick` system, e.g.:
https://github.com/MakieOrg/Makie.jl/blob/46d9c5e3cb7adacdd0dbe364005baa9a6c656f97/GLMakie/src/events.jl#L183-L222
instead of setting up something new.

egavazzi: Thanks for the pointer. So you mean using a new struct (e.g. `WindowResizeUpdater`) which store values + an update flag on GLFW resize callbacks, and then an `on(updater, scene, screen_render)` which checks the update flag and applies the resize?

egavazzi: Probably easier to review actual code, so I pushed some changes. Let me know if that's what you meant!

ffreyer: I meant that you should remove the callback alltogether and instead query `glfwGetWindowSize()`. But I maybe it's better not to. For me this benchmarks with a max time 200-300µs which seems like a lot for something that could just be a cached value lookup.
```julia
using GLMakie, GLFW, BenchmarkTools
f,a,p = scatter(rand(10))
screen = display(f)
@benchmark GLFW.GetWindowSize($(screen.glscreen))
```
I'd say leave it as is for now

egavazzi: Oh yes I also thought about something like that. It would be a nice simplification of the code, but having some kind of passive listener with `glfwSetWindowSizeCallback()` felt better for performance (even though I didn't benchmark anything).
Ran the benchmark a few times and I also often get a max time of around 300µs. I even got 1ms once. Median is 8µs though.

Issue #5433, opened 2025-11-25
empty!(ax::Axis3) deletes the axes, frame and grids


ophelia-adams:
- [x] what version of Makie are you running? v0.24.7
- [x] can you reproduce the bug with a fresh environment ? Yes
- [x] What platform + GPU are you on? Mac, Apple CPU with built in GPU
- [x] Which backend are you using? GLMakie
- [x] Please provide an minimal example of the failing code. Screenshots at the bottom.

```
using Makie

fig = Figure()
ax = Axis3(fig[1,1])
scatter!(ax,0,0,0)
empty!(ax)
```

The issue is that the axes are stored in ax.scene.plots, hence subject to removal by `empty!`. The 2d axis keeps its axes in `ax.xaxis`, `ax.yaxis` but has the dedicated `LineAxis` which includes ticks and the like. It would make sense to have those (and `ax.zaxis`) for an `Axis3` but it's not clear how that should be done if `LineAxis` can't be used.

At the moment, I work around it by removing the associated plots from `Axis3.scene.plots` right after creation.

<img width="712" height="590" alt="Image" src="https://github.com/user-attachments/assets/12e8eb7f-8109-4ac7-b3c9-49278e9d17ed" />
<img width="712" height="590" alt="Image" src="https://github.com/user-attachments/assets/2bbc9e8e-b7d3-44b5-8e17-4feaaa320f46" />

PR #5434, opened 2025-11-25
Generalizes the text Base.show from 2D Axis to any AbstractAxis.

ophelia-adams:
# Description

Takes the existing `Base.show` for a 2d `Axis` and uses it for any `AbstractAxis`. 

This makes an assumption that every `AbstractAxis` will have `ax.scene.plots`, which seems reasonable, and is the case for all the other subtypes (`Axis3`, `PolarAxis`, and `LScene`). If not, perhaps this tweak could be duplicated into each of them instead.

## Type of change

## Checklist

- [x] *not* logged in CHANGELOG.md. This is way too small count as a new feature (but it's not precisely a bug fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-- Improved handling of empty strings in text bounding boxes, fixing errors and broken layouting with empty strings in `Label`, `Menu` and `Textbox` [#5420](https://github.com/MakieOrg/Makie.jl/pull/5420)
-
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Sounds like a good idea!

ffreyer:
Axis3 is a bit of a problem since it has 12 decoration plots in `axis.scene` which would always be shown

ophelia-adams:
> Axis3 is a bit of a problem since it has 12 decoration plots in `axis.scene` which would always be shown

Yes it would, although that would have helped me realize what was going on in #5433 more quickly. Perhaps this can wait until after 5433 is resolved.

PR #5437, opened 2025-11-25
Implement equality for RichText

### LINKED ISSUE #5435: legend merging fails with rich labels
```julia
scatter([1], color=:red, label=rich("A", rich("B", color=:gray)))
scatter!([2], color=:red, label=rich("A", rich("B", color=:gray)))
scatter!([3], color=:red, label=rich("A", rich("B", color=:gray)))
axislegend(merge=true)
```
results in
Image
with multiple identical legend entries, instead of a single one.

- [ ] what version of Makie are you running? (`]st -m Makie`) v0.22 or v0.24, same behavior
- [ ] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie) any

henrik-wolf:
# Description

Fixes #5435 by adding `==` and `hash` for the `RichText` type.

## Type of change

### DIFF
--- a/Makie/src/makielayout/lineaxis.jl
+++ b/Makie/src/makielayout/lineaxis.jl
@@ -450,7 +450,7 @@ function LineAxis(parent::Scene, attrs::Attributes)
     end
 
     tickpositions = Observable(Point2f[]; ignore_equal_values = true)
-    tickstrings = Observable(Any[]; ignore_equal_values = true)
+    tickstrings = Observable(Any[]; ignore_equal_values = false)
 
     onany(
         update_tickpos_string, parent,


jkrumbiegel:
For some reason the reference test `Auto units for observables` fails, seemingly because the protrusion space for labels is not updated correctly. That example doesn't seem to use rich text explicitly but maybe unit suffixes are implemented that way? In that case maybe some update is not triggered because with the new equality method there doesn't seem to be a difference. But I haven't checked if this theory makes sense

henrik-wolf:
I did investigate a bit, and it seems like now, that `RichText`s are equal by content, the tickstrings use `ignore_equal_values=true` at
https://github.com/MakieOrg/Makie.jl/blob/46d9c5e3cb7adacdd0dbe364005baa9a6c656f97/Makie/src/makielayout/lineaxis.jl#L453

Which causes some updates not to happen. I am not super sure why this should be a problem though, as the relevant updates should have already happened in the last iteration when there was a non-equal new value for the tickstrings...

This would mean that the discrepancy to the reference image which I introduced in this PR should exist in the same way for plain strings as well. What I don't understand is why it makes a difference at all.

SimonDanisch:
We have a couple of problems like this and it would be nice to get to the bottom of it. I wonder if it could get fixed by moving Axis to computegraph, since I suspect some weird observable loops, which trigger a few recompuations which we somehow ended up relying on to get the correct layout.

henrik-wolf:
I was not able to find a smaller example that has the same behaviour as the animation loop, so I strongly suspect that we are looking at such a strange update loop/dependency. I think (without having extensively looked into performance) moving just about everything we can to compute graph will at least expose the problematic loops, if not fix them.

jkrumbiegel:
In the meantime it might make sense to remove the `ignore_equal_values` then, I think some of these were added later to try and improve performance but nobody could cut through my forest of observables anymore to find out if these updates were completely necessary or not. If I had had a compute graph available back then, it would definitely not be as convoluted but there's hidden dependencies all over the place (like, if A has updated then B must have also, so we can avoid lifting on B here...)

henrik-wolf:
I spent a bit of time yesterday manually figuring out the observable dependency graph of the `LineAxis`, and what I ended up with did not... fill me with confidence:
Untitled-2025-09-29-1014(1)

I guess that most of this can be untangled in some way or another, however, the one thing that I am unsure about is the fact that some of the observables/functions depend on the state of other plot objects:
image
Here, `labeltext = text!(scene, position; text=label, fontsize=labelsize, ...)`, the rest are observables. Updating either `label` or `labelsize` triggers a call to `calculate_protrusion`, which internally calls `boundingbox` on the `labeltext`.

Is there any guarantee that this call to `boundingbox` operates on the `labeltext` plot AFTER all the necessary changes to `labeltext` have been propagated? Or does this depend on the order in which the callbacks in the observables have been defined?

If I where to put the observables into a compute graph and capture the labeltext in the callback
```julia
g = ComputeGraph()
add_input!(g, :label, attrs.label)
add_input!(g, :labelsize, attrs.labelsize)

labeltext = text!(scene, position; text=g.label, fontsize=g.labelsize)
map!(g, [:label, :labelsize], :protrusion) do _, _  # inputs only to trigger changes
    bbox = boundingbox(labeltext)
    return calculate_protrusion(bbox)  # or something like that.
end
```
Would that cause correct update order? As the `labeltext` is implemented with a compute graph, I would expect that the `boundingbox` call sees the dirtyness of the label or labelsize inputs to the text, and then correctly resolves those, before returning the bbox. Is capturing mutable variables in this way at all safe, as long as I specify all the inputs that go to the `labeltext` as inputs to the compute edge? How much would I be relying on the internals of the compute graph?

What happens with other plots, like `lines`, whose boundingbox (probably?) depends on the layout?

I am asking in so much detail as getting the boundingboxes for subplots based on a compute graph and feeding them back into the same compute graph for further processing is something that is important, and potentially problematic for the text rendering refactor #5396.

jkrumbiegel:
Hahaha oh god, that puts into visual form my feeling about the complexity of Axis.

Regarding the boundingbox, wouldn't it be possible to add the boundingbox as a compute graph node to `text` if needed? So not always, as it may be expensive work that's unnecessary in many cases where text is used, but just as an add-on, like the functions we have for projecting stuff into different spaces which also add additional nodes. Like

```julia
labeltext = text!(scene, position; text=g.label, fontsize=g.labelsize)

map!(labeltext.attributes, [...], :boundingbox) do ...
    ...
end
```

I think most of the internals of Axis can probably be represented well with the compute graph, I'm not sure about the interface to the layout mechanism as `GridLayout` machinery only uses observables. And those need to be triggered currently which might or might not mean we can't have the Axis graph be as lazy as it normally would be. But I haven't thought it through in detail.

jkrumbiegel:
We should probably move that Axis discussion to a different place as I've just noticed this is the RichText equality PR. I'd be ok with merging this for now as the non-triggering of ticks seems to be clearly buggy and it's a much bigger lift to clean that up. I'd like to have RichText equality for AoG, too, before that is solved. @SimonDanisch or @ffreyer any issues with merging this one even though it might have a bit of a performance impact?

ffreyer:
> any issues with merging this one even though it might have a bit of a performance impact?

I think the main risk here is infinite loops. As long as those aren't happening I'd be fine with merging this.

> Regarding the boundingbox, wouldn't it be possible to add the boundingbox as a compute graph node to `text` if needed?

I already moved text bounding boxes to the compute graph in the refactor. There are a bunch of different versions to fill different needs too. I'd guess ticks just want to know how much space each string requires, so they could probably use
```julia
# excludes `position`, fully in markerspace, no projections
register_fast_string_boundingboxes!(textplot)
map!(..., textplot.fast_string_boundingboxes, :max_tick_widths) do bboxes
    return maximum(widths, bboxes)
end
```

PR #5439, opened 2025-11-26
turn off inspector for mirrored ticks

ophelia-adams:
# Description

DataInspector produces up a tooltip on mirrored tick marks, unlike their non-mirrored counterparts. This turns inspection off for them.

## Type of change


ffreyer:
This will be unnecessary with #5241

ophelia-adams:
> This will be unnecessary with #5241

Great! I was also going to adjust the inspector to not report a z-coordinate for 2D graphs - is that covered? It sounds like it should wait until after that overhaul anyway.

ffreyer:
Yea, that should be fixed to. With the pr tooltips are created based on (converted) plot arguments mostly, and don't need to be transformed anymore. So they should naturally be whatever the user passes to their plot now. The master version sometimes transforms pixel positions back which makes it hard to know what dimensionality things should have.

PR #5442, opened 2025-11-27
fix annotation: observable positions

manuelbb-upb:
# Description

Fixes no open issue (as far as I can tell) but is related to #5114

It seems that `annotation` does not properly update path positions for  `Observable` label or target positions.
On position changes, only the text label (if not empty) would move, but not the annotation path.
I am not too familiar with the bounding box offset computations, but these simple changes appear to resolve the issue for me.
If they don't trigger too many additional updates, it might be a cheap fix until #5114 is completed.

## Type of change

## Checklist


ffreyer:
Closed by #5448

PR #5443, opened 2025-11-28
empty!(ax::Axis3) deletes the axes, frame and grids bug

### LINKED ISSUE #5433: empty!(ax::Axis3) deletes the axes, frame and grids

ophelia-adams:
# Description

Fixes #5433.

Started to work on this, and have moved the gridlines, frame, and spines to the top scene with the ticks. Mostly a matter of mimicking how `add_ticks_and_ticklabels!` handles the coordinates.

One thought: for the front spines there was nothing to change, and also no complicated coordinate maneuvers because the `overlay` duplicates the main scene's camera. Perhaps it would make sense to simplify the ticks, gridlines, etc by adding a complementary `underlay`. Even though `overlay` is marked as a WGLMakie bandaid, this seems like a natural way to organize.

The xy/yz/zx panels are still not done -- there seems to be a separate issue with them (see attached).

## Type of change

## to do

- [x] gridlines
- [x] frames
- [ ] panel

## panel issue

The panels have a corner at their respective origins, rather than being centered, and the xy panel is at the top of the z-axis rather than the bottom.

panel problem

### DIFF
--- a/Makie/src/makielayout/blocks/axis3d.jl
+++ b/Makie/src/makielayout/blocks/axis3d.jl
@@ -433,10 +433,14 @@ function add_gridlines_and_frames!(topscene, scene, overlay, ax, dim::Int, limit
     d1 = dim1(dim)
     d2 = dim2(dim)
 
-
     tickvalues = @lift($ticknode[1])
 
-    endpoints = lift(limits, tickvalues, min1, min2, xreversed, yreversed, zreversed) do lims, ticks, min1, min2, xrev, yrev, zrev
+    endpoints = lift(
+    	topscene, limits, tickvalues, miv, min1, min2,
+    	scene.camera.projectionview, scene.viewport,
+    	xreversed, yreversed, zreversed
+    ) do lims, ticks, miv, min1, min2, pview, pxa, xrev, yrev, zrev
+    	o = origin(pxa) - origin(topscene.viewport[])
         rev1 = (xrev, yrev, zrev)[d1]
         rev2 = (xrev, yrev, zrev)[d2]
         f1 = min1 ⊻ rev1 ? minimum(lims)[d1] : maximum(lims)[d1]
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
An alternative would be to add a vector to Axis3 which keeps track of the plot objects which are internal to it and should not be deleted from `scene` on `empty!`. Although that makes understanding `ax3.scene.plots` a bit more complex, I wonder if that's actually considered public or not. In Makie, the lines tend to be pretty blurry because it's so useful to manipulate the internals that we often do it regardless.

ffreyer:
We could also mirror the camera settings to the decorations scene and use `space = :pixel` whenever needed, if those scenes are/can be the same size.

SimonDanisch:
Didnt we use another subscene for e.g. Axis for decorations etc? Seems like a natural way to separate things.

ophelia-adams:
> An alternative would be to add a vector to Axis3 which keeps track of the plot objects which are internal to it and should not be deleted from `scene` on `empty!`. Although that makes understanding `ax3.scene.plots` a bit more complex, I wonder if that's actually considered public or not. In Makie, the lines tend to be pretty blurry because it's so useful to manipulate the internals that we often do it regardless.

I think it is preferable for them not to be in `ax.scene.plots`. That's consistent with the other axis types and the comment at `makielayout/types.jl:746`. See below also.

> We could also mirror the camera settings to the decorations scene and use `space = :pixel` whenever needed, if those scenes are/can be the same size.

> Didnt we use another subscene for e.g. Axis for decorations etc? Seems like a natural way to separate things.

Looks like decorations for the 2D axis are also plotted to  `ax.blockscene`. It seems like it might be convenient to have two scenes, one for static decoration and another for dynamic decoration that transforms with the main scene. (this seems a little related to that data inspector thing I ran into? not sure)
In 2D, the ticks and axes are both static (and anyway it's easier to work out the coordinates) whereas in 3D the ticks and axes are dynamic.

So I think maybe the question for the main developers is how flexible the diagram here is:
[https://docs.makie.org/dev/explanations/architecture#Figures-and-Blocks](https://docs.makie.org/dev/explanations/architecture#Figures-and-Blocks)

Perhaps I should have made a discussion instead ☺

SimonDanisch:
Why cant we mirror this for the Axis3 decorations?
image

What do you mean by the ticks in 2d static? They also move around on camera interaction.

ophelia-adams:
> Why cant we mirror this for the Axis3 decorations?
> [picture]

Yes, I think we can/should do that, that's what this PR does.
I thought the other approach (a `decorationscene` mimicking the plot's camera) would make things simpler by reducing the amount of projecting between spaces, but after playing around with the other idea yesterday, I no longer think it does so.

> What do you mean by the ticks in 2d static? They also move around on camera interaction.

Yeah, this is essentially what I ran into.

-- --

So I think what's in this PR is ok, plus more consistent with the other axis-type blocks. I still haven't been able to figure out the geometry for the panels. I'll poke at it later today or tomorrow, but that may be a job for someone with a better sense of the third dimension + familiarity with Makie's transformations/projections.

PR #5448, opened 2025-12-01
Update annotation to use the compute graph

ffreyer:
# Description

I originally just wanted to review #5442 and maybe tweak observables a bit to remove some duplicate callbacks but ended up refactoring the plot instead. I also added another per-string text boundingbox that skips `offset` so that we no longer remove it afterwards.

- fixes `annotation(offsets, position_obs)` not updating paths when `position_obs` updates, closing #5442
- fixes `update!(p, arg1 = ..., arg2 = ..., text = ...)` failing when arrays get resized
- fixes `p__advance_optimization[] = 5` not erroring because of a typo

Also tested compatibility with float32convert.

## Type of change

## Checklist

### DIFF
--- a/Makie/src/basic_recipes/annotation.jl
+++ b/Makie/src/basic_recipes/annotation.jl
@@ -237,22 +237,22 @@ function plot!(p::Annotation)
     register_camera_matrix!(p, :data, :pixel)
     inputs = [
         :screenpoints_target, :labelspace, :label_offsets_or_positions,
-        :world_to_pixel, :f32c, :model, :transform_func
+        :world_to_pixel, :f32c, :model, :transform_func,
     ]
     register_computation!(
         p.attributes, inputs, [:screenpoints_label]
     ) do (tps, space, loffpos, proj, f32c, model, tf), changed, cached
         if space === :relative_pixel
             if isnothing(cached) || changed[1] || changed[2] || changed[3]
-                return (tps .+ loffpos, )
+                return (tps .+ loffpos,)
             else
                 # Skip updates from camera and transform func
-                return (nothing, )
+                return (nothing,)
             end
         else
             transformed_label_pos = apply_transform(tf, loffpos)
             f32c_mat = f32_convert_matrix(f32c)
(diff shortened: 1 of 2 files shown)


Makie/src/basic_recipes/annotation.jl
```diff
@@ -227,89 +222,101 @@ function Makie.plot!(p::Annotation{<:Tuple{<:AbstractVector{<:Vec4}}})
...
-        p.path,
-        p.clipstart,
-        p.shrink,
-        p.style,
-        p.color,
-        p.linewidth,
-    ) do text_bbs, pth, clipstart, shrink, style, color, linewidth
+    # create observable updating offsets in text plot
+    # This forces everything offsets rely on to update asap, before the backend pulls
+    on(offsets -> update!(txt, offset = offsets), p.offsets, update = true)
```
ffreyer, line 296: Since I added a new text bounding box that skips `offset` we could skip the Observable here without creating a loop if we added a way to add nodes without parents. Might be interesting to try since Axis and other blocks probably have similar cases?

ffreyer:
For some reason the direction of labels in one test flipped:

Old GLMakie:
annotation pointcloud
New GLMakie:
annotation pointcloud

I think most of the displacements here are based on self repelling forces (text bboxes of the labels vs their target positions), which are probably float sensitive because they should start off as equal. Since the result isn't the best either way (either Fig or Date overlaps its marker) I added some bias towards the center into the initial `offsets`. That should give every displacement that happens based on self-repulsion a bias away from the edge, where it would get bumped back. With that:

image

jkrumbiegel:
Nice, I still wanted to try out ways to improve border behavior, maybe this is already enough. Although I think labels can still get pushed between border and some point by other labels if it gets crowded. But less likely.

ffreyer:
Yea, this can still happen:
image
though it jumps a little bit later:
image
This seems to be more or less the same as master. Which was also my intention - I just wanted to resolve cases with no other forces.

PR #5454, opened 2025-12-03
Complex Recipes for facets etc

SimonDanisch:
This is a prototype and enables: 

```julia
using GLMakie

@recipe complex DualView (x, y) begin
    "Color for scatter plot"
    scatter_color = :blue
    "Color for line plot"
    line_color = :red
end

function Makie.plot!(cr::DualView)
    # cr[row, col] returns a RecipeSubfig that tracks created axes
    # cr.x, cr.y give access to converted arguments
    # cr.scatter_color gives access to attributes
    scatter(cr[1, 1], cr.x, cr.y; color=cr.scatter_color)
    lines(cr[1, 2], cr.x, cr.y; color=cr.line_color)
    return cr
end
fig, ax, cr = dualview(1:10, rand(10))  # Creates Figure + ComplexRecipe
dualview(fig[2, 1], 1:10, rand(10))     # Places in existing figure position
```
We still need to discuss the return type, corner cases and clean up the implementation.
I think theming already kind of works, but haven't really tested it.

### DIFF
--- a/ReferenceTests/src/tests/complex_recipes.jl
+++ b/ReferenceTests/src/tests/complex_recipes.jl
@@ -1,5 +1,5 @@
 
-@recipe complex AllBlocks (x, y) begin
+@recipe complex AllBlocks (positions,) begin
     "Color for scatter plot"
     scatter_color = :blue
     "Color for line plot"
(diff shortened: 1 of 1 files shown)


asinghvi17:
That seems broadly reasonable to me (looked over the code but have not yet tested it)

ffreyer:
We had some discussion internally about what these should return. `[fig,] ax, complex_recipe` matches up nicely with normal plots, but there can be multiple equally important axes to return. Or none. The other option would be to return `[fig,] complex_recipe`, where you are forced to grab the axis with `complex_recipe[i, j]`.

jkrumbiegel:
> The other option would be to return [fig,] complex_recipe, where you are forced to grab the axis with complex_recipe[i, j].

I like this better but like with Gridposition you'll have to use a `content(recipe_object[i, j])` as there can be multiple objects in the same grid location

PR #5456, opened 2025-12-04
add interactive zoom inset

SimonDanisch:
```julia
using GLMakie

fig = Figure()
ax = Axis(fig[1, 1], title="Interactive Zoom Inset Demo")

# Create some data
x = 1:100
y = sin.(x .* 0.1) .+ 0.5 .* cos.(x .* 0.3)
lines!(ax, x, y)

# Create an interactive zoom inset
# The rectangle defines the initial zoom region: Rect2f(x_origin, y_origin, width, height)
zi = zoom_inset!(ax, Rect2f(30, -0.5, 20, 1.5))
fig
```
https://github.com/user-attachments/assets/41e63320-8573-44ea-8a1c-4565d77312fd


ffreyer:
Changes made:
- inset axis now works with transformations (both transform functions and model transforms)
  - The inset axis also works with a different transform_func (xscale, yscale) than the parent Axis. So you can have a linear view into a log plot/axis. (This requires no model transformations. Bottom right in refimg does this with `xscale = log10` in the parent axis.)
- placement of the inset axis now dodges the zoom rect by default
- the dotted lines now connect corresponding corners. These lines are picked to never overlap with either the zoom rect or the inset
- added refimg test:

image

Issue #5259, opened 2025-08-19
Warning message of `arrows3d` with kwargs `rasterize` using `CairoMakie`


ytdHuang:
- [x] what version of Makie are you running? (`]st -m Makie`)
`Makie v0.24.5`, `CairoMakie v0.15.5`
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)
Yes, I created a new environment, added `Makie` and `CairoMakie`
- [x] What platform + GPU are you on?
`Rocky Linux`
- [x] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)
`CairoMakie`
- [x] Please provide an minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

```julia
fig = Figure()
lscene = LScene(fig[1,1])

v = [1, 0, 1]
a3d = arrows3d!(lscene, Point3f(0), Point3f(v), rasterize = 3)

fig
```

The above code can be displayed, but there is an extra warning message:

```
Warning: Arrows3D{Tuple{Vector{GeometryBasics.Point{3, Float32}}, Vector{GeometryBasics.Point{3, Float32}}}} is not supported by cairo right now
```

If `rasterize` is set to be `0` or `false`, then the warning message disappeared.

Does this warning message (under `rasterize != 0`) make sense ? The figure displays the arrow correctly and everything seems normal.

asinghvi17:
Was there a stack trace or a line number associated with that warning?

ytdHuang:
@asinghvi17 
```
@ CairoMakie /usr/local/julia-1.10.8/share/julia/packages/CairoMakie/mQKIJ/src/plot-primitives.jl:191
```

ytdHuang:
(this is a kind reminder)

ytdHuang:
Hi @asinghvi17 
any thoughts on this ?

ffreyer:
> ```
> @ CairoMakie /usr/local/julia-1.10.8/share/julia/packages/CairoMakie/mQKIJ/src/plot-primitives.jl:191
> ```

Oh that's literally all it says. I thought this was one line of a larger trace.

This happens because a plot with `rasterize != false` is considered a primitive/atomic by
https://github.com/MakieOrg/Makie.jl/blob/b2457025a63b5fe508d39c2fe2d1c2e78dce09d2/CairoMakie/src/plot-primitives.jl#L67

[...]

PR #5459, opened 2025-12-05
Special case Arrows2D in Cairo to avoid rasterizing SVG outputs

jmert:
# Description

Helps in part with issue comments:
1. https://github.com/MakieOrg/Makie.jl/issues/5259#issuecomment-3371919199
2. https://github.com/MakieOrg/Makie.jl/issues/4155#issuecomment-3593733361

Use CairoMakie's overrides mechanism to intercept the `Arrows2D` plot type and use the 2D polygon rendering path instead of the one for generic meshes.

Using the sample plot from comment (1), I observe a fully vector output instead of rasterized:
```bash
$ for img in arrows2d*.svg; do grep -qE '<image' $img && echo "$img: rasterized" || echo "$img: vector"; done                                                                                 
arrows2d_after.svg: vector
arrows2d_before.svg: rasterized
```
and a significant change in file size:
- `arrows2d_before.svg`: 1.3MiB
- `arrows2d_after.svg`: 58KiB

At the same time, improve the handling of arrow sets where arrows have different z coordinates by sorting the meshes in ascending z-order within the recipe.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Added loading spinner in WGLMakie that displays while the plot is being loaded [#5469](https://github.com/MakieOrg/Makie.jl/pull/5469)
 - Moved decoration plots in `Axis3` to `ax.blockscene` so they no longer show up as user plots in the Axis3 [#5463](https://github.com/MakieOrg/Makie.jl/pull/5463)
 - Fixed issue with `transformation` being applied multiple times when set by a user in a recipe that passes applicable attributes to child plots [#5464](https://github.com/MakieOrg/Makie.jl/pull/5464)
 - Fix `arrows2d[!]` plot elements causing CairoMakie SVGs to be rasterized. [#5459](https://github.com/MakieOrg/Makie.jl/pull/#5459)
--- a/CairoMakie/src/precompiles.jl
+++ b/CairoMakie/src/precompiles.jl
@@ -13,6 +13,9 @@ let
         base_path = normpath(joinpath(dirname(pathof(Makie)), "..", "precompile"))
         shared_precompile = joinpath(base_path, "shared-precompile.jl")
         include(shared_precompile)
+        # Cleanup globals to avoid serializing stale state (fonts, figures, tasks)
+        # Note: __init__ doesn't run during precompilation, so we must always clean up here
+        Makie.cleanup_globals()
     end
 end
 precompile(openurl, (String,))
--- a/GLMakie/src/precompiles.jl
+++ b/GLMakie/src/precompiles.jl
@@ -28,7 +28,6 @@ let
                 display(plot(x); visible = false)
             catch
             end
-            Makie.CURRENT_FIGURE[] = nothing
 
             screen = Screen(Scene())
             refresh_func = refreshwindowcb(screen)
@@ -52,6 +51,9 @@ let
             @assert isempty(SCREEN_REUSE_POOL)
             @assert isempty(ALL_SCREENS)
             @assert isempty(SINGLETON_SCREEN)
+            # Cleanup globals to avoid serializing stale state (fonts, figures, tasks)
+            # Note: __init__ doesn't run during precompilation, so we must always clean up here
+            Makie.cleanup_globals()
         end
     end
     nothing
--- a/Makie/src/Makie.jl
+++ b/Makie/src/Makie.jl
@@ -402,6 +402,9 @@ function __init__()
         @warn "The global configuration file is no longer supported." *
             "Please include the file manually with `include(\"$cfg_path\")` before plotting."
     end
+    # Register atexit for runtime cleanup (when Julia exits normally)
+    # Note: This doesn't affect precompilation since __init__ doesn't run during precompile
+    atexit(cleanup_globals)
     return
 end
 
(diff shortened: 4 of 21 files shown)


jmert:
I was pretty conservative here and applied the specialization to `Arrows2D` specifically since I only partially follow how the rendering pipeline works, but that raises two obvious questions:

1. Can 2D check be relaxed to any set of coordinates which are purely planar (z-coordinate is equal everywhere)?
2. Should this idea be moved to the generic mesh handling instead of being specific to `Arrows2D`? (And what would the correct way to do that be?)

jmert:
Assuming that I've found the right image comparisons in my local tests folder, then my subjective opinion is that for the three failing test cases, the proposed rendering of arrows in CairoMakie better matches the GLMakie reference images than the status quo.

1. `arrow min- and maxshaftlength scaling`
2. `arrows2d alignment`
3. `quiver`

|Test | Recorded vs Ref CairoMakie | Recorded vs Ref GLMakie | Ref CairoMakie vs Ref GLMakie |
|-:|:-:|:-:|:-:|
|1 | rec_diff arrow min- and maxshaftlength scaling | gl_diff arrow min- and maxshaftlength scaling | ref_compare arrow min- and maxshaftlength scaling |
| 2 | rec_diff arrows2d alignment | gl_diff arrows2d alignment | ref_compare arrows2d alignment |
| 3 | rec_diff quiver | gl_diff quiver | ref_compare quiver |

ffreyer:
> I was pretty conservative here and applied the specialization to `Arrows2D` specifically since I only partially follow how the rendering pipeline works, but that raises two obvious questions:
> 
> 1. Can 2D check be relaxed to any set of coordinates which are purely planar (z-coordinate is equal everywhere)?

Yea that's fine. The list of plots is sorted before this point, so sending arrows2d down the mesh path just makes them more expensive in that case. The mesh path only matters if the individual arrows need to be sorted.  Though you could render these with the Point path too. You just need to sort `meshes` based on their z values before reducing them to 2D points. (These meshes are always planar, they're just offset in z direction. So checking the first vertices z value of each mesh is enough for this.)

Might also be a good idea to directly call the inner `draw_poly` function and set `strokewidth = 0`. That's only on to improve anti-aliasing for W/GLMakie and may just make CairoMakie worse.

https://github.com/MakieOrg/Makie.jl/blob/f9b432dd4409e51962a4d331a74db6f3faab06b4/CairoMakie/src/overrides.jl#L90-L93

> 2. Should this idea be moved to the generic mesh handling instead of being specific to `Arrows2D`? (And what would the correct way to do that be?)

I think that would be hard. The `Vector{<:Point}` path assumes the points describe the outline of a polygon. That may not be true for a generic mesh, so you'd have test for it. Or convert a mesh into a (vector of) polygon(s) by merging faces and tracking the outline.

jmert:
Instead of applying the z-order sorting in the CairoMakie override, I've tried making the change directly to the `Arrows2D` recipe since it simultaneously fixes CairoMakie and GLMakie. I've added a new reference image test for this behavior, with the following before and after images:

| | CairoMakie | GLMakie |
|-|-|-|
|Before|arrows_zorder_before_cairomakie| arrows_zorder_before_glmakie |
|After|arrows_zorder_after_cairomakie|arrows_zorder_after_glmakie|

jmert:
Bump. Any additional comments?

Regarding,

> Might also be a good idea to directly call the inner `draw_poly` function and set `strokewidth = 0`. That's only on to improve anti-aliasing for W/GLMakie and may just make CairoMakie worse.

I'd like to defer that to future follow-up, as it'd make the CairoMakie and GLMakie outputs different — I can see a "ring" around the arrow when they're partially transparent where the inner polygon and the surrounding stroke overlap. I prefer that the backends produce consistent outputs, so I think the stroke issue should be handled in a separate PR.

jmert:
Bump again.

The WGLMakie failure appears to be the same one as already happens on master so is unrelated, and then the others are expected because reference images need to be updated to match the new behavior.

ffreyer:
> Instead of applying the z-order sorting in the CairoMakie override, I've tried making the change directly to the `Arrows2D` recipe since it simultaneously fixes CairoMakie and GLMakie. I've added a new reference image test for this behavior, with the following before and after images:

I'd rather have backends handle something like this, because it's more of a general rendering issue than a plot specific issue. But right now it's not something any backend handles, so moving it to CairoMakie would just make W/GLMakie worse. And we can always change it later if we implement something like depth peeling for GL backends. So I'm ok with this.

> I'd like to defer that to future follow-up, as it'd make the CairoMakie and GLMakie outputs different — I can see a "ring" around the arrow when they're partially transparent where the inner polygon and the surrounding stroke overlap. I prefer that the backends produce consistent outputs, so I think the stroke issue should be handled in a separate PR.

Again this is just something I added for GL backends to improve antialiasing. Thin arrows looked pretty bad without it and seemed like a fairly common case to me, so I added this as a band-aid. But you can also just turn that off manually by setting `strokemask = 0`, so it's not like you're forced to have stroke in CairoMakie if we include it there. (You could also argue that this is something backends should handle and maybe that'll be the case if we switch to MSAA in future.)

I'll merge this once the tests pass

jmert:
@ffreyer Is there a chance of getting a new patch release made with this commit included? I'd like to depend on it for some documentation I'm writing.

Issue #5427, opened 2025-11-22
[FR] Density line chart


ConnectedSystems:
I'm interested in contributing a density line chart method to Makie. Having a look around, I don't think this plot method exists in Makie yet, I didn't see any open issues to add it, and it would be very useful in my work to visualize many time series together (where "many" can be tens of thousands or more).

The method is described in [Moritz & Fisher (2018)](https://idl.cs.washington.edu/files/2018-DenseLines-arXiv.pdf) (a high-level graphical explanation can be found [here](https://idl.uw.edu/papers/dense-lines)).

In effect, the trajectory of each series is rasterized and the number of times a trajectory hit a pixel is normalized to create a density map. It is similar to, if not identical, to the implementation in [Python Datashader](https://datashader.org/user_guide/Timeseries.html#plotting-large-numbers-of-time-series-together). 

I have a working demonstrator, and while I'm comfortable submitting PRs, my main issue is how complex the Makie architecture is and the organisation of the monorepo itself - I'm not sure where I'm to actually make my changes.

Full disclosure: I found the documentation and examples around plotting recipes very hard to digest and understand so in the end I got things working with some trial and error and a lot of help from AI, so things in that end may not be idiomatic or optimal.

Here's an example output for 76 timesteps and 20,000 series discretized into 200 and 100 bins for x and y respectively.

<img width="582" height="430" alt="Image" src="https://github.com/user-attachments/assets/ad1efc3c-4d3c-40ca-8128-55f08d29c533" />

Current code below. Open to any and all suggestions of course:

```julia
using GLMakie
using Statistics
using Dates

"""
    bresenham_line(x1, y1, x2, y2)

Bresenham's line algorithm to get all pixels between two points.

# References
- Bresenham, J.E., 1965.  \
  Algorithm for computer control of a digital plotter. \
  IBM Systems Journal 4, 25-30. \
  https://doi.org/10.1147/sj.41.0025
"""
function bresenham_line(x1::Int, y1::Int, x2::Int, y2::Int)::Vector{Tuple{Int,Int}}
    pixels = Tuple{Int,Int}[]

    dx = abs(x2 - x1)
    dy = abs(y2 - y1)
    sx = x1 < x2 ? 1 : -1
    sy = y1 < y2 ? 1 : -1
    err = dx - dy

    x, y = x1, y1

    while true
        push!(pixels, (x, y))

        if x == x2 && y == y2
            break
        end

        e2 = 2 * err
[...]

PR #5461, opened 2025-12-07
Adds denselines recipe

ConnectedSystems:
# Description

Adds a new `DenseLines` recipe for visualizing time series data as a density line chart, based on the Moritz and Fisher method. It calculates and displays the density of overlapping lines, providing insights into data distribution over time.

## Type of change

## Checklist


ConnectedSystems:
Denselines implements the method of Moritz and Fisher (2018).
It plots, in essence, heatmaps of time series.

Closes #5427 

@jkrumbiegel suggested that this be part of the Datashader recipe, however I opted to keep this separate at least for now.

Usage example:

```julia
using Dates
using Statistics

using WGLMakie

function example_data()
    n_times = 100
    n_series = 10000
    t = range(0, 2π, n_times)

    time_series = zeros(n_times, n_series)

    # First group: constant frequency sine wave
    for i in 1:div(n_series, 2)
        time_series[:, i] = 2.0 .+ sin.(t) .+ 0.1 .* randn(n_times)
    end

    # Second group: increasing frequency and amplitude
    for i in (div(n_series, 2)+1):n_series
        freq = range(1, 10, n_times)
        amp = range(0.5, 2, n_times)
        time_series[:, i] = 1.0 .+ amp .* sin.(freq .* t) .+ 0.1 .* randn(n_times)
    end

    return time_series
end

f, ax, sp = denselines(example_data(); bins_x=200, bins_y=100, colormap=:plasma)
ax.xlabel = "Time"
ax.ylabel = "Value"
```

image

Makie/src/basic_recipes/denselines.jl
```diff
@@ -0,0 +1,186 @@
...
+        x2 = round(Int, time_to_bin[i + 1])
+        y2 = value_to_bin(series[i + 1])
+
+        pixels = bresenham_line(x1, y1, x2, y2)
+
+        for (px, py) in pixels
+            if 1 <= px <= bins_x && 1 <= py <= bins_y
+                dense_mat[px, py] = 1.0
+            end
+        end
```
ffreyer: Couldn't we also skip the `pixels` vector here and directly write to the matrix? That should save quite a bit of GC time I'd imagine.

Similarly, `compute_normalized_density` could probably clear and write to a matrix created once in `denseline_data` instead of creating a new one for every set of points

jkrumbiegel: I still think it might make sense to reuse the datashader infra for this if at all possible

ConnectedSystems: I've made changes for performance.

@jkrumbiegel , is a barrier function all you have in mind when you say "reuse the datashader infra"?

EDIT to add before and after:

Before: 1.797718 seconds (2.06 M allocations: 1.673 GiB, 29.31% gc time)

After: 0.377040 seconds (60.37 k allocations: 3.668 MiB)

ConnectedSystems:
@jkrumbiegel unfortunately the existing datashader infrastructure is complex and rigid - it was written for its own use case. I think it would take a lot of work to redo it to accommodate both use cases, and I'm not sure I can do it as I'd have to spend a lot of time to understand all the moving pieces here.

PR #5462, opened 2025-12-08
Allow GLFW platform selection from `GLMakie.activate!`

lpaulino07:
# Description

Introduces a `glfw_platform::Union{Nothing, GLFW.Platform` optional argument that allows selection of the underlying GLFW platform when calling `GLMakie.activate!`.

This change will ~~do nothing~~ not be useful until https://github.com/JuliaGL/GLFW.jl/pull/244 is merged. If/when it is, the default platform (ANY_PLATFORM)  will start picking Wayland, which some people might not like yet. This allows doing something like 
```
using GLFW
GLMakie.activate!(;glfw_platform=GLFW.PLATFORM_X11)
```
to force X11.

## Type of change

## Checklist

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -142,7 +142,7 @@ function activate!(; glfw_platform::Union{Nothing, GLFW.Platform} = nothing, inl
     end
     if !isnothing(glfw_platform)
         GLFW.Terminate()
-        GLFW.Init(;platform = glfw_platform)
+        GLFW.Init(; platform = glfw_platform)
     end
     Makie.inline!(inline)
     LAST_INLINE[] = inline


SimonDanisch:
Hm, should this be rather done as an ENV var in GLFW? Seems like something users would want to set permanently.

lpaulino07:
Wouldn't that complicate the init code a bit more in GLFW? Since envars come as strings, I feel like we'd need to handle more cases than here because we can just rely on the GLFW.Platform enum.

Also, maybe this change is not really needed since this is just three lines in GLFW:

```
using GLFW
GLFW.Terminate()
GLFW.Init(;platform=GLFW.Platform)
...
```

I'm fine either way though, just let me know which one makes the most sense!

lpaulino07:
> Hm, should this be rather done as an ENV var in GLFW? Seems like something users would want to set permanently.

Done [here](https://github.com/JuliaGL/GLFW.jl/pull/247)

PR #5465, opened 2025-12-13
Blocks as complex recipes

ffreyer:
# Description

This is an alternative to #5454 which extends `@Block` instead of `@recipe`. @jkrumbiegel suggested this because Blocks and complex recipes both fill the purpose of placing something in the layout. 

#### Changes relevant to existing Blocks

- all attributes are now in a ComputeGraph, rather than fields
- `block.layout` currently always exists, but isn't always initialized

#### Problems for existing Blocks

- moving to `ComputeGraph` broke ~~`notify(block.attribute)`~~ `block.attr[] = block.attr[]`
- `setfield!`, `hasfield`, `getfield` for attributes is broken since attributes aren't fields anymore

#### Blocks as complex recipes compared to `ComplexRecipe` from #5454

- uses `DualView(f[1, 1], args...; attributes...)` instead of `dualview(f[1, 1], args...; attributes...)`
  - ~~currently can't be used without a GridPosition~~
- uses `initialize_block!(b::DualView, args...; kwargs...)` instead of `plot!(cr::DualView)`
  - ~~currently this means that arguments aren't automatically in the compute graph and no `convert_arguments` is used~~ `@Block` now allows arguments and initialize_block will automatically apply argument conversions with `initialize_block!(block; kwargs...)` defined (instead of the `args...` version)
  - no arguments is allowed here, but not for ComplexRecipe
- attributes work, but are in a `@attributes begin ... end` block
- this returns a `::DualView` instead of `::AbstractAxis, ::ComplexRecipe`
- things like `Axis(b[1, 1])` and `heatmap(b[1, 1], ...)` work, including from outside the recipe
- `b.blocks` works with the list generated on demand
- ~~`b.plots` works with the list generated at init time~~ I decided to drop plots because it's rather confusing to have them with `Axis`
- `Block` also allows adding fields, which may be useful?

#### Example

This is the example from #5454 rewritten with this pr:

```julia
Makie.@Block DualView (x, y) begin
    @attributes begin
        "Color for scatter plot"
        scatter_color = :blue
        "Color for line plot"
        line_color = :red
    end
end

function Makie.initialize_block!(b::DualView)
    scatter(b[1, 1], b.x, b.y; color=b.scatter_color)
    lines(b[1, 2], b.x, b.y; color=b.line_color)
    return b
end

fig, b = DualView(1:10, rand(10))
DualView(fig[2, 1], 1:10, rand(10))
```

### TODO

- [x] watch half the refimages fail and fix them
- [x] add `Block()` constructor that generates a figure and places the block at `fig[1, 1]`
- [x] improve argument support, specifically automatic adding to the compute graph (not sure if convert_arguments is worth bothering with?)
- ~~consider adding more `onany/map!/map` methods for `Computed` inputs with N preceding `Observable` inputs. I'd be surprised if we don't run into issues with those...~~ doesn't seem to be necessary
- [x] export `@Block`
- [x] figure out what to do with `block.layout` for traditional blocks
- [x] consider types defined for attributes again
- [x] fix wrong layouting in "basic shading"

## Type of change

- Breaking change due to `ComputeGraph` breaking ~~some `notify()` calls~~ and similarly `obs[] = obs[]`, `setfield/hasfield/getfield` for attributes

## Checklist

### DIFF
--- a/ReferenceTests/src/tests/block_recipes.jl
+++ b/ReferenceTests/src/tests/block_recipes.jl
@@ -214,7 +214,7 @@ end
 
     # wait for Toggle to finish
     colorbuffer(fig)
-    sleep(0.5)
+    sleep(0.1)
 
     Makie.step!(st)
 
--- a/docs/src/explanations/specapi.md
+++ b/docs/src/explanations/specapi.md
@@ -423,7 +423,7 @@ function Makie.convert_arguments(::Type{<:Makie.Block}, multidata::MultiData)
         )
         plotsym = multidata.dim_distribution_plot_type
 
-        # We place the distibution axes in a 1 wide/tall GridLayout of fixed
+        # We place the distribution axes in a 1 wide/tall GridLayout of fixed
         # height/width, each with on plot
         column_axes = map(enumerate(multidata.sets)) do (i, data)
             color = colors[mod1(i, end)]


jkrumbiegel:
I would probably make the Block infrastructure more like the plot infrastructure in this case, to make it behave more like the other recipe code does. Rename it even. Because when I wrote them first, I didn't consider blocks in a recursive way like they would work now, where a complex recipe by some user would use some predefined blocks (then complex recipes) internally. But I think the similarities to the complex recipe idea have shown that it should have worked more like that from the beginning. I just didn't see the similarity between a button, a slider an axis and something like a whole facet plot with a colorbar.

ffreyer:
> I would probably make the Block infrastructure more like the plot infrastructure in this case, to make it behave more like the other recipe code does. Rename it even. Because when I wrote them first, I didn't consider blocks in a recursive way like they would work now, where a complex recipe by some user would use some predefined blocks (then complex recipes) internally.

What do you mean with that? Have `block.blocks` like we have `plot.plots`? Because I see lots of differences between them:
- blocks allow fields to be added
- plots have the whole argument conversion pipeline, while blocks just pass them along
- plots must have arguments, Blocks can have none
- plots don't allow kwargs that aren't attributes, block do
- plots are just one parametric type, while each Block is its own type
- plots are defined recursively in a tree like structure via `plot.plots`, while (complex recipe style) blocks currently connect through `block.layout`
- plots have `plot.transformations`, Blocks don't
- plots have `plot.kw`, Blocks don't
- plots are always added with lower case functions, Blocks always uppercase

And what would you rename it to? I kind of like `Block` for its simplicity.

ffreyer:
Only textbox fails now, because the bottom one resets its alignment differently. Imo that's better, because as far as I can tell it reflects the alignment without any translation (like the top one)
| Before | After |
| --- | --- |
| Textbox | Textbox |

Makie/src/makielayout/blocks.jl
```diff
@@ -437,6 +644,48 @@ function _block(T::Type{<:Block}, fig_or_scene::Union{Figure, Scene}, args, kwdi
...
+        `convert_arguments` ($(length(attr.converted[]))) does not match the
+        number of expected arguments ($(length(converted_names))).""")
+    end
+
+    # splat to defined names
+    map!(identity, attr, :converted, converted_names)
+
+    initialize_block!(block; kw_dict...)
+    return
+end
```
ffreyer, line 817: This is what handles arguments for blocks now.

The macro may define argument names via `@Block BlockType (argname1, argname2, ...)`. This builds `argument_names(::Type{<:Block})`.

If the method above is hit, `argument_names()` must exist. If it does, the normal conversion pipeline runs, excluding dim converts. This adds `:arg1, :arg2, ...`, `:args`, :converted` and finally `:argname1, :argname2, ...` to the blocks compute graph.

With this:
- 0 arguments is allowed, with or without an argument tuple in `@Block`
- arguments can still be handled manually by implementing `initialize_block!(block, args...; kwargs...)` (used only by SliderGrid I believe)
- an argument tuple must exist when automatically handling arguments

jkrumbiegel:
> What do you mean with that? Have `block.blocks` like we have `plot.plots`? Because I see lots of differences between them:

No not necessarily that implementation detail with `block.blocks`, although why not. I meant more in general terms that both have attributes via compute graph, both have a mutating version and a spec version (the GridLayout spec thing should become the spec version of Block basically). I wouldn't still call the init function `initialize_block!` if it's so similar to what the `plot!(plt, ...)` methods do.

> 
>     * blocks allow fields to be added

Currently there's some special logic there to initialize the struct fields after the fact, but I think those fixed type fields never really paid off in the grand scheme of things. Fields that are not compute graph attributes are nice though sometimes, so maybe plots could also get them. So far, if I needed some dynamically created field that shouldn't be themed, I just used `add_input!`.

> 
>     * plots have the whole argument conversion pipeline, while blocks just pass them along

Yeah that's a question but I guess blocks should have that once they are actually for plotting data. That's exactly why I didn't make the connection from blocks to complex recipes at the time, because the Blocks we have are not about plotting data. But they could be (some of them). For example, it could just as well be `Axis(some_data)` or `Axis(some_plot_specs)`, but of course in general I would think people do something like `my_facet_plot(...)` which then has a custom layout of axes, legends, etc.

> 
>     * plots must have arguments, Blocks can have none

I think it would be better if plots could also not have arguments, I needed that in my photobook app and just circumvented it by passing `nothing` and ignoring that arg.

> 
>     * plots don't allow kwargs that aren't attributes, block do

This is only used in one or two places or so I think, if it wasn't needed at all I'd be happier.

> 
>     * plots are just one parametric type, while each Block is its own type

Right, although that's not for any specific reason, it just came more naturally to me.

> 
>     * plots are defined recursively in a tree like structure via `plot.plots`, while (complex recipe style) blocks currently connect through `block.layout`

Yes the backend structure here will be different, but if you plot an `Axis` into a facet plot recipe, that is going to be recursive, too.

> 
>     * plots have `plot.transformations`, Blocks don't

I think the main problem for transformations is that you can't move the Scene canvas around, that's why you can't just rotate a whole`Axis` around. But it wouldn't be that much use anyway, usually, I think.

> 
>     * plots have `plot.kw`, Blocks don't
> 
>     * plots are always added with lower case functions, Blocks always uppercase

Maybe that's a good convention to keep and might simplify remembering whether `fig, ax, plt = ...` or `fig, block = ...` works. Although it's also somewhat arbitrary and still stems from very early decisions I made in MakieLayout.

> 
> 
> And what would you rename it to? I kind of like `Block` for its simplicity.

Yeah I don't mind the name either, but if it's a recipe then those should be tied together maybe? like `@block_recipe ...` or so? We've always been calling them complex recipes, recipe just being the word people are used to from Plots.

ffreyer:
> > What do you mean with that? Have `block.blocks` like we have `plot.plots`? Because I see lots of differences between them:
> 
> No not necessarily that implementation detail with `block.blocks`, although why not.  I meant more in general terms that both have attributes via compute graph, 

I realized later that `block.blocks` is already working like `plots.plots`. I was thinking it flattens out 

>both have a mutating version 

A mutating version seems needlessly breaking to me. That would imply `Axis()`/`axis()` or `axis!(f[1, 1]`? Or if it's just with arguments `legend!(f[1, 1], ax)`, next to `mycomplexrecipe!(f[1, 1], data)`?

> and a spec version (the GridLayout spec thing should become the spec version of Block basically).

A spec version - isn't that just BlockSpec? The complex recipe variant of a block still needs to be placed in the layout of a figure (though the figure creation can happen automatically now), so doesn't that just map to a BlockSpec in a GridLayoutSpec?

> I wouldn't still call the init function `initialize_block!` if it's so similar to what the `plot!(plt, ...)` methods do.

Just renaming it seems a little needlessly breaking to me. Maybe we can have a default `initialize_block!(block)` call another function for this? Would that be `plot!(block)` then? Or `block!(block)`. (I don't like that. This isn't blocking something.)

> Currently there's some special logic there to initialize the struct fields after the fact, but I think those fixed type fields never really paid off in the grand scheme of things. Fields that are not compute graph attributes are nice though sometimes, so maybe plots could also get them. So far, if I needed some dynamically created field that shouldn't be themed, I just used `add_input!`.

I kind of like them. Adding static/management things like a scene or interactions to the compute graph doesn't feel right to me. Though there are also quite a few things that would fit into there nicely, like intermediate limit observables, legend entry groups, maybe event stuff. I'd rather keep fields around and deal with updating traditional blocks later.

> 
> > * plots have the whole argument conversion pipeline, while blocks just pass them along
> 
> Yeah that's a question but I guess blocks should have that once they are actually for plotting data. That's exactly why I didn't make the connection from blocks to complex recipes at the time, because the Blocks we have are not about plotting data. But they could be (some of them). For example, it could just as well be `Axis(some_data)` or `Axis(some_plot_specs)`, but of course in general I would think people do something like `my_facet_plot(...)` which then has a custom layout of axes, legends, etc.

Hmm, the current setup would be awkward for something like `Axis(some_data)` because it requires `@Block Axis (converted1,) begin ... end` to work. Maybe I should relax that so that `@Block Axis begin ... end` doesn't error when arguments are passed (and automatically processed) and just processes them with default names. (You could still set up your own `initialize_block!(block, args...; kwargs...)` here and handle them yourself.)

> > * plots are just one parametric type, while each Block is its own type
> 
> Right, although that's not for any specific reason, it just came more naturally to me.

Simon and I though a bit about this, but `AbstractAxis` would make this difficult.

jkrumbiegel:
With regard to the specs I mostly meant that a block spec should not be a `plot` thing then, but whatever the new thing is. So that the return type matches FigureBlock or whatever, decomposable into two elements.

I guess top should always be the block itself, not a GridLayout. So the internal layout of the recipe would always be hidden inside the thing. The top level grid would usually just be populated at 1,1. If we don't do that, then there's not really such a thing as a complex recipe object (= Block) because a GridLayout with stuff in it is just such a loose concept. Within a block recipe you could of course use further gridspecs.

ReferenceTests/src/tests/specapi.jl
```diff
@@ -232,3 +232,75 @@ end
...
+    f, b1 = Box(S.GridLayout())
+    b2 = Box(f[1, 2], S.GridLayout())
+    st = Makie.Stepper(f)
+    sync_step!(st)
+    b1[1] = S.GridLayout(
+        [
+            S.Axis(; plots = [S.Lines(1:4; color = :black, linewidth = 5), S.Scatter(1:4; markersize = 20)])
+            S.Axis3(; plots = [S.Scatter(Rect3f(Vec3f(0), Vec3f(1)); color = :red, markersize = 50)])
+        ]
+    )
```
ffreyer, line 249: BlockSpec and GridLayoutSpec can now be passed to a Block, either directly or through convert_arguments. 

The way I set it up it's more or less irrelevant which block is used. If it doesn't implement an argument-catching `initialize_block!(block, args...; kwargs...)` and there is a BlockSpec or GridLayoutSpec present after `convert_arguments` it switches to SpecApi block construction. That then treats the block like a figure and defers to existing SpecApi code for GridLayoutSpec.

ffreyer: Seems weird to me that something like `Box(S.GridLayout())` works but I think that's fine if it's just a background quirk. 
If it's useful to have some container block to pass a SpecApi GridLayoutSpec or BlockSpec to, then I think we should make a dedicated one. I think it could also be useful in general to have an empty placeholder block.

jkrumbiegel:
The `FigureAxis` type should be `FigureBlock` I would say?

jkrumbiegel:
For your example with the `DualView` I get this with CairoMakie

grafik

So the layout is off but also the limits are not right. This must be because the `update_state_before_display!` step doesn't run for the nested blocks. Things look okay if I add

```julia
autolimits!.(dualview.blocks)
```

ffreyer:
Huh, still? I added a method to propagate update_state_before_display through nested blocks to fix this. Seems like that only works for `DualView(figpos, ...)` though?

jkrumbiegel:
The overload for gridlayout sides isn't implemented, yet:

```julia
Label(b[1, 1:2, Top()], "Hello", font = :bold)
```

ERROR: MethodError: no method matching getindex(::DualView, ::Int64, ::UnitRange{Int64}, ::Top)

jkrumbiegel:
Thanks, found another one:

```julia
Colorbar(b[end+1, :], vertical = false)
```

ERROR: MethodError: no method matching axes(::DualView, ::Int64)

Makie/src/basic_recipes/poly.jl
```diff
@@ -179,16 +179,25 @@ function plot!(plot::Poly{<:Tuple{<:Union{Polygon, MultiPolygon, Rect2, Circle,
...
     map!(to_lines, plot, :polygon, [:outline, :increment_at])
+<<<<<<< ff/complex-blocks
     map!(to_color, plot, :strokecolor, :rgb_strokecolor)
     map!(plot, [:outline, :increment_at, :rgb_strokecolor, :meshes], :computed_strokecolor) do outline, increment_at, sc, meshes
         if !(meshes isa Mesh) && meshes isa AbstractVector && sc isa AbstractVector && length(sc) == length(meshes)
+=======
+    map!(plot, [:outline, :increment_at, :strokecolor, :meshes], :computed_strokecolor) do outline, increment_at, sc, meshes
+        if meshes isa AbstractVector && sc isa AbstractVector && length(sc) == length(meshes)
+            new_colors = similar(sc, length(outline))
+>>>>>>> master
```
asinghvi17: Looks like unresolved merge markers?

asinghvi17: Fixed in https://github.com/MakieOrg/Makie.jl/pull/5465/commits/70a4f2465d2122383fa10b4ba87beed792d94bcb

PR #5473, opened 2025-12-31
Bump RecipesPipeline to v0.6

blegat:
No idea how breaking these releases were

### DIFF
--- a/MakieRecipes/Project.toml
+++ b/MakieRecipes/Project.toml
@@ -11,5 +11,5 @@ RecipesPipeline = "01d81517-befc-4cb6-b9ec-a95719d0359c"
 [compat]
 Colors = "0.9, 0.10, 0.11, 0.12"
 RecipesBase = "1"
-RecipesPipeline = "0.1"
+RecipesPipeline = "0.1, 0.2, 0.3, 0.4, 0.5, 0.6"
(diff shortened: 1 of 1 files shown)


SimonDanisch:
MakieRecipes is somewhat a zombie package... Its not actually used and likely doesnt even compile, but we haven't removed it yet in case we want to pick it up again...
Honestly, we should just remove it and if we pick it up again get it from the git history.

blegat:
I have packages like Polyhedra.jl or Dionysos.jl using RecipesBase so this would be useful for me to get it working with CairoMakie. I can take a look to see how hard it is to update it and make it work. Maybe, once it's released it will get used more ?

SimonDanisch:
Sure, that would be great :)

asinghvi17:
IIRC it was missing a bunch of features in Makie which we now have.  I think @ffreyer's complex recipe PR is going to be needed for us to get full coverage though.  And no idea what has changed on the Plots side - maybe we could set up a short call with @BeastyBlacksmith to discuss?

BeastyBlacksmith:
I don't think I remember much of these releases anyway 😅, but if you revive this I'd set the compat straight to 0.6 and don't hustle with supporting older versions

PR #5482, opened 2026-01-02
Nested ComputeGraph attributes

ffreyer:
# Description

Nested Attributes can be useful for categorizing attributes. For example, something like `Axis` could benefit from `Axis.xticks.(...)` to reduce the number of attributes you see when autocompleting. Recipes with a lot of passthrough attributes may also benefit from grouping them, e.g. `scatter.color` in `scatterlines`.

Complex recipes (#5454/#5465) may also become very unwieldy without this, as you may want to include all (relevant) block and plot attributes. That'll become a lot very quickly, and may require a lot of manual work to set up. (I.e. creating all the `axis_attribute` inputs and passing them along. It would be nice if you could just set up `axis = default_axis_attributes()` and pass that instead.)

For now this pr is a prototype for how nested attributes could work in the ComputeGraph. It doesn't add any real nesting, but translation layers instead. I.e. there is still just one graph with all the nodes, but nodes can be added and accessed as if they were part of a graph within the graph.

```julia
graph = ComputeGraph()

# add "nested" nodes graph.a.b.c and graph.a.b.z
# these actually add nodes Symbol("a.b.c") and Symbol("a.b.z")
add_input!(graph, :a, :b, :c, 1)
add_input!(graph, (:a, :b, :z), 2)
add_input!(graph.a.b, :z, 2)

# accessing a nested layer returns a ComputeGraphView
# which allows for further `getproperty` calls
# show collects all the nodes reachable from there (one level)
graph.a
# Nested view of ComputeGraph at graph.a containing:
#   b => nested nodes...

graph.a.b
# Nested view of ComputeGraph at graph.a.b containing:
#   c => Input(:a.b.c, 1)
#   z => Input(:a.b.z, 2)

# accessing the full path returns the node 
# (like with a normal node in the graph)
graph.a.b.c
# Computed:
#   name = :a.b.c
#   parent = Input(:a.b.c, 1)
#   value = 1
#   dirty = false

# nested nodes can be referred to with tuples or directly accessed for inputs
map!(+, graph, [graph.a.b.c, (:a, :b, :z)], (:a, :x))
# Symbols/tuples can also be relative to a nested view (this outputs to graph.a.y)
map!(+, graph.a, [(:b, :c), :x], :y)
```

TODO:
- [x] allow nested attributes to automatically connected in passthrough
- [x] figure out what to do with (theme) inherited nested attributes, e.g. fonts
- [x] test scene theme based overwrites more (i.e. not `@inherit` but `Scatter = Attributes(...)` in scene.theme)

## Type of change

  - ~~`graph.attr[].inner_name` currently errors  as it no longer grabs `Attributes()` from a compute node, but instead tries to fetch a non-existing node (could compat-fix this to just return the `ComputeGraphView`, i.e. the intermediate object containing tracking the partially resolved path)~~ `graph.attr[]` now just returns `graph.attr` again to avoid erors here
  - `Attributes()` can no longer be added as a container as they are always converted to nested nodes. This was an issue with `fonts` for example, which are currently hard-coded to ignore the nesting path. Might cause issues elsewhere

## Checklist

### DIFF
--- a/Makie/test/plots/inherited_attributes.jl
+++ b/Makie/test/plots/inherited_attributes.jl
@@ -101,7 +101,7 @@ end
         )
     ) do
         scatter(rand(Point3f, 10))
-    end;
+    end
     ax = a.scene[OldAxis]
     @test ax.padding[] == 0.3
     @test ax.ticks.gap[] == 12


ffreyer:
Test failures are a result of changes from #5241 which I already updated refimages for

ffreyer:
There is now some special casing for `fonts` to allow them to remain as unnested `Attributes` , some changes to `Attributes` to avoid Observable-wrapping `ComputeGraphView` and `Attributes` as well a some changes to deal with nested structures, and also some awkward code to deal with `Attributes` when merging user supplied attributes with defaults. I'm not really happy with that...

For `:fonts` specifically it may make sense to swap out `Attributes` for a `Dict`. `Attributes` lower to nesting in this pr which fonts can't have because they act as a lookup table for possible names in `:font`. `Dict` is treated as normal attribute, i.e. it doesn't generate nesting, so that would avoid the special casing. 

There is still a behavior question then though. For a normal "value" attribute, user input just replaces the default. For nested attributes it merges. On master, `:fonts` doesn't merge, see https://github.com/MakieOrg/Makie.jl/pull/5332#issuecomment-3824613455. In this pr it currently does merge. With `Dict` it again wouldn't, but that could be changed. I'm not sure anymore if it should though.

Perhaps it would be better to adjust the `@recipe` attribute generation to allow setting the behavior. E.g.
- `@nest name = DictLike(...)` marks `name` as a nested attribute (default if `DictLike = Attributes`)
- `@merge name = DictLike(...)` marks `name` as an unnested attribute that should merge its content with user input (maybe default for `Dict`?)
- `@nonest name = DictLike(...)` marks `name` as an unnested attribute that gets replaced entirely by user input (maybe default for `Dict`?)

I'm also wondering if we should either change `Attributes` or introduce something new for (plot, block?) attributes. It feels quite awkward to have unnecessary Observables everywhere and it made it harder to write correct code.

ffreyer:
On https://github.com/MakieOrg/Makie.jl/pull/5482/commits/2f6b0e957a06de753afb706bc2a95c46e3ea32ba: For 
```julia
register_computation!(graph, inputs, outputs) do args, changed, cached
    ...
end
```
I thought it would be better if the names in args, changed and cached were relative to the next nesting layer, not relative to the root layer, e.g. `:c` instead of `Symbol("a.b.c")`. That way working with a subgraph would make more sense:
```julia
function foo(graph)
    register_computation!(graph, [:b, :c], [:d]) do args, changed, cached
        # return isnothing(cached[Symbol("a.d")]) ? 0 : cached[Symbol("a.d")] + 1
        return isnothing(cached.d) ? 0 : cached.d + 1
        # return isnothing(cached[1]) ? 0 : cached[1] + 1 # also valid either way
    end
end

graph = ComputeGraph()
add_input!(graph, :a, :b, 1)
add_input!(graph, :a, :c, 1)
foo(graph.a)
```

This does however create a problem: If multiple subgraphs define the same name, e.g. `graph.xaxis.protrusion` and `graph.yaxis.protrusion`, then they can't both be used as inputs of one computation as their names would collide. You'd have to create a buffer node before, to rename one of the inputs. (E.g. `map!(identity, graph, (:xaxis, :protrusion), :xaxis_protrusion)`). 

Given that you can just index `args`, `changed` and `cached`, and that `register_computation!()` is fairly rare now I reverted back to full names.

PR #5487, opened 2026-01-03
Tweak ReferenceUpdater

ffreyer:
# Description

Basically what the commits say:

- `parseFloat()` returns nan for "NaN" and "Inf", which is intepreted as false for `parseFloat() || 0` so they become 0 scores. They should be Inf scores. (And because 0 is also false, we need to avoid `0 || Infinity`)
- (re-)adds toggle all buttons. I don't want to manually click on 94 refimages for #5241 
- adds a checkbox to switch between cycling `refimage - recording` and `refimage - glmakie refimage - recording`
- fixes missing refimages getting treated as new/changed refimages (not getting sorted into the delete list) 

## Type of change

- CI


jkrumbiegel:
> adds a checkbox to switch between cycling refimage - recording and refimage - glmakie refimage - recording

that's very good, this was annoying me as well

PR #5489, opened 2026-01-08
`band` docs enhancement: add alpha transparency

martincornejo:
Adding transparency to the band plot makes the lineplot (mean) also visible.

This was a regression compared to the docs from v0.23: 
- https://docs.makie.org/v0.23/reference/plots/band
- https://docs.makie.org/v0.24/reference/plots/band


jkrumbiegel (approved):

jkrumbiegel:
Thanks!

PR #5500, opened 2026-01-19
Add possibility to collect legend entries from multiple axes

jusack:
# Description

I added a convenient way to gather legend entries from multiple axes, i.e. 
```julia
using GLMakie

fig = Figure()
ax1 = Axis(fig[1,1])
ax2 = Axis(fig[1,2])
lines!(ax1, randn(10), label="Trace 1")
lines!(ax1, randn(10), label="Trace 2")
lines!(ax1, randn(10), label="Trace 3")
lines!(ax2, randn(10), label="Trace 1")
lines!(ax2, randn(10), label="Trace 2")
Legend(fig[1,3], [ax1, ax2], merge=true)
```
will produce a joint legend for both axes. 
img1

The main reason why this is relevant for me is that the interactivity is then linked as well:
img2

## Type of change

## Checklist

### DIFF
--- a/Makie/src/makielayout/blocks/legend.jl
+++ b/Makie/src/makielayout/blocks/legend.jl
@@ -1075,7 +1075,7 @@ get_plots(p::PlotList) = haskey(p.attributes, :label) && p.attributes[:label] is
 
 get_plots(ax::Union{Axis, Axis3}) = get_plots(ax.scene)
 get_plots(lscene::LScene) = get_plots(lscene.scene)
-get_plots(axes::AbstractVector) =  collect(Iterators.flatmap(get_plots, axes))
+get_plots(axes::AbstractVector) = collect(Iterators.flatmap(get_plots, axes))
 function get_plots(scene::Scene)
     plots = AbstractPlot[]
     for p in scene.plots
--- a/Makie/src/makielayout/blocks/legend.jl
+++ b/Makie/src/makielayout/blocks/legend.jl
@@ -997,13 +997,15 @@ end
 
 
 """
-    Legend(fig_or_scene, axis::Union{Axis, Scene, LScene}, title = nothing; merge = false, unique = false, kwargs...)
+    Legend(fig_or_scene, axis::Union{Axis, Axis3, Scene, LScene, AbstractVector{<:Union{Axis, Axis3, Scene, LScene}}}, title = nothing; merge = false, unique = false, kwargs...)
 
 Create a single-group legend with all plots from `axis` that have the
 attribute `label` set.
 
 If `merge` is `true`, all plot objects with the same label will be layered on top of each other into one legend entry.
 If `unique` is `true`, all plot objects with the same plot type and label will be reduced to one occurrence.
+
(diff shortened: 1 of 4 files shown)


jkrumbiegel (changes requested):
One of the tests is really enough, given the super simple mechanism not much is gained by piping many different plot types through. Rather, it could be tested once that the legend actually has the two expected entries.

Makie/src/makielayout/blocks/legend.jl
```diff
@@ -1005,7 +1005,7 @@ attribute `label` set.
 If `merge` is `true`, all plot objects with the same label will be layered on top of each other into one legend entry.
 If `unique` is `true`, all plot objects with the same plot type and label will be reduced to one occurrence.
 """
-function Legend(fig_or_scene, axis::Union{Axis, Axis3, Scene, LScene}, title = nothing; merge = false, unique = false, kwargs...)
+function Legend(fig_or_scene, axis::Union{Axis, Axis3, Scene, LScene, AbstractVector{<:Union{Axis, Axis3, Scene, LScene}}}, title = nothing; merge = false, unique = false, kwargs...)
```
jkrumbiegel, line 1010: Docstring should be updated here

docs/src/reference/blocks/legend.md
```diff
@@ -60,6 +60,8 @@ With the keywords `merge` and `unique` you can control how plot objects with the
 If `merge` is `true`, all plot objects with the same label will be layered on top of each other into one legend entry.
 If `unique` is `true`, all plot objects with the same plot type and label will be reduced to one occurrence.
 
+To create a joint legend for multiple axes it is also possible to pass a `Vector` of axis objects.
```
jkrumbiegel: Would be good to add a simple visual example here, maybe one that uses merge too

jusack:
Thank you for the comment and looking at the PR. I have adjusted the test. However I am not sure if I the indexing into all the layers is the best way to get to the legend content. Is there an easier way to achieve what I was trying?

jkrumbiegel:
two more comments got stuck in the pipeline, I've submitted them again now. The indexing in the test is fine, using internals internally is ok :)

StefanPofahl:
I would like share my view:
My opinion about the functionality is: It is very important to add such a feature to the function `Legend()`!
Thanks a lot Julius for your work on this! 
My proposal is: 
What about an automatism: if there is a vector of axis handles join the legends of the axes?

Makie/src/makielayout/blocks/legend.jl
ffreyer: This isn't really related to the pr, but shouldn't the types here be `Union{AbstractAxis, Scene}` so this can work with other axis types more easily? Like `GeoMakie.GeoAxis` or `PolarAxis`? (And maybe we should also have `get_plots(ax::AbstractAxis) = ax.scene`?)

jkrumbiegel: Sounds sensible

ffreyer:
I merged the docs and tests from here into #5551 and merged that into master

PR #5508, opened 2026-01-22
Add Clustering Extension for dendrogram

kescobo:
# Description

Extends #2755 to add a package extension for the Clustering.jl `HClust` type. Draft still because I wasn't sure how (or if I should) add the `useheight` parameter to enable usage of the heights in the `HClust`. 

## Type of change

## Checklist

### DIFF
--- a/Makie/Project.toml
+++ b/Makie/Project.toml
@@ -66,15 +66,18 @@ UnicodeFun = "1cfade01-22cf-5700-b092-accc4b62d6e1"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [weakdeps]
+Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
 DynamicQuantities = "06fc5a27-2a28-4c7c-a15d-362465fb6821"
 
 [extensions]
+MakieClusteringExt = "Clustering"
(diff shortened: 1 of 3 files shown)


jkrumbiegel:
Weak deps are still dependencies as far as dependency hell, update churn etc. are concerned. So I'm generally hesitant to add more to Makie. Would this fit as an extension to Clustering instead?

kescobo:
Fair enough - I'd be happy to ask over there. I saw the one extension here so thought it might be worth it, but Makie def. has more deps than Clustering. 

That said, Clustering.jl hasn't had a commit merged in over a year, so not sure it's actively maintained.

kescobo:
It could also be an option to just highlight the `hcl_nodes` function in the docs, not export it, and let people know how it works. It's not *so* complicated :shrug:

kescobo:
Well - that issue on Clustering.jl has not received any reply. Shall I just add this to the docs as an example?

ffreyer:
> Weak deps are still dependencies as far as dependency hell, update churn etc. are concerned. So I'm generally hesitant to add more to Makie. Would this fit as an extension to Clustering instead?

The only thing this uses is the `Hclust` struct that hasn't changed in 8 years, or since 0.12. If we set the compat bound to `Clustering = ">= 0.12.0"` this will probably require 0 work for quite a while.

Issue #4747, opened 2025-01-23
A hybrid between menu and textbox, and a table


sprig:
Hello,

From the widgets that Makie currently supports, there are a couple simple ones missing IMHO:

- A "large menu"/"search box" aka widget where completions are generated dynamically and are shown when user selects/is typing into the field. At least for the web it should be simple to implement, I think, and I'd be willing to try.
- A table widget. Here acceptable alternatives are available for webui. However, it would be preferable to have this built in for portable UIs (between web and local gl). This might take some effort and I'm not sure I want to spend the time myself so just throwing it out there :-)

bjarthur:
by "large menu / search box" do you mean sth like this?  https://www.w3schools.com/howto/howto_js_filter_dropdown.asp

sprig:
yes, exactly.

preferably allowing scroll (ie be able to limit the size for the menu). a more advanced version would allow multiple selection, and not automatically reset the selected elements when search text is changed.

my priorities changed since writing this and i haven’t worked much with makie since, so i haven’t implemented it for myself. i would still be willing to, provided there is interest and the potential for merged PRs.

bjarthur:
i would definitely be interested in a menu which would somehow permit filtering the options with a search box.  i've got a list of ~20k items i need the user to be able to choose from.  that's too much to handle with a normal menu.  i would imagine that combining makie's textbox and menu source code would be a good place to start.

bjarthur:
@sprig see https://github.com/MakieOrg/Makie.jl/pull/5642

bjarthur:
@sprig see https://github.com/MakieOrg/Makie.jl/pull/5510 too

PR #5510, opened 2026-01-28
add table widget

SimonDanisch:
Still needs some review etc.
image


Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        tellheight = true
+        "The horizontal alignment of the table in its suggested bounding box."
+        halign = :center
+        "The vertical alignment of the table in its suggested bounding box."
+        valign = :center
+        "The alignment mode of the table in its parent GridLayout."
+        alignmode = Inside()
+
+        "The tabular data to display. Can be a NamedTuple or Dict{Symbol, Any} where each value is a column vector."
+        data = (a = [1, 2, 3], b = ["x", "y", "z"])
```
jkrumbiegel, line 1447: I would just use a matrix as the source, not every (visual) table conforms to this header/column logic. Matrix is simple and you can do anything with that.

A header row is just a differently formatted first row, or first couple of rows. For that it would make sense to be able to "fix" some rows and/or columns when scrolling, these would then effectively be the always visible "headers" without forcing a header concept.

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        "Custom column names to display. If `automatic`, uses the data's keys."
+        column_names = automatic
```
jkrumbiegel, line 1449: same here, don't need column_names

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        "Column widths. Can be `:auto` (equal widths), `:fit` (auto-fit to content), a number (all same width), or a vector of widths."
+        column_widths = :auto
+        "Row heights. Can be `automatic` (uniform using row_height), a number, or a vector of heights per row."
+        row_heights = automatic
+
+        "Index of the currently selected row. 0 means no selection."
+        i_selected = 0
```
jkrumbiegel, line 1456: this kind of granular selection logic doesn't make sense I think, the logic people will want to implement is going to be too varied to have something simple like this. Better to just give them robust click events and let them do the styling based on that.

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        selection = nothing
+        "Selected cell as (row, col) tuple. (0, 0) means no cell selection."
+        i_selected_cell = (0, 0)
+        "The data of the currently selected cell. This is the output observable for cell-level selection."
+        cell_selection = nothing
+
+        "Index of the column to sort by. 0 means no sorting."
+        sort_column = 0
+        "Sort direction, either `:ascending` or `:descending`."
+        sort_direction = :ascending
```
jkrumbiegel, line 1467: it's super easy to sort a table yourself so I don't think this should be builtin

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+
+        "Maximum number of visible rows. If `nothing`, shows all rows."
+        max_visible_rows = nothing
+        "Current scroll offset (row index to start displaying from)."
+        scroll_offset = 0
+        "Scroll speed multiplier for mouse wheel scrolling."
+        scroll_speed = 3.0
+
+        "Background color of the header row."
+        header_color = RGBf(0.2, 0.2, 0.2)
```
jkrumbiegel, line 1479: All of these special colors could be handled much more easily by making `color` not just a matrix but also a function. typst does it like that and it makes it very easy to implement alternating row colors, header colors, etc. Like

```julia
color = function (i, j)
    i == 1 ? header_color : iseven(i) ? ...
```

this concept can also be used for paddings, cell border colors, etc

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        grid_color = RGBf(0.8, 0.8, 0.8)
+        "Width of grid lines."
+        grid_linewidth = 1.0
+        "Whether to show vertical grid lines."
+        show_vertical_lines = true
+        "Whether to show horizontal grid lines."
+        show_horizontal_lines = true
+
+        "Callback function `(table, row_index, row_data) -> nothing` called on row click."
+        on_row_click = nothing
```
jkrumbiegel, line 1520: If we remove the concept of row/column header, `on_cell_click` is enough here. Although maybe the others like `over`, `enter`, `out`, etc would also be good, I kind of would just want all the mousestatemachine options but with the work of determining i,j done for the user

Makie/src/makielayout/types.jl
```diff
@@ -1380,6 +1380,134 @@ end
...
+        cell_padding = (8, 8, 4, 4)
+
+        "Whether to show grid lines."
+        show_grid = true
+        "Color of grid lines."
+        grid_color = RGBf(0.8, 0.8, 0.8)
+        "Width of grid lines."
+        grid_linewidth = 1.0
+        "Whether to show vertical grid lines."
+        show_vertical_lines = true
```
jkrumbiegel, line 1515: better to use a function of cell border colors instead of `show_vertical`, `show_horizontal`, etc.. much more flexible

jkrumbiegel (changes requested):
my general comment is to make the logic much more generic and allow functions to style, like typst does it. makes it really easy to set up common styles and still avoids having to update a color matrix when the data changes. Conditional formatting can also be done like `(i, j) -> data[i,j] > some_value : "red" : "black"`

Makie/src/makielayout/types.jl
SimonDanisch: ah, but its a nice default user interaction when clicking on the header, I think

jkrumbiegel: one of many possible things one could do, maybe prepare it as some kind of transformation interaction one can choose rather than a special builtin thing?

bjarthur:
this PR would close half of https://github.com/MakieOrg/Makie.jl/issues/4747#issuecomment-3053787566 IIUC

BambOoxX:
I stumbled upon your discourse thread, and pulling over it led me here. Could this PR benefit from the latest developments of `PrettyTables` ? It currently has a pretty advanced table formatting system, no pun intended. Just my two cents.

SimonDanisch:
Not sure, the last time I looked into `PrettyTables`, I couldn't get it to work for other packages, since it's basically build for REPL/Terminal only and that assumption is everywhere. Correct me if I'm wrong, but the main primitive is IOBuffer(), and without reparsing the resulting strings, you can't really use it for anything else.

BambOoxX:
Yeah, you are probably right. I used it to feed `text!(...)` to `Makie`, but that is not very efficient i guess. There might be good ideas to borrow from there though such as the formatters / highlighters.

PR #5412, opened 2025-11-17
Fix log barplots

ffreyer:
# Description

Fixes #4549
Fixes #5440

Another attempt at #4911/#4849, i.e. fixing issues related to bars starting at 0 in a log-transformed scene/axis.

This removes the modification of `fillto` and instead tries to enable "infinite" bars. To avoid rendering complications the bars are really just larger than the visible area and occasionally grow or shrink to stay in a fitting range. This also adds a custom `boundinbox` method that should keep limits in check when the transform_func sends 0 to -Inf. (Same `log(0.5 * min_height)` as in #4849)

Also fixes #5320 because I switched from `Rect` to `Polygon`. (Because this applies transform function which may turn rects into other things)

## Type of change

Delete options that do not apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [x] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #5511, opened 2026-01-28
fix histogram in polaraxes in CairoMakie

### LINKED ISSUE #3175: PolarAxis histogram fails in CairoMakie
```julia
f = Figure()
ax = PolarAxis(f[1, 1])
p = hist!(ax, 2pi .* rand(300), direction = :x, strokewidth = 1, bins = range(0, 2pi, length=37))
f
```

### In GLMakie

Screenshot from 2023-08-23 17-57-04

### In CairoMakie

Screenshot from 2023-08-23 17-56-45

This implies that GLMakie is applying the transform function to the vertices of the rectangles generated in `hist!` while CairoMakie doesn't. It might be applying earlier or not at all.

This also needs an example in the docs once it works!

bjarthur:
fixes https://github.com/MakieOrg/Makie.jl/issues/3175

tests pass locally.

gemini wrote this.  seems reasonable but i'm not that familiar with this part of Makie so please review thoroughly.

### DIFF
--- a/CairoMakie/src/overrides.jl
+++ b/CairoMakie/src/overrides.jl
@@ -140,8 +140,48 @@ function draw_poly(scene::Scene, screen::Screen, poly, shapes::Vector{<:Union{Re
     space = poly.space[]::Symbol
     planes = poly.clip_planes[]::Vector{Plane3f}
 
+    # Check for non-linear transform (e.g. PolarAxis)
+    tf = Makie.transform_func(scene)
+    is_linear = (tf === identity || tf == (identity, identity))
+    is_nonlinear = !is_linear
+
     projected_shapes = map(shapes) do shape
-        clipped = clip_shape(planes, shape, space, model)
+        # [FIX START]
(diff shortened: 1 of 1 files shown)


ffreyer:
#5412 also fixes this iirc

bjarthur:
#5412 is close!  the r axis though doesn't start at 0.  is that maybe why it isn't merged yet?  is there a workaround using maybe `rlimits` or `rticks`?

Screenshot 2026-01-29 at 8 08 48 AM

with this code from #3175 

```
f = Figure()
ax = PolarAxis(f[1, 1])
p = hist!(ax, 2pi .* rand(300), strokewidth = 1, bins = range(0, 2pi, length=37)) # direction = :x for ax.theta_as_x = false
f
```

ffreyer:
Got around to fixing that too. It was caused by `data_limits` including transform_func, because it used the default implementation that checks child plots which use post-transform positions

bjarthur:
superseded by https://github.com/MakieOrg/Makie.jl/issues/3175

PR #5524, opened 2026-02-08
disambiguate from 撒き餌 (same romanization)

rokke-git:
# Description

the main readme says "the japanese word maki-e," but [maki-e](https://en.wiktionary.org/wiki/maki-e) is an english word.  
you should either put the actual word (蒔絵, not 撒き餌 which has the same romanization),  
or say something like "the word for the japanese technique of..."

## Type of change

- [x] documentation, I guess?

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The backend packages **GLMakie**, **WGLMakie**, **CairoMakie** and **RPRMakie** 
 You can use Makie to interactively explore your data and create simple GUIs
 in native windows or web browsers, export high-quality vector graphics or even raytrace with physically accurate lighting.
 
-The name Makie (we pronounce it Mah-kee) is derived from the Japanese word [<ruby><rb>蒔</rb><rp>(</rp><rt>まき</rt><rp>)</rp><rb>絵</rb><rp>(</rp><rt>え</rt><rp>)</rp></ruby>](https://en.wikipedia.org/wiki/Maki-e), which is a technique to sprinkle lacquer with gold and silver powder.
+The name Makie (we pronounce it Mah-kee) is derived from the Japanese decoration technique [maki-e](https://en.wikipedia.org/wiki/Maki-e), which uses lacquer to make designs with gold and silver powder.
 Data is the gold and silver of our age, so let's spread it out beautifully on the screen!
 
 To learn more, we invite you to visit the documentation at [docs.makie.org](http://docs.makie.org/stable/).


SimonDanisch:
image
I guess this doesnt work well internationally?
Same in the comments: 
image
Maybe we just do `the word for the japanese technique of`?

rokke-git:
> I guess this doesnt work well internationally? 

huh, that's kinda disappointing that browsers don't automatically load in a CJK font when you encounter characters needing one yet.

> Maybe we just do `the word for the japanese technique of`?

ok

just curious, if you visit for example [here](https://en.wiktionary.org/wiki/%E8%92%94%E7%B5%B5#Japanese) or [here](https://ja.wikipedia.org/wiki/%E8%92%94%E7%B5%B5) do you get just a ton of boxes?

rokke-git:
actually, if I do [<ruby lang="ja"><rb>蒔</rb><rp>(</rp><rt>まき</rt><rp>)</rp><rb>絵</rb><rp>(</rp><rt>え</rt><rp>)</rp></ruby>](https://en.wikipedia.org/wiki/Maki-e) does it render properly?

SimonDanisch:
Nope!

rokke-git:
looking around, it's possibly a side effect of firefox's resist fingerprinting; are you using that? I am as well but apparently my set of fonts includes cjk characters. also, was the nope only to my second attempt to do the link or are [these](https://en.wiktionary.org/wiki/%E8%92%94%E7%B5%B5#Japanese) [pages](https://ja.wikipedia.org/wiki/%E8%92%94%E7%B5%B5) also unreadable?

one more attempt: <span font-family="Noto Sans JP">[<ruby lang="ja" font-family="Noto Sans JP"><rb>蒔</rb><rp>(</rp><rt>まき</rt><rp>)</rp><rb>絵</rb><rp>(</rp><rt>え</rt><rp>)</rp></ruby>](https://en.wikipedia.org/wiki/Maki-e)</span>

SimonDanisch:
I'm using brave on opensuse... Could be my browser, but I think if that's possible, I'd rather go with the English description.

rokke-git:
I was mostly just curious; def not mozilla's rf then since brave is based on chromium. I would still kinda like to know if your browser is making [any](https://en.wiktionary.org/wiki/%E8%92%94%E7%B5%B5#Japanese) [pages](https://ja.wikipedia.org/wiki/%E8%92%94%E7%B5%B5) with japanese text unreadable or if it has some kind of metric that it's using

ffreyer:
Why not both, like wikipedia? "Maki-e (蒔絵)"

SimonDanisch:
Brave and Firefox look both broken

rokke-git:
> Brave and Firefox look both broken

oh, no; I meant that I thought maybe rf could be a reason that it doesn't load. firefox looks normal when I try it, it's just brave that looks broken. unless you tested both?

also, that's why I keep asking you if [these](https://en.wiktionary.org/wiki/%E8%92%94%E7%B5%B5#Japanese) [pages](https://ja.wikipedia.org/wiki/%E8%92%94%E7%B5%B5) look normal to you; either there's something we can do to make it work, or your browser just looks broken in the first place

SimonDanisch:
>also, that's why I keep asking you if [these](https://en.wiktionary.org/wiki/%E8%92%94%E7%B5%B5#Japanese) [pages](https://ja.wikipedia.org/wiki/%E8%92%94%E7%B5%B5) look normal to you;

That was actually my answer ;) So the pages look broken both on Firefox and Brave on my PC.

SimonDanisch:
On windows it looks fine for both ;)

ffreyer:
Everything is fine for me on ubuntu firefox.

I don't think we should avoid Japanese symbols here.  I would assume that they come from the same font that wikipedia etc uses, and that pretty much every windows and mac can display these. It's probably just some linux systems that can't render them, either because they use a minimal set of fonts or because they avoid proprietary fonts. But that's not our problem, that's a problem with the system setup.

rokke-git:
> the same font wikipedia etc uses

to be precise, it's the browser that chooses the font, not the website; but the website can influence this choice with the `lang` and `font-family` tags (that's what I tried adding in the earlier comments). typically, the browser will keep trying all installed fonts to fill in box-characters unless you have resist-fingerprinting enabled, which will limit attempts to a known set of fonts which (ideally) cover the set of characters that people use

it should never be a proprietary-ness problem, because [noto](https://en.wikipedia.org/wiki/Noto_fonts) did this amazing thing where they are trying to make a font which covers [all of unicode](https://notofonts.github.io/overview/) under an open font license (broken up by language). while it doesn't cover everything; it basically covers what people actually use.

commit 2 or 3 are equally correct, you can use whichever you want. would prob be good to add `lang=ja` to the `<ruby>` tag [for accessibility](https://www.w3.org/International/questions/qa-lang-why.en) if you choose commit 2

PR #5534, opened 2026-02-21
Render LinePattern as vector graphics in CairoMakie

KristianHolme:
CairoMakie previously rasterized Makie.LinePattern to a bitmap. This change draws LinePattern fills as vector hatch lines (clip path + stroke) so output stays fully vector for SVG/PDF. It also improves resolution of the pattern when rendering to PNG.

I am not sure if the reference test added is strictly necessary in addition to the "Color patterns in recipes" test and others?

Locally, the following reference tests are failing, due to the new rendering.

- "Color patterns in recipes"
- "Color patterns"
- "textlabel"

<details>

<summary> Reference test fails diffs </summary>

- "Color patterns in recipes"

step-2

- "Color patterns"

step-2

- "textlabel"
textlabel

</details>

Coded using Codex 5.3/Opus 4.6. Description below also written by the agents.

## Description

CairoMakie previously rasterized `Makie.LinePattern` to a bitmap for SVG/PDF, producing embedded `<image>` elements. This change draws LinePattern fills as vector hatch lines (clip path + stroke) so that SVG/PDF stay fully vector and scale without quality loss.

**Motivation:** Exporting figures with hatched fills (barplot, poly, band, textlabel background) should yield vector paths/lines in SVG/PDF, not embedded rasters.

**Summary of changes:**
- **CairoMakie:** LinePattern is no longer rasterized; `to_cairo_color(::LinePattern)` passes it through, and a new vector path draws hatches via clip + stroke (`linepattern_spacing`, `draw_linepattern_hatch!`, `draw_linepattern_fill!`). Poly, mesh 2D (when `uv_transform` is automatic), and Band use this path; Band reuses a single path builder for all fill types. 
- **Makie:** New Poly attribute `project_point3_to_2d` so backends can opt into flattening Point3 polygons to 2D (e.g. screen-space shapes). TextLabel sets it for its background poly so CairoMakie can draw it with the vector LinePattern path.
**Testing:** New CairoMakie tests in `svg_tests.jl` ensure LinePattern outputs contain no `<image>` tags. Added reference image test `@reference_test "LinePattern in recipes"` in `ReferenceTests/src/tests/examples2d.jl` to lock in visual behavior across recipes.

Fixes # (leave blank or add issue number if applicable)

## Type of change

## Checklist

- [x] Added unit tests for new algorithms, conversion methods, etc. *(CairoMakie `svg_isnt_rasterized` tests)*
- [x] Added reference image tests for new plotting functions, recipes, visual options, etc. *(`@reference_test "LinePattern in recipes"` in `ReferenceTests/src/tests/examples2d.jl`)*

## Images
(Links to pdfs, easier to see difference with vector graphics there)
Before:
[linepattern_examples_4996d7dfa.pdf](https://github.com/user-attachments/files/25461448/linepattern_more_examples_4996d7dfa.pdf)

linepattern_more_examples_4996d7dfa

After:
[linepattern_examples_ce285671f.pdf](https://github.com/user-attachments/files/25463341/linepattern_more_examples_ce285671f.pdf)

linepattern_more_examples_ce285671f

<details>

<summary>Reference test image</summary>

ref_linepattern_in_recipes_ce285671f

</details>


ffreyer (approved):

ffreyer:
Now that I'm looking at this again I'm noticing that this cause some color changes: 
(GLMakie | new CairoMakie, see top left)
image

The reason for this is that the way Makie mixes lines and background is different from how Cairo(Makie) does it.

Makie only mixes the line color and the background color where the line and background meet, i.e. in the AA (anti-aliasing) region. The core of the line is just the line color, the space between just the background color.

CairoMakie (with the state of the pr) draws the background first and then draws lines on top. If those lines have transparency they will blend with the background, core included. If we change the blending operator to `Cairo.OPERATOR_SOURCE` we would not blend the core with the background, but we would also not blend the AA region with the background. This ends up looking even darker in the test above.

I don't think there is a way to fix this other than changing the other backends to match. I also don't think this is a big problem. I'd imagine most patterns are opaque or have a fully transparent background and even if they don't you can just tweak colors to work for CairoMakie. 

Should we ignore the color differences and merge this regardless? Imo the cleaner AA is worth it.

KristianHolme:
I agree that this trade-off is very much worth it. We could maybe add a section to the docs on [Transparency](https://docs.makie.org/stable/explanations/transparency#Transparency) but i don't know if it is really necessary, as the difference does not look that big?

ffreyer:
The worst case is something like this:
```julia
pat = Makie.LinePattern(
    linecolor = RGBAf(1,0,0,0),
    backgroundcolor = RGBAf(0,0,1,1),
    width = 5
)
mesh(Rect2f(0, 0, 1, 1), color = pat)
```
(GLMakie | CairoMakie pr)

image

(Here you could just swap background and line colors)

KristianHolme:
That was a bit bigger difference...

At least to me the Cairo version seems more correct with respect to the Makie syntax of lines + background. We ask for "invisible lines on blue background" which is what Cairo does, whereas the GL version is more like "alternating invisible and blue lines".
So I agree with your comment above that the best approach is to merge this and then maybe change how the other backends work.

jkrumbiegel:
I can imagine some niche cases in which you might want transparent lines and visible background, and in the crosshatch you can't just switch colors and adjust line thicknesses to get the same result as in previous GLMakie. On the other hand the gain from sharp hatch lines seems valuable. So for me it would be ok to change this I think.

ffreyer:
In that case you can still manually convert to an image pattern and use that:
```julia
pat = Pattern('x', linecolor = RGBAf(1,0,0,0), backgroundcolor = RGBAf(0,0,1,1))
ipat = Makie.ImagePattern(Makie.to_image(pat))
mesh(Rect2f(0, 0, 1, 1), color = ipat)
```
image

This probably isn't obvious though and needs some documentation. I'll add it later and then merge this.

PR #5536, opened 2026-02-26
Fix rendering of LaTeXString when using bracket

KristianHolme:
# Description
Passing LaTeXString as bracket text converted to regular String. This now preserves the LaTeXString without conversion to regular String.

## Type of change

## Checklist

- [x] Tests run locally

## Images
**Before:**
test_bracket_latex_output_before

**After:**
test_bracket_latex_output_after


jkrumbiegel:
I think the better test would be whether the types end up unconverted in the text child plot. Short of using a reference image to confirm the final output.

Makie/src/basic_recipes/bracket.jl
```diff
@@ -131,7 +131,7 @@ function plot!(pl::Bracket)
 
         # TODO: add a broadcast/map version of broadcast_foreach doing this:
         bps = BezierPath[]
-        text_pos = Tuple{String, Point2f}[]
+        text_pos = Tuple{Union{String, LaTeXStrings.LaTeXString}, Point2f}[]
```
henrik-wolf:
```suggestion
        text_pos = Tuple{Union{String, LaTeXStrings.LaTeXString, RichText}, Point2f}[]
```
to avoid the same issue with `RichText` strings in the future.
(To allow for any type of "string renderable" to be passed (once that is possible), maybe even
```suggestion
        text_pos = Tuple{Any, Point2f}[]
```)

KristianHolme: Thanks!

KristianHolme:
> I think the better test would be whether the types end up unconverted in the text child plot. Short of using a reference image to confirm the final output.

Updated test to check `:input_text` attribute of the child text plots, let me know if that was not what you intended @jkrumbiegel .
Are these tests sufficient or should i add a reference test?

jkrumbiegel:
I think it's fine, if the inputs of the text plots have the correct types we can rely on reftests of text itself. Reftests are heavier so it's nice not to make them too freely if unit tests can suffice. Alternatively, you could edit one of the existing bracket reftests and change a string to Latex.

KristianHolme:
Replaced two texts in reftest "bracket scalar" to use one LaTeXString and one RichText. Does it look ok? If ok, is there anything else I should do?

jkrumbiegel:
Yeah looks ok, the reftests will fail and a maintainer has to update refimages before merging

PR #5542, opened 2026-03-02
Implement text glow rendering in CairoMakie

### LINKED ISSUE #4625: no glow in CairoMakie
Running `text((0,0), text="abc", glowwidth=5, glowcolor=:red)` shows glow with the GLMakie backend, but just shows the raw text (without glow) with CairoMakie.
Would be nice to support it for Cairo as well, including pdf exports :)

- [x] what version of Makie are you running? (`]st -m Makie`) latest, 0.21.16
- [x] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`) yes
- [x] What platform + GPU are you on? any

aplavin:
## Summary
- Implement `glowwidth`/`glowcolor` support for `Text` plots in CairoMakie
- Builds a combined glyph path per text block, then strokes with decreasing widths using `stroke_preserve` to approximate a smooth glow falloff

Fixes #4625

## Example

```julia
using CairoMakie
f = Figure()
ax = Axis(f[1,1])
text!(ax, (0,0), text="abc", glowwidth=5, glowcolor=:red, fontsize=30)
f
```
text_glow_test2


asinghvi17:
Seems reasonable to me, thanks!

ffreyer (approved):
Seems fine to me too. Shouldn't interfere with any existing code and if we find a better solution we can always just change it in the future.

The test failures are from reference images that now correctly include glow. No real problems there, and no need to add more refimages

PR #4984, opened 2025-05-15
allow Legend to be created from multiple Axis

TomRottier:
# Description

Adds the ability to create a Legend from a vector of axes. 

Modified from https://discourse.julialang.org/t/how-to-combine-the-legend-of-multiple-axes-on-same-figure-into-one-in-makie-jl/97299/3?u=tomrottier

```julia
using GLMakie

f = Figure()
ax1 = Axis(f[1,1])
ax2 = Axis(f[1,2])

ps = [0.1, 0.2, 0.3, 0.4]
xs = range(0, 5, 100)
f1(x, p) = p * x^2
f2(x, p) = p * sin(x)

for p in ps
    lines!(ax1, xs, f1.(xs, p), label=string(p))
    lines!(ax2, xs, f2.(xs, p), label=string(p))
end

scatter!(ax1, xs[1:5:end], f1.(xs[1:5:end], ps[1]), label=string(ps[1]), )

Legend(f[2,:], [ax1, ax2], orientation=:horizontal, merge=true, unique=false)
```
tmp

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [ ] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

PR #5551, opened 2026-03-05
allow Legend to be created from multiple Axis (re-submit of #4984)

TomRottier:
resubmitting as accidentally closed


jusack:
If this is merged we can close my PR #5500. Maybe you can also check if there is anything you want to include from that version, I was not able to finalize it in the last months.

Issue #5214, opened 2025-07-24
Heatmap colorrange does not accept `nothing`


ederag:
```julia-repl
(jl_r2FXiV) pkg> st -m Makie
Status `/tmp/jl_r2FXiV/Manifest.toml`
  [ee78f7c6] Makie v0.24.4

julia> versioninfo()
Julia Version 1.10.10

julia> using GLMakie

julia> heatmap(rand(10, 10); colorrange = nothing)
```

<details>
<summary>Error message</summary>

```julia-repl
julia> err
1-element ExceptionStack:
Failed to resolve gl_renderobject:
[ComputeEdge] gl_renderobject = #168((x_transformed_f32c, y_transformed_f32c, space, alpha_colormap, scaled_color, scaled_colorrange, lowclip_color, highclip_color, nan_color, model_f32c, instances, uniform_clip_planes, uniform_num_clip_planes, depth_shift, visible, fxaa, resolution, projection, projectionview, view, upvector, eyeposition, view_direction, ), changed, cached)
  @ /usr/local/build/julia/depot/packages/GLMakie/WBusz/src/plot-primitives.jl:238
[ComputeEdge] scaled_colorrange = (::MapFunctionWrapper(#459))((colorrange, colorscale, scaled_color, ), changed, cached)
  @ unknown method location
  with edge inputs:
    colorrange = nothing
    colorscale = identity
    scaled_color = Float32[0.7449372 0.94140023 … 0.4109077 0.2630752; 0.9797123 0.96500045 … 0.48283303 0.77538806; … ; 0.57946336 0.2471568 … 0.91459095 0.23615626; 0.50757957 0.39451396 … 0.85274863 0.37235495]
Triggered by update of:
  colorrange, colorscale, arg1, colorscale or alpha
Due to ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type Float32

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:84
  convert(::Type{T}, ::ColorTypes.Colorant{T, 1} where T) where T<:Real
   @ ColorTypes /usr/local/build/julia/depot/packages/ColorTypes/L7biZ/src/conversions.jl:110
  convert(::Type{T}, ::CartesianIndex{1}) where T<:Number
   @ Base multidimensional.jl:135
  ...

Stacktrace:
  [1] (::GeometryBasics.var"#28#42"{Float32, Nothing})(i::Int64)
[...]

PR #5555, opened 2026-03-10
Allow adaptive colorrange in one direction

### LINKED ISSUE #3992: Possibility to set low or high bound of colorrange  and let the other one be adaptive
### Feature description
It would be good to be able to set only one of the bounds of the colorrange and let the other one be adaptive. 
We had a discussion on slack and one possible API would be to use `colorrange=(0, nothing)` to set the lower bound of the colorrange and let the upper be adaptive and `colorrange=(nothing, 100)` to set the upper bound and let the lower bound be adaptive.
@jkrumbiegel mentioned that nothing could be replaced here by automatic.

felixcremer:
# Description

Fixes #3992 

This allows to set the low or high bound of the colorrange explicitly and let the other side be adaptive. My main motivation for this is in PyramidScheme.jl to set the low bound but still change the upper bound based on the actual data of the zoomed in value. 

I added an entry to the cheat sheet as documentation and didn't see another place to add it. 
I added a test in primitives.
 

## Type of change

## Checklist


felixcremer:
Bump. The tests go through and I also just formatted the file with runic so that should also work.

Makie/src/compute-plots.jl
```diff
@@ -243,6 +243,10 @@ function register_colormapping!(attr::ComputeGraph, colorname = :color)
             return nothing
         elseif colorrange === automatic
             return autorange
+        elseif first(colorrange) == automatic
+            return Vec2f((first(autorange), last(colorrange)))
+        elseif last(colorrange) == automatic
+            return Vec2f((first(colorrange), last(autorange)))
```
ffreyer, line 249: I guess the only question is whether we want this to be `nothing` instead, since that's more accessible. Or maybe allow both?

felixcremer: I used automatic, because according of this discussion https://github.com/MakieOrg/Makie.jl/issues/5214#issuecomment-3113309779

PR #5561, opened 2026-03-19
fix(CairoMakie): Batch glyphs in PDF/SVG text export

### LINKED ISSUE #3369: Output axes numbers and text as one single text object
### Feature description

A follow-up to this discourse post: https://discourse.julialang.org/t/how-to-make-numbers-texts-appear-as-one-single-text-object-in-makie/105977

It would be be nice if the text or numbers in the exported plot (vector graphics format exported as pdf) can appear as one single text object when the figure is inspected in third-party software such as illustrator or inkscape, which will make any post-editing of those text much easier. Right now it appears rather random how individual alphabets or symbols in the original string are grouped.

N/A

### LINKED ISSUE #3862: CairoMakie: export `text` as one many-letter object, not multiple single-letter objects
When exporting vector-graphics plots containing `text`, the individual glyphs are each stored as a separate object, rather than having a single object for each piece of text. E.g., `text("foo")` will produce three text objects "f", "o", and "o", with their own locations etc., rather than a single object "foo". Example below (each red box is a separate object):
image

This is rather disruptive for workflows that edit the exported figure post-export, e.g., in Illustrator or Inkscape, because it becomes impossible to resize the text or change fonts without the kerning getting ruined.

I previously asked about this on Slack a while back, and got a the following nice answer from @jkrumbiegel:
> currently that happens here
> https://github.com/MakieOrg/Makie.jl/blob/746b03f46282f2385d438b01b486a74ea10f023d/CairoMakie/src/primitives.jl#L735
> I assume one would have to draw multiple glyphs at once with https://www.cairographics.org/manual/cairo-text.html#cairo-show-text-glyphs

I had hoped to figure it out from those hints but haven't been able to grok it, so I'm just registering this here for posterity.
My understanding is that it requires replacing `cairo_glyph_path` with `cairo_show_text_glyps`, but that it also requires a different way of inputting the various text transformations.

---

(Related to https://github.com/MakieOrg/Makie.jl/issues/2015, but not the same issue)

jkrumbiegel:
## Summary

- Batch consecutive glyphs with identical rendering properties (font, color, font matrix, stroke) into a single `cairo_show_glyphs` call instead of one call per glyph
- This makes PDF/SVG editors (Inkscape, Illustrator) treat glyphs from one text string as a single text object, enabling post-export text editing without broken kerning
- Per-character styling (e.g. per-char colors) correctly splits into separate batches; 3D perspective text where the font matrix varies per glyph falls back to individual rendering

Closes #3862, closes #3369

## How these changes were tested

- Generated PDF and SVG with uniform text, per-character colored text, stroked text, and multiple labels
- Inspected PDF stream: verified `text("Hello World")` produces `(Hello World)Tj` (one op) instead of 11 separate `Tj` ops
- Inspected SVG: verified all glyphs of a string are in one `<g>` group
- Verified per-char colored text correctly splits into separate BT/ET blocks
- Visual comparison of PDF/SVG/PNG output

Created with the help of [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - CairoMakie now batches glyphs from the same text string into a single PDF/SVG text object, so that text can be selected and edited as a unit in vector editors like Inkscape and Illustrator [#5561](https://github.com/MakieOrg/Makie.jl/pull/5561)
+- Fixed `annotation` not showing lines/arrows when `text` is blank [#5560](https://github.com/MakieOrg/Makie.jl/pull/5560)
 
 ## [0.24.9] - 2026-03-04
 
--- a/Makie/src/basic_recipes/annotation.jl
+++ b/Makie/src/basic_recipes/annotation.jl
@@ -230,8 +230,10 @@ function plot!(p::Annotation)
     )
 
     # still without offset
+    # Empty strings produce non-finite Rect3d() bounding boxes, replace with zero-size rects
+    _guard_nonfinite(bb) = isfinite_rect(bb) ? bb : Rect2d(0, 0, 0, 0)
     map!(p, [txt.raw_string_boundingboxes, p.screenpoints_target], :text_bbs) do bboxes, px_pos
-        return Rect2d.(bboxes) .+ px_pos
+        return _guard_nonfinite.(Rect2d.(bboxes)) .+ px_pos
     end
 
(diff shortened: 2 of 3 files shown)


kaandocal:
This is fantastic, thank you!

thchr:
Indeed, fantastic stuff: thanks a ton..!

PR #5562, opened 2026-03-20
refactor(ffmpeg): Switch from FFMPEG_jll to FFMPEG_nogpl_jll

jkrumbiegel:
The main motivation here is that while Makie is MIT, the FFMPEG_jll dependency with the gpl and nonfree components that it has built in muddies the ability to use Makie as its license intends. I think the `record` function is central enough that it's preferable to keep it by default, so an extension on FFMPEG_jll or so does not really work. On the other hand, I don't think most users care about the exact codec they're getting. I can't say that I'm super knowledgeable about the settings Claude suggested here, but it seems that `libaom-av1` is a modern alternative to the ubiquitous `libx264`, so some older devices might not play it well but since the 2020s, most things should. Modern devices also generally have hardware support. I think that's acceptable in the context Makie is used in, also, people can still convert to any codec they want if they choose to, or use webm, mkv or even gif.

## Summary

- Replace `FFMPEG_jll` with `FFMPEG_nogpl_jll` to avoid distributing GPL-encumbered codecs (`libx264`, `libx265`, `libfdk-aac`)
- MP4 encoding now uses `libaom-av1` (AV1) instead of `libx264` (H.264), with `-cpu-used 8` for fast encoding
- MKV encoding now explicitly uses `libvpx-vp9` instead of relying on ffmpeg defaults (which could fail without libx264)
- WebM and GIF encoding unchanged
- `profile` and `pixel_format` options are kept for backwards compatibility but no longer apply (a warning is issued if set)
- `compression` (CRF) now also applies to MKV format

## How these changes were tested

- Recorded test MP4 (AV1) and MKV (VP9) animations via CairoMakie and verified playback
- Updated warning tests in `Makie/test/record.jl` to match new allowed-format rules

Created with the help of [Claude Code](https://claude.com/claude-code)


jkrumbiegel:
The failing backend tests are just slight changes in some mp4 references, which makes sense as this is a format with lossy compression, so you're not going to have exactly the same result with two different codecs.

Makie/src/ffmpeg-util.jl
```diff
@@ -154,13 +142,17 @@ function to_ffmpeg_cmd(vso::VideoStreamOptions, xdim::Integer = 0, ydim::Integer
...
+        `
     elseif format == "mp4"
-        `-profile:v $(profile)
+        `-c:v libaom-av1
          -crf $(compression)
-         -preset slow
-         -c:v libx264
-         -pix_fmt $(pixel_format)
+         -b:v 0
+         -cpu-used 8
```
asinghvi17: That's the worst quality (`-cpu-used 8`), right?

jkrumbiegel: Currently that's the least of my problems :D It's more that support for the modern AV1 codec is not that great, yet, even though it seems to be where things are heading. But there also seems to be no clear alternative to libx264 to get h.264 videos. There's openh264 but it seems to focus on livestreaming and is generally considered much less featureful than libx264. I haven't tried it yet, I don't think it's enabled in FFMPEG_jll

jkrumbiegel:
Closing in favor of https://github.com/MakieOrg/Makie.jl/pull/5587/changes

PR #5568, opened 2026-03-25
Annotation stackoverflows when having empty strings

### LINKED ISSUE #5567: Annotation stackoverflows when having empty strings
- [x] what version of Makie are you running? (]st -m Makie)
0.24.9
- [x] can you reproduce the bug with a fresh environment ? (]activate --temp; add Makie)
yes
WSL(Ubuntu 24.04) and Windows
- [x] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)
Cairo Makie
```julia
using CairoMakie   # or GLMakie

# 1. Minimal dummy data
x = [1.0, 2.0, 3.0]
y = [1.0, 2.0, 3.0]

# 2. Labels containing LaTeXStrings (including an empty string, matching the original case)
labels = [L"$H^+$", "", L"$O_2^+$"]

# 3. Setup figure
fig, ax, plt = scatter(x, y)

# 4. Trigger the bug using `annotation!`
# This will cause a StackOverflowError due to infinite recursive updates
# in ComputePipeline.jl / calculate_best_offsets!
annotation!(ax, x, y, text=labels)

# Attempt to display (will likely crash before rendering)
display(fig)
```

hz-xiaxz:
<!--
* Any PR that is not ready for review should be created as a draft PR.
* Please don't force push to PRs, it removes the history, creates bad notifications, and we will squash and merge in the end anyways.
* Feel free to ping for a review when it passes all tests after a few days (@simondanisch, @ffreyer). We can't guarantee a review in a certain time frame, but we can guarantee to forget about PRs after a while.
* Allowing write access on the PR makes things more convenient, since we can make edits to the PR directly and update it if it gets out of sync with `master` (should be automatic if not disabled).
* Please understand, that some PRs will take very long to get merged. You can do a few things to optimize the time to get it merged:
    * The more tests you add, the easier it is to see that your change works without putting the work on us.
    * The clearer the problem being solved the easier. A PR best only addresses one bug fix or feature.
    * The more you explain the motivation or describe your feature (best with pictures), the easier it will be for us to priorize the PR.
    * Changes with more ambigious benefits are best discussed in a github [discussion](https://github.com/MakieOrg/Makie.jl/discussions) before a PR.
* What deserves a unit test or a reference image test isn't easy to decide, but here are a few pointers:
   * Makie unit tests are preferable, since they're fast to execute and easy to maintain, so if you can add tests to `Makie/src/test`
   * For new recipes or any changes that may get rendered differently by the backends, one should add a reference image test to the best fitting file in `Makie\ReferenceTests\src\tests\**` looking like this:
   ```julia 
   @reference_test "name of test" begin
        # code of test
        ...
        # make sure the last line is a Figure, FigureAxisPlot or Scene
    end
    ``` 
    Adding a reference image test will let your PR fail with the status "n reference images missing", which a maintainer will need to approve and fix. 
    Ideally, a comment with a screenshot of the expected output of the reference image test should be added to the PR.
    We prefer one reference image test with many subplots over multiple reference image tests.
-->
# Description

Fixes #5567 (issue)

# Bug Report: annotation! StackOverflowError with empty strings                                                                                   
                                                                                                                                                  
 ## Reproduction                                                                                                                                    
   ```julia                                                                                                                            
  using CairoMakie, LaTeXStrings
  x = [1.0, 2.0, 3.0]; y = [1.0, 2.0, 3.0]
  fig, ax, plt = scatter(x, y)
  annotation!(ax, x, y, text=[L"$H^+$", "", L"$O_2^+$"])
  display(fig)  # StackOverflowError
```

  ## Root Cause

  The bug has two interacting components:

  1. normalize(zero_vector) produces NaN in calculate_best_offsets!

  In annotation.jl, the label placement algorithm initializes offsets by pushing labels toward the viewport center:
```julia
  v = normalize(center - bb_center)
  offsets[i] = 0.1 * algorithm.repel * v
```

  When a label has a zero-size bounding box (from an empty string ""), its bb_center equals its screen position. If that screen position coincides
   with the viewport center (as with [1,2,3] data where point 2 is at center), then center - bb_center ≈ (0, 0) and normalize((0, 0)) = (NaN, 
  NaN).                                                                                                                                           
                                                            
  The NaN propagates through the repulsion iterations, making all offsets NaN.

  2. NaN != NaN defeats the ComputePipeline's convergence check

  The annotation creates a feedback loop:
  - offsets → on callback → update!(txt, offset=offsets) → text recomputes → bounding boxes re-resolve → offsets recomputed

  The ComputePipeline's onchange handler is supposed to break this cycle by comparing values:

  elseif val != obs[]  # skip if unchanged

  But != uses ==, and NaN == NaN is false in IEEE 754. So [NaN, ...] != [NaN, ...] always returns true, meaning the Observable is always
  re-notified. This creates infinite re-entrant Observable notifications that overflow the call stack.

  Investigation Process

  1. Reproduced the crash — confirmed StackOverflowError with 50K+ recursive update! calls
  2. Isolated the trigger — tested combinations:
    - ["A", "B", "C"] with [1,2,3] → PASS (no empty string)
    - ["A", "", "C"] with [1,2,3] → OVERFLOW (empty string at center)
    - ["A", "", "C"] with [1,5,9] → PASS (empty string not at center)
  3. Not LaTeX-specific — regular strings with "" also overflow; LaTeX is incidental
  4. Traced the recursion — _setproperty! showed old=nothing, new=Attributes() with is_same=false repeating endlessly, confirming the feedback
  loop never converges                                                                                                                            
  5. Identified NaN source — normalize(Vec2d(0,0)) in the initial bias step of calculate_best_offsets!
  6. Identified convergence failure — val != obs[] in ComputePipeline.jl:341 uses == which fails for NaN                                          
                                                            
 ## Fixes

  Fix 1 (primary) — Makie/src/basic_recipes/annotation.jl: Guard normalize against zero vectors
```julia
  # Before (line 448):
  v = normalize(center - bb_center)
  offsets[i] = 0.1 * algorithm.repel * v

  # After:
  v = center - bb_center
  n = norm(v)
  offsets[i] = n > 0 ? (0.1 * algorithm.repel / n * v) : zero(eltype(offsets))
```

  Fix 2 (safety net) — ComputePipeline/src/ComputePipeline.jl: Use isequal for NaN-safe comparison
```julia
  # Before (line 341):
  elseif val != obs[]

  # After:
  elseif !isequal(val, obs[])
```
  isequal(NaN, NaN) returns true (unlike ==), so even if NaN values somehow enter the pipeline, the convergence check works correctly and breaks
  the feedback loop.

  Why both fixes?

  - Fix 1 prevents the NaN from being generated — the correct solution for the annotation algorithm
  - Fix 2 makes ComputePipeline robust against any NaN-producing computation, preventing infinite Observable notification loops in all compute
  graphs, not just annotation

## Type of change

## Checklist


hz-xiaxz:
An unrelated change: eliminating a boxed function introduced in previous fix

jkrumbiegel:
The Makie part looks good to me, @SimonDanisch or @ffreyer is the `isequal` ok, too?

ffreyer:
I don't see why not. That part is supposed to mimic `ignore_equal_values` from observables, which uses isequal as well.

PR #5574, opened 2026-03-31
feat(axis): Add Pretransformed scale for pre-transformed data

jkrumbiegel:
Sometimes it's quite difficult to make a plot work with `scale = log10` because some plot types don't play well with these transformations, like ABLines. So I thought what might make sense is to make it easier to pretransform the data and then show it in a way that matches `scale = log10`. This is all about the ticks, but I've found it's a bit simpler to set up to work with other tick types if `Pretransformed` is a scale and not a tick object.

## Summary

- Add `Pretransformed(f)` scale type that displays ticks as if scale `f` were active, while keeping the axis linear
- Enables plot types like `ablines!` that require identity transforms, while still showing meaningful (e.g. log-spaced) tick labels
- Tick specifications (manual arrays, tuples, functions) are interpreted in the original data space and mapped to axis positions automatically
- Makes `is_identity_transform` recursive for tuples so `Pretransformed` is correctly recognized

## Example

```julia
using CairoMakie

fig = Figure(size=(700, 500))

ax1 = Axis(fig[1, 1], xscale=Pretransformed(log10), title="Pretransformed(log10)")
xs = range(0, 3, length=100)
lines!(ax1, xs, sin.(xs .* 2) .+ 2)
ablines!(ax1, 1.5, 0.3, color=:red, linewidth=2, label="ablines! works")
axislegend(ax1, position=:lt)

ax2 = Axis(fig[1, 2], xscale=Pretransformed(log10),
    xticks=[1, 5, 10, 50, 100, 500, 1000], title="Manual ticks in data space")
lines!(ax2, xs, xs .^ 2)

ax3 = Axis(fig[2, 1], xscale=Pretransformed(log2), title="Pretransformed(log2)")
xs2 = range(1, 10, length=100)
lines!(ax3, xs2, 2 .* xs2 .+ 1)

ax4 = Axis(fig[2, 2], xscale=Pretransformed(log10),
    xticks=([1, 10, 100, 1000], ["1", "10", "100", "1k"]), title="Tuple ticks")
lines!(ax4, xs, xs .^ 1.5)

fig
```

Pretransformed scale examples showing log10, log2, manual ticks, and tuple ticks

## Checks

- Verified automatic ticks produce correct positions and superscript labels for log10 and log2
- Verified manual tick arrays are interpreted in original data space (`[1, 10, 100, 1000]` → positions `[0, 1, 2, 3]`)
- Verified tuple ticks `([values], [labels])` transform values correctly and preserve labels
- Verified `ablines!` works (no identity transform error)
- Verified custom tick formatters work

### DIFF
--- a/Makie/src/layouting/transformation.jl
+++ b/Makie/src/layouting/transformation.jl
@@ -602,9 +602,8 @@ end
 inverse_transform(F::Tuple) = map(inverse_transform, F)
 inverse_transform(s::ReversibleScale) = s.inverse
 
-function is_identity_transform(t)
-    return t === identity || t isa Tuple && all(x -> x === identity, t)
-end
+is_identity_transform(t) = t === identity
+is_identity_transform(t::Tuple) = all(is_identity_transform, t)
 
 
 ################################################################################
--- a/Makie/src/makielayout/MakieLayout.jl
+++ b/Makie/src/makielayout/MakieLayout.jl
@@ -19,6 +19,7 @@ include("defaultattributes.jl")
 include("lineaxis.jl")
 include("interactions.jl")
 include("blocks/axis.jl")
+include("pretransformed.jl")
 include("blocks/axis3d.jl")
 include("blocks/polaraxis.jl")
 include("blocks/colorbar.jl")
@@ -55,7 +56,7 @@ export Textbox
 export linkxaxes!, linkyaxes!, linkaxes!
 export AxisAspect, DataAspect
 export autolimits!, limits!, reset_limits!, rlims!, thetalims!
-export LinearTicks, WilkinsonTicks, MultiplesTicks, IntervalsBetween, LogTicks, AngularTicks
+export LinearTicks, WilkinsonTicks, MultiplesTicks, IntervalsBetween, LogTicks, AngularTicks, Pretransformed
 export hidexdecorations!, hideydecorations!, hidezdecorations!, hidedecorations!, hidespines!
 export hiderdecorations!, hidethetadecorations!
 export tight_xticklabel_spacing!, tight_yticklabel_spacing!, tight_ticklabel_spacing!, tightlimits!
--- a/Makie/src/makielayout/pretransformed.jl
+++ b/Makie/src/makielayout/pretransformed.jl
@@ -1,0 +1,76 @@
+"""
+    Pretransformed(f)
+
+A scale for axes where data has already been transformed by `f` before plotting.
+
+The axis remains linear (so all plot types like `ablines!` work), but ticks are
+computed and labelled as if the original scale `f` were active.
+
+# Example
+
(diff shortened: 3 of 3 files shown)


ffreyer:
If I understand correctly, this is to allow `plot!(log10.(data))` to still have log-like ticks?
That can also be done with `transformation = :inherit_model`:
```julia
f = Figure()
ax = Axis(f[1, 1], xscale = log10)
lines!(ax, log10.([1, 5, 10]), 1:3, transformation = :inherit_model)
```
image

`ablines!()` currently errors when you do this because of the `is_identity_transform` assertion on the scenes transform func. It would also need some changes to switch from `axis_limits` to `axis_limits_transformed` when the transform_func isn't inherited. For that we probably need to delay the deletion of the `transformation` kwarg when constructing plot (i.e. in compute_plots.jl), so the recipe can still see it. Maybe we should just put it in `plot.kw`?

Either way, `:inherit_model` would be another path to do this. And maybe a better path since your axis can still be transformed then and still have automatically transformed plots?

ffreyer:
Oh, I overcomplicated this. For ablines you really just need to swap out `transform_func(scene)` for `transform_func(p)` and use `:axis_limits_transformed` in the computation.

```julia
fig = Figure(size=(700, 500))
ax1 = Axis(fig[1, 1], xscale=log10, title="inherit_model")
xs = range(0, 3, length=100)
lines!(ax1, xs, sin.(xs .* 2) .+ 2, transformation = :inherit_model)
ablines!(ax1, 1.5, 0.3, transformation = :inherit_model, color=:red, linewidth=2, label="ablines! works")
axislegend(ax1, position=:lt)
fig
```

image

jkrumbiegel:
A linear equation in log space should curve though, no?

ffreyer:
> So I thought what might make sense is to make it easier to pretransform the data and then show it in a way that matches scale = log10.

I'm just trying to do that with the transformation options we have. `:inherit_model` means to ignore the transform_func of the parent axis and only inherit the model matrix. So it leaves applying log10 to the user, but on a per plot basis. 

If you wanted to do a linear fit in log space, you would do
```julia
fig = Figure()
ax = Axis(fig[1, 1], xscale=Pretransformed(log10), yscale = Pretransformed(log10))
scatter!(ax, log10.(xdata), log10.(ydata))
y0, m = linfit(log10.(xdata), log10.(ydata))
ablines!(ax, y0, m)
```
with `Pretransformed` and this with `transformation = :inherit_model`:
```julia
fig = Figure()
ax = Axis(fig[1, 1], xscale=log10, yscale = log10)
scatter!(ax, xdata, ydata)
y0, m = linfit(log10.(xdata), log10.(ydata))
ablines!(ax, y0, m, transformation = :inherit_model)
```

PR #5577, opened 2026-04-02
docs: add AGENTS.md with development guide

jkrumbiegel:
Added some info I find useful for claude etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ metrics/ttfp/benchmark-projects/*
 benchmark_results
 WGLMakie/test/electron.log
 reference_images_site/build/
+.scratch/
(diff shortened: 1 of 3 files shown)


ffreyer:
Does this close #5200?

jkrumbiegel:
Ah I didn't see that one. Yeah, I'm personally not a big fan of duplicating all kinds of info about the repo that could go stale into such agent files. In my experience claude for example just needs a bit of nudging where to look for stuff and then it gets by fine on its own. That way you also waste less context window. Like I did it here, mention some docs pages etc. and only explicitly add things that are otherwise in my head when I work on Makie code.

asinghvi17:
Yeah, I think that PR was made when they needed a lot more guidance :D but this looks good to me.

PR #5581, opened 2026-04-06
Add argument-level `preferred_axis_attributes` hook

jariji:
This adds to the new `args_preferred_axis` feature a single-argument `preferred_axis_attributes(arg)` hook, so that data types passed as plot arguments can provide axis defaults.                                                        
                                           
Motivation: In MakieExtra.jl, `FPlot` is a data type (not a recipe) that carries axis metadata — labels derived from accessor functions, scales, limits. It participates in plotting via `convert_arguments`, not via `@recipe`. With the current PR, the only way for `FPlot` to influence axis attributes is either type piracy (overriding `preferred_axis_attributes(::Plot, ::Type{<:Block})` for all plots) or wrapping every plot call in a custom recipe.

The single-argument form lets MakieExtra define `Makie.preferred_axis_attributes(::FPlot)` — no    
  piracy, no wrapper:
         
```jl                                                                                                  
# MakieExtra side — one method on its own type
function Makie.preferred_axis_attributes(fplt::FPlot)
    # compute labels/scales from FPlot's accessor functions
end

# user code — just works
scatter(fig[1,1], fplt)  # axis labels set automatically
```                                                                                                   
  Implementation: Mirrors the existing `args_preferred_axis(arg)` pattern for axis type selection. In `create_axis_for_plot`, after applying recipe-level hints, iterate `plot.args[]` and call `preferred_axis_attributes(arg)` on each. Recipe-level hints and user axis keywords both take priority. The fallback `preferred_axis_attributes(::Any) = NamedTuple()` is a no-op.

NB: The names might be more intuitive as `axis_hint` and `attribute_hint` or something.
  
---

Note: Claude was used for this PR.

cc @aplavin

### DIFF
--- a/Makie/src/figureplotting.jl
+++ b/Makie/src/figureplotting.jl
@@ -154,6 +154,21 @@ NamedTuple).
 """
 preferred_axis_attributes(p::Plot, ::Type{<:Block}) = NamedTuple()
 
+"""
+    preferred_axis_attributes(arg)
+
+Returns axis attributes for a single plot argument. Data types that are passed
+as arguments to plot functions can define this method to provide default axis
+labels, scales, limits, etc.
+
+Analogous to `args_preferred_axis(arg)` which determines the axis *type* from
+arguments — this determines axis *attributes*.
+
+The return type should be a NamedTuple or other iterable of key-value pairs.
+The default returns an empty NamedTuple.
+"""
+preferred_axis_attributes(::Any) = NamedTuple()
+
 to_dict(dict::Dict) = convert(Dict{Symbol, Any}, dict)
 to_dict(nt::NamedTuple) = Dict{Symbol, Any}(pairs(nt))
 to_dict(attr::Attributes) = attributes(attr)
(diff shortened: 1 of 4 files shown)


aplavin:
Interface does look nice!
And it's not just MakieExtra's FPlot as a motivating example, but also stuff like KeyedArray – there, to get axis labels, one has to overload `plot!` – https://github.com/JuliaAPlavin/AxisKeysExtra.jl/blob/master/ext/MakieExt.jl#L61 – to fill in axis attributes. But that's clearly hacky, a cleaner way would be 👍!

ffreyer:
I think the args version should continue to include the axis type. Otherwise you may need copy code from `args_preferred_axis()` to figure out what axis you're plotting to again, and then branch based on the result as different axis types may use different attribute names or have different amounts (x, y vs x, y, z).

I don't think you should run this per argument. Typically the argument index tells you something about the dimension an argument applies to. For an argument type that affects ticks, for example, that would be important to know, right? I'd just pass `plot.args[]...` directly.

Is there a point in allowing both versions of `preferred_axis_attributes`  to run?  Instead of having one 
```julia
preferred_axis_attributes(AxisType, plot, args...) = NamedTuple()
# or defaulting:
preferred_axis_attributes(AxisType, plot) = preferred_axis_attributes(AxisType, plot.args[]...)
preferred_axis_attributes(AxisType, args...) = NamedTuple()
```

SimonDanisch:
I'm on board with that... I think we initially wanted this to support returning observables to update from the data.

PR #5584, opened 2026-04-08
WGLMakie improvements

SimonDanisch:
The WGLMakie half of the Bonito 5 release. Shader fixes already merged
separately (#5634, #5640).

- Switch testing/display to ElectronCall (old Electron backend still works)
- First-plot ~5.5s → ~0.6s: workload re-caches the serve path Makie
  invalidates, browser-gated display tasks precompiled, tiered workload, lazy
  TeX/RectangleZoom children, drop the `FilePaths` dep
- ComputePipeline: lazy `TypedEdge`, displayed-but-never-updated plots skip
  the specialized resolve (30–75% less first-plot serialization compile)
- `insert!`/`delete!` round-trip on a live connection (plot is there when
  they return), but when nothing's connected (still loading, disconnected,
  static export) they just flag the scene for a full re-serialize on the next
  (re)connect instead of blocking on a connection that may never come; a connection dropping mid-send no longer throws into user code
- Targets the HTTP/2 Bonito stack; CI pins it (Bonito `sd/http2`, HTTP 2.1.1,
  Reseau, MsgPack `perf/write-be`, ElectronCall).

Will need all deps merged and tagged before we can merge this PR


ComputePipeline/src/ComputePipeline.jl
```diff
@@ -660,14 +666,77 @@ function _resolve!(computed::Computed)
...
+    dynamic_resolve!(edge::ComputeEdge)
+
+Runs the edge callback for the very first resolve without building the
+`TypedEdge{InputTuple, OutputTuple, F}` island. Compiled once for all edges;
+the callback itself is invoked via dynamic dispatch (which compiles only the
+callback specialization, exactly as the TypedEdge constructor would). Output
+semantics match `TypedEdge(edge, f, inputs)`.
+"""
+function dynamic_resolve!(edge::ComputeEdge)
+    if edge.callback === compute_identity
```
ffreyer, line 679: Might be good to have `dynamic_resolve!(callback, edge)` so other specialized resolves can be implemented.

Might also be faster to do that. To my understanding `edge.callback` needs to specialize twice in this function, once for `===` and once for `edge.callback(...)`. With `dynamic_resolve!(edge) = dynamic_resolve!(edge.callback, edge)` it would only specialize once.
It might also be good to process `result` in another function (or do the invokelatest thing I added in TypedEdge(...)). As is, the type of `result` should be unknown and every usage specializes it again

PR #5587, opened 2026-04-09
feat!: move FFMPEG_jll to a package extension

jkrumbiegel:
This PR changes `FFMPEG_jll` from a hard dependency to a weak dependency / package extension (`MakieFFMPEGExt`). This avoids pulling in GPL-licensed libraries (e.g. libx264) and lowers the number of deps for users who don't need video recording. A new env variable `MAKIE_FFMPEG` is an escape hatch to support custom ffmpeg binaries because someone may want to use special codecs etc.

The helper `get_ffmpeg_path()` checks the env var first, then whether the extension loaded a method, and otherwise throws a clear error explaining what changed and what to do. Doing `]add FFMPEG_jll` in an existing Makie env would not change any existing dep versions I think, so usually one should even be able to continue in a running session without problems, making this extra step relatively painless for users. It does add some annoyance vs the previous all-batteries-included situation, but it does remove a big hurdle for using Makie in MIT license scenarios.

For context, I did try first to use the recent FFMPEG_NoGPL_jll but the AV1 and MPEG4 codecs it has were not compatible enough with the devices I tried and the openh264 codec I also tried had too bad of a quality especially for moving textures like animated heatmaps. So this is deemed the simplest way to keep the ffmpeg and h.264 capabilities while cleaning up the license situation.


asinghvi17:
I'm not really sold on this, having to do `using FFMPEG_jll` is a very weird thing to ask a new user to do if they just want to record a plot.  

Can we not hard dep on FFMPEG_nogpl_jll, but then have an extension on FFMPEG_jll that causes us to default to that if it's already loaded?  Then at least you can do something out of the box, and we can still recommend `using FFMPEG_jll` before record everywhere in our docs so you'll be more likely to do that

jkrumbiegel:
> having to do using FFMPEG_jll is a very weird thing to ask a new user to do if they just want to record a plot

I'd say it's a _little_ weird, but not terribly. Also there's a clear error if one hasn't done it yet which was important to me. Not much of a chance to get lost there.

> Can we not hard dep on FFMPEG_nogpl_jll, but then have an extension on FFMPEG_jll that causes us to default to that if it's already loaded? Then at least you can do something out of the box

I mean you _could_ but what I found was that mp4 with h264 is still pretty much the most compatible codec out there so it's a great default recommendation. We don't want people going back to gifs etc, and if you worry about new users not understanding `using FFMPEG_jll`, try educating about the intricacies of video codecs... I'd rather take the small win of making default Makie a bit more lightweight in terms of jll dependencies I think, but it's a judgment call.

ffreyer:
My thought on that was that we could have persistent settings, e.g. with Preferences.jl, so it's a one time thing. With that you'd run `Makie.set_ffmpeg_backed!("FFMPEG_jll")` once and Makie just auto-loads it after that (and maybe also auto-installs it). That would be a little more convenient than having to call `using FFMPEG_jll` every time. Though I also think that typing out `using FFMPEG_jll` is not a big deal.

ffreyer (approved):

jkrumbiegel:
Yeah that's why I thought auto-requiring FFMPEG_jll if available would be convenient, then it would only have to be installed. For the ENV override, a preference could also be used instead. I just didn't use it because we weren't using Preferences.jl in Makie, yet.

jkrumbiegel:
Ok I've switched this to an auto-loading mechanism if FFMPEG_jll is in the active env, plus a binary path override either at runtime with a function, or long-term through a preference. I think this covers all bases?

ffreyer:
> Yeah that's why I thought auto-requiring FFMPEG_jll if available would be convenient, then it would only have to be installed. For the ENV override, a preference could also be used instead. I just didn't use it because we weren't using Preferences.jl in Makie, yet.

ComputePipeline uses it to control some debug checks

jkrumbiegel:
> ComputePipeline uses it to control some debug checks

Yep, also seen this in the meantime

PR #5594, opened 2026-04-14
Remove unnecessary trailing comma

fredrikekre:
Before version 1.6.0 Runic left these alone but this was not consistent with how named functions were formatted.


jkrumbiegel:
thanks!

PR #5599, opened 2026-04-16
feat(ci): parallelize reference tests via bucketing

jkrumbiegel:
## Summary

Parallelize reference tests by splitting them into buckets that run on separate CI machines, with a setup job that precompiles once per backend+version and shares the depot via artifact.

**Pipeline per backend+version:**

```
Setup (install + precompile) → bucket 1 ┐
                              → bucket 2 ├→ consolidation
                              → bucket N ┘
```

- **Bucketing** controlled by `REFTEST_BUCKET` / `REFTEST_NBUCKETS` env vars in the `@reference_test` macro. Uses contiguous chunks (tests 1-N in bucket 1, N+1-2N in bucket 2, etc.) so each bucket touches fewer distinct code paths, reducing JIT overhead. Total count is obtained via a regex scan of `@reference_test` occurrences in the included files.
- **Setup job** runs `Pkg.test` with `PRECOMPILE_ONLY=true` env var (test scripts `exit(0)` immediately). This resolves the full test environment and precompiles everything with the same flags the bucket jobs will use.
- **Depot transfer** via artifact: `~/.julia/{compiled,packages,artifacts,scratchspaces}` + `monorepo/Manifest.toml` is packed into a zstd tarball and uploaded. Bucket jobs download, extract, and go straight to `Pkg.test` (no `pkg"dev"`, no `Pkg.update()`). This keeps the environment byte-identical between setup and buckets so pkgimages aren't invalidated.
- **JULIA_CPU_TARGET** set to the portable multi-target string Julia itself uses (`generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)`) so pkgimages work across GitHub's heterogeneous runner fleet.
- **Per-version independence**: each Julia version runs as a separate reusable-workflow call, so Julia 1.10 buckets don't wait for Julia 1.12 setup (and vice versa).
- **Single parameterized workflow** `_backend-tests.yml` replaces the three near-identical backend workflows. Backends differ only in `backend` name, `nbuckets`, `system-packages`, and `test-prefix` (for `xvfb-run` wrapping).

**Bucket counts:** CairoMakie: 2, GLMakie: 2, WGLMakie: 4 (WGLMakie is the slowest).

**Other changes:**

- Made a few WGLMakie reference tests independent of execution order (they previously relied on state left by preceding tests).
- Dropped `coverage=true` from test runs: there was no active codecov integration (`.codecov.yml` had everything disabled), and on Julia 1.10 `coverage=true` forces `use_pkgimages=false`, preventing pkgimage reuse.
- Set `JULIA_PKG_PRECOMPILE_AUTO=0` around the install step so `pkg"dev"` doesn't kick off a stray precompile pass before `Pkg.test` resolves the real test env.
- The compilation benchmark workflow is now triggered by a collaborator posting `/benchmark` on a PR (instead of running on every push). Sample count bumped 20→60.

**Wall-clock impact** (critical path: slowest backend+version job):

| | Unbucketed (master) | Bucketed, first run (no cache) |
|---|---|---|
| Wall-clock | ~65m (WGLMakie Julia 1) | ~27m (GLMakie Julia 1.10) |

Speedup ~2.4x on first run. Subsequent runs on the same PR benefit from the `julia-actions/cache` restore in the setup job.

## Checks

- [x] CI green on all backends × versions × buckets
- [x] Consolidated `ReferenceImages` artifact produced with same structure as before (downstream `reference_images_site.yml` unchanged)
- [x] Local runs unaffected when env vars are unset
- [x] Benchmark workflow tested via `/benchmark` comment


SimonDanisch:
Cool! Do we make sure, that we re-use compilation?

jkrumbiegel:
> Cool! Do we make sure, that we re-use compilation?

It should compile once and then copy those files to the bucket jobs as an artifact. The compilation step is also cached so in the next run it should reuse that.

PR #5600, opened 2026-04-16
Support emojis

### LINKED ISSUE #4901: Emoji in text/labels/titles
- [x] what version of Makie are you running? (`]st -m Makie`); [CairoMakie](https://github.com/MakieOrg/Makie.jl.git) v0.13.2
- [x] What platform + GPU are you on? Ubuntu24; this is CairoMakie
---
```julia
let
f,ax,h = scatter([1];figure=(;fonts = (;regular="Noto color emoji")))
ax.xlabel = "🎈"
end
```

I expected this to work², but it throws the following error:

```julia
Invalid text boundingbox GeometryBasics.HyperRectangle{3, Float64}([NaN, NaN, NaN], [-Inf, -Inf, -Inf])
```

Any simple fix to this? Is this an issue of FreeType?

Image

² it didnt find a font with emoji support first, so I maually specified one

aplavin:
# Description

https://github.com/user-attachments/assets/5eeeb1c8-d78a-4973-9375-d06ea59b8838

Fixes #4901

Support emojis in Cairo and GL, in text labels and scatter markers. Should work with any single-codepoint emoji!
Three commits: cairo -> GL text -> GL scatter.

Can cleanup/restore/add comments/tests if the overall approach is ok.
Also, reference images should be added I guess?..

## Type of change

## Checklist


jkrumbiegel:
Seemed to work well when I tried it, although it didn't find the Noto font and I had to add that manually to the fallbacks. Maybe needs an asset update?

Makie/src/utilities/texture_atlas.jl
```diff
@@ -56,6 +56,132 @@ function Base.show(io::IO, atlas::TextureAtlas)
     return println(io, "  font_render_callback: ", length(atlas.font_render_callback))
 end
 
+# Color texture atlas for color emoji (COLRv1, etc.)
+# Stores RGBA bitmaps instead of signed distance fields.
+struct ColorTextureAtlas
```
SimonDanisch, line 61: Do we really need another texture atlas, or could we just use `color=array_of_images`?

jkrumbiegel: I was also wondering about that, in a way a color image could just be four grayscale images in the sdf atlas

Makie/src/utilities/texture_atlas.jl
```diff
@@ -249,6 +376,8 @@ function find_font_for_char(glyph, font::NativeFont)
     error("Can't represent character $(glyph) with any fallback font nor $(font.family_name)!")
 end
 
+_is_color_font(font::NativeFont) = (font.face_flags & FreeTypeAbstraction.FreeType.FT_FACE_FLAG_COLOR) != 0
```
SimonDanisch: `is_color_font` seems to be a perfectly  normal function to me, not exporting it already marks it internal, so I really don't see the value `_` is adding ;) 

Actually, going forward, I want to be more strict about `_*` function names.
It should get documented somewhere, and we still have lots of `_*` names in the code base, but every day I hate this more :D 
Especially AIs intense usage of `_*` function names has showed me even more, how arbitrary that convention is. It would just be noise, to make every non exported function start with `_` and what other criterion is there, besides "this feels somehow like a helper function", which is arbitrary, in the end every function is "helping" :D 

And marking a function as "dubios unexported helper" at *call site* is also weird, since that code IS actually calling that function to a hopefully good effect - does that mean that code shouldn't call that function? That reviewers should be more careful about that piece of code? That would signal to me, that one should try harder to not write a function that would not need this poisoning prefix.  I think functions should always be written with a certain contract in mind, which it should fulfill, however diffuse that contract is. It seems to me the `_` prefix is something to ease the mind, to say, hey I didn't think as much about this and I want to be honest about it. I think that's a good intention, but it's also pretty normal to me which doesn't need a special marker, instead the process should be to document that in a comment, name the function normally and see how it can be used, and then refactor it whenever it is needed to sharpen its purpose. If one intends to mark the use of unexported functions in the code base from other packages, one should just use `Pkg.unexported_function`. 


@aplavin sorry for the rant,  this is the first time I'm writing this down and doesn't strictly apply to this function, and its fueled by having lots of "DONT USE `_*` FUNCTION NAMES!!!11" in my claude.md, but still getting them sneak back in ;)
 
@jkrumbiegel, what do you think, would you be on board? I think you started doing this?

Makie/src/utilities/texture_atlas.jl
```diff
@@ -56,6 +56,132 @@ function Base.show(io::IO, atlas::TextureAtlas)
...
+        g = unsafe_load(data_ptr, offset + 2)
+        r = unsafe_load(data_ptr, offset + 3)
+        a = unsafe_load(data_ptr, offset + 4)
+        af = a / 255f0
+        if af > 0
+            result[x, y] = RGBAf(r / 255f0, g / 255f0, b / 255f0, af)
+        else
+            result[x, y] = RGBAf(0, 0, 0, 0)
+        end
+    end
```
SimonDanisch: I _think_ this should be equivalent:
```julia
data_ptr = ccall((:cairo_image_surface_get_data, Cairo_jll.libcairo), Ptr{UInt8}, (Ptr{Cvoid},), surface)
stride = Int(ccall((:cairo_image_surface_get_stride, Cairo_jll.libcairo), Cint, (Ptr{Cvoid},), surface))
@assert stride == sz * 4 
bgra = unsafe_wrap(Matrix{BGRA{N0f8}}, Ptr{BGRA{N0f8}}(data_ptr), (sz, sz))
convert(Matrix{RGBAf}, bgra)
```
Maybe also do the clean up in a try finally?

Makie/src/utilities/texture_atlas.jl
jkrumbiegel: I don't use them so much for saying "this is internal" but more for "this is a sub-part of something bigger and needs to be understood as such". Like if I see `is_color_font` I'm thinking about what interface this could belong to and if I see `_is_color_font` I'm thinking, what other functionality does this support. But sure, it's pretty arbitrary. I don't share your hate for the underscore but I'm completely fine with having Makie not use these :) Also, with the `public` functionality which we have these days, you can at least differentiate between unexported public and nonpublic functions, which you couldn't before.

Makie/src/utilities/texture_atlas.jl
SimonDanisch: Yeah, using RGBA{UInt8} as the font atlas could also work nicely.
We could just compress the float32 signed distance field values into the 4 uint8s - there are also papers that compress more information into RGBA{UInt8} for the sdf, for higher font rendering quality. Then it would be pretty smooth.
Maybe use color=array_of_equally_sized_images for now, and once we have time, try the RGBA sdf compression, which can also be simply used for pure images with a flag?
@ffreyer

Makie/src/utilities/texture_atlas.jl
SimonDanisch: Isn't every function written to belong to something bigger? 
I see every program as a Matrjoschka of interfaces...
I don't really see a hard limit to "now this is small enough to not be an interface" - I guess I do try to think of every function as an interface/helper to another function. Of course there's a gradient of API strictness and how much it's thought through.
Also, I think the best marker for belonging is proximity to other functions.
`_` doesn't give me any idea about what else it belongs to, but putting it next to where it's used and grepping it and finding it nowhere else gives me the biggest indicator. 
I guess that's also what AIs usage of `_` names showed me. It tried really hard to stick to the convention, and then the function were so useful it started calling them all over the place. And once we're back at `_` being a score of "usefulness" we're at my original argument...
Anyways, I guess you don't even need convincing and this is just a bit of a braindump of the last weeks of fighting `_` :D

Makie/src/utilities/texture_atlas.jl
ffreyer: The information you pull out of and sdf is just inside vs outside. If you map that to colors it's 0 or 1 red, green and blue, with some blur at the edges. If you try to encode the color values into the minimum (inside) on the sdf, you'll probably have issues with anything that relies on those distance values, e.g. scaling, stroke, glow, anti-aliasing, anisotropic corrections... 
I'd guess you probably need to separate the emoji based on color and have an sdf for each. If there are gradients you'd also need a way to describe that.

Makie/src/utilities/texture_atlas.jl
aplavin: done, I have no preference in this regard

Makie/src/utilities/texture_atlas.jl
aplavin: Not sure if I understand the pros and cons here, what do you guys think is preferable?

aplavin:
@jkrumbiegel yes, an emoji-supporting font is needed, and Noto seems like a good choice. Anything I should do on my side so that it's included into assets?

PR #5604, opened 2026-04-21
Format docstrings with Runic

fredrikekre:
Since you have adopted Runic I am using your code base for testing things and now to solicit input on new features :p

I have implemented docstring formatting in Runic. This is the results. Runic assumes that fenced code blocks are valid Julia but for indented blocks (typically for function signatures in docstrings) the syntax isn't always valid Julia code, in particular when it comes to using `[...]` for optional arguments. I initially thought we should filter out such docstrings but I settled on: "if it parses we format it". This leads to some quirks (see inline comments) but I think it might be better to rewrite those cases to valid Julia code anyway.

### DIFF
--- a/GLMakie/src/screen.jl
+++ b/GLMakie/src/screen.jl
@@ -756,7 +756,7 @@ function destroy!(screen::Screen)
 end
 
 """
-    close(screen::Screen; reuse=true)
+    close(screen::Screen; reuse = true)
 
 Closes screen and empties it.
 Doesn't destroy the screen and instead frees it to be reused again, if `reuse=true`.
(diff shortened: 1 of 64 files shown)


GLMakie/src/screen.jl
```diff
@@ -1149,7 +1149,7 @@ end
 export plot2robjs
 
 """
-    render_asap([callback::Function, ]screen, N)
+    render_asap([callback::Function]screen, N)
```
fredrikekre, line 1152: This look a bit odd now for example. I would (manually) rewrite this as 
```
    render_asap([callback::Function], screen, N)
```
or
```
    render_asap(screen, N)
    render_asap(callback::Function, screen, N)
```

fredrikekre:
Also imlemented formatting of julia code in markdown files.

fredrikekre:
Runic 1.7 is released with these changes so I updated the CI files etc too. Probably need some manual edits (but I will let someone else handle that).

ffreyer:
Does this require a function signature like `foo(x, y)` as the first line? Because that's something we automated away in 0.25/#5484/#5389 for recipes.

fredrikekre:
> Does this require a function signature like foo(x, y) as the first line?

No.

Issue #4407, opened 2024-09-25
Compose different spaces – data, relative, pixel, ...


aplavin:
Would be convenient and generally useful to enable defining coordinate with a mix of different "spaces". Is it feasible to add a type for that or something?
Potential examples include:
```julia
(data(123), relative(0.5))
(data(123) + pixel(3), data(456))
(data(123) + relative(0.01), data(456))
(data(123), data(456)) .+ (pixel(2), pixel(2))
```

ffreyer:
I tried adding this in #1730. Iirc I wasn't happy with how specific the conditions get - you can't mix coordinates until they are in the same space, i.e. no rotations, practically no 3D, and transform functions probably need to be per-element.

PR #5612, opened 2026-05-01
feat(text): Allow mixed per-axis `space` like `(:data, :relative)`

### LINKED ISSUE #3038: Add text to lines and spans
### Feature description

A nice feature would be to be able to add some sort of text or label to a `h/v-line/span` .
The way I see it, there are two possibilities there :
- 1 : one should be able to add some `text` attribute to these recipes and this text would be positionned relatively to the line / span
- 2 : `text` could be modified to allow two different spaces for x and y directions, while currently only one is possible (e.g. x in data space and y in relative space, defaulting to both in the same if the `space` attribute is used.)

The idea would be to allow for figures as below, while supporting zooming (otherwise its just a matter of axis limit settings)

image

jkrumbiegel:
## Summary

`space` can now be a tuple of symbols, with each entry applying to a single axis (axes not covered default to `:data`). The natural use case is annotating `vlines` / `hlines` where one coordinate is meaningful in data space and the other is most easily expressed as a fraction of the viewport (or a pixel offset). Mixed-space plots only contribute to axis autolimits on their `:data` axes, so a `(:data, :relative)` scatter dot at relative y=0.95 doesn't blow up the y limits.

The implementation builds a per-axis combined projection matrix from each axis's individual matrix (taking the diagonal entry only), so it's exact for orthographic, axis-aligned cameras (the standard 2D `Axis`) and ignores rotation/perspective coupling. The same path is used for both `text` and `scatter` (and any other plot whose positions go through `register_camera_matrix!` / `register_positions_projected!`).

## Example

```julia
using Random
Random.seed!(2)
n = 200
y = cumsum(randn(n))
y .-= (minimum(y) + maximum(y)) / 2

f = Figure()
ax = Axis(f[1, 1], xgridvisible = false, ygridvisible = false)
lines!(ax, 1:n, y)

events = [(40, "Event A"), (110, "Event B"), (165, "Event C")]
for (x, name) in events
    vlines!(ax, [x]; color = (:black, 0.4))
    text!(ax, [Point2(x, 1)];
        text = [name],
        align = (:right, :top),
        rotation = π / 2,
        offset = (0, -4),
        space = (:data, :relative),
    )
end

for (yv, label, color) in (
        (0, "zero", :crimson),
        (minimum(y), "min = $(round(minimum(y); digits = 2))", :seagreen),
    )
    hlines!(ax, [yv]; color = color)
    text!(ax, [Point2(4, yv)];
        text = [label], color = color,
        align = (:left, :bottom),
        space = (:pixel, :data),
    )
end

f
```

Closes #3038. Partial progress on #4407 (covers the `(data, relative)` style from the request, doesn't cover arithmetic composition like `data(123) + pixel(3)`).

## Checks

- Added `@testset "mixed-space data_limits"` in `Makie/test/boundingboxes.jl` — asserts which axes contribute to `data_limits` for `scatter` and `text` under `(:data, :relative)`, `(:relative, :data)`, `(:data, :pixel)`, plus homogeneous-tuple corner cases (12 assertions).
- Added `@reference_test "mixed-space annotations"` in `ReferenceTests/src/tests/text.jl` covering `(:data, :relative)`, `(:relative, :data)`, `(:data, :pixel)` with text + scatter on top of vline / hline.
- Added a "Mixed per-axis space" section to `docs/src/reference/plots/text.md` with the example shown above.
- Manually verified that homogeneous-`space` plots, `Axis3`, and `space = :relative`-only plots are unchanged.


Makie/src/camera/camera.jl
```diff
@@ -327,16 +327,46 @@ end
...
+    # values are intentionally not supported yet — markerspace is expected to be a Symbol.
+    rep_space(s) = s isa Tuple ? first(s) : s
+
     # merging Symbols is somewhat expensive so we shouldn't do it repetitively
-    if haskey(plot_graph, :markerspace)
-        map!(plot_graph, [:space, :markerspace], :camera_matrix_names) do space, markerspace
-            return get_projectionview_name(markerspace), get_projection_name(markerspace),
-                get_view_name(markerspace), get_camera_matrix_name(space, markerspace)
+    if has_markerspace
+        if space_is_tuple || markerspace_is_tuple
```
ffreyer: Having branching here means you can't switch between uniform spaces and mixed spaces dynamically. I think if allow switching between e.g. pixel and data space we should also allow switching between e.g. data and (data, relative).

On how to handle this - I reworked this code in #5556 to just use normal `map!()`s whose inputs get replaced when space/markerspace changes. Maybe it would be useful to generalize this and replace the whole edge instead here? (Maybe that's also something I should do?)

I've also been wondering if we should deprecate dynamically switching between spaces entirely. That would remove the need for the awkward camera matrix handling here, and I doubt it's something people actually use dynamically...

ffreyer:
I want to rework this. Specifically pull in and generalize the changes I made in #5556, which get rid of this weird `CameraMatrixCallback` workaround and instead just use map!()'s that get replaced dynamically. 

And then:
- [x] stop `space` from defaulting in `_register_common_camera_matrices!()`. That should only be called when space is dynamic, i.e. exists
- [x] make tuple spaces dynamically changeable (to other tuple spaces and Symbol spaces)
- [ ] probably allow `markerspace` to be a tuple space (I don't think that is really allowed atm?)
- [x] allow `space` to be a tuple space without `markerspace` existing, i.e. allow projection, view, projectionview to be mixed too. (As far as I can tell it only currently handles preprojection right now?)
- [ ] allow input and output spaces in `register_camera_matrix!()` and `register_projected_positions!()` to be tuple spaces

jkrumbiegel:
Okay, sounds good. You can just continue this one if you want. I don't know enough about the internals anyway.

ffreyer:
> I want to rework this. Specifically pull in and generalize the changes I made in #5556, which get rid of this weird `CameraMatrixCallback` workaround and instead just use map!()'s that get replaced dynamically.
> 
> And then:
> 
>   * [x]  stop `space` from defaulting in `_register_common_camera_matrices!()`. That should only be called when space is dynamic, i.e. exists
>   * [x]  make tuple spaces dynamically changeable (to other tuple spaces and Symbol spaces)
>   * [ ]  probably allow `markerspace` to be a tuple space (I don't think that is really allowed atm?)
>   * [x]  allow `space` to be a tuple space without `markerspace` existing, i.e. allow projection, view, projectionview to be mixed too. (As far as I can tell it only currently handles preprojection right now?)
>   * [ ]  allow input and output spaces in `register_camera_matrix!()` and `register_projected_positions!()` to be tuple spaces

I guess I'm done with that for now. Not sure if mixed markerspace would be useful, so I didn't bother adding it.  `register_camera_matrix!()` supports mixed input spaces now. `register_projected_positions!()` kind of does, but not with transformations and float32convert (and model?). That's still a larger problem to solve.

jkrumbiegel:
> register_projected_positions!() kind of does, but not with transformations and float32convert (and model?). That's still a larger problem to solve.

In what scenario would one need that?

ffreyer:
If you have a log-log axis the transformations apply regardless of space: (300px should be past the center)
```julia
xs = 101:1100
data = 100 .+ cumsum(randn(1000))
f,a,p = lines(xs, data, axis = (xscale = log10, yscale = log10))
mini, maxi = extrema(data)
hlines!(a, [mini, maxi])
t = text!(
    a, [300, 300], [mini, maxi], text = ["minimum", "maximum"],
    space = (:pixel, :data), align = (:left, :bottom)
)
f
```
image

I guess `handle_transformation!()` should also allow inheriting transformation per dimension (if possible) and the space compatibility check should check and inherit per dimension?

float32convert also always applies, so other spaces get shifted out of the viewport:
```julia
vs = 1e8 .+ (1:10)
f,a,p = lines(vs, vs)
hlines!(a, [1e8 + 1, 1e8 + 10])
t = text!(
    a, [5, 5], [1e8 + 1, 1e8 + 10], text = ["minimum", "maximum"],
    space = (:pixel, :data), align = (:left, :bottom)
)
f
```
image

This just needs some extra branching in `register_positions_transformed_f32c!()` and an `f32convert` that masks out dimensions. "backend-functionality.jl" also has some special cases for f32converts 

`register_project_positions!()` is more of a completion thing I guess. I would like for it to be able to do anything our transformation-projection pipeline does so we don't go back to plots including their own projection code. Might be useful for `textlabel` with mixed spaces? It might also be something that just works once transformations and float32convert work.

PR #5615, opened 2026-05-01
crop decorations in Axis3 viewmode :free properly

### LINKED ISSUE #5614: rotation of Axis3 with viewmode = :free causes overlap into neighboring block
GLMakie v0.13.10, MacOS M2, fresh env

seems related to https://github.com/MakieOrg/Makie.jl/issues/5292.  though couldn't a crop somehow be used to tell it not to render outside its box?

Image

```
julia> using GLMakie

julia> f = Figure()

julia> ax3 = Axis3(f[1,1])
Axis3 with 0 plots:


julia> sc = scatter!(ax3, rand(Point3f, 10) .* Point3f(3,1,1))
Scatter{Tuple{Vector{Point{3, Float32}}}}

julia> ax3.aspect[] = :data
:data

julia> ax2 = Axis(f[2,1])
Axis with 0 plots:


julia> ax3.viewmode[]
:fitzoom

julia> ax3.viewmode[] = :free
:free
```

bjarthur:
fixes https://github.com/MakieOrg/Makie.jl/issues/5614

Screenshot 2026-05-01 at 2 27 38 PM

### DIFF
--- a/Makie/src/makielayout/blocks/axis3d.jl
+++ b/Makie/src/makielayout/blocks/axis3d.jl
@@ -133,9 +133,9 @@ function initialize_block!(ax::Axis3)
         get_ticks(ax.scene.conversions[3], ticks, identity, format, minimum(lims)[3], maximum(lims)[3])
     end
 
-    add_panel!(blockscene, ax, 1, 2, 3, finallimits, mi3)
-    add_panel!(blockscene, ax, 2, 3, 1, finallimits, mi1)
-    add_panel!(blockscene, ax, 1, 3, 2, finallimits, mi2)
+    add_panel!(scene, ax, 1, 2, 3, finallimits, mi3)
+    add_panel!(scene, ax, 2, 3, 1, finallimits, mi1)
+    add_panel!(scene, ax, 1, 3, 2, finallimits, mi2)
 
     # This exists as a bandaid for WGLMakie. See add_gridlines_and_frames!()
     overlay = Scene(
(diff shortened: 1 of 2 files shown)


bjarthur:
it works for me for this use case.  my knowledge of makie internals is very superficial.  had a lot of help from gemini (tried claude too but it's response was garbage.). please review carefully.

bjarthur:
should note that gemini also advised to

1. Remove the broken clip_decorations logic (approx. lines 20–26). This logic incorrectly attempted to "clip" decorations by modifying the viewport of the shared blockscene, which can disrupt other layout blocks.
2. Clean up the camera update loop (approx. lines 72–96). Remove the section that mirrors the camera to blockscene, as decorations will no longer be plotted there.

bjarthur:
just saw this in the docstring:  "Zooming affects the whole axis rather than just the content. This allows you to zoom in on content without it getting clipped by the 3D bounding box of the Axis3."  does that mean the desired behavior is to rather not have anything cropped?  and so the bug here is not that the decorations were not clipped, but that the points were?  ugh.

Makie/src/makielayout/blocks/axis3d.jl
```diff
@@ -133,9 +133,9 @@ function initialize_block!(ax::Axis3)
         get_ticks(ax.scene.conversions[3], ticks, identity, format, minimum(lims)[3], maximum(lims)[3])
     end
 
-    add_panel!(blockscene, ax, 1, 2, 3, finallimits, mi3)
-    add_panel!(blockscene, ax, 2, 3, 1, finallimits, mi1)
-    add_panel!(blockscene, ax, 1, 3, 2, finallimits, mi2)
+    add_panel!(scene, ax, 1, 2, 3, finallimits, mi3)
```
ffreyer, line 136: This is something we moved away from so that `ax.plots` only has user plots in it

ffreyer:
> just saw this in the docstring: "Zooming affects the whole axis rather than just the content. This allows you to zoom in on content without it getting clipped by the 3D bounding box of the Axis3." does that mean the desired behavior is to rather not have anything cropped? and so the bug here is not that the decorations were not clipped, but that the points were? ugh.

This is trying to explain how the decorations are affected differently. When you zoom with `viewmode != :free` zooming affects user plots and moves ticks around. The frame remains static. If user plots pass the boundary of the frame, they get cut off. With `viewmode = :free` the axis frame, ticks, etc act like user plots when zooming. Everything gets larger or smaller together, and also translates together. The plot never gets cut off by the frame boundary of the axis. Ticks don't move relative to the frame/plot either. The clipping of the viewport, i.e. the 2D area the Axis3 block lives in, is something else.

> Remove the broken clip_decorations logic (approx. lines 20–26). This logic incorrectly attempted to "clip" decorations by modifying the viewport of the shared blockscene, which can disrupt other layout blocks.

I don't think that the blockscene viewport has anything to do with layouting atm. I think the update just loses to window resize (i.e. window resize overwrites it). Imo blockscenes should just use the GridLayout assigned area as their viewport in the first place, so that this code would be redundant.

bjarthur:
turns out adding `ax3.clip_decorations=true` is all i need.  introduced in https://github.com/MakieOrg/Makie.jl/pull/4131.  is there a reason the default is `false`?

PR #5616, opened 2026-05-01
implement :cursor zoommode for Axis3 with :free viewmode

bjarthur:
per the title, this PR implements `Axis3(zoommode = :cursor, viewmode = :free)`.

gemini wrote it.  tried claude and it's response didn't work.  please review carefully.


bjarthur:
with this PR:

https://github.com/user-attachments/assets/bf2a36e1-a962-49ed-91b4-350420d1c6ec

```
julia> using GLMakie
julia> f = Figure()
julia> a = Axis3(f[1,1], zoommode = :cursor, viewmode = :free)
julia> s = scatter!(a, rand(Point3f, 10))
```

not sure how to make a test for a video.

ffreyer:
The current implementation isn't accurate for `perspectiveness > 0`, i.e. with perspective projection which adds some depth/z based scaling that affects the translation you need to keep the point at the same pixel position. I'm not sure how hard that is to fix. Maybe it makes more sense to just have a note that `perspectiveness > 0` adds drift to `zoommode = :cursor`? (DragPan has this issue too btw)

bjarthur:
i'm fine with fixing perspectiveness later in a separate PR.

PR #5618, opened 2026-05-02
Fix links in themes.md

abhro:
# Description

Without parentheses it linked to the sections in the page instead of the docstrings in the API page

## Type of change

## Checklist

  N/A
  N/A
  N/A


asinghvi17:
Hm shouldn't they link to the page here though?  Arguably the API docs should also be here, just condensed...

abhro:
Since the text just before talks about the functions, I figured the docstrings are probably more appropriate. Also, the section headings show up on the right-hand side of the page anyway as part of the table of contents

SimonDanisch:
Thanks!

PR #5622, opened 2026-05-04
fix(GLMakie): chunk large GL_LINE_STRIP_ADJACENCY draws on macOS

### LINKED ISSUE #5304: GLMakie.jl does not plot long vectors correctly
- [ ] what version of Makie are you running? (`]st -m Makie`)

```
  [e9467ef8] GLMakie v0.13.6
  [ee78f7c6] Makie v0.24.6
```

- [ ] can you reproduce the bug with a fresh environment ? (`]activate --temp; add Makie`)

Yes

- [ ] What platform + GPU are you on?

Apple M series

- [ ] Which backend are you using? (GLMakie, CairoMakie, WGLMakie, RPRMakie)

GLMakie.jl

- [ ] Please provide an minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

I would have expected to see only the parabola (which I get for smaller vectors) instead of the artificial line artifact.

```julia
julia> using GLMakie

julia> lines(range(0.0, 1.0, length = 3 * 10^6).^2)
```

Image

### LINKED ISSUE #5491: GLMakie hallucinates lines for large datasets
**Description:**
Using `GLMakie` I have encountered a rendering glitch where `lines!` draws an incorrect horizontal line connecting the end of the data trace back to the start (or near the y-axis origin) when the number of points is large.

In the example below, I plot subsets of a random walk. The artifact consistently appears when the number of points $N$ exceeds approximately 2 million. Smaller subsets render correctly. The data vectors are sorted by time (index) and contain no NaNs or duplicates.

**MWE (Minimal Working Example):**

```
using GLMakie #Changing this to CairoMakie gets rid of the artefact
using Random

# Reproducible data
Random.seed!(42)

# Configuration
max_n = 5_000_000
step_size = 500_000

# Generate Data
step_pool = collect(-1:0.05:1)
# Create a large random walk
data = rand(step_pool, max_n)
datac = cumsum(data)

# Define Subsets (Reverse order to plot largest first)
subset_sizes = reverse(collect(step_size:step_size:max_n))

f = Figure(size = (1000, 600))
ax = Axis(f[1, 1], title = "Artifact Demo: Horizontal Line appears at high N")

# Use a color palette to distinguish lines
colors = cgrad(:viridis, length(subset_sizes), categorical=true)

for (i, n) in enumerate(subset_sizes)
    # Plotting larger datasets first
    # The artifact appears as a straight line connecting the last point to the start
    lines!(ax, 1:n, datac[1:n], 
        label = "N=$(n)", 
        color = colors[i], 
        linewidth = 2
    )
end

axislegend(ax)
display(f)
```

**Observed Behavior:**
For traces where $N \ge 2.5 \times 10^6$, a distinct straight line artifact appears, connecting the end of the trace back towards the beginning. This line does not exist in the data (see attached image).

**Expected Behavior:**
The plot should render the continuous random walk from start to finish without adding any connecting lines between the endpoints.

**Environment:**

Julia Version: Julia 1.12.4+0.aarch64.apple.darwin14

Makie Version: v0.24.8

GLMakie Version: v0.13.8

OS: macOS Sequoia 15.6.1 (24G90)

Image

jkrumbiegel:
## Summary

Fixes #5491 fixes #5304. On macOS, `GLMakie`'s `lines!` plots draw a phantom segment between the first and last point of the line whenever the plot exceeds ~2.4M points. Empirically bisected: a single `GL_LINE_STRIP_ADJACENCY` draw call breaks at `count >= 2_418_341` indices (= 2,418,338 primitives) and is fine below that. The bug only reproduces under Apple's OpenGL → Metal layer (Metal has no native geometry shaders, so they're emulated) — Linux and Windows GL drivers handle the same draw fine.

Workaround: when the bound element buffer would exceed the threshold, the renderer now splits the single `glDrawElements` call into chunks of 2M indices, with consecutive chunks overlapping by 3 indices in the same buffer to give the joints at the chunk boundary their adjacency context. With `LINE_STRIP_ADJACENCY` a primitive `(p, q, r, s)` only draws segment `q→r` and uses `p, s` as joint hints, so the 3-index overlap is adjacency-only — every segment is still drawn exactly once, no overdraw.

- No buffer or shader changes; the index buffer that the rest of GLMakie produces is reused as-is.
- macOS-only branch gated on `Sys.isapple() && mode == GL_LINE_STRIP_ADJACENCY`. Other platforms keep the single-draw fast path.
- ~17% headroom below the bisected threshold to absorb any drift in the Apple driver-side cutoff (it's an opaque internal constant we can't observe).

## Alternatives considered

Two other approaches reproduce-tested as fixes; both were dropped in favour of chunked draws because they cost more for no visible benefit:

- **NaN sentinel padding.** Append one NaN entry to the position buffer (and matching zero entries to `valid_vertex` / `lastlen`), and one extra index pointing at it; the spurious primitive then reads NaN at one end and is culled. Works in practice, but requires coordinated edits in 3–4 places across the compute graph (positions for both the FAST_PATH and the linestyle path, plus `valid_vertex` and `lastlen`), and depends on Apple's spurious primitive always reading the buffer tail. The added bookkeeping and the cross-pipeline coupling is more fragile than the single-function chunked-draw branch.
- **Switch `GL_LINE_STRIP_ADJACENCY` → `GL_LINES_ADJACENCY`** with the index buffer expanded into independent 4-tuples (the geometry shader doesn't need to change because `layout(lines_adjacency)` accepts both primitive types). Reproduce-tested as a clean fix on macOS, but the index buffer grows by ~4× (e.g. ~120 MB extra at N=10M) — a real cost for animated/streaming line plots. The `LINE_STRIP_ADJACENCY` strip already shares indices implicitly, and chunked draws preserve that sharing while avoiding the bug.

Chunked draws win on memory (zero added bytes), code locality (one `if` in `render`), and platform isolation (the branch is only taken on Apple).

## Example

```julia
using GLMakie

N = 2_500_000
y = zeros(Float64, N); y[end] = 1.0
fig = Figure(size = (1000, 600))
ax = Axis(fig[1, 1], title = "L-shape, N=$N")
lines!(ax, 1:N, y, color = :blue, linewidth = 2)
fig
```

**Before**

L-shape with phantom diagonal across the figure (triangle instead of L)

**After**

Clean L-shape, no phantom

The original issue's MWE (10 overlaid `cumsum` random walks up to N=5M):

```julia
using GLMakie, Random
Random.seed!(42)
max_n = 5_000_000
step_size = 500_000
data = rand(-1:0.05:1, max_n)
datac = cumsum(data)
subset_sizes = reverse(collect(step_size:step_size:max_n))
fig = Figure(size = (1000, 600))
ax = Axis(fig[1, 1], title = "Issue #5491 MWE (random walks)")
colors = cgrad(:viridis, length(subset_sizes), categorical = true)
for (i, n) in enumerate(subset_sizes)
    lines!(ax, 1:n, datac[1:n], label = "N=$(n)", color = colors[i], linewidth = 2)
end
axislegend(ax)
fig
```

**Before**

Original issue MWE: spurious diagonal lines connecting first and last points of each long subset

**After**

Original issue MWE: clean overlaid random walks with no diagonal artifacts

## Checks

- L-shape (`zeros(N); y[end]=1`) at N = 2,500,000 / 5,000,000 / 10,000,000: clean.
- Original issue's `cumsum` MWE: clean.
- Existing line behaviors unchanged: closed loops (circle, square), NaN-split lines, dashed/joints lines.
- Chunk-boundary cases verified: N just below chunk size (1,999,999), just over (2,000,001), exactly 2 chunks (4,000,000) — no visible discontinuity at chunk seams.
- `GLMakie/test/isolated_tests.jl` (line indices) passes 20/20.
- `julia tooling/formatter/format.jl` clean.
- Bisection (with workaround temporarily disabled) confirms the bug starts at exactly count = 2,418,341 indices and the workaround at chunk size 2,000,000 prevents recurrence at every tested N up to 10M.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed `lines` rendering a phantom segment between the first and last point on macOS for line plots above ~2.4M points, by splitting `GL_LINE_STRIP_ADJACENCY` draw calls into chunks of 2M indices on macOS. Each chunk stays below the threshold at which Apple's OpenGL geometry-shader emulation emits the spurious primitive; consecutive chunks overlap by 3 indices to preserve the strip's adjacency context. No buffers change and non-Apple platforms are unaffected. [#5622](https://github.com/MakieOrg/Makie.jl/pull/5622)
 - Fixed memory-aliased arrays not propagating in ComputePipeline [#5605](https://github.com/MakieOrg/Makie.jl/pull/5605)
 
 ## [0.24.10] - 2026-04-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fixed `lines` rendering a phantom segment between the first and last point on macOS for line plots above ~2.4M points, by splitting `GL_LINE_STRIP_ADJACENCY` draw calls into chunks of 2M indices on macOS. Each chunk stays below the threshold at which Apple's OpenGL geometry-shader emulation emits the spurious primitive; consecutive chunks overlap by 3 indices to preserve the strip's adjacency context. No buffers change and non-Apple platforms are unaffected. [#5622](https://github.com/MakieOrg/Makie.jl/pull/5622)
+- Fixed `lines` rendering a phantom segment between the first and last point on macOS for line plots above ~2.4M points [#5622](https://github.com/MakieOrg/Makie.jl/pull/5622)
 - Fixed `Legend` not reflecting `linecap` and `joinstyle` set on `lines!`/`linesegments!` plots [#5621](https://github.com/MakieOrg/Makie.jl/pull/5621)
 - Fixed memory-aliased arrays not propagating in ComputePipeline [#5605](https://github.com/MakieOrg/Makie.jl/pull/5605)
 
--- a/GLMakie/src/GLAbstraction/GLRender.jl
+++ b/GLMakie/src/GLAbstraction/GLRender.jl
@@ -145,6 +145,10 @@ function render(vao::GLVertexArray{T}, mode::GLenum = GL_TRIANGLES) where {T <: 
     return nothing
 end
 
+# Empirically bisected: GL_LINE_STRIP_ADJACENCY breaks at count >= 2_418_341.
+# Picked with ~17% headroom since the exact driver-side cutoff may drift.
+const MAC_LINE_STRIP_CHUNK = 2_000_000
+
 # using indexbuffer (faces)
 function render(vao::GLVertexArray{GLBuffer{T}}, mode::GLenum = GL_TRIANGLES) where {T <: Union{Integer, AbstractFace}}
     # Note: not discarding draw calls with 0 indices may cause segfaults even if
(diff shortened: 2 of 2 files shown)


ffreyer:
Should also fix #5304

ffreyer (approved):
The offset/overlap looks good to me.

Just as a note: 
On master, we could also move this code to a custom postrender function and have it more directly relate to line plots. But on 0.25 we can't, because I moved `render` out of the postrender function.

PR #5623, opened 2026-05-06
feat!: Support matrix data with units

icweaver:
# Description

Follow-up to <https://discourse.julialang.org/t/juliahub-raises-65m-series-b-and-launches-dyad-3-0/136941/20>

Allows for things like this to work now:

```julia
using CairoMakie, DynamicQuantities

fig, ax, p = heatmap(rand(3, 4)u"kg")

Colorbar(fig[1, 2], p; label = "Mass")
Colorbar(fig[1, 3], p; label = "Mass", dim_convert = Makie.DQConversion(us"g"))
Colorbar(fig[1, 4], p; label = "Mass", label_suffix = "unit: ({})")

fig
```

display

Changes listed here: https://github.com/icweaver/Makie.jl/blob/units-matrix/CHANGELOG.md#unreleased

Would something like this make sense to add? At any rate, thanks all for the exciting developments that are going into v0.25 and beyond!

## Type of change

## Checklist

- [ ] Implement `format` for DQ


Makie/src/dim-converts/argument_dims.jl
```diff
@@ -88,9 +88,9 @@ function _argument_dims(args; direction::Symbol = :y, orientation::Symbol = :ver
 end
 
 
-argument_dims(::ImageLike, x, y, z) = (1, 2)
-argument_dims(::VertexGrid, x, y, z) = (1, 2) # contour, contourf
-argument_dims(::CellGrid, x, y, z) = (1, 2)
+argument_dims(::ImageLike, x, y, z) = (1, 2, 3)
+argument_dims(::VertexGrid, x, y, z) = (1, 2, 3) # contour, contourf
+argument_dims(::CellGrid, x, y, z) = (1, 2, 3)
```
ffreyer, line 98: With this, the heatmap/image/etc. data will be forced¹ to synchronize with other plots that map to dim 3. E.g. if you add a z component to scatter, that will have to synchronize. (You can use z values for ordering/layering in Axis.) Or if you plot multiple heatmaps. 
Is that something we want? Or don't want? Or don't care about?

¹ Breaking also added `force_dimconverts` as a plot kwarg which can be set to `false` to allow unitless data to pass through. So it's not entirely forced, but still what happens by default

icweaver: Oh cool, `force_dimconverts` is a really nice escape hatch:

```julia
fig = Figure()
ax = Axis3(fig[1, 1])

scatter!(ax, 1:3, 1:3, (4:6)u"cm")
heatmap!(ax, 1:3, 1:3, rand(3, 3); force_dimconverts = false)

fig
```

Would adding a unit-aware method for `transform!` be feasible too? Then, being able to do something like this seems intuitive to me:

```julia
 fig = Figure()
 ax = Axis3(fig[1, 1])

 scatter!(ax, 1:3, 1:3, (4:6)u"cm")
 heatmap!(ax, 1:3, 1:3, rand(3, 3); transformation = (:xy, 5u"cm"),  force_dimconverts = false)

 fig
```

I'm still exploring the internals, but it seems like more invasive changes would need to be made if we were to go with implementing a totally separate color dim convert instead of just re-using the machinery that's already in place, so I feel inclined to just stick with the above design decision and document the `force_dimconverts` option for folks that need it.

I rarely if ever work in 3D, so apologies if I am missing other common use cases where this might be a problem.

ffreyer: You don't need to be in 3D to use z values. You can use them in 2D too, for ordering. Or for colors in surface plots.

Alternatively to treating and synchronizing heatmap values as z coordinates we could also:
- just strip units and not synchronize at all. Maybe by marking a dimension as -1?
- add a fourth "color" channel to dim converts that groups heatmap values with things like (numeric) scatter colors. 

> Would adding a unit-aware method for `transform!` be feasible too? 

I'd rather not since it would require transformations to update with changes to dim converts. Currently they just resolve once at init time. I also think that units are mostly used in 2D where plots generally rely on `direction/orientation` instead of transformations to swap axes. Might be useful (and easy) to allow users to overwrite/set `arg_dims` themselves though, so you can have a `:yz` transformed heatmap with `(2, 3)` for synchronization

icweaver:
Looking forward to this PR being replaced by https://github.com/MakieOrg/Makie.jl/pull/5673, thanks Frederic!

PR #5624, opened 2026-05-10
fix: make axislegend's margin themeable

### LINKED ISSUE #2415: axislegend ignores theme's margin
`axislegend` currently ignores the legend margin set by the theme (See below) in favor of its default kwarg value.

I believe this is set here:
https://github.com/MakieOrg/Makie.jl/blob/f2bc3e93650883f82a91a93b347dc706304541e4/src/makielayout/blocks/legend.jl#L598-L604

Would it be possible to update this to use the theme's margin, if set, and fall back to the kwarg or current default otherwise?

### MWE

```julia
using CairoMakie

function axislegend_mwe()
    my_theme = Theme(
        Legend = (;
            margin = (1, 1, 1, 1)
        )
    )
    with_theme(my_theme) do
        f = Figure()

        # This won't match the themed margin
        ax1 = Axis(f[1, 1]; title="Margin from theme")
        scatter!(ax1, rand(10), rand(10); label="foo")
        axislegend(ax1)

        # This one will match the kwarg margin
        ax1 = Axis(f[1, 2]; title="Margin from kwarg")
        scatter!(ax1, rand(10), rand(10); label="foo")
        axislegend(ax1, margin=(1, 1, 1, 1))

        f
    end
end
```
plot_56

awadell1:
# Description

Fixes #2415

Update axislegend to allow for a themeable margin, falling back to the existing default. Kwarg still takes priority. 

I'm suspect that this is not the ideal way to get the theme's Legend.margin entry, but I couldn't find a clear alternative.

## Type of change

## Checklist


Makie/src/makielayout/blocks/legend.jl
```diff
@@ -1128,7 +1128,7 @@ function axislegend(ax, args...; position = :rt, kwargs...)
     return Legend(
         ax.parent, args...;
         bbox = ax.scene.viewport,
-        margin = (6, 6, 6, 6),
+        margin = something(theme(:Legend), (; margin = (6, 6, 6, 6))).margin,
```
asinghvi17: It should probably also try to get it from `kwargs`

ffreyer (approved):

jkrumbiegel:
Sorry only looking at this now that it's merged. I'm not sure if this is ideal because margin in a theme would apply to all Legends, not just axislegend specifically, and for the others it's not useful.

So I was thinking that maybe the axislegend needs to become a bit more of an actual thing, for example by introducing a second margin attribute that is only relevant when the legend is set to "in axis mode". That would make both themeable separately. This could also mean attaching the axis to the legend so that deleting the axis can delete the legend as well. Currently both are completely separate and there's nothing special about the axislegend Legend.

ffreyer:
Priority wise this seems correct to me. It's preferring kwargs, then the overwrite `theme[:Legend]` and then the replacement default `(6, 6, 6, 6)`. This also runs as part of `axislegend` so it doesn't affect other `Legend` constructors. I guess it would be nicer to have Block defaults for `axislegend` directly, but it doesn't seem necessary to me.

A problem I'm seeing now that I look at it again is that this will error if `theme[:Legend]` is defined with `:margin`.

jkrumbiegel:
Doesn't margin add unnecessary space for a non axislegend Legend? That's what I mean would happen if you themed this globally

ffreyer:
If you do `set_theme!(Legend = (margin = ..., ))` it will set the margin for both now, sure. I'd say that's not optimal but better than not being able to theme `axislegend` at all.

PR #5626, opened 2026-05-12
refactor(ticks): render scientific tick labels as RichText, drop Showoff

jkrumbiegel:
## Summary

Default linear tick labels used `Showoff.showoff(..., :auto)` to produce scientific notation like `1.0×10⁻⁵` with unicode superscript glyphs. Makie's default font (TeXGyreHerosMakie) does not fully cover the unicode superscript block, so some exponents rendered in Dejavu instead. Switch to a `RichText` `superscript` span, the same way `LogTicks` already builds its labels.

Inline the small subset of Showoff.jl Makie used, drop the `Showoff` dependency entirely.

## Example

```julia
using CairoMakie
f = Figure(size = (600, 300))
Axis(f[1, 1]; limits = ((-1e7, 1e7), nothing))
Axis(f[2, 1]; limits = ((0, 1.5e-7), nothing))
save("ticks.png", f)
```

**Before**

default ticks with Showoff's unicode superscripts (top label clipped at 1.0×10, bottom shows 5.00×10⁻⁸ etc.)

**After**

ticks with RichText superscripts (top reads 1×10⁷, bottom reads 1.5×10⁻⁷ with proper exponent)

## Checks

- `Makie/test/isolated/showoff.jl` ports Showoff's own unit tests to the new public API (precision heuristic, special values, auto-style threshold, `.0`-strip rule, mixed-precision padding, `Formatters` backward compat).
- Existing `Tick functions` / `MultiplesTicks strip_zero` testsets still pass.
- Manual visual check against `LogTicks`, `MultiplesTicks(pi, "π")`, `pseudolog10` automatic ticks, and plain linear ranges to confirm no regression elsewhere.


ffreyer:
Does updating the formatter in the old axis3d! still work with this? (And did it before?)

jkrumbiegel:
I don't know, I just tried to leave it looking as it was but didn't check that case

PR #5628, opened 2026-05-13
feat: Theme-driven color scheme for Block widgets

jkrumbiegel:
The point of this PR is to make it easier to theme all the UI elements with the colors one likes, this system needs three colors only and the rest are derived automatically. Examples at the bottom

------

## Summary

- Every interactive Block (`Button`, `Checkbox`, `Toggle`, `Slider`, `IntervalSlider`, `Menu`, `Textbox`) now reads its default fill, border, and text colors from a nested `colors` Attributes block on the theme. State conventions are uniform across blocks: idle → `surface`, hover → `accent_subtle`, active/checked/focused → `accent`.
- `Makie.derive_colors(; accent, gray, background)` turns three inputs into a full nine-role scheme (`background`, `surface`, `surface_subtle`, `border`, `text`, `text_muted`, `text_on_accent`, `accent`, `accent_subtle`) via Oklab-space mixing, so equal weights give perceptually-even steps under tinted gray poles and non-default backgrounds.
- `set_theme!(colors = Makie.derive_colors(accent = :crimson))` recolors every widget consistently; per-role overrides (`set_theme!(colors = (text = :navy,))`) work too.
- The `@inherit` macro inside `@Block` now accepts a tuple of symbols (`@inherit((:colors, :accent), default)` or just `@inherit((:colors, :accent))`) to walk nested theme dicts.
- The internal `COLOR_ACCENT` / `COLOR_ACCENT_DIMMED` `Ref`s are removed; the textbox cursor blink now sources its colour from the textbox's own `bordercolor_focused`.
- Fixed a pre-existing menu bug where the dropdown button stayed in its hover state once the cursor entered the open dropdown.

Visible shifts in light-mode neutrals from the historical literals: `surface` 0.94 → 0.92, `border` 0.80 → 0.74, `text_muted` unchanged. Other roles match within rounding.

## Example

The same showcase figure is used for all three screenshots:

```julia
function widget_showcase()
    fig = Figure(figure_padding = 70)
    Label(fig[0, 1:3], "Block showcase", fontsize = 16, tellwidth = false)
    btn = Button(fig[1, 1], label = "Idle")
    Button(fig[1, 2], label = "Hover preview", buttoncolor = btn.buttoncolor_hover)
    Button(fig[1, 3], label = "Active preview",
        buttoncolor = btn.buttoncolor_active, labelcolor = btn.labelcolor_active)
    Checkbox(fig[2, 1], checked = false); Checkbox(fig[2, 2], checked = true)
    Label(fig[2, 3], "Checkboxes", tellwidth = false)
    Toggle(fig[3, 1], active = false); Toggle(fig[3, 2], active = true)
    Label(fig[3, 3], "Toggles", tellwidth = false)
    Slider(fig[4, 1:3], range = 0:0.1:10, startvalue = 5)
    IntervalSlider(fig[5, 1:3], range = 0:0.1:10, startvalues = (2, 7))
    Menu(fig[6, 1], options = ["alpha", "beta", "gamma"], width = 100)
    Textbox(fig[6, 2], placeholder = "type here")
    Textbox(fig[6, 3], stored_string = "with text")
    resize_to_layout!(fig)
    fig
end
```

### Default scheme

Default scheme

### Warm light

```julia
bg   = RGBf(0.96, 0.92, 0.84)
warm = RGBf(0.20, 0.13, 0.08)
acc  = RGBf(0.70, 0.36, 0.18)

with_theme(
    backgroundcolor = bg,
    textcolor = warm,
    colors = Makie.derive_colors(accent = acc, gray = warm, background = bg),
) do
    widget_showcase()
end
```

Warm light theme

### Dark green

```julia
bg    = RGBf(0.08, 0.11, 0.09)
light = RGBf(0.86, 0.96, 0.87)
acc   = RGBf(0.40, 0.85, 0.55)

with_theme(
    backgroundcolor = bg,
    textcolor = light,
    colors = Makie.derive_colors(accent = acc, gray = light, background = bg),
) do
    widget_showcase()
end
```

Dark green theme

## Checks

- [x] `Makie/test/isolated/derive_colors.jl` covers default/dark/tinted-gray/contrast cases (37 tests).
- [x] Pixel diff of the widget showcase vs. base branch on default theme: 0 differing pixels at PixelMatch's default perceptual threshold.
- [x] New docs page at `docs/src/explanations/theming/block_colors.md`.
- [x] Formatter clean.
- [ ] Reference image suite to be checked in CI.


ffreyer:
Probably best to not merge this until I'm done with #5620 since that's revamps all of the attribute inheritance stuff

ffreyer:
All the changes in blocks.jl should be unnecessary with #5620. `@inherit((:outer, :inner))` is valid syntax in #5620. The rest of the changes should be independent

jkrumbiegel:
Yeah that's why I put it on top of your branch in the first place, because I was vaguely remembering you had done this kind of thing here (but it's not merged yet as I see now). Will change this one once it goes in

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Added a unified color scheme for Block widgets. `MAKIE_DEFAULT_THEME` now contains a nested `colors` Attributes block with 9 role tokens (`background`, `surface`, `surface_subtle`, `border`, `text`, `text_muted`, `text_on_accent`, `accent`, `accent_subtle`) that every interactive Block (Button, Checkbox, Toggle, Slider, IntervalSlider, Menu, Textbox) reads its defaults from. A new helper `Makie.derive_colors(; accent, gray, background)` produces a full scheme from three inputs in the style of Radix UI's custom palette tool, so `set_theme!(colors = Makie.derive_colors(accent = :crimson))` recolors all widgets consistently. Mixing happens in Oklab so equal weights produce perceptually-even steps, which means default surfaces shift slightly from the historical literals: surface ≈ 0.92 (was 0.94), border ≈ 0.74 (was 0.80), muted text ≈ 0.50 (unchanged). The `@inherit` macro inside `@Block` now accepts a tuple of symbols (`@inherit((:colors, :accent), default)`) to walk nested theme dicts. The internal `COLOR_ACCENT` and `COLOR_ACCENT_DIMMED` `Ref`s have been removed. [#5628](https://github.com/MakieOrg/Makie.jl/pull/5628)
```
ffreyer: Also these AI generated changelog entries seem way too long to me. We used to just have something short like "Updated color theming of blocks" there, not a full paragraph.

jkrumbiegel: Yeah I didn't go over this fully, yet, mostly wanted to check CI and docs. I also mislike overly verbose entries like that a lot and delete stuff where I can :)

bjarthur:
with this PR can the theme be changed dynamically?  meaning, after the layout has been rendered, can i switch from dark to light mode and have everything update without having to recreate the scene?

jkrumbiegel:
No, the mechanism is still the same. At the time we decided against this because it creates so many more observable connections that just accumulate. Not sure if things could be different now with the compute graph

Makie/src/theming.jl
```diff
@@ -37,6 +99,7 @@ const MAKIE_DEFAULT_THEME = Attributes(
         italic = "TeX Gyre Heros Makie Italic",
         bold_italic = "TeX Gyre Heros Makie Bold Italic",
     ),
+    colors = Attributes(DEFAULT_COLORS),
```
ffreyer, line 102: Would probably be good to be a bit more descriptive than just `colors` here, since colors are everywhere. Maybe just `blockcolors`?

jkrumbiegel: It would actually be a good umbrella name if textcolor and the others would be part of this but I guess they have to stay top level

PR #5629, opened 2026-05-14
Validate Figure and Legend keyword arguments

### LINKED ISSUE #4217: Make blocks error on wrong keyword arguments
Below calls with wrong kwargs don't error, except for Axis, which throws an unsupported keyword error - I'd expect this in all cases. Or even better get the nice suggestions already introduced for plots.

I'll paypal 50€ to the person that fixes this in general :)

```julia
fig = Figure(;noproblem=false)
ax = Axis(fig[1,1])
scatter!(ax, [1], [1]; label="foo")
Legend(fig[2,1], ax; thisshoulderror="but it does not")
ax2 = Axis(fig[3,1]; butthiserrors="and I don't know why it errors here but not for Figure or Legend")
```

mshipley:
## Summary
- reject unknown `Figure` keyword arguments before forwarding to `Scene` theme kwargs
- validate specialized `Legend` constructors before applying block defaults
- add regression coverage for `Figure` and both specialized `Legend` paths

Fixes #4217.

## Tests
- `JULIA_PKG_PRECOMPILE_AUTO=0 /tmp/codex-julia/julia-1.10.10/bin/julia --project=Makie/test -e ...` targeted smoke assertions for invalid `Figure` kwargs, both invalid specialized `Legend` kwargs, and valid `Figure` theme kwargs.

If accepted for the bounty mentioned in #4217, PayPal receive link: https://paypal.me/mikeshipley

### DIFF
--- a/Makie/src/figures.jl
+++ b/Makie/src/figures.jl
@@ -120,6 +120,7 @@ function Figure(; kwargs...)
 
     kwargs_dict = Dict(kwargs)
     padding = pop!(kwargs_dict, :figure_padding, theme(:figure_padding))
+    _check_figure_kwargs(kwargs_dict)
     scene = Scene(; camera = campixel!, clear = true, kwargs_dict...)
     padding = convert(Observable{Any}, padding)
     alignmode = lift(Outside ∘ to_rectsides, padding)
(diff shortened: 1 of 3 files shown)


ffreyer:
Legend is also fixed in #5620. It may take a while until that breaking gets released though, so I wouldn't mind fixing this on master too.

The figure case is a bit awkward. Everything that is not a named kwarg of `Scene` gets merged into `scene.theme`. Assuming we want to continue to allow theming through `Figure(; theme_kw...)` we should accept:
- the result of every `Makie.plotsym` for overwrite themes `Scatter = (color = :red, )`
- each `nameof(block_type)` where `block_type` is a leaf subtype of `Makie.Block` (this may need multiple calls to `subtypes` because things like `Axis <: AbstractAxis <: Block` are allowed) for Block overwrite themes `Axis = (xgridvisible = false, )` 
- every inheritance sources defined in the default attributes of plots, e.g. `colormap = @inherit colormap` within `@recipe`. This would require searching every `Makie.documented_attributes` method for `Inherit`. On master this won't capture sources from old style recipes (`@recipe ... do scene ... end`) unless you also update attribute parsing. On #5620 it will because that pr reworked and merged the default attribute infrastructure
- every inheritance source defined for Blocks. More or less has the same problem as old recipes
- everything that is in `Makie.current_default_theme()` 
- the name of every backend. The current ones are all in the default theme, but the default theme can change to not include them, and technically someone could also define their own backend which is not in the default theme we manage.

PR #5630, opened 2026-05-14
Fix convert_arguments for StatsBase.Histogram with non-uniform binning

### LINKED ISSUE #2422: Histogram with non-uniform binning / variable binning fails
```julia
julia>  h = fit(Histogram, rand(1000), [0.0, 0.1, 0.2, 0.5, 1.0])
Histogram{Int64, 1, Tuple{Vector{Float64}}}
edges:
  [0.0, 0.1, 0.2, 0.5, 1.0]
weights: [90, 88, 311, 511]
closed: left
isdensity: false

julia>  p = stairs(h)
ERROR: MethodError: no method matching step(::Vector{Float64})
Closest candidates are:
  step(::StepRange) at range.jl:678
  step(::AbstractUnitRange{T}) where T at range.jl:679
  step(::LinRange) at range.jl:682

julia>  p = hist(h)
ERROR: MethodError: no method matching step(::Vector{Float64})
Closest candidates are:
  step(::StepRange) at range.jl:678
  step(::AbstractUnitRange{T}) where T at range.jl:679
  step(::LinRange) at range.jl:682

```

qwjyh:
# Description

Fixes #2422 

Replaced `step` with `diff` as `StatsBase.Histogram.edges`is not always a `AbstractRange`.

## Type of change

## Checklist

image

Note that this doesn't fix plotting as `Stairs` which was used in the original issue #2422 , as it requires very different arguments than `BarPlot`.
Edit: I added another method of `convert_arguments` in #5631 .


ffreyer:
I noticed that the 2D Histogram case was not correctly spaced unless edges edges were spaced uniformly so I fixed that. I also changed the 3D representation to Voxels (from Volume) because histograms are a discrete thing.

PR #5632, opened 2026-05-14
feat(text): Add text_handler extension hook

jkrumbiegel:
Refactors the `text` recipe so external text renderers can plug in via a `text_handler` attribute, and so GPU backends can rasterize opaque markers at upload time only when needed. Companion MakieTeX PR (`LaTeX` and `Typst` handlers): https://github.com/MakieOrg/MakieTeX.jl/pull/73.

Real-LaTeX axis labels, title, and Roman-numeral xticks

## What's new

1. **`text_handler` plot attribute** — themable, scoped, per-plot. `nothing` keeps the built-in MathTeXEngine path; setting a handler routes matching content through it.
   ```julia
   set_theme!(text_handler = MakieTeX.LaTeX())
   with_theme(text_handler = MakieTeX.LaTeX(preamble = "…")) do … end
   text!(scene, L"…"; text_handler = MakieTeX.LaTeX(…))
   ```
2. **Live handler swap.** The text recipe owns a single `plotlist!` child fed by a reactive `text_specs::Vector{PlotSpec}` channel; `plot.text_handler = …` (or any text content change) re-runs the handler and PlotList diffs the children.

## Extension hooks

- `compile_text(handler, src, color, fontsize, lineheight)` — returns the payload `place_text!` will consume, or `nothing` to fall through to FreeType.
- `place_text!(handler, outputs, i, N, compiled, …)` — pushes `PlotSpec`s into `outputs.text_specs`, with parallel `text_spec_block_indices` / `text_spec_bboxes`.
- `is_text_input(::Type) -> Bool` — opt-in trait for stringlike inputs that don't conform to `AbstractString`.
- `convert_text_string!(outputs, str, …)` — alternative entry that bypasses the `(compile_text, place_text!)` protocol.
- `shift_text_spec(::Val{type}, spec, offset)` — for `PlotSpec` types whose positional data isn't `args[1]`.
- `rasterize_marker_for_gpu(marker, scale) -> Union{Nothing, Matrix, Vector{Matrix}}` — hook in `compute_marker_attributes` so GPU backends can rasterize opaque markers at upload time. CairoMakie keeps its native vector dispatch.

## Checks

- [ ] regression: existing `text!` with `String` / `LaTeXString` / `RichText`
- [ ] end-to-end with the MakieTeX PR on CairoMakie (vector) and GLMakie (raster)
- [ ] live handler swap on already-rendered text plots
- [ ] reference image tests

### DIFF
--- a/Makie/src/basic_recipes/text.jl
+++ b/Makie/src/basic_recipes/text.jl
@@ -372,6 +372,7 @@ end
 function compute_glyph_collections!(attr::ComputeGraph)
     inputs = [
         :input_text,
+        :latex_handler,
         :fontsize,
         :selected_font,
         :align,
@@ -395,7 +396,7 @@ function compute_glyph_collections!(attr::ComputeGraph)
         :linesegments, :linewidths, :linecolors, :lineindices,
         :text_primitives, :text_primitive_block_indices,
     ]
-    return register_computation!(attr, inputs, outputs) do (input_texts, _inputs...), changed, cached
+    return register_computation!(attr, inputs, outputs) do (input_texts, latex_handler, _inputs...), changed, cached
         _outputs = (
             glyphcollections = GlyphCollection[],
             glyphindices = UInt64[],
@@ -419,7 +420,13 @@ function compute_glyph_collections!(attr::ComputeGraph)
 
         N = length(input_texts)
         for (block_index, str) in enumerate(input_texts)
-            convert_text_string!(_outputs, str, block_index, N, _inputs...)
+            # `LaTeXString` is themable: if `latex_handler` is set, route the
+            # block through it instead of the built-in MathTeXEngine path.
+            if str isa LaTeXString && latex_handler !== nothing
+                latex_handler(_outputs, str, block_index, N, _inputs...)
+            else
+                convert_text_string!(_outputs, str, block_index, N, _inputs...)
+            end
         end
 
         return values(_outputs)
--- a/Makie/src/basic_recipes/text.jl
+++ b/Makie/src/basic_recipes/text.jl
@@ -4,6 +4,61 @@ function check_textsize_deprecation(@nospecialize(dictlike))
     end
 end
 
+# Text primitives: pluggable graphical elements that a string layout can emit
+# alongside (or instead of) glyphs. Built-in: glyphs and MathTeX line segments
+# have dedicated channels for performance; arbitrary primitives go through
+# the `text_primitives` channel below and are dispatched to child plots via
+# `register_text_primitive_plot!`.
+
+"""
+    AbstractTextPrimitive
+
+Supertype for graphical elements that a `text` plot can render at the
+markerspace position of a text block. Built-in subtypes include
+[`ImageTextPrimitive`](@ref); downstream packages (e.g. MakieTeX) define
+their own to plug arbitrary content into `text`.
+
+A subtype `P` must implement:
+
+* `text_primitive_bbox(p::P)::Rect3d` — markerspace bounding box relative to
+  the text block origin (used by the text plot's bounding-box pipeline).
+* `register_text_primitive_plot!(plot::Plot, ::Type{P}, items::Vector{P}, block_indices::Vector{Int})`
+  — register a child plot that renders `items`, observing the text plot's
+  current markerspace block positions.
+
+Items are pushed by `convert_text_string!` overloads (one per input text type)
+into the `outputs.text_primitives` channel.
+"""
+abstract type AbstractTextPrimitive end
+
+function text_primitive_bbox end
+function register_text_primitive_plot! end
+
+"""
+    ImageTextPrimitive(image, marker_offset::Vec3f, size::Vec2f, rotation::Quaternionf)
+
+A text primitive that renders `image` as a single scatter marker at the
+text block's position. Used by e.g. MakieTeX to embed pre-rendered LaTeX
+into a Makie `text` plot.
+
+* `image`     — any marker accepted by `scatter` (typically a `Matrix{<:Colorant}`)
+* `marker_offset` — markerspace offset from the block origin to the marker's
+                    lower-left corner (after rotation has been applied)
+* `size`      — markerspace marker size
+* `rotation`  — applied to the marker by scatter
+"""
+struct ImageTextPrimitive{I} <: AbstractTextPrimitive
+    image::I
+    marker_offset::Vec3f
+    size::Vec2f
+    rotation::Quaternionf
+end
+
+function text_primitive_bbox(p::ImageTextPrimitive)
+    bb = Rect3d(to_ndim(Point3d, p.marker_offset, 0), Vec3d(p.size..., 0))
+    return rotate_bbox(bb, p.rotation)
+end
+
 # We sort out position vs string(-like) vs mixed arguments before convert_arguments,
 # so that we only get positions here
 conversion_trait(::Type{<:Text}, args...) = PointBased()
@@ -316,6 +371,7 @@ function compute_glyph_collections!(attr::ComputeGraph)
         :text_color, :text_rotation, :text_scales,
         :text_strokewidth, :text_strokecolor,
         :linesegments, :linewidths, :linecolors, :lineindices,
+        :text_primitives, :text_primitive_block_indices,
     ]
     return register_computation!(attr, inputs, outputs) do (input_texts, _inputs...), changed, cached
         _outputs = (
@@ -334,6 +390,8 @@ function compute_glyph_collections!(attr::ComputeGraph)
             linewidths = Float32[],
             linecolors = RGBAf[],
             lineindices = Pair{Int, Int}[],
+            text_primitives = AbstractTextPrimitive[],
+            text_primitive_block_indices = Int[],
         )
         # strokewidth = Float32[] # TODO: Skipped?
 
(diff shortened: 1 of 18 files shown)


ffreyer:
Related: #5396

jkrumbiegel:
Got both MakieTeX and this branch green now, so I think at this point it would be good to do a review. Questions are whether the mechanism on the Makie side is sound, and then also for @asinghvi17 whether he can get on board with the sweeping changes I made to MakieTeX (basically a completely new package at this point to simplify the API).

henrik-wolf:
Having studied the text rendering pipeline for a while to get #5396 working, I feels somewhat qualified to comment here. First of all, I am very happy to see some momentum happening here to open up the text rendering pipeline for extension!

Design-wise I am not really sure about simply adding another API onto the internal "APIs" that already exist for the three hard-coded string types instead of designing a single API that is powerful enough to implement all the functionality that we currently have within it. That way, we would not need to maintain multiple, slightly different implementations in parallel but just one API which is then consumed by a default implementation for the currently supported string types.

Additionally, there seems to be some dead, dead-ish or nearly duplicate code floating around the text implementation that is often only called in one place, which in many cases happens to be the (semi deprecated) 3d Axis? For reasons of future maintainability, I argue that designing/adding a text extension API is a great chance to take stock and clean up the historically grown code the text rendering currently runs on.

For the specific design here, I think it makes a lot of sense to split the rendering of the text/string payload from the actual placement. In order to make that work, we need to decide on the attributes that should be exposed to the consumer and which should not. Just having `color`, `fontsize` and `lineheight` is in my opinion not sufficient. (From a quick scroll through the text docs, I think `font`, `glowcolor`, `glowwidth`, `justification`, `strokecolor`, `strokewidth` and `alpha` should be available in some way. (Also, passing a vector into `rotation` seems to act per glyph?))

Maybe exposing as many of these attributes as possible is not a huge problem if we pass the `inputs, changed, cached` to the consumer and have them decide which changes in attributes require a recompilation?

(Small caveat: While I am trying to be objective here, I am of course to some extent partial to my solution. I hope that this does not invalidate my points. :) )

jkrumbiegel:
I have not studied your code yet as I thought it was somewhat orthogonal to my purpose here, which was just to get _some_ way of having custom objects render in the text pipeline, not to do it such that the default pipeline works the same way. I assumed that the plot spec way would always be a bit slower than the special path and that this was acceptable. But I'll have to look into your changes to see how it compares to what I did.

Makie/src/basic_recipes/text.jl
```diff
@@ -36,17 +82,18 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
-        elseif args isa Tuple{<:AbstractVector{<:AbstractString}}
-            return ((a_pos,), Ref{Any}(args[1]))
-        elseif args isa Tuple{<:AbstractVector{<:Tuple{<:Any, <:VecTypes}}}
+        # Text always comes from the `text` keyword/attribute and positions from
+        # positional args. The only positional shorthand recognized here is the
+        # `[(text, pos), ...]` form. Always emit a freshly-allocated vector for
+        # `:input_text` so ComputePipeline's `is_same(::Array, ::Array)`
+        # pointer-distinctness check can confirm structural equality on refires
+        # triggered by unrelated inputs (e.g. position layout writes).
+        if args isa Tuple{<:AbstractVector{<:Tuple{<:Any, <:VecTypes}}}
```
ffreyer, line 91: `f,a,p = text("test", position = Point2f(0))` works and does not depwarn on master and breaking-0.25. We should probably have a 0.x version with depwarns before we remove it? (And probably do the same for tooltip, textlabel, etc)

We should also remove the `position` attribute if we do.

Makie/src/basic_recipes/text.jl
```diff
@@ -356,6 +418,25 @@ function register_text_computations!(attr::ComputeGraph)
...
+    ) do texts, handler, colors, fontsizes, lineheights
+        handler === nothing && return Any[]
+        return Any[
+            compile_text(
+                    handler, str,
+                    sv_getindex(colors, i),
+                    sv_getindex(fontsizes, i),
+                    sv_getindex(lineheights, i),
+                ) for (i, str) in enumerate(texts)
+        ]
```
ffreyer, line 437: Not sure about lineheight, but fontsize and color should probably support per char values as well. I.e. use `per_glyph_block`. (Or maybe we should extend `PerCharIterator` to avoid array fills?)

Makie/src/basic_recipes/text.jl
```diff
@@ -417,30 +498,37 @@ function calculated_attributes!(::Type{Text}, plot::Plot)
...
+        isempty(specs) && return PlotSpec[]
         markerspace_positions = _project(preprojection * model_f32c, positions, clip_planes, space)
-        return map(linesegments, indices) do seg, (block_idx, glyph_idx)
-            return seg + markerspace_positions[block_idx]
+        return map(specs, indices) do spec, idx
+            shifted = shift_text_spec(spec, markerspace_positions[idx])
+            kw = copy(shifted.kwargs)
+            kw[:space] = markerspace
+            get(kw, :markerspace, nothing) === nothing || (kw[:markerspace] = markerspace)
+            return PlotSpec(shifted.type, shifted.args...; kw...)
```
ffreyer, line 527: Is this not something that `place_text` should handle? Or maybe some kind of layout function? Or the recipe used in the plotspec?
Otherwise these projections are always calculated, even if they get discarded/ignored by `shift_text_spec()` later. And they also trigger SpecApi (`diff_plotlist!()`) on camera movement. (This is about 1/10 the cost of updating ticks for a single text plot with three MathTexEngine latexstrings.)

Makie/src/basic_recipes/text.jl
ffreyer: If I understand correctly this step is supposed to do the expensive work of lowering input text into something useful with as few inputs as possible. Shouldn't those inputs depend on the text_handler then though? E.g. anything outputting glyphs probably doesn't care about colors, but may care about e.g. fonts.
Maybe the text handler should create the computation here, rather than just being involved in it?

Makie/src/basic_recipes/text.jl
ffreyer: I think this is also the only thing that forces PlotSpecs to be per block. That would also be good to avoid since it's cheaper to draw everything with one plot.

ffreyer (commented):
Since I always think of plotspecs as expensive, I tried to benchmark the update cost a bit using
```julia
a = L"\sqrt{1-x}"
b = "test" # L"\frac{1}{x}"
f,ax,p = text(Point2f(0), text = a);

@benchmark begin
    $p.text = $a
    foreach(kv -> kv[2][], $p.plots[1].attributes.outputs)
    $p.text = $b
    foreach(kv -> kv[2][], $p.plots[1].attributes.outputs)
end
```
This just uses MathTexEngine with https://github.com/Kolaru/MathTeXEngine.jl/pull/157.
When switching between two LaTeXStrings this pr dropped from 270µs -> 255µs. So just ~6%. When switching between a LaTeXString and a normal String it dropped from 225µs -> 390µs. That's about ~73% slower.

For create time of plotlist vs direct:
```julia
# breaking 0.25:    360µs
# nesting branch:   270µs
@benchmark plotlist!(scene, [S.Text(Point2f(0), text = "test")]) setup = (scene = Scene())

# breaking 0.25:    290µs
# nesting branch:   220µs
@benchmark text!(scene, Point2f(0), text = "test") setup = (scene = Scene())

@recipe SuperText (positions::Makie.PointVector,) begin
    Makie.documented_attributes(Makie.Text)...
end
Makie.conversion_trait(::Type{<:SuperText}, args...) = PointBased()
Makie.plot!(p::SuperText) = text!(p, p.attributes, p.positions)

# breaking 0.25:   470µs
# nesting branch:  320µs
@benchmark supertext!(scene, Point2f(0), text = "test") setup = (scene = Scene())
```
So here plotlist is actually faster than a recipe statically adding a plot. ~~Probably because it skips `cycle`.~~ Cycle doesn't affect the benchmark

dawbarton:
I don't know the Makie internals well enough to have a properly formed opinion, but following the (brief) discussion about `TeXLayout.jl` in Slack I got Claude to review this PR (including comments) and iterated on it a bit myself to refine things (and learn a bit more about the pipeline myself). The doc it produced is 
[makie-text-pipeline.md](https://github.com/user-attachments/files/28419192/makie-text-pipeline.md). In short, it iterated to the point where

1. `compile_text` - Remove `:computed_color` from the compile node, and add `font`/`fonts`, `justification`, and `word_wrap_width` so font information is available for layout and so line-breaking and embedded-text handlers work in the general case. This can produce its own custom payload as before *or* `CompiledText`. 
2. `CompiledText` - A backend-neutral representation of the layout that would be applicable to the likes of `MathTeXEngine.jl` or `TeXLayout.jl`. Backends that use this representation can avoid having to implement a boilerplate `place_text!`. Essentially it assumes that attributes like font size, colour, rotation, etc, are independent of the text layout and can be applied generically to the `CompiledText` representation. The main point is simply to reduce boilerplate in other packages since they would just have to implement the same code themselves.
3. `place_text!` - Add a generic handler for `CompiledText`. For those backends using a custom payload (e.g., font size dependent layout), they would still have to implement this

To me, this makes the `compile_text` and `place_text!` split more useful.

I got Claude to implement this in

- <https://github.com/dawbarton/Makie.jl/tree/new_text_handler> - Makie.jl (quite a small change, branched off this PR - note it added some exports that I'd probably remove)
- <https://github.com/dawbarton/TeXLayout.jl/tree/new_text_handler> - TeXLayout.jl
- <https://github.com/dawbarton/MathTeXEngine.jl/tree/new_text_handler> - MathTeXEngine.jl

Use or ignore as you see fit!

dawbarton:
Has there been any progress on getting this mechanism (or something similar) into Makie v0.25? It'd be great to land a proper extension mechanism rather than my somewhat janky type piracy that I use at the moment. I've also added HarfBuzz support to [TeXLayout.jl](https://github.com/dawbarton/TeXLayout.jl/) so you can get nice font kerning, ligatures, etc, in a relatively lightweight way. Regardless of the outcomes, thanks for all of your Makie-related efforts!

jkrumbiegel:
No progress yet but I want to get into it again when I'm back from vacation

jkrumbiegel:
Superseded by https://github.com/MakieOrg/Makie.jl/pull/5717

Would be good if you could take another look at that @henrik-wolf and @dawbarton and see if there's any big pain points left for your goals. I tried to simplify the mechanism and extension API while taking note of your suggestions but there's always a chance that things fall by the wayside

PR #5636, opened 2026-05-25
rework cycle init

ffreyer:
# Description

To figure out the cycle index, the previous implementation iterates through all plots in a scene and counts each of the same type. This requires runtime type inference and scales with the number of plots. Together with other runtime types (e.g. `plot.cycle[]`) this cost becomes noticeable with tens of plots.

Deriving cycle indices like this also leads to some awkward results when deleting plots:
```julia
f,a,p1 = scatter(1:10, color = :blue) # skipped
p2 = scatter!(2:11, 1:10) # 0 plots =>  cycle index 1
p3 = scatter!(3:12, 1:10) # 1 plot  =>  cycle index 2
p4 = scatter!(4:13, 1:10) # 2 plots =>  cycle index 3
delete!(a, p3) 
p5 = scatter!(5:14, 1:10) # 2 plots => cycle index 3
f
```

| Before | After |
| --- | --- |
| image | Screenshot from 2026-05-25 21-34-10 |

Cycling now works like this:
- scenes track cycle counters per `plotsym` (again)
- `Plot()` now only adds an input for `plot.cycle`. This is to allow cycle to be passed as an observable.
- `connect_plot!` -> `add_theme!` now does the bulk of the work:
  - adds a non-input node that fetches a cycle index from the parent scene on first resolve. This increments the counter in the scene for the fetches plot type
  - does theme resolution + init for `plot.cycle`
  - replaces input callbacks for all cycled attributes (doing basically the same as before)
  - runs cycle-aware theme resolution for all attributes. A cycled attribute can now be overwritten by `theme[:Plot]` but not plot defaults (value and `@inherit`)

The `Cycle` type also got changed a bit to have more useful data formats.

## Benchmarks

This isn't expected to show a significant difference since it has 0 plots to search through. Still seems a little better

```julia
using Makie, BenchmarkTools
@benchmark scatter!(scene, rand(Point2f, 10)) setup = (scene = Scene())

#                        min time      Allocations
# ff/nested-attributes:  29   - 31µs   48.71KiB   827
# ff/optimize-cycle:     28.5 - 29µs   47.51KiB   801
```

With multiple plots in a scene improvements should start to become more noticeable:

```julia
function nplots(scene, N)
    for _ in 1:N
        scatter!(scene, rand(Point2f, 10))
    end
    return
end

@benchmark nplots(scene, 30) setup = (scene = Scene())

#                        min time       Allocations
# ff/nested-attributes:  1.05 - 1.11ms  1.46MiB   26 329
# ff/optimize-cycle:     0.83 - 0.92ms  1.39MiB   23 875
```

## Functional Changes

- cycle can now be overwritten by `theme[:Plot]`
- cycle can now be inherited from scene themes (rather than just `theme(nothing)`)
- cycled attributes can now be overwritten by `theme[:Plot]` (not sure if that worked at all before?)
- plots inside recipes now cycle attributes that aren't set (should they?)
    - this causes a test failure with `Axis Hints`
- deleting a plot no longer decrements cycle indices of plots added later
- SpecApi probably acts a bit differently... still need to figure out how that actually behaved

## Type of change

## Checklist


ffreyer:
> SpecApi probably acts a bit differently... still need to figure out how that actually behaved

I tested how SpecApi behaves on master vs this branch when adding and removing SpecApi plots between normal plots:

```julia
begin
    inputs = [
        [S.Scatter(fill(i+1, 5), 1:5) for i in 1:3],
        [S.Scatter(fill(i+1, 5), 1:5) for i in 1:7],
        [S.Scatter(fill(i+1, 5), 1:5) for i in 1:3],
        [S.Scatter(fill(i+1, 5), 1:5) for i in 1:2],
        [S.Scatter(fill(i+1, 5), 1:5) for i in 1:3],
    ]

    f = Figure(size = (500, 1000))
    Label(f[0, 1], "update", tellwidth = false)
    Label(f[0, 2], "init", tellwidth = false)

    pls = []

    for j in 1:5
        scatter(f[j, 1], fill(1, 5), (1:5), marker = Rect)
        p = plotlist!(inputs[1])
        push!(pls, p)
        scatter!(fill(9, 5), (1:5), marker = Rect)

        scatter(f[j, 2], fill(1, 5), (1:5), marker = Rect)
        plotlist!(inputs[j])
        scatter!(fill(9, 5), (1:5), marker = Rect)

    end

    display(f)

    for i in 1:5
        for j in i:5
            pls[j].arg1 = inputs[i]
        end
    end
end
```

The left column shows a successively updated plotlist. First with 3 plots (top), then updated to 7 plots (axis below), then reduced back to 3 plots, then to 2, then back to 3. The right column shows the same setup without updates, i.e. with the list of plotspecs being used immediately.

| master | pr |
| --- | --- |
| Screenshot from 2026-06-04 16-16-17 | Screenshot from 2026-06-04 16-20-20 |

For reference, this is the color palette:

Screenshot from 2026-06-04 16-22-06

Notes:
- both master and pr advance cycle indices per plot spec and don't consider the plotlist itself. I.e. in the init column the plotlist plots all start at cycle index 2 (light orange) and the second normal scatter plot always uses cycle index `2 + length(inputs[i])`. So the displayed plots follow the palette order without gaps or jumps
- neither master nor pr change colors in normal scatter plots when the plotlist updates
- both master and pr skip the cycle index of the final plot when adding more specs (3 -> 7)
  - master cycle indices: `[1] 2 3 4 9 11 13 15 [5]`. (`[ ]` indicates normal scatter plots here) Adding more plotspecs seems to reset the cycle counter, add one for each cycled plot past the plotlist and then increment in steps of 2. (Checked with 0-3 plots at the end) 
  - pr cycle indices: `[1] 2 3 4 6 7 8 9 [5]`. Adding plots continues counting from the largest cycle index in steps of 1. 
- neither master nor pr cleanly recover from plotspec removal (last entry, going from 3 -> 2 -> 3)
  - master cycle indices: `[1] 2 3 7 [5]`. With normal plots after the plotlist, the added plotspec gets `max_cycle_index + 2`. Without I get `[1] 2 3 6`...?
  - pr cycle indices: `[1] 2 3 6 [5]`. The index chosen here is `max_cycle_index + 1`. If there are no plots past the plotlist, the cycle index is 4.

To me it seems like SpecApi cycling was already behaving weirdly on master and this pr tends to improve it a bit.

PR #5642, opened 2026-05-27
add searchable kwarg to Menu for use with lots of options

bjarthur:
fixes part of https://github.com/MakieOrg/Makie.jl/issues/4747#issuecomment-3053787566

```
using GLMakie

fig = Figure()

fruits = ["Apple", "Apricot", "Banana", "Blackberry", "Blueberry",
          "Cherry", "Cranberry", "Date", "Elderberry", "Fig",
          "Grape", "Grapefruit", "Honeydew", "Kiwi", "Lemon",
          "Lime", "Mango", "Nectarine", "Orange", "Papaya"]

sm = Menu(fig[1,1],
          options = fruits,
          searchable = true,
          search_placeholder = "type to filter...",
          prompt = "Searchable menu...")
```

Screenshot 2026-05-27 at 4 04 58 PM


bjarthur:
lots of options (100s or more, even 10s really) are hard to scroll through.  this PR filters the list of options with a user-supplied string down to a manageable size.

not sure why the reference tests are failing, but i don't think it is caused by this PR.

bjarthur:
@mkitti you might be interested in this

Makie/src/makielayout/blocks/menu.jl
```diff
@@ -99,13 +107,71 @@ function initialize_block!(m::Menu; default = 1)
...
+            placeholder = m.search_placeholder,
+            fontsize = m.fontsize,
+            textpadding = m.textpadding,
+            boxcolor = m.cell_color_inactive_even,
+            boxcolor_focused = m.cell_color_inactive_even,
+            boxcolor_hover = m.cell_color_inactive_even,
+            bordercolor_focused = m.cell_color_active,
+            tellwidth = false, tellheight = false,
+            width = lift(a -> Float32(widths(a)[1]), blockscene, textbox_area),
+        )
```
ffreyer: We could also consider replacing the `selectiontext` with an `editabletext` plot instead of adding a `Textbox`.

jkrumbiegel: Yeah I was going to say the same, that one should be used now I think

bjarthur: done!  see latest commit

bjarthur:
not sure why codespell failed, but i don't think it's this PRs fault, and the button to re-run it in the actions tab doesn't seem to be there.

bjarthur:
all tests pass now!  can you please take another look?

ffreyer (approved):

jkrumbiegel:
I added some minor tweaks to the hover behavior (readding some fixes that somehow got removed) and coloring

jkrumbiegel:
Wouldn't it make sense to set this to `true` by default? I don't see why it would hurt, seems like a nice general usability improvement

bjarthur:
i'm fine with a `true` default.  i chose to make it `false` simply for backwards compatibility.  but if this PR doesn't get released until the in-progress v0.25 then maybe that doesn't matter.

jkrumbiegel:
even then, it's just an additional feature, in my opinion it doesn't break any existing workflows

Makie/src/makielayout/blocks/menu.jl
```diff
@@ -1,11 +1,25 @@
...
+function _option_text_colors(hovered, optionstrings, m, filtered_indices)
+    return map(eachindex(optionstrings)) do idx
+        global_idx = filtered_indices[][idx]
+        if global_idx == m.i_selected[]
+            return to_color(m.textcolor_active[])
+        elseif idx == hovered
+            return to_color(m.textcolor_hover[])
+        else
+            return to_color(m.textcolor[])
+        end
```
ffreyer, line 10: These could also be type annotations in `@Block Menu` since we have

https://github.com/MakieOrg/Makie.jl/blob/a2ee515d868579592bb4c0ee16102886e2bc59d0/Makie/src/makielayout/blocks.jl#L615

SimonDanisch:
Btw, I already merged this into https://github.com/MakieOrg/Makie.jl/pull/5702 ...
This is still based against master, so I guess I'm happy to merge it here, and then just get the changes via master onto sd/breaking-gui

PR #5648, opened 2026-05-29
EXPERIMENT: Check Runs API with output.images

jkrumbiegel:
Throwaway PR to evaluate the Check Runs UI for benchmark output. Generates placeholder SVGs (no real benchmark) and publishes them as a check run on the PR head SHA via the Check Runs API — no PR comment, no commit statuses.

CI / changelog enforcer triggers are temporarily disabled on this branch to keep iteration cheap. Restore before merging anywhere.

Comment `/benchmark` here to test.


jkrumbiegel:
Abandoning — the github-actions app token doesn't surface manually-created check_runs in the PR's checks UI, so this approach would require a custom GitHub App with installation token JWT signing. Too much infrastructure for the upside. Going back to the comment + status approach on #5640.

PR #5654, opened 2026-05-30
Fix contourf not rendering hollow bands around extrema (#5651)

### LINKED ISSUE #5651: contourf drops closed interior minima (rendered correctly by surface) — CairoMakie
Image

### MWE

```julia
using CairoMakie

z = [cosd(3l) * cosd(y) for l in -180:4:180, y in -90:4:90]

fig = Figure(size = (1000, 430))
ax1 = Axis(fig[1, 1]; title = "contourf", aspect = DataAspect())
contourf!(ax1, z; levels = -1:0.2:1, colorrange = (-1, 1))
ax2 = Axis(fig[1, 2]; title = "surface", aspect = DataAspect())
surface!(ax2, z; colorrange = (-1, 1), shading = NoShading)
Colorbar(fig[1, 3]; limits = (-1, 1))
fig
```

In the **`contourf`** panel (left), the **closed interior minima** (the purple troughs at the interior columns) appear overdrawn by the surrounding `≈ 0` band. The **`surface`** panel (right) shows all the interior minima and maxima.

### Version

```
Makie       v0.24.10
CairoMakie  v0.15.10
Julia       v1.12.6
```

Maybe related to #4126 for `tricontourf` on GLMakie. But not a duplicate: This is `contourf` in CairoMakie. (cc @jkrumbiegel) 

Is this a regression? Did `contourf` render these correctly at some point? Maybe there is an existing test that would have caught this?

briochemc:
## Description

Fixes #5651. `contourf` failed to render hollow cut-outs around certain extrema, so lower-level bands were overdrawn by higher-level bands that should be hollow. The effect was most visible at interior **minima** (they disappeared), while maxima looked fine.

image

## Root cause

`contourf` builds each filled band from `Isoband.isobands`, which returns a flat, unordered set of rings (a mix of outer rings and hole rings). `_group_polys` reassembles them into `[outer, hole1, …]` using a containment matrix that tested **only the first vertex** of each ring:

```julia
PolygonOps.inpolygon(first(p1), p2) == 1
```

`PolygonOps.inpolygon` returns `1` strictly-inside, `-1` **on the boundary**, `0` outside. Marching-squares rings frequently share vertices at grid-edge crossings — especially where contours touch the data boundary (non-closed contours). When a hole ring's first vertex landed on its container's boundary, `inpolygon` returned `-1`, the containment was missed, and the hole was promoted to an independent outer polygon and **filled solid**.

The minima-vs-maxima asymmetry follows from draw order: bands are drawn in level order and later (higher) bands paint over earlier ones. At a maximum the innermost band is drawn last/on-top so the wrongly-solid band is hidden; at a minimum the innermost band is drawn first and gets overdrawn.

## Fix

Since Isoband rings never intersect, ring `inner` is contained in ring `outer` iff *any* representative point of `inner` is strictly inside `outer`. The new helper `_is_ring_contained` tests all vertices (with early-exit) and falls back to edge midpoints for the degenerate case where every vertex lies on a shared boundary vertex. This fixes both the rectilinear and curvilinear code paths (both route through `_group_polys`).

image

## Tests

- Added reference test `"contourf hollow bands around minima #5651"` in `ReferenceTests/src/tests/examples2d.jl` using the issue MWE.
- The existing central-minimum example in `generic_components.jl` (`[1 2 1; 2 0 2; 1 2 1]`) also exercises the hole path and now renders its minimum correctly, so its reference image will change too.
- Verified no regression on a maximum/saddle dataset and the curvilinear (x/y-matrix) path.


briochemc:
AFAIU the CI failure is because I added some image references. Maybe they are not needed, but if they are, can a dev upload them? 

As I said in the issue, this PR was generated by Claude Code, but I checked everything and it was a small diff and it looked good to me. However I don't have the deep understanding of Makie needed so it would be great for one of the Makie devs to take a good look knowing that an LLM wrote this. Probably @jkrumbiegel 🧙 given that this is about hollow filled contours! 😃

jkrumbiegel:
What I'd like is a minimal example where isoband returns lines that share points. I thought that wasn't possible when I wrote this code but apparently that assumption doesn't always hold.

Then rather than one reftest I'd like to have a couple unit tests that the algorithm works as intended with manually crafted polygons. And preferably one unit test where such polys come out of isoband itself.

Ah and the comments are too verbose, classic claude.

briochemc:
> What I'd like is a minimal example where isoband returns lines that share points. I thought that wasn't possible when I wrote this code but apparently that assumption doesn't always hold.

image


Here are 2 MWEs with a big of extra fluff to show the heatmap, the contourf, and the isobands individually, where the shared points are shown in red. Code:

```julia
using CairoMakie
const Makie = CairoMakie.Makie
const Isoband = Makie.Isoband

# all band rings for a grid over a set of consecutive level edges
function band_rings(z, edges)
    xs = collect(1.0:size(z, 2)); ys = collect(1.0:size(z, 1))
    lows = Float64.(edges[1:end-1]); highs = Float64.(edges[2:end])
    return lows, highs, Isoband.isobands(xs, ys, Float64.(z), lows, highs)
end

# points shared between distinct rings of one band
function shared_points(g)
    pts = collect(zip(g.x, g.y)); out = Point2f[]
    for p in unique(pts)
        length(unique(g.id[findall(==(p), pts)])) >= 2 && push!(out, Point2f(p))
    end
    out
end

function plot_iso3d!(ax, z, edges)
    lows, highs, isos = band_rings(z, edges)
    centers = (lows .+ highs) ./ 2
    cmap = Makie.compute_contourf_colormap(edges, :viridis, nothing, nothing)  # same discrete cmap as contourf
    crange = extrema(edges)
    for (center, g) in zip(centers, isos)
        for i in unique(g.id)
            sel = g.id .== i
            x = g.x[sel]; y = g.y[sel]; push!(x, first(x)); push!(y, first(y))  # close ring
            lines!(ax, x, y, fill(Float32(center), length(x)); color = center,
                colorrange = crange, colormap = cmap, linewidth = 3)
        end
        sp = shared_points(g)
        isempty(sp) || scatter!(ax, [Point3f(p[1], p[2], center) for p in sp];
            color = :red, markersize = 11)
    end
    # z ticks at band centers, labelled with the band range
    labels = [string(round(l, digits = 2), "–", round(h, digits = 2)) for (l, h) in zip(lows, highs)]
    ax.zticks = (centers, labels)
    return centers
end

function column!(fig, col, z, edges, title)
    ax_h, hm = heatmap(fig[1, col], z; colormap = :viridis, colorrange = extrema(edges),
        axis = (title = title, aspect = DataAspect()))
    Colorbar(fig[1, col + 1], hm)
    ax_c, cof = contourf(fig[2, col], z; levels = edges,
        axis = (title = "contourf", aspect = DataAspect()))
    Colorbar(fig[2, col + 1], cof)
    ax3 = Axis3(fig[3, col]; title = "isobands by level (red = shared pts)", zlabel = "band")
    plot_iso3d!(ax3, z, edges)
end

zA = [0 1 0; 1 2 1; 0 1 0]; edgesA = [0.5, 1.0, 1.5, 2.5]
zB = [0 2 0; 2 0 2; 1 1 1]; edgesB = [-0.5, 1.0, 1.5, 2.5]

fig = Figure(size = (1100, 1180))
column!(fig, 1, zA, edgesA, "[0 1 0; 1 2 1; 0 1 0]")
column!(fig, 3, zB, edgesB, "[0 2 0; 2 0 2; 1 1 1]")
rowsize!(fig.layout, 3, Relative(0.42))
fig
```

I'll work on updating the PR accordingly :)

jkrumbiegel (approved):

PR #5656, opened 2026-06-01
Volume improvements

ffreyer:
# Description

Some relatively small improvements for volume plots.

#### Color mapping

Adds the standard color mapping code to volume. This enables `lowclip`, `highclip` and `nan_color` to work. Kind of surprised this wasn't already there. Probably pretty useful to have to mask out sections of a volume:

```julia
data = [(x*x + y*y - z*z) for x in 0:10, y in 10:-1:0, z in 0:10]
f,a,p = volume(
    data, algorithm = :absorption,
    colorrange = (10, 30), lowclip = :transparent, highclip = :transparent,
    absorption = 10, transparency = true
)
```

image

##### mip colorrange

Mip uses the maximum value along a ray. When `highclip = :transparent` this leads to any column with values beyond colorrange to be transparent. I think it's more reasonable to instead use the highest visible value, so I changed the code accordingly. (If highclip is not transparent, values are still allowed to go past the colorrange)

#### Blending/Color accumulation

I checked how colors should be blended assuming that `over` operations are meant to be used. This lead to two changes:
1. color summation (which uses transmittance) now happens before the transmittance update
2. the final color is now weighted by `1 / (1 - transmittance)`

Without (1) each color sample gets weighted by an additional `1 - color_sample.alpha`, reducing it's impact. This effectively makes things darker, because it gives weight to the initial `vec3(0)` value of the color sum.

Without (2) the final alpha value of the color (which gets blended with the scene background and other plots) gets applied to the color twice, effectively darkening it. This should impact more transparent regions more.

These effects can be seen easily when dark colors are not involved in the colormapping of volume:

```julia
samples = fill(0.0, 101, 101, 101);
for _ in 1:100_000
    speed = 40rand()
    dir = rand(-1:1, 3) .+ 0.1 .* rand(3) .- 0.05
    i, j, k = 51 .+ round.(Int, speed .* dir)
    samples[i, j, k] += 1
    samples[i-1:i+1, j-1:j+1, k-1:k+1] += rand(3,3,3)
end

f = Figure(backgroundcolor = RGBf(0.5, 0.5, 0.5))
a,p = volume(
    f[1, 1],
    -1..1, -1..1, -1..1, samples,
    algorithm = :absorption,
    colormap = RGBAf.(to_color.([:white, :white]), [0.1, 0.1]),
    colorrange = (50, 200),
    lowclip = :transparent, highclip = :red,
    absorption = 50
)
```

The rays should brighten the scene, as they add white or get removed via :transparent.  (`:transparent` being `RGBAf(0,0,0,0)` should not matter here, as the transition to lowclip is not color-interpolated.)

| master | pr |
| --- | --- |
| Screenshot from 2026-06-01 18-06-22 |  Screenshot from 2026-06-01 18-06-38 |

#### Local `step_size`

The volume shader sends out a ray for each pixel that needs to be drawn. The rays are restricted to the bounding box of volume, starting at one end and stopping at the other. Each ray goes through a fixed number of steps, meaning the real step size depends on where the ray is within the volume bounding box.

Most volume algorithms weigh their color sample by the step size (~Riemann sum). They used to use a global constant `step_size` for this, which give too much weight to samples near corners/edges. Now they just use `length(dir)` instead, which is the local step size.

```julia
cube = [x+y+z == 10 for x in 0:10, y in 0:10, z in 0:10]
f,a,p = volume(
    -1..1, -1..1, -1..1, cube,
    algorithm = :absorption,
    colormap = [RGBAf(1,0,0,0), RGBAf(0,1,0,1)],
    absorption = 5, transparency = true
)
```

| before | after |
| --- | --- |
| Screenshot from 2026-06-01 18-53-51 | Screenshot from 2026-06-01 18-54-02 |

#### Add `samples` as an attribute

Made the sample count a uniform/attribute so it can be adjusted. Because why not?

#### Add `absorption` as a sample scaling attribute to `:additive`

The current version of `additive` typically doesn't produce anything useful without scaling your data (typically down). This change allows you to do that with an attribute instead of manually multiplying `0.01 .* data`. Doesn't change the baselines plot at all. Just convenience.

## Type of change

I think these can be seen as fixes/extensions to how volume should work, even if they change the visuals.

## Checklist


asinghvi17:
I'd tend to agree here, these look generally better than what we had before!

PR #5662, opened 2026-06-07
Weigh volume samples for `:additive` by step size

ffreyer:
# Description

When throwing normal RGBA values at `:additive` you usually get junk. Typically just a white blob, sometimes weird interference patterns. With the same data on master:

Screenshot from 2026-06-07 19-38-27

This happens because samples colors directly and overshoots `RGB(1, 1, 1)` (white) by a lot. Adding a `step_size = length(dir)` multiplier to the sampled color (like all absorption algorithms) adjusts this to more reasonable ranges, producing:

Screenshot from 2026-06-07 19-38-18

I'm separating this from #5656 because I have absolutely no clue what `additive` is trying to do and thus I can't say this is fixing anything.

## Type of change

## Checklist


SimonDanisch (approved):

ffreyer:
Still unsure whether this should be breaking or not. Adding `step_size` means all samples are scaled down by a factor of roughly 0.01, which means any plot that produced results before could very well end up mostly transparent after this.

ffreyer:
Eh, there are only 2 or 3 issues that include `:additive` and those don't even use it correctly (with RGBA volumes) so the impact of this pr is probably pretty low...

PR #5669, opened 2026-06-15
fix issue with linestyles changed via cycles and explicitly set to solid

### LINKED ISSUE #5267: axislegend displays wrong linestyle
- [x] what version of Makie are you running? v0.24.6 (Julia v1.11.6)
- [x] can you reproduce the bug with a fresh environment ? Yes. (If I add CairoMakie or GLMakie)
- [x] What platform + GPU are you on? MacOS, ARM
- [x] Which backend are you using? CairoMakie v0.15.6  (the same is with GLMakie v0.13.6)
- [x] Please provide a minimal example of the failing code. If the problem is with the visualization, please provide a picture or video of it.

I use `set_theme!` to `Cycle` through linestyle. Then while plotting I specify the style of the second line to `:solid`, but on the legend it's still `:dash`.

```julia
using CairoMakie

set_theme!(Lines=(cycle=Cycle([:linestyle]),))

fig = Figure()
ax = Axis(fig[1, 1])

lines!(ax, [0.0, 0.3], [0.0, 0.2], label="default (1)")
lines!(ax, [0.15, 0.3], [0, 0], color=:green, linestyle=:solid, label="green solid (2)")

axislegend(ax)
fig
```

Image

I haven't tried other backends than CairoMakie and GLMakie. 
Maybe this is related to #5241,  which I found through bug #4866 (but I don't know :))

ranocha:
# Description

Fixes https://github.com/MakieOrg/Makie.jl/issues/5267

There is a bug when a cycle changes the linestyle but the users sets the linestyle to solid explicitly for a plot.

## Type of change

## Checklist

## Disclosure

I used an LLM to help me find the issue and a possible fix, but reviewed and improved the code as much as I can. However, I would be very grateful for a review by people familiar with this codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed `Legend` showing the wrong linestyle for plots that explicitly set `linestyle = :solid` when a `Cycle([:linestyle])` theme is active [#5267](https://github.com/MakieOrg/Makie.jl/issues/5267).
 - `text` now validates `align` and errors with a clear message for invalid values like `align = :center` [#4651](https://github.com/MakieOrg/Makie.jl/pull/4651).
 - Fixed `contourf` not rendering non-closed contours [#5651](https://github.com/MakieOrg/Makie.jl/issues/5651).
 - Fixed WGLMakie flickering while continuously resizing a canvas by re-rendering right after the resize.
--- a/Makie/src/makielayout/blocks/legend.jl
+++ b/Makie/src/makielayout/blocks/legend.jl
@@ -418,6 +418,9 @@ function legendelement_plots!(scene, element::MarkerElement, bbox::Observable{Re
     attrs = element.attributes
     fracpoints = attrs.markerpoints
     points = lift((bb, fp) -> fractionpoint.(Ref(bb), fp), scene, bbox, fracpoints)
+    # We need to explicitly pass `cycle = nothing` to avoid issues with cycles, see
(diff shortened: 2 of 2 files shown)


SimonDanisch (approved):

ffreyer:
This is not specific to `Legend`. We should fix it more generally.

Another example:
```julia
with_theme(
    Theme(
        palette = (color = [:red, :blue], linestyle = [:dash, :dot]),
        Lines = (cycle = Cycle([:color, :linestyle], covary = true),)
    )) do
    lines(fill(5, 10), linestyle = nothing)
    lines!(fill(4, 10), linestyle = nothing)
    lines!(fill(3, 10), linestyle = nothing)
    lines!(fill(2, 10), linestyle = nothing)
    lines!(fill(1, 10), linestyle = nothing)
    current_figure()
end
```
image

To fix this we just have to not use `nothing` as a marker for "derive this from cycling" here:
https://github.com/MakieOrg/Makie.jl/blob/31de53c4643b41605b3a4e23ad68449e13de14de/Makie/src/compute-plots.jl#L634-L649

E.g. `get(kwargs, k, :cycled)` and `if value !== :cycled` instead of `if !isnothing(value)`. 

A small test would also be nice.

ranocha:
I agree, a general fix would be nice. However, that's beyond my current capacity, so please go ahead with a PR following the approach you proposed.

ranocha:
@ffreyer I implemented your suggestion in #5677

PR #5672, opened 2026-06-16
Allow `Legend(..., ax)` to update with plots created after Legend

### LINKED ISSUE #3207: `axislegend` could update when new plots are added to the axis
### Feature description

After creating an axis, plotting, and running `axislegend`, it would be good if adding another plot updated the legend.

```julia
p = scatter(rand(10); label="first scatter")
axislegend(p.axis)
```
We get as expected a scatter plot with a single entry in the legend.

But then if we add another scatter:

```julia
scatter!(p.axis, rand(10); label="other scatter")
```

Its not updated with the new label, but would be nice if it did.

The workaround is we have to delete the old one and add another one. In a recipe context this means we have to find it first before we can delete it.

ffreyer:
# Description

This enables creating the legend first and adding plots later:

```julia
f = Figure()
a = Axis(f[1, 1])
l = Legend(f[0, 1], a, tellwidth = false, tellheight = true)

scatter!(a, rand(10), label = "label1")
scatter!(a, rand(10), label = "label2")
```

When executed line by line, this starts off with an empty legend (occupies a little bit of space, maybe because of gaps?), then updates to add "label1" then updates again to add "label2".

This allows for auto-legend to be implemented like this:

```julia
@Block DecoratedAxis <: Makie.AbstractAxis begin
    axis::Axis
    legend::Legend
end

function Makie.initialize_block!(b::DecoratedAxis)
    b.axis = Axis(b[1, 1])
    # maybe with conditions...
    b.legend = Legend(b[0, 1], b.axis)
    return
end

Makie.preferred_axis_type(::Scatter, x, y) = DecoratedAxis

# required forwards for AbstractAxis. There may be more
Makie.plot!(ax::DecoratedAxis, p::Makie.AbstractPlot) = Makie.plot!(ax.axis, p)
Makie.reset_limits!(ax::DecoratedAxis) = Makie.reset_limits!(ax.axis)

scatter(rand(10), label = "label")
```

- fixes #3207

## Type of change

## Checklist


ffreyer:
I think this is done up to docs, tests, changelog. Do we want this feature? I.e. that calling `plot!(ax, ... label = ...)` adds legend entries to a `Legend(..., ax)` after the creation of that legend?

ffreyer:
Todo: update on delete as well

PR #5678, opened 2026-06-20
Update Colorbar to compute graph

ffreyer:
# Description

This contains functional changes with respect to how colormapping data/attributes are extracted from plots and used by `Colorbar`.

#### Before

`extract_colormap(plot)` is called recursively, returning a `ColorMapping` if implemented. Only one child is allowed to produce an output when recursion happens. Primitive plots produce `ColorMapping` without recomputing. Recipe plots, by default, try to build their of `ColorMapping` using default names like `plot.color`, `plot.colormap` etc. Custom implementation are added when non-default names are used and/or more processing is needed. All of these cause compute node -> Observable transitions.

In Colorbar most (but not all) the content of `ColorMapping` is used to build the Colorbar. At the some of fields overwrite `Colorbar` attributes (limits, colormap, highclip, lowclip, scale). (This makes limits post-(color)scale if Colorbar derives from a plot and pre-(color)scale if it does.)

When constructing a Colorbar without a plot, a `ColorMapping` is created using Colorbar attributes.

#### After

Colorbar now always processes color(mapping) using `register_colormapping_without_color!()` which is also used for plots. To allow this, `alpha` and `values` (color values) have been added. Color values + colorrange are handled separately from plots because plots apply colorscale too early for Colorbar.

When constructing a `Colorbar` from a plot `extract_colormap(_recursive)` is still used to gather the relevant attributes. It now primarily expects a dict containing the relevant attributes, but `ColorMapping` still works with a depwarn. The default `extract_colormap` now grabs all the relevant colormap attributes from the plot (with default names). If that set of attributes is incomplete, a set is generated from each child plot. If that yields one set of attributes it's returned, otherwise an error is thrown (like before).

Users can overwrite `extract_colormap` or `_extract_colormap` to define the colormap attributes themselves. The latter method will use a default `extract_colormap` that adds missing attributes with default names. The former should do that itself. If it doesn't, attributes can be sourced from different plots when one implements `extract_colormap(plot) = extract_colormap(plot.plots[1])` which might rarely be problematic? `extract_colormap_recursive` will also try to fill out missing attributes using default names and an error will be thrown if that does not yield a full set.

Any attributes that were grabbed set the matching Colorbar attributes and trigger an error if they are manually set. Any attributes that are not extracted fall back to Colorbar defaults. (Currently all but alpha are required)

#### Other functional changes

- highclip and lowclip triangles are now part of the same `poly` plot (using two Polygons)
- `limits` has been marked as deprecated in favor of the duplicate Colorbar attribute `colorrange`

#### TODO

- [x] reconsider changes to `extract_colormap`
- ~~maybe fix slight shift of Colorbars with lowclip/highclip~~

## Type of change

## Checklist

### DIFF
--- a/Makie/src/makielayout/blocks/colorbar.jl
+++ b/Makie/src/makielayout/blocks/colorbar.jl
@@ -7,6 +7,19 @@ function colorbar_check(keys, kwargs_keys)
     return
 end
 
+# compat
+extract_colormap(::AbstractPlot) = nothing
+function extract_colormap_recursive(plot::AbstractPlot)
+    result = extract_colormap(plot)
+    if isnothing(result)
+        for child in plot.plots
+            child_result = extract_colormap_recursive(child)
+            isnothing(child_result) || return child_result
+        end
+    end
+    return result
+end
+
 function extract_colorbar_attributes(plot::Arrows2D)
     map!(plot, [:tailcolor, :shaftcolor, :tipcolor, :color], :raw_merged_color) do a, b, c, d
         return [default_automatic(a, d); default_automatic(b, d); default_automatic(c, d)]
@@ -17,6 +30,7 @@ end
 extract_colorbar_attributes(plot::Voxels) = (color = plot.chunk, colorrange = plot.value_limits)
 extract_colorbar_attributes(plot::StreamPlot) = extract_colorbar_attributes_with_defaults(plot.plots[1])
 extract_colorbar_attributes(plot::VolumeSlices) = (color = plot[4],)
+extract_colorbar_attributes(plot::Hexbin) = (color = plot.count_hex,)
 
 _normalize_clipcolor(x) = x in (nothing, :auto, automatic) ? automatic : x
 function extract_colorbar_attributes(plot::Union{Contourf, Tricontourf})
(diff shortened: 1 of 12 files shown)


Makie/src/makielayout/blocks/colorbar.jl
```diff
@@ -7,241 +7,232 @@ function colorbar_check(keys, kwargs_keys)
...
-            return (lv * height(box), hv * height(box))
-        end .* sin(pi / 3)
+    map!(x -> x !== automatic, cb, :lowclip, :lowclip_tri_visible)
+    map!(x -> x !== automatic, cb, :highclip, :highclip_tri_visible)
+
+    map!(
+        cb, [:highclip_tri_visible, :lowclip_tri_visible, :framebox, :vertical], :tri_heights
+    ) do hv, lv, box, vertical
+        return (lv, hv) .* ifelse(vertical, width(box), height(box)) .* sin(pi/3)
     end
```
ffreyer, line 267: Tests fail because of slight shift in the Colorbar size when lowclip and/or highclip are used. The reason for this is the change here. Before `sin(pi/3)` was not included because of the return in the branch, now it is. That should be a bug fix

PR #5685, opened 2026-06-25
Support non-identity axis scales in ablines

jkrumbiegel:
`ablines` used to error on any non-identity axis scale (`ABLines is only defined for the identity transform`). But there's no real reason for that: `f(x) = slope * x + intercept` is defined in data coordinates, so under a transform it just becomes a curve in screen space. This draws it as a subdivided curve (256 points sampled evenly across the visible range), and keeps the cheap 2-point line for the identity transform.

A few details fall out of this:
- On log-log axes a zero-intercept line maps back to a straight line, as you'd expect; with a nonzero intercept it's genuinely curved.
- The line value can leave the y-scale domain (e.g. go negative on a log y axis) even when the axis limits are valid. Those points become `NaN`, so the line breaks there. Since the visible y-range is always inside the domain, the break is always off-screen.
- Per-line `color`/`linewidth` vectors still work, and it now draws with `lines` rather than `linesegments`.

```julia
f = Figure(size = (600, 500))

ax = Axis(f[1, 1], title = "identity")
ablines!(ax, [0.0, 1.0, 2.0], [1.0, -0.5, 0.25], color = [:orange, :red, :purple], linewidth = 3)
limits!(ax, 0, 10, -2, 6)

ax = Axis(f[1, 2], xscale = log10, title = "log x")
ablines!(ax, 0.0, 0.002, color = :orange, linewidth = 3)
limits!(ax, 1, 1000, 0, 3)

ax = Axis(f[1, 3], yscale = log10, title = "log y, crosses zero")
ablines!(ax, -50.0, 50.0, color = :orange, linewidth = 3)
limits!(ax, 1, 4, 1, 200)

ax = Axis(f[2, 1], xscale = log10, yscale = log10, title = "log-log, zero intercept")
ablines!(ax, [0.0, 0.0, 0.0], [1.0, 3.0, 0.3], color = [:orange, :red, :purple], linewidth = [2, 4, 6])
limits!(ax, 1, 100, 0.1, 1000)

ax = Axis(f[2, 2], xscale = log10, yscale = log10, title = "log-log, nonzero intercept")
ablines!(ax, [1.0, 10.0, 100.0], 1.0, color = [:orange, :red, :purple], linewidth = 3)
limits!(ax, 1, 100, 1, 1000)

ax = Axis(f[2, 3], xscale = sqrt, title = "sqrt x")
ablines!(ax, 0.0, 1.0, color = :orange, linewidth = 3)
limits!(ax, 0, 100, 0, 100)

f
```

ablines under different axis scales

### DIFF
--- a/Makie/src/basic_recipes/ablines.jl
+++ b/Makie/src/basic_recipes/ablines.jl
@@ -3,30 +3,74 @@
 
 Creates a line defined by `f(x) = slope * x + intercept` crossing a whole `Scene` with 2D projection at its current limits.
 You can pass one or multiple intercepts or slopes.
+
+Under a non-identity axis scale the line is drawn as a subdivided curve, so it stays correct on e.g. log axes.
 """
 @recipe ABLines (intercept, slope) begin
-    documented_attributes(LineSegments)...
+    documented_attributes(Lines)...
 end
 
-function Makie.plot!(p::ABLines)
-    scene = Makie.parent_scene(p)
-    transf = transform_func(scene)
-    is_identity_transform(transf) || throw(ArgumentError("ABLines is only defined for the identity transform, not $(typeof(transf))."))
+const N_ABLINE_SUBDIVISIONS = 256
 
+function Makie.plot!(p::ABLines)
     add_axis_limits!(p)
 
-    map!(p.attributes, [:axis_limits, :intercept, :slope], :points) do lims, intercept, slope
+    map!(
+        p.attributes,
+        [:axis_limits, :axis_limits_transformed, :intercept, :slope, :transform_func],
+        :points
+    ) do lims, lims_transformed, intercept, slope, transf
         points = Point2d[]
         broadcast_foreach(intercept, slope) do intercept, slope
-            f(x) = intercept + slope * x
-            xmin, xmax = first.(extrema(lims))
-            push!(points, Point2d(xmin, intercept + slope * xmin))
-            push!(points, Point2d(xmax, intercept + slope * xmax))
+            append_abline_curve!(points, lims, lims_transformed, transf, intercept, slope)
         end
         return points
     end
-    linesegments!(p, Attributes(p), p.points)
+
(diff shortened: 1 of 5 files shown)


Makie/src/basic_recipes/ablines.jl
```diff
@@ -3,32 +3,76 @@
...
 @recipe ABLines (intercept, slope) begin
-    documented_attributes(LineSegments)...
+    documented_attributes(Lines)...
 end
 
-function Makie.plot!(p::ABLines)
-    scene = Makie.parent_scene(p)
-    transf = transform_func(scene)
-    is_identity_transform(transf) || throw(ArgumentError("ABLines is only defined for the identity transform, not $(typeof(transf))."))
+const N_ABLINE_SUBDIVISIONS = 256
```
ffreyer, line 13: This should probably be an attribute?

jkrumbiegel: Yeah good point!

SimonDanisch:
At some point we wanted to make this work generically by adapting the apply_transform machinery to be able to add points to lines:
```julia
apply_transform(func, vector_of_points)::Vector{Point}
```
Maybe worth a shot with claude being able to do the annoying interface refactor?

ffreyer:
Dendrogram has something specifically for `Polar`:
https://github.com/MakieOrg/Makie.jl/blob/eff2c6f12bb351a35a5cf1c174c5cdef716c0454/Makie/src/stats/dendrogram.jl#L93-L136

If we added resampling in an automated way it would definitely need an on/off switch. Discrete plots like scatter definitely shouldn't resample. Something like lines may also have situations where it shouldn't. For example if you make a spider chart by reducing the number of points in a PolarAxis.

SimonDanisch:
Yeah, it'd need to dispatch on the plot type as well...
We had a concept somewhere sometime I think, which thought this through a bit more than my comment :sweat_smile:

PR #5688, opened 2026-06-29
[draft/test] WGLMakie docs interactivity via DocumenterVitepress SPA script-execution fix

asinghvi17:
**Draft — do not merge. CI test only.**

This points the docs environment at a DocumenterVitepress branch that fixes interactive `text/html` outputs not rendering after client-side (SPA) navigation in VitePress.

### Background
DocumenterVitepress renders `show(::IO, ::MIME"text/html", x)` output through Vue's `v-html` (`el.innerHTML = …`). Browsers never execute `<script>` tags inserted via `innerHTML`. On a hard page load this is masked because VitePress server-renders the markup and the parser runs the baked-in scripts — but VitePress is an SPA, so on client-side navigation (clicking a sidebar link) the scripts never run. WGLMakie/Bonito figures bootstrap from `<script>` tags, so they show only the Bonito loading spinner. This is the long-standing "WGLMakie examples are not being built correctly" issue noted on `explanations/backends/wglmakie.html`.

### What this PR does
- `docs/Project.toml`: `[sources]` entry pointing `DocumenterVitepress` at [`LuxDL/DocumenterVitepress.jl@as/wglmakie-vitepress-script-execution`](https://github.com/LuxDL/DocumenterVitepress.jl/tree/as/wglmakie-vitepress-script-execution), and widen compat to allow `0.3`.

The DV branch re-executes the `<script>` tags inside `text/html` outputs after each client-side navigation (preserving order; only on navigation, not initial load, so SSR-baked scripts aren't double-run).

### How to check
After the docs preview deploys, navigate (don't hard-reload) to **Explanations → Backends → WGLMakie** and confirm the example figures render instead of spinning.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -34,7 +34,7 @@ WGLMakie = "276b4fcb-3e11-5398-bf8b-a0c2d153d008"
 [sources]
 CairoMakie = {path = "../CairoMakie"}
 ComputePipeline = {path = "../ComputePipeline"}
-DocumenterVitepress = {url = "https://github.com/LuxDL/DocumenterVitepress.jl", rev = "d3dd92302124f74a5b7ad0d8ad719785433525eb"}
+DocumenterVitepress = {url = "https://github.com/LuxDL/DocumenterVitepress.jl", rev = "as/bonito-plugin"}
 GLMakie = {path = "../GLMakie"}
 Makie = {path = "../Makie"}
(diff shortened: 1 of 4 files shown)


lazarusA:
unfortunately, not all examples preserve their interactive nature: https://docs.makie.org/previews/PR5688/explanations/backends/wglmakie

good progress on the 3D plots, those look good.

PR #5693, opened 2026-07-04
WGLMakie: make scene lookup retry window configurable

samtkaplan:
## Summary
- make WGL scene lookup retry timing configurable via `WGLMAKIE_SCENE_RETRY_DELAY_MS` and `WGLMAKIE_SCENE_RETRY_TOTAL_MS`
- derive `max_retries` from configured delay and total wait with sane clamping
- replace hard-coded JS retry constants with interpolated values from Julia config

## Motivation
Large scenes in app mode can initialize more slowly than a fixed retry window, causing `Scene not found after retries` errors.

## Notes
- defaults remain conservative (`100ms` retry delay and `60000ms` total wait)
- this PR keeps configuration scoped to `WGLMAKIE_*` env vars only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- WGLMakie: made scene lookup retry timing configurable via `WGLMAKIE_SCENE_RETRY_DELAY_MS` and `WGLMAKIE_SCENE_RETRY_TOTAL_MS` to avoid timeout failures for slow scene initialization.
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Thanks! I will merge this as is even though this is not the greatest mechanism - would be nice to have this work without any configuration.
But solving this nicely with e.g. announcing what plots there are etc would be a bigger refactor, so this is a good compromise!

PR #5695, opened 2026-07-05
Add high-level schematic to the Figures and Blocks architecture docs

### LINKED ISSUE #2866: High level schematic overview of Makie's elements/structure
This outline was drawn on a blackboard today at MakieCon and the explanations that went with it seemed really helpful for the participants to understand how Makie's "building blocks" are structured and how they relate to each other.

How about putting this into some kind of digital form (e.g. with drawio or maybe even Makie itself) so that we can put it on the introductory pages of the docs?

aminehd:
Closes #2866.

At MakieCon, a blackboard sketch of Makie's building blocks (`Figure → Scene / Layout`, and how `Axis(f[1,1])` connects in) helped participants understand the structure; #2866 asked to digitize it for the introductory docs.

This adds a small Graphviz diagram and a short explanation under the [**Figures and Blocks**](https://github.com/MakieOrg/Makie.jl/blob/master/docs/src/explanations/architecture.md#figures-and-blocks) section of `architecture.md` — the location suggested by @asinghvi17 in the issue. It shows how a `Figure`'s `Scene`, `Layout` and `content[]` connect to an `Axis`'s `blockscene` and `GridContent`:

- the axis's `blockscene` attaches to the figure's root `Scene` (so it's drawn)
- the axis's `GridContent` is added to the layout's `content[]` (so it's positioned)

It reuses the docs' existing `@graphviz` block style, so it renders in the same monochrome look as the neighboring diagrams. The change is docs-only and purely additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - WGLMakie: fixed a `Cannot destructure property 'geometry' of 'mesh'` JS error and allow `Bonito@v5` [#5683](https://github.com/MakieOrg/Makie.jl/pull/5683)
 - Adjusted cycled attributes to be marked as `:cycled` instead of `nothing` so that `nothing` doesn't get overridden. This allows e.g. `linestyle = nothing` to be set when `linestyle` is cycled. [#5267](https://github.com/MakieOrg/Makie.jl/issues/5267)
+- Added a high-level schematic to the "Figures and Blocks" section of the architecture docs [#5695](https://github.com/MakieOrg/Makie.jl/pull/5695).
 
 ## [0.24.12] - 2026-06-18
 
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
Hi, thanks for the PR! Have you checked the other diagrams on that page? The layout/scene connections of Block objects are already covered there

aminehd:
Hi @jkrumbiegel , thanks for reviewing.
That's correct.
I've read the page again and ran it with Claude Code. Based on Claude code a higher-level overview of ecosystem (plotting API and all the backends such as GLMakie...) is missing. Do you think it would be useful if I add that paragraph and schema under Architecture heading? Something like below

image

 
Or any documentation part that you think is missing.
I am pretty new to open source contribution.

jkrumbiegel:
I don't think such a graphic would be that informative to be honest. As we're trying to make documentation that is understandable for humans (that often also means for agents) the best thing you can do is note where you had trouble understanding something and trying to improve that bit. Not pointing an agent with the prompt "do some improvement to the docs"

aminehd:
HI @jkrumbiegel , Yeah it makes sense. Thanks for the advice, i think reading the README by human and adding to it is better. I'll do so. Anything you think is missing so I can start with that?

jkrumbiegel:
Not really to be honest. The best way I think is to use Makie, whenever you struggle with something consult the docs, if the docs are not good think about how they could be improved. It's quite hard to do this without familiarity with the tool and docs that comes from having used it for a while.

PR #5699, opened 2026-07-09
Fix `PlotSpec` constructor

topolarity:
# Description

Fixes these errors:
```julia
julia> PlotSpec(1:4)
ERROR: ArgumentError: new: too few arguments (expected 3)

julia> PlotSpec(1:4; color = :red)
ERROR: MethodError: no method matching tuple(::Symbol, ::UnitRange{Int64}; color::Symbol)
This method does not support all of the given keyword arguments (and may not support any).
```

and this warning:
```julia
┌ Warning: PlotSpec objects are supposed to be title case. Found: plot. Please use plot instead.
└ @ Makie ~/repos/Makie.jl/Makie/src/specapi.jl:34
```

## Type of change

## Checklist

### DIFF
--- a/Makie/src/specapi.jl
+++ b/Makie/src/specapi.jl
@@ -26,7 +26,7 @@ struct PlotSpec
         if type_str[end] == '!'
             error("PlotSpec objects are supposed to be used without !, unless when using `S.$(type)(axis::P.Axis, args...; kwargs...)`")
         end
-        if !isuppercase(type_str[1])
+        if type !== :plot && !isuppercase(type_str[1])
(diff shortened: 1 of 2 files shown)


asinghvi17:
Hmm I don't know if that actually works in practice (you probably want some resolved plot type before you hit the function which constructs the plot from the spec).  Did you happen to test this out @topolarity?

I suspect `PlotSpec(:plot, ...)` probably also breaks the diffing

topolarity:
> Did you happen to test this out @topolarity?

Only to build the object, but not to actually use it

This is just a drive-by from a PkgEval that caught the bad `new(...; kwargs...)` constructor

Issue #208, opened 2018-11-02
FR: Add save button to the window opened up by Makie


Datseris:
Super - duper - convenient :) 

When a plot is shown, it is very helpful to add a save button to this interactive window opened up by makie!

mschauer:
I suppose this will save the window with the current camera settings? Is there a corresponding functionality to `save` already?

Datseris:
> I suppose this will save the window with the current camera settings? 

Yeap this is why it is super convenient, because it uses the figure I got after moving around / zooming / rotating, and I believe it is so much easier to do this with the mouse than to start testing numbers upon numbers with a save command

SimonDanisch:
you mean save a png or serialize the scene to a file or something yet different?

mkborregaard:
Matplotlib has this as a button on the window, which to my aesthetics makes for a rather clunky plot window. Plotly has it as an invisible mouseover in the top right corner, if you do decide to implement this that would be prettier. But could an alternative be to have `save(scene)` be sensitive to the current state of the scene window? Would make sense I think.

Datseris:
I mean save a png @SimonDanisch

mschauer:
Rather, `save(scene)` can have a default to do one thing and an option to do the other.

Datseris:
I think you are misjudging the power that this button provides... and in general the power of the buttons of the window of matplotlib should not be understated. Just yesterday I was trying to solve a question in a quantum transport simulation. Both the save button as well as the "undo" button, which undoes the last zoom action of he user, where very useful for my work.

This goes way beyond development and touches the user's work, so you should look at it from a different perspective as well.

mkborregaard:
I'm not very familiar with matplotlib, but should things like zooming not be handled easily in makie with mouse interaction?

Datseris:
To the best of my knowledge you can't (not yet) zoom in a rectangle with make. You can only use the scroll wheel. Of course this is trivial to undo with rolling the wheel back, but if you zoom in a rectangle that is also displaced an undo button is much more powerful.

SimonDanisch:
Yes, you can, space + left drag ;)
It's very easy to implement those buttons.
Just add it and hook it up to `save(scene)` ;) 

History is a bit more involved, but basically boils down to add a vector of camera states to the camera, and save the old state in this function:
https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/camera/camera2d.jl#L65
and connect a button to it as well ;)

[...]

PR #5702, opened 2026-07-16
All the new GUI features combined

SimonDanisch:
This merges:
https://github.com/MakieOrg/Makie.jl/pull/5510
https://github.com/MakieOrg/Makie.jl/pull/3491
https://github.com/MakieOrg/Makie.jl/pull/5650
https://github.com/MakieOrg/Makie.jl/pull/5628
While it's not amazing to have them all in one big PR, i did want to heavily test and document all of them together, and make sure we really nail the event system and polish some other features we may need for more complex GUIs.

https://github.com/user-attachments/assets/e33d1987-6139-44d3-8467-a267bb1fa015


hz-xiaxz:
This is so damn cool!

Makie/src/basic_recipes/editabletext.jl
```diff
@@ -601,7 +601,7 @@ function plot!(plot::EditableText)
         # Hide selections when the editor isn't focused — matches typical
         # text-input UX where the highlight disappears on click-away.
         visible = plot.focused,
-        space = :pixel, transformation = :nothing, inspectable = false,
+        space = plot.space, transformation = :nothing, inspectable = false,
```
ffreyer, line 604: Does this work with any space other than pixel?

Makie/src/camera/camera.jl
```diff
@@ -119,11 +119,9 @@ Returns true if the current mouseposition is inside the given scene.
...
 is_mouseinside(x) = is_mouseinside(get_scene(x))
 function is_mouseinside(scene::Scene)
-    return scene.visible[] && in(Vec(scene.events.mouseposition[]), viewport(scene)[])
-    # Check that mouse is not inside any other screen
-    # for child in scene.children
-    #     is_mouseinside(child) && return false
-    # end
+    scene.visible[] || return false
+    in(Vec(scene.events.mouseposition[]), viewport(scene)[]) || return false
+    return receives_events(scene)
```
ffreyer, line 124: Why would a scene need to receive events for the mouse to be inside it? If there are multiple scenes where only one has focus it seems pretty natural to me to switch between them by clicking on an inactive one with a `if is_mouseinside(inactive_scene)` check

Makie/src/camera/camera2d.jl
```diff
@@ -329,8 +329,13 @@ struct UpdatePixelCam
...
 
 function (cam::UpdatePixelCam)(window_size)
+    # Project absolute window pixels (not viewport-local) to NDC so a coord
+    # `(X, Y)` in a campixel scene renders at window pixel `(X, Y)` regardless
+    # of where the scene sits in its parent. For a scene whose viewport is at
+    # the window origin this is identical to the old `(0..w, 0..h)` projection.
+    vx, vy = Float64.(minimum(window_size))
     w, h = Float64.(widths(window_size))
-    projection = orthographicprojection(0.0, w, 0.0, h, cam.near, cam.far)
+    projection = orthographicprojection(vx, vx + w, vy, vy + h, cam.near, cam.far)
```
ffreyer: So with this you wouldn't be plotting to a (0..w, 0..h) range which is fully defined by the local scene, but a (x..x+w, y..y+h) range which is only defined in the context of the whole scene tree? That does not seem like a good change to me. And unnecessarily breaking. I'd rather see a new camera for this if it's needed

Makie/src/GUI/hovermenu.jl
```diff
@@ -0,0 +1,102 @@
...
+    # Create background box using layout (for @forwarded_layout blocks)
+    g.box = Box(
+        g.layout[1, 1:3];
+        height = g.height,
+        width = g.width,
+        color = g.bar_color,
+        cornerradius = g.corner_radius,
+        strokewidth = 1,
+        strokecolor = g.bar_strokecolor
+    )
```
ffreyer, line 29: With complex/block recipes this can just be `Box(g[1, 1:3], ...)`. `@forwarded_layout` is gone now, iirc.

Note that with nested attributes we can now have `g.box.color` etc. as attributes. That would take priority over g.box as a field in getproperty, iirc, but that shouldn't be a problem if all the relevant attributes are part of the HoverMenu. You can also do
```julia
@attributes begin
    box = @attributes begin
        documented_attributes(Box)...
    end
end
```
within a `@Block`. `filtered_attributes` should work too. You can still get the blocks from `g.blocks` or the internal layout outside `initialize_block!()` if you need them.

Makie/src/interaction/events.jl
```diff
@@ -54,6 +54,104 @@ function disconnect_screen(scene::Scene, screen)
...
+    for field in CONTEXT_EVENT_FIELDS
+        s = getfield(src, field)
+        d = getfield(dst, field)
+        push!(
+            obsfuncs, on(s; priority = priority) do value
+                d[] = value
+                return Consume(false)
+            end
+        )
+    end
```
ffreyer: This means that all events of a child/target happen at the same priority in the parent/source. With default `priority=0` you'd have an order like this:
```
30 parent interaction 1
1 parent interaction 2
0.20 child interaction 1
0.-200 child interaction 2
-1 parent interaction 3
```
Adding a parent interaction with priority 0 could happen before, after or between any of the child interactions.

That's what you naturally get by just forwarding events to another Observable and it never seemed right to me. I always though that priorities should be preserved, so you'd get this order:
```
30 parent interaction 1
  20 child interaction 1
1 parent interaction 2
-1 parent interaction 3
  -200 child interaction 2
```
That's what we currently have by just using the identical events object in every scene. I guess if we want to keep that we'd need a new Observable type that manages a shared pool of listeners alongside a local one?

Makie/src/makielayout/blocks/legend.jl
```diff
@@ -1110,11 +1111,58 @@ function get_plots(scene::Scene)
...
+    isempty(plots) && error("There are no plots with labels in the given axis that can be put in the legend. Supply labels to plotting functions like `plot(args...; label = \"My label\")`")
+    pos_kw = legend_position_to_aligns(position)
+    return Legend(
+        ax.parent, plots, labels, title;
+        bbox = ax.scene.viewport,
+        margin = margin,
+        pos_kw...,
+        kwargs...
+    )
+end
```
ffreyer, line 1160: Just fyi, I removed a bunch of `Legend` constructors on breaking because they were using attributes before the default attribute processing of blocks takes place and replaced them with `initialize_block!` methods. Overwrites that don't use attributes (like this one) should be fine.

Makie/src/scenes.jl
```diff
@@ -437,6 +446,55 @@ function root(scene::Scene)
...
+    # `intersect` on `Rect2i` returns negative widths for disjoint rects, which
+    # turns `glScissor` into `GL_INVALID_VALUE` (and Cairo's clip into a no-op).
+    # Clamp to an empty rect at the intersection origin so callers get a sane
+    # "draw nothing" instead.
+    w = widths(rect)
+    if w[1] < 0 || w[2] < 0
+        return Rect2i(minimum(rect), Vec2i(0, 0))
+    end
+    return rect
+end
```
ffreyer, line 499: > Intersection of every ancestor's viewport (not including `scene`'s own).

This seems like a bad idea to me. With this plots could draw outside the the viewport, potentially anywhere in the window. Depending on the scene tree, you could get different clipping results. It could also be that some drivers require or set scissor <= viewport.

Makie/src/makielayout/blocks/table.jl
```diff
@@ -0,0 +1,693 @@
...
+        cell_fontsize = t.cell_fontsize,
+        row_height = t.row_height,
+        cell_padding = t.cell_padding,
+        show_grid = t.show_grid,
+        grid_color = t.grid_color,
+        grid_linewidth = t.grid_linewidth,
+        show_vertical_lines = t.show_vertical_lines,
+        show_horizontal_lines = t.show_horizontal_lines,
+        inspectable = false
+    )
```
ffreyer: Looks like a case for just calling `tableplot!(scene, t.attributes, t.data, inspectable = false)`

Makie/src/makielayout/types.jl
```diff
@@ -1470,6 +1474,308 @@ end
...
+        "Controls if the parent layout can adjust to this element's width."
+        tellwidth = false
+        "Controls if the parent layout can adjust to this element's height."
+        tellheight = false
+        "The horizontal alignment of the subfigure in its suggested bounding box."
+        halign = :center
+        "The vertical alignment of the subfigure in its suggested bounding box."
+        valign = :center
+        "The alignment of the subfigure in its suggested bounding box."
+        alignmode = Inside()
```
ffreyer: These are added automatically (with the same values) by `mixin_block_layout_observables`. Could remove them.

Makie/src/makielayout/types.jl
```diff
@@ -1470,6 +1474,308 @@ end
...
+        "Controls if the parent layout can adjust to this element's width."
+        tellwidth = true
+        "Controls if the parent layout can adjust to this element's height."
+        tellheight = true
+        "The horizontal alignment of the table in its suggested bounding box."
+        halign = :center
+        "The vertical alignment of the table in its suggested bounding box."
+        valign = :center
+        "The alignment mode of the table in its parent GridLayout."
+        alignmode = Inside()
```
ffreyer: Only `width` is needed here

Makie/src/makielayout/types.jl
```diff
@@ -1470,6 +1474,308 @@ end
 end
 
 
+"""
+    Subfigure(fig_or_scene; kwargs...)
+
+A clipped, optionally event-isolated, scrollable region with its own
+`Scene` paired with a `GridLayout` — the same shape as a `Figure`, scoped
+to a sub-region. Place blocks via `Axis(subfig[1, 1])` etc., or plot
+directly into `content_scene(subfig)` (in viewport-local pixel coords).
```
ffreyer: `Subfigure` makes me think this is just an area of a figure that acts as figure itself, like the stupidly simple `Container` block I added. But this also has a bunch of plots and the scroll interactions too... Makes me wonder if this is more than it should be, or if it should be split up into multiple things. I.e. should the plots be something that is added to a Subfigure (e.g. via Label, Box, ...) rather than being part of it directly? And should the scrolling be an add-on? Or perhaps the difference between Container and Subfigure? (which should be named more appropriately then)

Makie/src/bezier.jl
```diff
@@ -116,7 +116,19 @@ function bbox(commands::Vector{PathCommand})
         else
             endp = endpoint(prev)
             _bb = cleanup_bbox(bbox(endp, comm))
-            bb = bb === nothing ? _bb : union(bb, _bb)
+            # NB: don't use `union(bb, _bb)` here. GeometryBasics considers any
+            # Rect with a zero-width dimension "empty" and returns the other
+            # operand, so accumulating the bboxes of axis-aligned segments
+            # (whose segment bbox is legitimately zero-width/height, e.g. a
+            # rectangle marker) would drop earlier segments and yield a
+            # degenerate path bbox. Grow the corners directly instead.
```
ffreyer: Also fixed by https://github.com/JuliaGeometry/GeometryBasics.jl/pull/276 / GeometryBasics 0.5.11

Makie/src/event_routing.jl
```diff
@@ -0,0 +1,86 @@
...
+
+Scenes that paint a translucent backdrop via plots instead of an opaque
+`clear = true` background (e.g. a modal dialog's overlay) can opt in
+explicitly with `scene.captures_mouse = true`.
+"""
+function covers_pointer(scene::Scene)
+    scene.visible[] || return false
+    scene.captures_mouse && return true
+    return scene.clear[] && z_world(scene) > 0
+end
```
ffreyer, line 15: I don't get this function. The mouse pointer/cursor is always on top of what's drawn, isn't it?

> True when `scene` is the topmost layer over its viewport

It doesn't check that?

> visible

Whether a scene is visible or not doesn't really matter for event processing? We could forbid it, but I don't see why we would

> `clear[] == true`

If clear worked as expected you could say that you can't interact with a parent scene if it's covered by a child scene that clears everything the parent scene draws. But that's not how clear works atm, at least not in all backends. See #4150, #4724. It also only makes sense when you're talking about interacting with visual elements, which may not always be the case.

> positive world-z

This scene having positive z does not guarantee that it covers things from it's parent. The parent can still draw at higher z, or plots in this scene could leave gaps. Negative z in this scene may also still draw over its parent. And in 3D this doesn't work, because z != depth.

Also what about sibling scenes? They could cover this scene too, if that's what this function or more generally the functions in this file are trying to figure out. Note though that the render order of scenes can be inconsistent between backends so any what-covers-what analysis is risky atm. (Again, see #4150, #4724)

Makie/src/camera/camera2d.jl
jkrumbiegel: I think it was the other way around, campixel seemed to plot to x,y in window coordinates regardless of scene position, and I thought it should be relative to the scene viewport instead. So as far as I remember I changed campixel to work that way and a fallback for the old behavior was added

ffreyer: Without this pr, on breaking:

```julia
using GLMakie
scene = Scene(camera = campixel!)
left = Scene(scene, viewport = Rect2f(0, 0, 100, 100), camera = campixel!, clear = true, backgroundcolor = :lightblue)
scatter!(left, Rect2f(5, 5, 90, 90), color = :blue)
right = Scene(scene, viewport = Rect2f(300, 300, 100, 100), camera = campixel!, clear = true, backgroundcolor = RGBf(1, 0.8, 0.8))
scatter!(right, Rect2f(5, 5, 90, 90), color = :red)
scene
```

image

jkrumbiegel: maybe I'm misremembering, it was a while ago..

chriswaudby:
Super exciting!

ffreyer:
Maybe closes #208

Makie/src/makielayout/blocks/spinner.jl
```diff
@@ -0,0 +1,57 @@
...
+    # Each `running = true` bumps `gen`; the loop reads its own captured
+    # generation and exits when a newer animator takes over, so rapid
+    # true/false/true toggles can't leave overlapping tasks running.
+    gen = Ref(0)
+    start_animator! = () -> begin
+        gen[] += 1
+        my_gen = gen[]
+        @async while sp.running[] && gen[] == my_gen
+            sp.visible[] && (frame_idx[] = frame_idx[] + 1)
+            sleep(sp.frame_interval[])
```
jkrumbiegel: this kind of stuff is generally much nicer with ticks now, instead of spinning up async tasks. it's also better for generating videos at non-real-time

PR #5664, opened 2026-06-08
add option to style WGLMakie's Tooltip just like GLMakie's DataInspector

bjarthur:
the offline WGLMakie ToolTip has a grey background, rounded corners, and no triangle:

Screenshot 2026-06-08 at 10 33 21 AM

whereas the GLMakie DataInspector looks like this:
Screenshot 2026-06-08 at 10 22 49 AM

this PR adds two fields to the WGLMakie ToolTip struct and two corresponding keyword arguments to permit the developer to style it as they wish, and specifically provides a stylesheet which matches GLMakie DataInspector.  the default stays the same though so as not to be breaking.

so with this PR it is possible to make WGLMakie ToolTip look like this:

Screenshot 2026-06-08 at 10 42 59 AM

- [x] New feature (non-breaking change which adds functionality)

### Checklist

- [x] Added an entry in CHANGELOG.md (for new features and breaking changes)
- [x] Added or changed relevant sections in the documentation
- [ ] Added unit tests for new algorithms, conversion methods, etc.
- [ ] Added reference image tests for new plotting functions, recipes, visual options, etc.

should this PR be merged (hopefully!), i have a follow on PR which adds the option to show the tooltip when hovering, just like in GLMakie, instead of having to click.

PR #5715, opened 2026-07-27
add option to display tooltips in WGLMakie when hovering

bjarthur:
as [promised](https://github.com/MakieOrg/Makie.jl/pull/5664), this PR brings more feature parity to WGLMakie by adding an option to display offline tooltips when hovering over a point, just like in GLMakie.

## Checklist

try it out for yourself!

```
using WGLMakie, Bonito

# -----------------------------------------------------------------------------
# Test 1: Click Trigger (trigger = :click)
# Tooltip ONLY appears when clicking on a point.
# -----------------------------------------------------------------------------
fig1, ax1, sc1 = scatter(rand(10));

tt_click = WGLMakie.ToolTip(
    fig1,
    js"(plot, index) => 'Clicked point #' + (index + 1)",
    plots = [sc1],
    trigger = :click,
    range = 3,
);

App(DOM.div(fig1, tt_click))


# -----------------------------------------------------------------------------
# Test 2: Hover Trigger (trigger = :hover)
# Tooltip appears as you hover near a point (within 15px).
# -----------------------------------------------------------------------------
fig2, ax2, sc2 = scatter(rand(10));

tt_hover = WGLMakie.ToolTip(
    fig2,
    js"(plot, index) => 'Hovered point #' + (index + 1)",
    plots = [sc2],
    trigger = :hover,
    range = 3,
);

App(DOM.div(fig2, tt_hover))
```


bjarthur:
runic fails perhaps because i used v1.7 and CI uses 1.4?  https://github.com/MakieOrg/Makie.jl/blob/master/.github/workflows/_formatting.yml#L17

bjarthur:
until https://github.com/MakieOrg/Makie.jl/pull/5716 is merged, when running the above example scripts you'll need to execute `App(...` twice.

bjarthur:
ok, so, #5716 actually didn't fix the need to call `App` twice.  but adding a semicolon to the call to `scatter` did.

bjarthur:
all checks pass!  ready for review.

SimonDanisch:
Thank you! Two things before merge:

As far as I can tell, the hover `requestAnimationFrame` loop never stops. It keeps running after the figure
leaves the DOM and its closure pins `scene`/`canvas`/`popup`, so the scene never gets
collected. And since `requestAnimationFrame(frame)` comes *after* `process(e)`, one throw breaks it. 
Scheduling a frame per mousemove instead fixes both:
```js
    let scheduled = false;
    canvas.addEventListener("mousemove", (e) => {
        pending = e;
        if (scheduled) return;
        scheduled = true;
        requestAnimationFrame(() => {
            scheduled = false;
            if (pending && !pressed) { const ev = pending; pending = null; process(ev); }
        });
    });
```

`pick_closest` needs a null check. When the scene can't be rendered, `pick_native`
returns `undefined`. The `range == 0` path handles that (`if (!picked) return;`), but
`pick_closest` destructures the result straight away and throws. Hover
always goes through `pick_closest`, so this can now fire on any frame. and together with
the loop issue above, one occurrence ends hovering for good. Adding
`if (!picked) return [null, 0];` in `pick_closest` covers it.

bjarthur:
those suggestions look good to me.  i went ahead and committed them. CI is green

PR #5716, opened 2026-07-27
WGLMakie: fix a DOM-mount race in three_display, add AwaitedNode

bjarthur:
# Description

  - Interpolating a DOM `Node` into `js"..."` (used for `three_display`'s `wrapper`/`canvas`) compiled to a one-shot `document.querySelector(...)` with no retry — unlike `Scene` interpolation, which already retries via a configurable window (#5693). On a slow/cold first page load (fresh browser tab, first WGLMakie display in a process), the JS module could finish loading before Bonito had actually mounted that specific node, silently querying `null` and failing scene initialization inside `setup_scene_init`'s own `try/catch` — with no error surfacing on the Julia side.
  - Adds `AwaitedNode`, wrapping a node so its lookup retries until the node is mounted, reusing the same `WGLMAKIE_SCENE_RETRY_DELAY_MS`/`WGLMAKIE_SCENE_RETRY_TOTAL_MS` env vars already used for `Scene` lookups. `three_display` now awaits `wrapper`/`canvas` alongside the JS module, instead of interpolating them directly.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Fixed a WGLMakie race where a slow/cold first page load (fresh browser tab, first
+  display in a process) could cause `three_display` to silently fail to initialize the
+  canvas: interpolating a DOM node into `js"..."` compiled to a one-shot
+  `document.querySelector(...)` with no retry, which could run before Bonito had actually
+  mounted the node. Added `AwaitedNode`, which retries until the node appears, mirroring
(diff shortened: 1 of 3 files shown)


bjarthur:
this actually didn't fix the race condition problem in #5715, but claude thinks it's a good hardening measure anyway.  i'll leave it as a draft for the maintainers to consider.

SimonDanisch:
Hm, this seems like something which should live in Bonito.
Also, bonito already has:
```
Bonito.Sessions.on_node_available($(id), 100)
```

PR #5717, opened 2026-07-28
feat(text)!: Pluggable text layout via Makie.layout_text

jkrumbiegel:
Supersedes #5632, opened separately so that PR's review history stays readable. It refactors the text pipeline so other packages can hook in their own way of rendering text or text-like objects: real LaTeX through MakieTeX (companion PR MakieOrg/MakieTeX.jl#73), or something like [TeXLayout.jl](https://github.com/dawbarton/TeXLayout.jl), which currently has to resort to type piracy for this.

The extension point is `Makie.layout_text`, which returns glyph data, other plot specs, or both, for a handler type and/or a text type:

```julia
struct MyHandler end

function Makie.layout_text(::MyHandler, src::LaTeXString, attributes::Makie.TextAttributes)
    return Makie.TextLayout(...)
end
```

Makie's own layout is the `handler === nothing` implementation of that same function, so there is one protocol rather than an extension API bolted next to the built-in string types, and a type a handler has no method for falls through to Makie without the handler doing anything.

Picking the handler in the theme means a whole figure can be rendered with LaTeX text without threading handlers through every call:

```julia
with_theme(text_handler = MakieTeX.LaTeX(render_strings = true), fontsize = 15) do
    fig = Figure()
    ax = Axis(fig[1, 1],
        title = L"y(t) = A\,e^{-\lambda t}\cos(\omega t + \varphi)",
        xlabel = L"time $t$ (s)", ylabel = L"amplitude $y$", xticks = 0:2:10)
    t = range(0, 10, length = 400)
    lines!(ax, t, exp.(-0.3 .* t) .* cos.(2 .* t), label = L"\lambda = 0.3")
    lines!(ax, t, exp.(-0.3 .* t), linestyle = :dash, label = L"e^{-\lambda t}")
    axislegend(ax, position = :rt)
    fig
end
```

figure with title, labels, ticks and legend rendered by a real TeX run

Title, labels, tick labels and legend entries all come from a real TeX run here.

## Glyphs and specs

Text is now a scatter-like plot of glyphs, and that idea gets its own plot type: `Glyphs`. `Text` becomes a container with a `Glyphs` child for the glyph batch and a plotlist for everything that isn't a glyph, which is how MathTeXEngine's rules and MakieTeX's image markers get drawn. Glyphs keep a fast path; the specs are slower, but they are flexible enough to switch between text and plot types dynamically, and these text systems aren't meant for performance sensitive work anyway. The point of MakieTeX is high quality static plots, not animating huge equations in a live GUI.

Specs are placed through their plot's `model` matrix rather than by rewriting their arguments, so any plot type can be a spec (an `image`, say), a spec's arguments stay in the block's layout frame, and moving the camera or the text updates one matrix per spec. Markers rasterized for the GPU backends (MakieTeX's PDF labels) are rasterized at the screen's `px_per_unit` in GLMakie and re-rasterized when it changes, so they stay sharp on hidpi screens and in high-resolution saves of an already displayed figure.

Moving text around doesn't relayout, because `align`, `rotation` and `offset` are applied downstream of layout, which for a LaTeX engine is the difference between moving a label and recompiling it. Almost every other attribute does relayout, which seems fine given how rarely they are animated.

The split also makes display-only updates cheap for plain text, since glyph geometry is reused instead of laid out again. This holds with a `text_handler` set: only blocks the handler actually claims (has a `layout_text` method for) re-run layout on a recolor, fall-through blocks keep the fast path.

```
# 4000 strings, 26893 glyphs, CairoMakie, realizing the render outputs
#                     before     after
# create              36.1 ms   34.0 ms
# color update        32.4 ms    5.0 ms
# fontsize update     33.7 ms   32.7 ms
# position update     33.8 ms    0.1 ms
```

## Breaking

Per-character styling goes away. A vector `color`, `strokecolor`, `strokewidth`, `fontsize` or `rotation` on `text` is one value per string now, and a length that doesn't fit errors. This is for shaping systems like HarfBuzz, where several code points become one glyph, so there is nothing for a per-character vector to index; we can style glyphs, not characters. `rich` text already covers the use case, since it styles ranges over a string rather than characters.

- `Text` is a container plot, so code that read glyph render attributes (`sdf_uv`, `quad_offset`, `quad_scale`, ...) off a `Text` has to read them from its `Glyphs` child.
- `pathtext` draws into a `Glyphs` child instead of a `Text` one.
- `align = (halign, :baseline)` now puts `rich` text and `LaTeXString`s on their baseline instead of their bottom edge, matching plain strings; for multi-line `rich` text that is the last line's baseline, also matching plain strings.
- With `justification = automatic`, a fractional `halign` justifies by that fraction instead of falling back to `0.5`.

Also fixed on the way: `rich` text in `pathtext` ignored the plot's `color` for the parts it didn't style itself, and applied `alpha` twice.

# Caveats

- Non-glyph specs are one plotlist entry per text block, so N LaTeX labels are N `LineSegments` plots. A camera move only updates each spec's `model` matrix, but merging same-type specs across blocks would still reduce the plot count; only text that emits specs pays anything at all, so keeping the implementation simple seems worth more until that turns out to be wrong.
- `MathTeXHandler` doesn't handle `word_wrap_width` yet; the built-in LaTeX path still does.
- Removing the old `text("string", position = ...)` form needs a deprecation cycle, so it stays as is here, per @ffreyer's note on #5632.
- Per-glyph `strokewidth` still collapses to one value in GLMakie and WGLMakie, whose shaders bind it to a uniform float. Same as on master, CairoMakie draws it per glyph.


ffreyer:
> The split also makes display-only updates cheap for plain text, since glyph geometry is reused instead of laid out again:
> 
> ```
> # 4000 strings, 26893 glyphs, CairoMakie, realizing the render outputs
> #                     before     after
> # create              36.1 ms   34.0 ms
> # color update        32.4 ms    5.0 ms
> # fontsize update     33.7 ms   32.7 ms
> # position update     33.8 ms    0.1 ms
> ```
> 

Is the before the other pr or the breaking branch (or master)?

Makie/src/basic_recipes/text.jl
```diff
@@ -38,7 +38,11 @@ convert_attribute(str::AbstractString, ::key"text", ::key"text") = Ref{Any}([str
 convert_attribute(rt::RichText, ::key"text", ::key"text") = Ref{Any}([rt])
 convert_attribute(x::AbstractVector, ::key"text", ::key"text") = Ref{Any}(vec(x))
 
-to_string_arr(text::AbstractVector) = text
+# `copy` so the producer emits a fresh `input_text` array each run. An aliased
+# array makes `is_same` (which can't tell whether a shared array was mutated in
+# place) report a change, re-running text layout on every position update while
+# layout solves, which for image handlers (LaTeX) means recompiling every pass.
+to_string_arr(text::AbstractVector) = copy(text)
```
ffreyer: I also introduced `ExplicitUpdate` for this kind of thing when I reworked Axis:

```julia
map!(plot, ..., :output) do ...
    ...
    return ExplicitUpdate(data, has_updated ? :force : :deny)
end

# consumers need to remove the wrapper
map!(plot, :output, ...) do output
    value = unwrap_explicit_update(output) # or output.data
end
```

Makie/src/basic_recipes/text.jl
```diff
@@ -435,33 +676,54 @@ get_text_type(::T) where {T} = T
...
+    return plotlist!(plot, plot._shifted_text_specs)
+end
 
-    return linesegments!(
-        plot, plot.linesgments_shifted; linewidth = plot.linewidths,
-        color = plot.linecolors, space = plot.markerspace
+function register_glyphs!(plot)
+    return glyphs!(
+        plot, plot.per_glyph_positions;
+        glyphindices = plot.glyph_indices,
```
ffreyer: Would be good to not name this `register` since we use that a lot for "add computation to compute graph". Probably better to just have this as part of `plot!(::Text)`

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
-    map!(attr, [:text_blocks, :positions_transformed_f32c], :per_char_positions_transformed_f32c) do blocks, pos
-        if length(blocks) != length(pos)
-            error("Text blocks and positions have different lengths: $(length(blocks)) != $(length(pos)). Please use `update!(plot_object; arg1/arg2/text/position/color/etc...) to update multiple attributes together.")
-        end
-        return [p for (b, p) in zip(blocks, pos) for i in b]
+    # One data-space anchor position per glyph (repeated within a string). The
+    # Glyphs child transforms these; per-string transform reuse is task #8.
+    map!(attr, [:positions, :text_blocks], :per_glyph_positions) do positions, blocks
+        return Point3f[to_ndim(Point3f, sv_getindex(positions, i), 0) for (i, r) in enumerate(blocks) for _ in r]
     end
```
ffreyer, line 717: This should probably still use `positions_transformed_f32c` as an input with the child plots using `transformation = :nothing` so transform_func, f32c and model don't apply per char but per string.

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
 end
 
-function map_per_glyph(text_blocks::Vector{UnitRange{Int}}, Typ, arg)
-    isscalar(arg) && return fill(arg, last(last(glyphs)))
-    result = Typ[]
-    per_glyph_attributes(text_blocks, (arg,)) do a
-        push!(result, a)
-    end
-    return result
+spec_position_bboxes(specs) = Rect3d[Rect3d(first(spec.args)) for spec in specs]
```
ffreyer: This seems too simplified to me. If the spec plot is something like scatter this would fail to capture the size of markers, for example.

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
-            x, y = position
-            p0 = rotation * to_ndim(Point3f, fontsize .* Point2f(x, y) .- tex_offset, 0) .+ offset
-            p1 = rotation * to_ndim(Point3f, fontsize .* Point2f(x + h.width, y) .- tex_offset, 0) .+ offset
-            push!(outputs.linesegments, p0, p1)
-            push!(outputs.linewidths, fontsize * h.thickness, fontsize * h.thickness)
-            push!(outputs.linecolors, color, color)
-            push!(outputs.lineindices, block_idx => pos_idx, block_idx => pos_idx)
-        end
+function layout_text(::MathTeXHandler, str::LaTeXString, attributes::TextAttributes)
+    (; color, strokecolor, strokewidth) = attributes
```
ffreyer: Why does this exist alongside the `layout_text(::Nothing, ::LaTeXString, attributes)` path? From what I can tell this skips word_wrap_width and otherwise does the same?

jkrumbiegel:
I'll have it redo these, I thought it was the other branch but I might be wrong 

Edit:

This is the table where claude ran all four

```
At 26893 glyphs (CairoMakie, realizing the render outputs, best of 7):

┌─────────────────┬──────────┬──────────────────┬──────────────┬──────────┐
│                 │  master  │ ff/breaking-0.25 │ old PR #5632 │ this PR  │
├─────────────────┼──────────┼──────────────────┼──────────────┼──────────┤
│ create          │ 38.97 ms │ 37.04 ms         │ 36.16 ms     │ 32.77 ms │
├─────────────────┼──────────┼──────────────────┼──────────────┼──────────┤
│ colour update   │ 32.35 ms │ 31.90 ms         │ 33.32 ms     │ 5.25 ms  │
├─────────────────┼──────────┼──────────────────┼──────────────┼──────────┤
│ fontsize update │ 34.09 ms │ 33.14 ms         │ 34.79 ms     │ 32.03 ms │
├─────────────────┼──────────┼──────────────────┼──────────────┼──────────┤
│ position update │ 34.21 ms │ 32.82 ms         │ 0.022 ms     │ 0.099 ms │
└─────────────────┴──────────┴──────────────────┴──────────────┴──────────┘
```

Makie/src/compute-plots.jl
```diff
@@ -664,8 +664,11 @@ function register_marker_computations!(attr::ComputeGraph)
...
 end
 
+# Text is a container (Glyphs + plotlist children) and never rendered directly, but it
+# stays here because it converts its own attributes and needs register_camera! to project
+# its plotlist block positions. The leaf traversal (is_atomic_plot) skips the container.
 const PrimitivePlotTypes = Union{
-    Scatter, Lines, LineSegments, Text, Mesh,
+    Scatter, Lines, LineSegments, Text, Glyphs, Mesh,
     MeshScatter, Image, Heatmap, Surface, Voxels, Volume,
 }
```
ffreyer: Seems awkward to keep Text as a primitive plot type. Maybe we should instead reconsider when `convert_attribute` applies and make it something any recipe can opt into? (`register_camera` can easily be called in a normal recipe plot)

Makie/src/interfaces.jl
```diff
@@ -1,5 +1,7 @@
+# `text` has no `plot!` body; it builds its Glyphs + plotlist children in
+# `calculated_attributes!`, so it lives here to satisfy the atomic-plot contract.
 const atomic_functions = (
-    text, meshscatter, scatter, mesh, linesegments,
+    text, glyphs, meshscatter, scatter, mesh, linesegments,
     lines, surface, volume, heatmap, image, voxels,
 )
```
ffreyer: I'd rather have `plot!(::Text)` than keep it here

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+
+        buffer = cached === nothing ? GlyphBuffer() : GlyphBuffer(cached)
+
+        if cached !== nothing && text_handler === nothing && all(display_independent_layout, input_text) &&
+                !changed.input_text && !changed.fontsize && !changed.selected_font &&
+                !changed.resolved_justification && !changed.lineheight &&
+                !changed.word_wrap_width && !changed.fonts
+            refill_display_attributes!(buffer, computed_color, strokecolor, strokewidth)
+            return node_outputs(buffer)
+        end
```
ffreyer: Can't this be its own separate computation, like with `register_glyph_placement`?

Makie/src/interfaces.jl
jkrumbiegel: yeah good point, seems more standard

Makie/src/basic_recipes/text.jl
jkrumbiegel: true, good catch

Makie/src/basic_recipes/text.jl
jkrumbiegel: it got kind of on a tangent with MathTeXHandler, seems refactorable

Makie/src/layouting/text_boundingbox.jl
```diff
@@ -21,8 +21,8 @@ end
 # Utility
 function text_bb(str, font, size)
     rot = Quaternionf(0, 0, 0, 1)
-    layout = glyph_collection(str, font, size, 0.0f0, 0.0f0, 0.0f0, 0.0f0, -1, rot)
-    return unchecked_boundingbox(layout.glyphindices, layout.char_origins, size, layout.glyph_extents, rot)
+    layout = layout_string(str, font, size, 0.0f0, 0.0f0, -1)
+    return unchecked_boundingbox(layout.glyphindices, layout.origins, size, layout.extents, rot)
 end
```
henrik-wolf: This function is (only) called once in `Axis3D`, to get the width of the ticklabels, I believe(?).
Maybe `Axis3D` can also get the size from the `layout_text` result and we can remove this function?

Makie/src/layouting/text_layouting.jl
```diff
@@ -90,34 +88,29 @@ function create_lineinfos(charinfos, word_wrap_width)
 end
 
 """
-    glyph_collection(str::AbstractString, font_per_char, fontscale_px, halign, valign, lineheight_factor, justification, rotation, color, word_wrap_width)
+    layout_string(str::AbstractString, font_per_char, fontscale_px, lineheight_factor, justification, word_wrap_width)
```
henrik-wolf, line 91: With the layouting code for rich text and latex strings living in `basic_recipes/text.jl`, I think it would make sense to either put this function (or this whole file) over there as well, or (probably helpful for readability) separate out the individual layouting functions into their own files.

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+"""
+    display_independent_layout(text) -> Bool
 
+Whether a text value's laid-out glyph geometry is independent of the display
+attributes (color, strokecolor, strokewidth). Plain strings apply those per-glyph
+*after* layout, so their geometry can be reused when only display attributes change.
+`RichText` and `LaTeXString` bake color into their layout, so they return `false`
+and recompute. Custom text types default to `false` (conservative).
+"""
+display_independent_layout(::AbstractString) = true
```
henrik-wolf, line 399:
```suggestion
display_independent_layout(::AbstractString) = true
display_independent_layout(::LaTeXString) = false
```

Makie/src/basic_recipes/text.jl
henrik-wolf: It seems to be a worse version of whatever it is the `nothing` handler does by default. I guess it is here to have a "worked example" to reference in the `explanations/latex.md`

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+        N = length(input_text)
+        for (name, value) in [
+                (:fontsize, fontsize), (:font, selected_font), (:lineheight, lineheight),
+                (:word_wrap_width, word_wrap_width), (:color, computed_color),
+                (:strokecolor, strokecolor), (:strokewidth, strokewidth),
+            ]
+            validate_per_string(name, value, N)
+        end
+
+        for (i, str) in enumerate(input_text)
```
henrik-wolf, line 549: Would it make sense to be able to set the `text_handler` per `input_text`?

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +86,579 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
-        :text_strokewidth, :text_strokecolor,
-        :linesegments, :linewidths, :linecolors, :lineindices,
+    outputs = collect(fieldnames(GlyphBuffer))
+    return register_computation!(attr, inputs, outputs) do inputs, changed, cached
+        (; input_text, text_handler, fontsize, selected_font, resolved_justification) = inputs
+        (; lineheight, word_wrap_width, fonts, computed_color, strokecolor, strokewidth) = inputs
+
+        buffer = cached === nothing ? GlyphBuffer() : GlyphBuffer(cached)
+
+        if cached !== nothing && text_handler === nothing && all(display_independent_layout, input_text) &&
```
henrik-wolf: Should `display_independent_layout` be a function of `text_handler` as well as `input_text`?
```julia
all(x->display_independent_layout(x...), zip(text_handler, input_text))
```
this way something like `HarfBuzz` could opt in to the fast path as well.

henrik-wolf (commented):
Very nice to see this taking shape. Having an explicit glyphs plot type as well as the ability to return arbitrary plot specs is probably all that will ever be needed.

In terms of implementation, I guess there is no way around merging glyphs into a global buffer if we want to maintain drawing performance at least for the default `text_handler`? I was playing with the idea of having the drawing backends collect all the characters from potentially multiple glyph plots in each text before plotting them. That would still leave the overhead of registering and diffing the plot specs when neccessary, but since I have no intuition nor any idea how to reasonably benchmark this compared to keeping around a special path/case, I did look into it much. :)

Makie/src/basic_recipes/text.jl
dawbarton: I think that would be a good idea. At the moment, I've kept TeXLayout.jl deliberately colour agnostic but if `display_independent_layout(::LaTeXString) = false`, then it'll keep recomputing with different colours.

dawbarton:
I've had a brief look through this in relation to TeXLayout.jl and it looks pretty good. Thanks for your work on this.

Being able to select which attributes force a recompute would be good, e.g., TeXLayout.jl is agnostic to colour, stroke, etc. 

Though that said, there is one aspect that leaks through - the previous integration with MathTeXEngine.jl had separate HLine and VLine implementations as well all the glyphs (though VLine was broken/ignored) which were automatically filled with the appropriate colour. Without those as additional primitives (or just a simple Rect), we have to use a PlotSpec that bakes in a colour/other attributes unnecessarily. If it was possible to have a simple line primitive, that would be helpful. (E.g., for fractions, roots, drawing bars for matrices, etc.)

jkrumbiegel:
I think I've addressed most of your comments now @henrik-wolf and @dawbarton. In the end, I've decided for now against a "baking in" fast path because I'm not sure yet how much that'll matter in practice, and I want to keep my options open. I assume most uses of the new API will be more for static texts like in MakieTeX and not for huge amounts of changing text, so while nice in principle, optimizations are now a bit premature, too, if they make the API surface larger. The most important thing, position shifting, already doesn't recompute, and this would be annoying because ticks and axes shift around on layout updates. Color changing happens much less often and usually not in tight animation loops, and again it would probably be fast enough for normal amounts of texts as it is anyway.

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +91,615 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+    map!(attr, inputs, :baked_display_attributes) do text, handler, color, strokecolor, strokewidth
+        # here rather than downstream so a bad length is reported when the plot is
+        # created, not when its colors are first pulled
+        for (name, value) in [(:color, color), (:strokecolor, strokecolor), (:strokewidth, strokewidth)]
+            validate_per_string(name, value, length(text))
         end
+        baked = !isempty(text) && any(str -> bakes_display_attributes(handler, str), text)
+        baked || return nothing
+        return (color, strokecolor, strokewidth)
     end
```
ffreyer, line 512: `return nothing` is probably a bug here. 

If the output is not initialized, `return nothing` will cause it to be initialized to `nothing`. I think that's what you want here.

If the output is already initialized `return nothing` causes the update to be discard. I.e. the old value remains and propagates as "up to date". So `baked_display_attributes` can not dynamically switch to `nothing` like this. I think after init `nothing` is always considered as "nothing changed here", so we'd need to disambiguate it.

Tangentially, if you have multiple outputs and `return nothing` in an init run, all outputs become nothing. I would've expected an error there. Maybe that should be changed?

Makie/src/basic_recipes/text.jl
```diff
@@ -56,22 +64,27 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+            return ((a_pos,), text_update(args[1], args_changed[1]))
         elseif args isa Tuple{<:AbstractVector{<:Tuple{<:Any, <:VecTypes}}}
             # [(text, pos), ...] argument
-            return ((last.(args[1]),), Ref{Any}(first.(args[1])))
+            return ((last.(args[1]),), text_update(first.(args[1]), args_changed[1]))
         else # assume position data
-            return (args, Ref{Any}(to_string_arr(a_text)))
+            return (args, text_update(to_string_arr(a_text), text_changed))
         end
     end
```
ffreyer, line 84: We're still not doing any kind of depwarn here or elsewhere for `text(..., position = ...)`. Also, do we want to deprecate the tuple version as well?

Makie/src/basic_recipes/text.jl
```diff
@@ -81,343 +94,619 @@ function register_arguments!(::Type{Text}, attr::ComputeGraph, user_kw, input_ar
...
+
+function Makie.layout_text(::MyHandler, str::AbstractString, attributes)
+    (; fontsize, color) = attributes
+    # ...
+end
+```
+"""
+struct TextAttributes
+    font::NativeFont
+    fonts::Any
```
ffreyer, line 297: Can we not fix the type to a `Dict`?

WGLMakie/src/plot-primitives.jl
```diff
@@ -266,7 +269,7 @@ function create_shader(scene::Scene, plot::Scatter)
         return (dict,)
     end
 
-    map!(attr, [:marker, :scaled_color], :scatter_color) do marker, color
+    map!(attr, [:image, :scaled_color], :scatter_color) do marker, color
         if marker isa AbstractMatrix
```
ffreyer, line 279: This fixes the rasterized marker thing in WGLMakie but still doesn't consider px_per_unit. I don't know how to include that without it being a mess...

For context, I moved the ppu handling into `all_marker_computations` so GLMakie doesn't rasterize twice. So `all_marker_computations` is adding a dummy `px_per_unit = 1f0` node here and uses that for rasterization.

ffreyer:
My commits are mostly cleanup. I'll push another to fix tests once they finish locally

ffreyer:
Heh, alright, benchmarking `Text` wasn't very representative. Benchmarking `text!(scene, 0, 0)` shows around 37% each in `plotlist!` and `glyphs!`. 
```julia
@profview for _ in 1:100
    scene = Scene()
    for _ in 1:100
        text!(scene, 0, 0)
    end
end
```
To optimize `glyphs` we could probably write a specialized `register_arguments!(Glyphs, ...)` which skips the work we have already done in `text!`. I think that's literally everything, i.e. the function could just `ComputePipeline.alias!` all the stages, ignoring expand_arguments, convert_arguments, dim converts etc?

We could also look into making a specialized `Plot{plotlist}(...)` method, since it doesn't need the same argument converts as usual. That might also be helpful for normal plots. Iirc there were some checks/branching that slowed things down in case we had specs... I think I'll look into that a bit.

It might also be possible to just skip the plotlist plot entirely until there are specs to process?

jkrumbiegel:
Have you actually diagnosed that these are what makes this slower? Or is it some compilation thing when plotting? Or is it what you say and it's just the larger number of text plots in a normal plot that cause this?

ffreyer:
I didn't directly compare breaking to this pr but since I've done a bunch of benchmarking on plot creation on breaking already I think I have a decent feeling for it. 

`register_arguments!()` is always a good chunk of the plot create time (here too) and this pr goes from 2 runs (text and linesegments) to 3 (text, glyphs, plotlist). Cutting that out for glyphs should do a fair bit. I'll probably try it later since I'm already done with plotlist.

ffreyer:
Ugh, nvm, most of the time in `register_arguments` is spent on resolving. That needs to happen either way to fill out `Plot{func, typeof(graph.converted[])}`

Makie/src/basic_recipes/text.jl
ffreyer: Did this move into `layout_text`? I guess that would be fine since a new text handler/string type could define more verbose spec bboxes if needed.

PR #5727, opened 2026-07-31
Add shader-based edge stroking to mesh

jkrumbiegel:
Mesh with edge lines is currently not well represented by mesh faces plus a separate `wireframe`/`lines` overlay: the lines z-fight with the face colors, which makes the apparent stroke widths irregular, edges and joints don't connect correctly, and the wireframe draws duplicated lines because it is not pruned, so adjacent faces produce double edge lines (and triangulation diagonals show up even where the mesh is conceptually made of quads or polygons).

This PR pulls stroking into the mesh shader itself, making mesh with stroke a primitive. `mesh` gets `strokewidth`, `strokecolor` and `strokeedges` (`:boundary` strokes only edges belonging to exactly one face, e.g. a polygon outline; `:all` additionally strokes shared edges). Each fragment shows either stroke or face color, decided by its screen-space distance to the visible edges of its own triangle, so there is nothing layered on top that could fight or double-blend. Each face gets its stroke drawn inward: shared edges get half the width from each adjacent face, so the halves combine into one full-width line. Edges with only one face currently get the full width inward instead of half, so a polygon outline is exactly `strokewidth` wide and lies fully inside the shape (this is a design decision that is up for discussion, the alternative would be a uniform half width everywhere). Since the neighboring face of a silhouette edge on a closed surface faces away, those edges show just the visible face's half.

This works for triangle, quad and mixed face meshes. Non-triangular face types now survive `convert_arguments`, and edges introduced by triangulating them are never stroked (**minor breaking**: `plot.mesh[]` can now contain e.g. `QuadFace` faces; backends consume the triangulated `:faces` compute attribute). To keep corners intact, each triangle corner also carries up to two roughly coplanar "wing" edges from neighboring triangles, so stroke bands continue across triangulation seams instead of leaving notches.

```julia
ps = [Point2f(x, y) for y in 0:1 for x in 0:2]
quads = [QuadFace{Int}(1, 2, 5, 4), QuadFace{Int}(2, 3, 6, 5)]
mesh(GeometryBasics.Mesh(ps, quads), color = :lightblue, shading = NoShading,
    strokewidth = 10, strokecolor = :black, strokeedges = :all)
```

showcase: icosphere, quad torus, Möbius strip boundary, colormapped terrain, all with transparency

## Comparisons with the overlay approach

Sphere pole, where the line-based wireframe gets especially artifact-y:

sphere pole zoom comparison

Z-fighting up close on a quad torus:

torus zoom comparison

With semi-transparent strokes, the overlay double-blends where the line overlaps the fill and at every join, while the shader stroke is a single uniform band (also note the miter spikes changing the footprint on the left):

star polygon comparison, opaque and semi-transparent strokes

Adjacent shapes keep separate strokes instead of overlapping into darker bars:

adjacent rects comparison

A fun consequence of the stroke being part of the surface: a fully transparent stroke color punches the edges out of the mesh, leaving floating faces. The equivalent overlay just draws nothing:

transparent stroke color comparison

This also makes `wireframe` mostly unnecessary: with `color = :transparent` and `transparency = true` only the strokes remain, giving the familiar see-through wireframe but with uniform line widths and without doubled edges:

wireframe comparison

## CairoMakie

Cairo's canvas model cannot express per-fragment stroking, so stroked meshes render through a new software rasterizer: a Julia port of the mesh shaders (perspective-correct interpolation, FastShading, pattern/matcap colors, strokes) that renders one mesh plot into one supersampled RGBA image with a private depth buffer, which is dropped after rendering, and composites it like an image plot covering the mesh's projected bounding box. Faces render back to front against the depth buffer, so semi-transparent meshes composite correctly within a plot. A depth buffer for meshes is desirable beyond stroking anyway, since drawing triangles with Cairo without one is bad; routing more mesh rendering through this (and combining plots into a shared depth buffer) could follow later. Meshes without stroke keep the existing vector drawing path, so existing output is unchanged. Matching the two backends surfaced that GLMakie sampled pattern colors in physical pixels, so pattern density changed with `px_per_unit`; that is fixed here by sampling in logical pixels like CairoMakie does. This is the same showcase as above rendered by CairoMakie, including the transparency:

CairoMakie rendering of the showcase figure

## WGLMakie

WebGL has no `gl_PrimitiveID`, so WGLMakie renders meshes de-indexed (one vertex per triangle corner with a trivial index buffer), which makes `gl_VertexID / 3` the primitive index for fetching the per-triangle stroke data from a 2D data texture. It also has no `noperspective` interpolation, so the fragment position comes from `gl_FragCoord` mapped back into viewport-local logical pixels. Picking keeps the original vertex indices through a dedicated attribute, and the output matches GLMakie. The cost is that meshes lose vertex sharing in WGLMakie (roughly 3x vertex memory); this could be gated on stroking being active at plot creation if that is a concern.

# Caveats / open questions

- Strokes wider than the local triangle size get clipped at triangulation seams, since each fragment only knows the edges of its own triangle plus the corner wings. Fine for reasonable widths, degrades for very wide strokes on fine triangulations.
- Corner joins at reflex corners are round, not mitered. A `joinstyle` equivalent would need extra work.
- `GeometryPrimitive` conversions (`Sphere`, `Rect3`, ...) still triangulate at conversion time, so `strokeedges = :all` gives a triangle wireframe for those instead of quads. Could be changed by keeping their native face types too.
- WGLMakie de-indexes all mesh plots, stroked or not, to keep one code path. If the memory tradeoff matters, de-indexing could be decided at plot creation based on `strokewidth`.
- A follow-up could lower `poly`'s stroke onto this instead of its `lines` overlay.


ffreyer:
> This also makes `wireframe` mostly unnecessary

Mostly meaning for triangle and quad meshes, but not faces with > 4 vertices?

jkrumbiegel:
@ffreyer I meant mostly as in "seems to cover the functionality but I didn't think carefully about all the edge cases"

ffreyer:
image

The k = 1 wings can create inwards folding lines, e.g. on a sphere (on both sides, i = 0, 1 on the left, i = 1, 2 on the right)

ffreyer:
> > > This also makes `wireframe` mostly unnecessary
> > 
> > Mostly meaning for triangle and quad meshes, but not faces with > 4 vertices?
>
> I meant mostly as in "seems to cover the functionality but I didn't think carefully about all the edge cases"

I didn't really understand what the wings were doing and thought their number was related to the amount of triangles something decomposes into. Or something like that...

Anyway, found an edge case that doesn't work:

```julia
ps = [Point2f(0), [Point2f(cos(x), sin(x)) for x in range(0, 3pi/2, 30)]...]
push!(ps, Point2f(0.01, -0.1)) #  This causes issues
fs = [NgonFace(eachindex(ps)...)]
m = GeometryBasics.Mesh(ps, fs)
f,a,p = mesh(
    m,
    color = :lightblue, strokewidth = 50, strokecolor = :black, strokeedges = :all
)
wireframe!(triangle_mesh(m), depth_shift = -1f-3)
f
```

image

(I changed the shader to output reds for the first 3, greens for the second 3 and blues for the third 3 stroke data entries. And also white for the mesh color.)

jkrumbiegel:
@ffreyer yes the stroke/wing mechanism degenerates for very wide strokes or the case that you showed. I updated so that the wings are only drawn up to a max distance, that prevents the long fold-backs that you showed above I think.

In general, this looks better than `wireframe` for the super common cases and meshes, but users will still be able to call `wireframe!` directly if they need the old rendering method. Maybe that could be a keyword on mesh that can only be set at creation time and then reroutes the attributes to a `wireframe` as before?

And I would still not use it for `poly` as there the lines should really be on top of the edges, and this mechanism can only draw as far as the triangle fragments go, so it moves the outer line a bit inward to compensate.

And just to keep it here, this is an image claude generated for me that shows why the wings are necessary, it's the adjacent triangles that share a bit of the edge stroke of their neigbor:

image

CairoMakie/src/mesh.jl
```diff
@@ -521,8 +431,10 @@ function draw_atomic(scene::Scene, screen::Screen, plot::Makie.Surface)
     attr = plot.attributes
     Makie.add_computation!(attr, Val(:surface_as_mesh))
     Makie.register_pattern_uv_transform!(attr)
+    Makie.register_surface_stroke!(attr)
+    Makie.compute_colors!(attr)
 
-    draw_mesh3D(scene, screen, attr)
+    draw_mesh_rasterized(scene, screen, attr, grid_size = size(attr.z[]))
```
ffreyer, line 437: Should we really entirely remove vector rendering for surface? Even 2D surfaces in a 2D Axis?
```julia
f = Figure()
a = Axis(f[1, 1])
surface!(a, zeros(10, 10), color = rand(10, 10))
f
```

CairoMakie/src/mesh.jl
```diff
@@ -33,105 +33,44 @@ end
...
-    end
-    color = compute_colors(attr)
-    cols = per_face_colors(color, nothing, fs, nothing, uv)
-    if cols isa Cairo.CairoPattern
-        align_pattern(cols, scene, attr.model[])
-    end
-    return draw_mesh2D(screen, cols, vs, fs)
+    raw_uv_transform === Makie.automatic || return false
+    return Makie.cameracontrols(scene) isa Union{Camera2D, Makie.PixelCamera, Makie.EmptyCamera}
 end
```
ffreyer, line 56: A mesh is rasterized if:
1. It has stroke
2. It does not have a hatch pattern
3. It has a user given uv_transform
4. It is not in a scene with Camera2D, PixelCamera or EmptyCamera

(2) hits pretty much every mesh already, since most meshes are not hatched. E.g. a `Circle(...)` with a flat color, no shading and no 3D transformations drawn in an Axis will be rasterized because of this

(3) From what I remember uv_transform is just a 2D rotation + scale + translation (+ maybe mirror/shear) matrix. Even if the user throws a random matrix at a mesh here, shouldn't that still be resolvable as a vector pattern?

(1) I'm still wondering if that can be done with Cairo commands as well. Maybe by stroking mesh patterns instead of patches?

To me vector rendering is the main selling point of CairoMakie. That includes meshes as well (2D and 3D) even if rasterized drawing is often better. Maybe we can formally add `rasterize` as an attribute for this (it already kind of exists in plot-primitives.jl), have Cairo decide the default (as well as veto, if necessary)?

CairoMakie/src/mesh_rasterizer.jl
```diff
@@ -0,0 +1,626 @@
...
+        ylo = min(ylo, p[2])
+        yhi = max(yhi, p[2])
+    end
+    isfinite(xlo) && isfinite(ylo) || return
+    x0 = clamp(floor(Int, xlo) - ss, 0, ceil(Int, viewport_w))
+    y0 = clamp(floor(Int, ylo) - ss, 0, ceil(Int, viewport_h))
+    x1 = clamp(ceil(Int, xhi) + ss, 0, ceil(Int, viewport_w))
+    y1 = clamp(ceil(Int, yhi) + ss, 0, ceil(Int, viewport_h))
+    W = cld(x1 - x0, ss) * ss
+    H = cld(y1 - y0, ss) * ss
```
ffreyer, line 569: This seems to be padding for subpixel rendering... Shouldn't it be using +- px_scale?

CairoMakie/src/mesh.jl
jkrumbiegel: I mean yes vector graphics are the main thing, but this mostly relates to lines, and text which are thin and look bad quickly when zooming in otherwise. Also note that the mesh drawing In cairomakie so far also always rasterized because it used these coons patches for triangles which are not part of svg. So you wouldn't have gained crispness from that. The only thing that makes sense to vector draw in the mesh space are basically polygons I think, or if you deconstructed a mesh into polygonal faces that all have a single color.

ffreyer: Hmm, I assumed it was vectorizing because the (sometimes massive) difference in file size. But yea, svgs are actually just creating many smaller images. PDFs on the other hand do seem to vectorize. If I import one into inkscape I can see each individual triangle as a path which can be moved, recolored and stroked.

jkrumbiegel: Oh right PDF's spec is larger than svg, in this case it does support the primitive. But for most meshes it would still not be good to use it, as separate triangle rendering without a depth buffer gives incorrect overlaps a lot of the time, plus you see these white fringes between the triangles because each triangle is rendered separately and includes some background in its antialiased edge. I might be wrong about this, but I'd say nobody wants meshes as vector graphics :) 2D polygons are a different category for me even though GLMakie etc draw them as meshes.

asinghvi17: Eh personally I do use a lot of vector rendered surfaces.  There are some cool things in the svg spec that would allow direct surface writing without rasterization in the 2d case, but unfortunately you can't get that through Cairo....

jkrumbiegel: do you have an example for a surface that works well rendered as a vector? I assume pdf given we established svg rasterizes this path anyway?

CairoMakie/src/mesh_rasterizer.jl
jkrumbiegel: As far as I understand it's for making a ring around the image so that Cairo's extend_pad mode doesn't smear the pixels of the mesh at the border. So it's one pixel of the buffer, not depending on screen scale

ffreyer: Hmm, yea, it's probably fine. Stepping away from an edge pixel by +-ss should always move you to a new supersampled cluster of pixels that does not contain data, i.e. give you a 1px border.

CairoMakie/src/mesh.jl
ffreyer: > gives incorrect overlaps a lot of the time

I think the per triangle depth sort was actually pretty ok? And it's not an issue in 2D

> plus you see these white fringes between the triangles because each triangle is rendered separately and includes some background in its antialiased edge

This depends on how the mesh is rendered. If every triangle is a new `CairoPatternMesh` you will see borders. If each triangle is a "patch" within a `CairoPatternMesh` you will not see borders. I did some testing with that in #5446, see the torus examples. All of them are just different triangle sorting and patch vs mesh pattern. (Maybe this is also viewer dependent with svg and pdf?)

ffreyer:
Some more rendering edge cases I found (GLMakie):

Another poke-through/folding issue with surface:
```julia
r = range(-3, 3, length = 11)
f,a,p = surface(
    [sin(x) * cos(y) for x in r, y in r], 
    strokewidth = 5, strokecolor = :black, strokeedges = :all
)
```
Screenshot from 2026-08-17 16-05-37

And the bottom of a Quad-faced sphere mesh seems to struggle with linewidths:
```julia
s = Sphere(Point3f(0), 1f0)
m = normal_mesh(s,facetype = QuadFace{UInt32})
f,a,p = mesh(m, strokewidth = 5, strokecolor = :black, color = :yellow, strokeedges = :all)
```
Screenshot from 2026-08-17 16-06-54
I'm guessing this is because one of the triangles is degenerate/0 volume. Maybe this is also fixed in your GeometryBasics branch? I can't check it out atm because github is having trouble

(I don't care that much if these edge cases get fixed. I mostly just want to log what I find)

jkrumbiegel:
The last one looks exactly like a case I hit when preparing the presentation, and there it was because the pole vertex wasn't bit exact the same for all triangles

ffreyer:
Can confirm the sphere thing is fixed, but the poke through on the surface still happens. My gut feeling is that the poke through is not (entirely) fixable, that it's something like z-fighting, so I'm not too concerned with getting it perfect.

Do we want to bring back (some of) the CairoMakie vector mesh paths? And are there any other TODO's left?

jkrumbiegel:
I'm not sure about cairomakie meshes. I'd like to avoid the coons patch stuff to be honest. Always felt hacky

SimonDanisch:
This came from my review: https://github.com/MakieOrg/Makie.jl/pull/5781 . After we merged that I'm happy to merge this pr.

SimonDanisch (approved):
happy to merge this once CI is green.

ffreyer:
My plan is to get the text refactor merged first. After that I'll probably take another look at rasterized vs vector meshes since there have been some fixes on master

ffreyer:
I restored vector paths for 2D flat colored meshes and 2D Makie.ImagePattern meshes. The latter only rasterizes the pattern, which then gets sampled across the vectorized mesh, if I understand correctly. So edges should be cleaner with this, and memory usage might be lower.

I did not restore the surface vector path, 2D vertex color vector path, 2D texture vector path or any 3D vector path.

I think it might also be worth considering running stroked, flat colored and patterned 2D meshes through vectorized paths, but I'm not sure how to do that. For `strokeedges = :boundary` we'd need to figure out the boundary path (which I think Cairo is doing already?) and for `:all` we'd need to not rely on triangle faces. That probably complicates the path merging though.

ffreyer:
The "Stage Camera" test shows off some issues with transparency. The rasterizer tries to render back to front by sorting faces, but that can go wrong. E.g.:
```julia
f,a,p = mesh(Rect3f(-5, -5, -0.1, 10, 10, 1), color = :lightgray, alpha = 0.8)
```
produces
image
I guess this doesn't really need to be fixed here because it's still closer to what it should be than before. A correct version would probably also be significantly more expensive to run...

PR #5736, opened 2026-08-04
Fix log-scale Axis zoom limits

### LINKED ISSUE #5735: Log-scale Axis zoom can produce invalid limits
## Problem

Two Axis interactions can construct limits outside the finite, positive domain required by `yscale = log10`.

1. **Rectangle zoom:** after zooming out, dragging a nonzero rectangle from above a positive curve to below it can produce a lower limit of `0.0`. The rectangle is visibly inside the axis, but the interaction warns and refuses the zoom.
2. **Scroll zoom:** repeated wheel zoom-out expands the transformed range until inverse transformation underflows to `0.0`. `ScrollZoom` assigns that invalid limit directly, so the window event callback throws.

Both cases should keep the axis within its current valid domain: a valid visible rectangle should clamp to the visible positive region and zoom; an unrepresentable scroll proposal should leave the last valid limits unchanged.

## Minimal WGLMakie example

```julia
using Bonito
using Makie
using WGLMakie

Bonito.Page(listen_port = 8002)
Bonito.browser_display()
WGLMakie.activate!()

figure = Figure(size = (900, 650))
axis = Axis(
    figure[1, 1];
    title = "Log-axis zoom MWE",
    yscale = log10,
    limits = (-20, 30, 1.0e-28, 2.0e6),
)
x = range(0, 10; length = 500)
lines!(axis, x, 1.0e-8 .* exp.(1.5 .* x); linewidth = 3)

display(Bonito.App() do _
    WGLMakie.WithConfig(figure)
end)
```

### Reproduction

1. Wheel-scroll down repeatedly to zoom out. Before the fix, eventually the callback reports invalid `log10` y-limits with a lower bound of `0.0`.
2. Start from the zoomed-out view and left-drag a rectangle whose first point is above the curve and whose second point is below it. Before the fix, the interaction can reject a visibly nonzero selection with an invalid `0.0` lower y-limit.

## Expected

- Rectangle selection is direction-independent and clamps to the currently visible valid log domain.
- Scroll zoom never assigns invalid/nonfinite limits or throws from a window-event callback.
- If a numerical proposal is unrepresentable, preserve the current limits.

## Environment

Observed with WGLMakie over VS Code Remote SSH; the failure is in Makie's Axis interaction code and is backend-independent once the input event reaches the Axis.

jacobleft:
Fixes #5735.

## Summary

- Normalize both RectangleZoom start endpoints and drag updates through the visible transformed limits before inversion.
- Clamp the completed rectangle selection to the current visible data limits, so a log-domain crossing zooms to the positive visible boundary instead of being rejected.
- Validate every Axis interaction limit proposal before mutation and apply the same guard to ScrollZoom, preventing inverse-transform underflow/overflow from reaching `targetlimits`.
- Include rejected selection and visible-limit context in the RectangleZoom warning.

## Root cause

RectangleZoom clamped only move events, not both drag-start endpoints. ScrollZoom directly assigned inverse-transformed limits even after a log-scale endpoint had underflowed to zero.

## Verification

- `Makie/test/interactivity/Axis.jl`: 85 passed / 85 total.
- Added synthetic regressions for direction-independent log rectangle selection, lower-domain clamping, invalid selection rejection, and log ScrollZoom underflow preservation.


ffreyer:
From trying the pr, the zoom out fix seems good. The rect select fix seems weird, as it forces the rectangle to the axis boundary. Still better than erroring though. Outside of that, translation (after zoom out) also causes these kinds of issues. We could probably fix it with the same reject-bad-limits fix that zooming has?

jacobleft:
Implemented the pan guard from the review comment. `DragPan` now keeps the current limits when its inverse-transformed proposal falls outside the active axis scale domain, matching `ScrollZoom` behavior.

Added a log10 regression where a full-height pan from an extreme-but-valid range would previously underflow to `0.0`; it now completes without a warning or exception and retains the prior limits.

Validation: `Makie/test/interactivity/Axis.jl` — 87/87 passed.

ffreyer:
The RectangleZoom snapping to the axis boundary when you're fully zoomed out was caused by `Rect` saving widths instead of two points. That causes float precision issues when going back to `maximum = origin + widths`. I adjusted some code to keep the minimum and maximum around longer and that seems to fix the problem.

PR #5526, opened 2026-02-11
Fix Resampler for RGB-images

johroj:
# Description

Fixes calling `heatmap(Resampler(img::Matrix{RGB}))`, which is used in the example at https://docs.makie.org/dev/reference/plots/heatmap#Plotting-large-Heatmaps. Running this exampled resulted in a conversion error from `RGB{}` to `Float32` [here](https://github.com/johroj/Makie.jl/blob/master/ComputePipeline/src/ComputePipeline.jl#L586). 

I believe this was probably broken [here](https://github.com/johroj/Makie.jl/commit/90bff47ee745bc351c3e3d1fcf7782541a670ace) or [here](https://github.com/johroj/Makie.jl/commit/09fcab1da0150f2008a075b6bae6e084bf1702cb#diff-e5d29117c4a61ff872d1749369ca4c0c465854ee7b99fa7fb1ef542a25de9c01). 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

PR #5738, opened 2026-08-07
Enable axis scaling for Resampler

johroj:
# Description

Made `Resampler` not ignore axis scaling settings. 

Its a follow-up on https://github.com/MakieOrg/Makie.jl/pull/5526 that got stuck on my computer for some time, sorry about the delay. 

The implementation should not change anything when identity scaling (default) is used. I made some quick tests to catch any regression in the future. When I tried to write better tests, I also found that I wanted to change some of the computations, which would make everything much more complicated. So I'm keeping it simple. Let me know if you need something more rigorous. 

## Type of change

## Checklist

### DIFF
--- a/Makie/src/basic_recipes/datashader.jl
+++ b/Makie/src/basic_recipes/datashader.jl
@@ -526,6 +526,7 @@ If the array doesn't support this, it will be converted to an interpolation obje
 * `method` is the interpolation method used, defaulting to `Interpolations.Linear()`.
 * `update_while_button_pressed` will update the heatmap while a mouse button is pressed, useful for zooming/panning. Set it to false for e.g. WGLMakie to avoid updating while dragging.
 * `lowres_background` will always show a low resolution background while the high resolution image is being calculated.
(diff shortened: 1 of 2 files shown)


Makie/src/basic_recipes/datashader.jl
```diff
@@ -641,7 +642,16 @@ function resample_image(x, y, image, max_resolution, limits)
...
         len = min(resolution, max_resolution[i])
-        return LinRange(max(1, indices[1]), min(indices[2], si), len)
+        transformed_inds = apply_transform(
+                inverse_transform(transform_func[i]), 
+                LinRange(
+                    apply_transform(transform_func[i], max(1, indices[1])),
+                    apply_transform(transform_func[i], min(indices[2], si)), 
+                    len
+                )
+            )
```
ffreyer, line 652: Indices aren't the right thing to scale. You can see this if you compare, for example:
```julia
img = Makie.loadasset("cow.png")
f,a,p = heatmap(1..10, 1..10, img, axis = (xscale = log10,))
a2,p2 = heatmap(f[2, 1], 1..10, 1..10, Resampler(img), axis = (xscale = log10,))
f
```
Instead this should act on the x and y coordinates in some way and then derive indices from that. (Might also be useful to know that `add_axis_limits` adds `:axis_limits_transformed` if those are needed.)

A very simple way to fix this would also be to just switch to `heatmap` as the child plot type but that might be bad for performance.

PR #5781, opened 2026-09-10
Reduce mesh stroke overhead

SimonDanisch:
Three performance fixes for #5727, measured against `4c433f6ca`.

- **`strokewidth` no longer invalidates the adjacency.** The width was an input to
  `stroke_edge_data` but only used for its `iszero` check, so every change recomputed the
  whole edge and wing search. It now depends on a `Bool` `:stroke_enabled` node.
  179k triangles: 185 ms → 0 ms per change.
- **`corner_wings!` is allocation free.** One pass tracking the best two candidates,
  written into the caller's scratch arrays, instead of `filter` + two `filter!` +
  `partialsort` per corner. `:all` on 179k triangles: 151 ms / 204 MiB → 92 ms / 69 MiB.
- **WGLMakie only de-indexes when stroking**, like `Surface` already did. Unstroked meshes
  were paying ~4x the vertex bytes over the websocket and on the GPU; a 576-vertex sphere
  now sends 576 vertices instead of 3174.

Wing slot order can differ from before. It isn't part of the contract — both consumers
`min` over all six slots and the tests compare with `Set` — and the wings are identical as
a per-corner set on seven meshes in both `strokeedges` modes, `edge_widths` bit-identical.
Reference scenes render identical at `px_per_unit` 1 .

### DIFF
--- a/WGLMakie/src/plot-primitives.jl
+++ b/WGLMakie/src/plot-primitives.jl
@@ -698,7 +698,7 @@ function create_shader(scene::Scene, plot::Makie.Mesh)
 
     # Stroking needs the de-indexed mesh path, which replaces the shared vertices by one
     # per triangle corner, so it is only taken when stroking is enabled at plot creation.
-    if !iszero(plot.strokewidth[])
+    if plot.stroke_enabled[]::Bool
         register_wgl_mesh_expansion!(attr)
         register_wgl_mesh_stroke!(attr)
         return create_wgl_renderobject(
--- a/WGLMakie/src/plot-primitives.jl
+++ b/WGLMakie/src/plot-primitives.jl
@@ -618,6 +618,54 @@ function register_wgl_mesh_stroke!(attr)
     return
 end
 
+# Inputs of the de-indexed (stroked) mesh and surface programs.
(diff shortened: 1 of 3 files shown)


WGLMakie/src/plot-primitives.jl
```diff
@@ -647,35 +695,19 @@ function create_shader(scene::Scene, plot::Makie.Mesh)
...
-        :wgl_mesh_uv => :texturecoordinates,
-        :wgl_mesh_vertex_color => :vertex_color,
-        :wgl_stroke_data => :stroke_data,
-        :wgl_viewport_origin => :viewport_origin,
-    )
-    return create_wgl_renderobject(mesh_program, attr, inputs; rename_updates)
+
+    # Stroking needs the de-indexed mesh path, which replaces the shared vertices by one
+    # per triangle corner, so it is only taken when stroking is enabled at plot creation.
+    if !iszero(plot.strokewidth[])
```
ffreyer:
```suggestion
    if plot.stroke_enabled[]::Bool
```

ffreyer (approved):

PR #5682, opened 2026-06-23
Add tricontour recipe, fix contour Colorbar and constant scalar field crashes

vincmarks:
# Description

Fixes #5680. Fixes #5681.

While utilizing contour plotting in [Trixi.jl](https://github.com/trixi-framework/Trixi.jl) (see [PR #3088](https://github.com/trixi-framework/Trixi.jl/pull/3088)), I noticed some missing features and crashes in Makie's contour infrastructure. This PR addresses all of them.

### 1. `Colorbar` support for 2D `contour`

`Colorbar(fig[1,2], plt)` did not work for contour plots. This now works:

```
xs = LinRange(0, 10, 100)
ys = LinRange(0, 15, 100)
zs = [5 * cos(x) * sin(y) for x in xs, y in ys]

f, ax, h = contour(xs, ys, zs)
Colorbar(f[1, 2], h)
f
```
download-11

### 2. New recipe: `tricontour`

`tricontour` (unfilled isolines on unstructured/scattered data) was not yet implemented, despite `tricontourf` already existing. The new recipe accepts the same inputs as `tricontourf ((xs, ys, zs)` or a pre-built `DelaunayTriangulation.Triangulation`) and uses `TriplotBase.tricontour` to compute connected polylines. `Colorbar(fig[1,2], tr)` is supported.

```
xr = range(-π, π, length = 40)
yr = range(-π, π, length = 40)
xs = vec([x for x in xr, _ in yr])
ys = vec([y for _ in xr, y in yr])
zs = @. sin(xs2) + cos(ys2)

f, ax, tr = tricontour(xs, ys, zs; levels = 12)
Colorbar(f[1, 2], tr)
f
```
download-12

### 3. Constant scalar field handling

Plotting constant scalar fields (e.g. `fill(c, n)`) caused crashes across several plot types. This PR makes all of the following work correctly:

- `contour`/`tricontour`: constant fields now produce no lines (empty levels) instead of crashing with `Can't interpolate in a range where cmin == cmax`.
- `heatmap` and all colormapped plots: an explicit `colorrange = (c, c)` is now automatically expanded to `(c - δ, c + δ)`. 

example:

```
xr = range(-π, π, length = 40)
yr = range(-π, π, length = 40)
xs2 = vec([x for x in xr, _ in yr])
ys2 = vec([y for _ in xr, y in yr])
zs2 = ones(Int, length(xs2)) .* 5 

f2, ax2, tr2 = tricontourf(xs2, ys2, zs2; levels = 12)
[...]

PR #5787, opened 2026-09-14
Fix `contourf` drawing a constant field in two colors

### LINKED ISSUE #5786: `contourf` draws a constant field in two colors
### Description

For a constant field, `contourf` builds levels symmetric around the constant value
(`register_contourf_computations!`, added in #5682):

With the default `levels` = 10 that gives 11 edges, so the middle edge is exactly the
constant value. Isoband puts everything below that edge into one band and everything at or
above it into the next, so one ulp of noise in the data is enough to draw a uniform field
in two colors that are a colormap step apart.

Image

### Version
CairoMakie v0.15.14

vincmarks:
# Description

Fixes #5786

Follow-up to #5682 . 
For a constant field the levels are built symmetric around the constant value. With an even
number of levels the middle edge falls exactly on that value, so a single ulp of noise in
the data splits the field over two neighbouring bands and a uniform field comes out
speckled. See #5786

## Type of change

## Checklist


ffreyer (approved):
Not sure what's going on with ci, but it doesn't seem related to this pr.

vincmarks:
> Not sure what's going on with ci, but it doesn't seem related to this pr.

Thanks for checking!

PR #5792, opened 2026-09-16
Make plot output independent of Dict iteration order

jkrumbiegel:
🤖:

---

Julia 1.13 changed hashing, so `Dict` and `Set` iteration order changed with it. Several recipes take visible output from that order, which is why the reference tests fail on 1.13 in #5790, with the same 12 images failing on all three backends:

- Contours.jl traces cells in `Dict` order, so a closed contour line can start at any of its vertices. Labels sit at the middle of the vertex list, so they move. Closed lines now start at a fixed vertex.
- `voronoiplot` built polygons in `each_generator` order but colors in point index order and paired them by position, so tiles got the wrong colors. Generators are iterated sorted now.
- `triplot`'s convex hull is a closed cycle whose start vertex shifted the dash pattern of the outline.
- `hexbin` iterated its bin `Dict`, which changed the draw order of overlapping hexagon strokes.
- `datashader` took category colors, image draw order and legend entries from the order of the category `Dict`.

Rendering the affected figures with CairoMakie under 1.12 and 1.13 and scoring them the way the reference tests do:

```
                                             before      after
Voronoiplot with empty polygons              0.924       0.0
contour labels 2D                            0.356       1.4e-5
Triplot with ghost edges and convex hull     0.276       2.8e-5
hexbin colorrange highclip lowclip           0.424       0.017
```

The hexbin test had a second problem: 7 bins have counts exactly equal to its colorrange minimum of 10, where a float difference between the versions flips a tile to `lowclip`. Its colorrange is moved off the integer grid.

The sorting is over bins, generators and category names, never over data points: 0.84 ms of the 46 ms `hexbin` call for 1e6 points in 40k bins, 67 µs of the 79 ms `voronoiplot` call for 20k generators. Rotating closed contour lines costs one array copy per closed line, about 1% on a 400x400 contour.

The 12 affected images are pinned in `ReferenceTests/refimage_updates/` for all three backends. The set comes from running the CairoMakie reference tests and the affected GLMakie tests locally against the release references, and it covers every image that failed on 1.13 in #5790. WGLMakie was not run locally (it needs Electron), so its pins are taken from the same set.


ffreyer:
Guess this is good to go?

PR #5793, opened 2026-09-17
fix(GLMakie, WGLMakie): Center marker strokes and fix sdf marker geometry

jkrumbiegel:
Ok this one is for you to review @ffreyer as I don't 100% understand the fixes Claude did here. I went off the observation that GLMakie hexbin plots looked kind of uneven, which led to the realization that stroke doesn't match what CairoMakie does (centered on the lines) which I think is what you usually expect. So that's changed here. But also there were slight problems with the way the sdfs were computed under some conditions which was especially noticeable with the hexbin. So that's the other part of this PR which I don't really understand so well. But the end result of the process appeared to match CairoMakie much better, and we have even hexbins

---

## Example

`hexbin colorrange highclip lowclip` from the reference tests, GLMakie.

## Before

image

hexbin in GLMakie before

## After

image

hexbin in GLMakie after

10x zoom, GLMakie before | GLMakie after | CairoMakie:

hexbin seams zoomed, before, after and CairoMakie

Scoring the GLMakie and WGLMakie renderings against the CairoMakie one of the same figure:

```
             before   after
GLMakie      0.131    0.010
WGLMakie     0.125    0.010
```

### Strokes were drawn outside the outline instead of centered on it

`distance_shape.frag` put the stroke band at `-strokewidth < d < 0`, i.e. fully outside the shape, while CairoMakie strokes the path with `set_line_width`, which straddles it. Growing `strokewidth` from 0 to 8 on a hexagon marker grew its ink bounding box by 16px in GLMakie and by 8px (plus antialiasing) in CairoMakie. For hexbin that meant each tile's stroke reached a full `strokewidth` into its neighbours, so neighbouring strokes overlapped and whichever tile was drawn last won, which is also why that reference test was so sensitive to draw order.

The band is now `|d| < strokewidth/2`. The stroke is also composited over the fill with a proper `over` blend rather than `mix`, which was losing coverage where the stroke overlapped the fill's antialiasing ramp (strokes came out 9 to 16% too light).

### Stroke width was scaled by the sprite's aspect ratio

`sprites.geom` converts distances from uv units to pixels with one scalar, `min(abs(o_w.z), abs(o_w.w))`, which can only be right along the shorter sprite axis. For a sprite that is not square in pixels the stroke came out too wide along the other axis by exactly the aspect ratio. With a `:rect` marker at `markersize = Vec2f(300, 100)` and `strokewidth = 4`, the band measured 11.4px on one pair of edges and 3.8px on the other, against 4.0px everywhere in CairoMakie. hexbin passes `markersize = size .* (1/sqrt(3), 1/2)`, so its tiles are in exactly that situation, at a factor of 1.57.

Distance field markers now divide the distance by its local gradient. That gradient is already computed for the antialiasing radius (`aspect_corrected_local_aa_radius`), so this costs no extra texture samples, and it makes `aa_radius` collapse back to the plain constant.

Measured stroke ink over the analytically known perimeter, against CairoMakie:

```
                   strokewidth = 1   = 2     = 4
square polygon          0.995        0.994   0.965
hexagon                 0.996        0.984   0.971
circle                  0.998        0.992   0.960
```

(0.84 to 0.91 before, depending on marker size.)

### Marker sdfs were off-center in their atlas cell

`render` rasterized bezier paths at `floor(256/downsample)*downsample = 255` px while `render_path` quantizes the path's bounding box to 1/64 of its long side. Those don't line up, so the bitmap could be half a pixel wider than the path, and after padding its size usually wasn't divisible by `downsample` either. `sdistancefield` rounds up in that case and adds all of the extra space on the top/right, so markers whose bounding box isn't square ended up off-center in their cell and were drawn shifted along their narrow axis.

Ink centroid of a single marker at the center of a 400x400 figure (the exact center is 200.5, 200.5):

```
                    before            after
square polygon   (200.50, 200.50)  (200.50, 200.50)
wide 2x1         (200.50, 200.13)  (200.50, 200.50)
tall 1x2         (199.41, 200.50)  (200.50, 200.50)
hexagon          (199.77, 200.50)  (200.50, 200.50)
```

Rendering at `pix_per_glyph * downsample` with a matching quantization makes both bitmap dimensions exact multiples of `downsample`, so nothing is rounded and the sdf comes out symmetric. Markers now use `pix_per_glyph` texels along their long axis like glyphs do, instead of 51.

## Notes

- All bezier marker sdfs change, so the atlas serialization version goes v7 -> v9. v8 is skipped because a `v8_texture_atlas_*.bin` already exists on the v0.21.0 release and would be downloaded and used. `v9_texture_atlas_*.bin` needs uploading there, otherwise every user regenerates the atlas locally on first plot.
- Reference images shift in GLMakie and WGLMakie; the pins still need to be filled in from this PR's CI run.
- Text is unaffected: glyph sprites are square, so the gradient normalization is a no-op there. The tick label area of the figure above differs from master by 4e-5 mean absolute.


ffreyer:
I remember discussing the stroke differences before, not sure if it was just with Simon or with you too. Iirc Simon was partial towards additive stroke like W/GLMakie has, though I also thought CairoMakie was subtractive rather than centered.

In the context of hexbin I think subtractive stroke could also be worth considering since it would avoid overlaps in the transparent case. Though it's probably worse for the opaque case...

Maybe this is something we should just introduce an attribute for? `strokealign = :inside / :center / :outside`?

jkrumbiegel:
Cairo doesn't have the ability to do anything but centered, I guess that's the main reason why I find this the most consistent option. I just never realized there's a difference before noticing the hexbin problems

ffreyer:
I definitely thought about fixing this before, by matching CairoMakie to W/GLMakie. Naively you could just adjust markersize to mimic additive/subtracive stroke, but I remember some edge cases getting complicated. Probably something with `markerspace != :pixel` or perspective corrections.

Looking at refimages, a point in favor of additive stroke is that it doesn't take away anything from the original shape. This is nice for the very niche case of stroked, scattered images ("scatter Billboard and transform_marker for image markers") and stroked text which is often naturally thin (e.g. "Threading Test", "scatter with stroke", "scatter Billboard and transform_marker for Char markers", ...). Though stroked text also gets more self-overlap issues with additive stroke (e.g. "textlabel"?).

There is also something wrong with "3D text"

jkrumbiegel:
I think specifically for text, CairoMakie is configured to draw the text on top of the stroke, exactly like you say it becomes harder to read otherwise. It's an inconsistency in Makie, but stroke-on-center is otherwise the completely dominant paradigm everywhere out there. SVG can't even do anything else, you can see some reasons why not (has been considered for SVG2) here https://wiki.selfhtml.org/wiki/Zeichnen_in_SVG/SVG_2#stroke-align

ffreyer:
> I think specifically for text, CairoMakie is configured to draw the text on top of the stroke, exactly like you say it becomes harder to read otherwise. 

Doesn't seem to be true in refimages. CairoMakie and (changed) W/GLMakie match here: (WGLMakie shows the old refimg for reference)
image

CairoMakie without stroke:
image

Btw, I'm not against using centered stroke. If Simon is fine with it I'd be happy with it too. I care more about backends matching than which specific stroke alignment we use.

jkrumbiegel:
hm maybe I was misremembering this about CairoMakie then. In your example, WGLMakie looks the best of course, it's true that for small and detailed shapes, having a centered vs outline stroke takes away too much of the actual glyph. Although it's exacerbated by white stroke on white background, which is like cutting away glyph. Maybe black would look ok here?

ffreyer:
There are a couple other refimages that show the same issue with scatter. E.g.: 
(old GLMakie | CairoMakie which matches new GLMakie)
image
Screenshot from 2026-09-18 10-18-22

jkrumbiegel:
Hm yeah GLMakie looks preferable in those. You can get a similar effect in CairoMakie with drawing the stroke first, then layering the text on top. Might even be possible to avoid alpha layering problems if they are drawn in a group where the inside replaces the stroke layer, not sure if that makes sense? Because otherwise a half-transparent fill will half cover a stroke. Just like currently, a half-transparent stroke will half cover a fill.

Another tricky case here is stroke-only markers though, a common case for scatter plots because they can be read better under overplotting. With stroke-only, probably centered looks better than outer.

